PR #125, opened 2021-02-04
Add sibling GltfWriter library into cesium-native

Samulus:
This is a preliminary draft PR, but I wanted to open it asap to give cesium-native people the chance to point out anything especially egregious or misaligned with the philosophy of the cesium-native libraries. The goal of this library is to provide a general purpose GLTF writer for the Cesium ecosystem (primarily the tilers) that accepts a constructed `CesiumGltf::Model` object and serializes that to a glTF or GLB `std::vector<std::uint8_t>` (without the proliferation of `tinygltf:XXX` types throughout our codebases).

This PR is still in an alpha state, but has support for writing glTFs with embedded uris and glbs without references to external binary content. Self-contained glbs are of primary concern for the terrain tiler and we plan to have full support for those first (i.e we won't wait for the library to support every possible configuration of serialized glTF before merging this).

### Design Decisions
* I started initially by mimicking the JavaScript functions that generate Cpp code for parsing, but I realized that it would actually be faster to just use https://kakoune.org/ and https://neovim.io/ macros to generate the writing classes as I go. The primary benefit of this is that the (semi-automatically) generated serialization code is very easy to read and debug, but tightly coupled to the rapidjson API. I should clarify that only the PRIVATE api right now is coupled to rapidjson, if `simdjson` added a sax writing API tomorrow we could change the classes around without changing the public API at all. I don't know how much of a real issue this is, as `rapidjson` appears to be the fastest C++ library in existence for serialization of structs to a json string.
* I'm using `magic_enum` in a few places to automatically convert enums to strings in a few places. I'm not attached to this library at all, but just wanted something that would let me automatically convert the `CesiumGltf` enums to strings without adding additional code to generate these functions in the JS code.
* I added the `JsonObjectWriter.cpp` class to deal with the potential `extras` object that can be placed inside almost any glTF object. It's a simple recursive implementation but I'm willing to change it to iterative if we're worried about pathological inputs causing stack overflow errors. I've never seen an extras objects with more than 4-5 layers of nesting so I decided not to optimize this part yet.
* We are planning on writing *a lot* of glb assets as quickly as possible, so I was considering making the writing API move only to minimize allocations. E.g when you want to take your `CesiumGltf::Model` and serialize it to a glTF or glb byte array the `CesiumGltf::Model` becomes unusable to reduce redundant allocations. I plan on adding move support no matter what, but do people think it's excessive to mandate moving in our API to force people to use the more performant path?

### Upcoming Builder
The last thing we want to do is include our builder pattern from the terrain tiler in this library (to make the construction of a `CesiumGltf::Model` object less verbose, e.g:

```cpp
  auto watermaskGltfImage = GltfImageBuilder()
      .name("watermask")
      .imageType(GltfImageType::PNG)
      .buffer(std::move(asPNGByteArray), GL_UNSIGNED_BYTE)
      .filter(GL_NEAREST, GL_NEAREST)
      .dimensions(waterMask.MaskWidth, waterMask.MaskHeight)
      .wrapST(GL_CLAMP_TO_EDGE, GL_CLAMP_TO_EDGE)
      .toImage();
  
  CesiumGltf::Model model = GltfBuilder()
      .addVertexBuffer(VertexAttribute("POSITION", positions, GL_FLOAT, GltfType::VEC3))
      .addVertexBuffer(VertexAttribute("NORMAL", normals, GL_FLOAT, GltfType::VEC3))
      .addIndices(Indices(indices, GL_UNSIGNED_SHORT))
      .addImage(std::move(watermaskGltfImage))
      .buildInterleaved()
   
   std::vector<std::uint8_t> gltfArray = CesiumGltf::writeModelToByteArray(std::move(model), WriteFlags::glTF | WriteFlags::PrettyPrint);
   
   // write gltfArray to disk
```
We started this in the terrain tiler, I'm currently porting it to CesiumNative and swapping all the uses of raw OpenGL enums to the corresponding CesiumGltf enums. (Haven't pushed any of these commits yet).

### Extensions
@IanLilleyT and I had a discussion about how to handle extensions this morning. Ideally we would want to be able to do something like
```cpp
 GltfBuilder()
    .addVertexAttribute(...)
    .addVertexAttribute(...)
    .addExtension([](CesiumGltf::Model& model) {
      model.extensionsUsed.emplace_back("blah");
        // add whatever we need for this extension
     })
     .buildConcatenated();
```
The problem with this arises when we want to serialize the extension to a json string. If extensions are implemented like `KHR_draco_mesh_compressionJsonHandler` where the attributes are hardcoded as attributes on structs, it makes it impossible for the GltfWriter to serialize arbitrary extensions / user provided extensions. We think that making it more like the `JsonValue::Object` type would solve this issue. In other words we don't think cesium-native's writer should be responsible for being aware of all extensions in existence, and that tilers should be able to plug in their own tiler specific extensions as necessary.


CesiumGltfReader/src/Reader.cpp
```diff
@@ -352,10 +353,13 @@ CesiumGltf::readImage(const gsl::span<const std::byte>& data) {
       &channelsInFile,
       image.channels);
   if (pImage) {
-    image.pixelData.assign(
+    const int lastByte =
```
jtorresfabra, line 356: Can you add an explanation or small comment for this code change? I see current code clearer and more concise than the proposed change. It's a personal opinion though.

CesiumGltfWriter/include/CesiumGltf/WriterException.h
```diff
@@ -0,0 +1,28 @@
...
+class AmbiguiousDataSource : public std::runtime_error {
+public:
+  AmbiguiousDataSource(const char* msg) : std::runtime_error(msg) {}
+};
+
+class MissingDataSource : public std::runtime_error {
+public:
+  MissingDataSource(const char* msg) : std::runtime_error(msg) {}
+};
+} // namespace CesiumGltf
```
jtorresfabra, line 28: Missing white line at the end of the file.

CesiumGltfWriter/src/AccessorSparseWriter.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+#include "CesiumGltf/AccessorSparse.h"
+#include "JsonWriter.h"
+
+namespace CesiumGltf {
+void writeAccessorSparse(
+    const AccessorSparse& accessorSparse,
+    CesiumGltf::JsonWriter& jsonWriter);
+}
```
jtorresfabra, line 9: missing white line at the end of the file

CesiumGltfWriter/src/AccessorWriter.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+#include "CesiumGltf/Accessor.h"
+#include "CesiumGltf/AccessorSpec.h"
+#include "JsonWriter.h"
+
+namespace CesiumGltf {
+void writeAccessor(
+    const std::vector<Accessor>& accessors,
+    CesiumGltf::JsonWriter& jsonWriter);
+}
```
jtorresfabra, line 10: Missing white line at the end of the file. Seems there is a bunch of them. Can you review all the missing files to add the newline? Thanks!

CesiumGltfWriter/test/TestJsonObjectWriter.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+  SECTION("[0,1,2.5]") {
+    CesiumGltf::JsonWriter writer;
+    const auto extrasObject =
+        Array{std::int64_t(0), std::uint64_t(1), double(2.5)};
+    writeJsonValue(extrasObject, writer, false);
+    REQUIRE(writer.toStringView() == R"([0,1,2.5])");
+  }
+
+  SECTION("[👀]") {
```
jtorresfabra, line 34: Nice emoji trick! :joy:

jtorresfabra (approved):
Nice work @Samulus, I just added some minor code reviews/comments/questions.

CesiumGltfReader/src/Reader.cpp
Samulus: `std::transform` is necessary here because we decided to change all of our uses of `std::uint8_t` to `std::byte`. `std::uint8_t` is not implicitly convertible to `std::byte` (intentional choice by the cpp powers that be). Transform allows us to use a lambda to perform the explicit conversion here. I added a clarifying comment.

Cesium3DTiles/src/SkirtMeshMetadata.cpp
```diff
@@ -47,31 +52,30 @@ SkirtMeshMetadata::parseFromGltfExtras(const JsonValue::Object& extras) {
...
+  try {
+    pWestHeight = gltfSkirtMeshMetadata.getSafeNumericalValueForKey<double>(
+        "skirtWestHeight");
+    pSouthHeight = gltfSkirtMeshMetadata.getSafeNumericalValueForKey<double>(
+        "skirtSouthHeight");
+    pEastHeight = gltfSkirtMeshMetadata.getSafeNumericalValueForKey<double>(
+        "skirtEastHeight");
+    pNorthHeight = gltfSkirtMeshMetadata.getSafeNumericalValueForKey<double>(
+        "skirtNorthHeight");
+  } catch (const JsonValueMissingKey&) {
```
baothientran, line 69: Should we throw in this case if no key is found or value is not correct type? It seems to me returning `nullptr` is a better option, but I may miss something

Samulus: I think I wrote this code to mimic the way it worked before, but I think throwing if the SkritMeshMetadata exists but is malformed isn't a bad idea, that way clients will know immediately their input is malformed. Curious what the majority consensus on this is.

Samulus:
@jtorresfabra Thanks for review, feedback addressed (with exception of @baothientran 's comment on the exception handling in SkirtMeshMetadata, unsure what I should do for that).

Also merged with `origin/main`

IanLilleyT:
Before merging add CesiumGltfWriter to https://github.com/CesiumGS/cesium-native#card_file_boxlibraries-overview

Cesium3DTiles/src/SkirtMeshMetadata.cpp
baothientran: I maybe over-concern about the exception in this case. In fact, the exception here probably won't throw as all the fields are double when first created. So it won't cost much. If @jtorresfabra is okay with this, then it's a thumb up for me too

Samulus: I've decided to keep it as is, the current test expects a `std::nullopt` to be returned  if the data is malformed. Keeping it this way unless anyone has strong objections.

Samulus:
> Before merging add CesiumGltfWriter to https://github.com/CesiumGS/cesium-native#card_file_boxlibraries-overview

https://github.com/CesiumGS/cesium-native/pull/210 Created a PR for that.

Samulus:
Thank you @baothientran !

CesiumGltfWriter/include/CesiumGltf/Writer.h
```diff
@@ -0,0 +1,86 @@
...
+ * @brief Write a glTF or glb asset to a byte vector.
+ *
+ * @param model Final assembled glTF asset, ready for serialization.
+ * @param flags Bitset flags used to control writing glTF serialization
+ * behavior.
+ *
+ * @details Serializes the provided model object into a byte vector using the
+ * provided flags to convert. Must provide WriteFlags::GLTF or WriteFlags::GLB.
+ * Cannot provide both simultaneously. There are a few special scenarios with
+ * serializing \link CesiumGltf::Buffer \endlink and \link CesiumGltf::Image
```
kring, line 22: Nitpicky, but the rest of our code use the jsdoc style: `{@link CesiumGltf::Buffer}` rather than `\link CesiumGltf::Buffer`.

CesiumGltfWriter/src/AnimationWriter.cpp
```diff
@@ -0,0 +1,104 @@
...
+    return;
+  }
+
+  auto& j = jsonWriter;
+  j.Key("animations");
+  j.StartArray();
+  for (const auto& animation : animations) {
+    j.StartObject();
+    if (animation.channels.empty()) {
+      throw std::runtime_error("Must have at least 1 animation channel");
```
kring, line 77: We've generally worked on the assumption that cesium-native will be used in some environments where exceptions are disabled. And so exceptions shouldn't be the primary means of reporting errors. Especially in situations like this one where it might indicate user error more than a truly exceptional circumstance.

IMO, something like this should be reported as a warning, and say "The generated glTF will not be spec-compliant" rather than refusing to write a glTF at all.

extern/base64_encode/base64_encode.hpp
```diff
@@ -0,0 +1,71 @@
+/*
+* Base64 encoding/decoding (RFC1341)
```
kring, line 2: We already have `modp_b64` and it can do encoding, too.

modules/CodeCoverage.cmake
```diff
@@ -0,0 +1,683 @@
+# https://github.com/bilke/cmake-modules/blob/master/CodeCoverage.cmake
```
kring, line 1: Can we include this via submodule rather than copying it into our project?

extern/magic_enum/magic_enum.hpp
```diff
@@ -0,0 +1,1131 @@
+//  __  __             _        ______                          _____
```
kring, line 1: Can we include this by submodule rather than copying it into our project?

kring (commented):
I wish I had gotten around to reviewing this properly before it was merged. My plan was to move to it right after the reader/extension stuff, but that took longer than I had hoped. So here are some (better late than never?) comments.

> Modified JsonValue to support storing integers internally. I noticed a narrowing conversion bug while implementing this. Previously any std::uint64_t / std::int64_t would be automatically static casted to double when the JsonValue(foo) constructor was invoked. This would cause silent truncation for integer values larger than 2^53

IMO this is a lot of added complexity to solve a non-problem. JSON cannot reliably represent integers larger than 2^53, because it comes from JavaScript where every number is a double.  For example, in NodeJS:

```
$ node
> JSON.parse('{"n":18014398509481984}')
{ n: 18014398509481984 }
> JSON.parse('{"n":18014398509481985}')
{ n: 18014398509481984 }
```

(note that `18014398509481984` is `2^54`)

Chrome agrees:

image

So if a glTF file had such a large integer, it couldn't be read by javascript clients, anyway.

Silently truncating is bad, but we can just warn and truncate rather than imposing unnecessary complexity on clients of `JsonValue`.

> I started initially by mimicking the JavaScript functions that generate Cpp code for parsing, but I realized that it would actually be faster to just use https://kakoune.org/ and https://neovim.io/ macros to generate the writing classes as I go.

This seems like a big problem to me. "It saved me time" isn't a good enough reason by itself. As I said in a previous review comment:

> But this is one reason that it's very important that anyone be able to regenerate the data classes, the reader classes, and (now) the writer classes: to make it easy to add support for extensions. I'm not necessarily opposed to the editor-based approach rather than javascript if it's easier to work with, but I would like to see everything that's needed to generate the writers checked into the source tree so that anyone can regenerate them at any time.

But it seems like this didn't happen? Should the writer classes by considered hand-written at this point? If so, that's unfortunate because the extension changes in #208 require a small change to every writer class.

Samulus:
@kring I added an issue addressing your concerns / discussing what we'll change going forward: https://github.com/CesiumGS/cesium-native/issues/213

CesiumGltf/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ target_include_directories(
     SYSTEM PUBLIC
         ${CESIUM_NATIVE_RAPIDJSON_INCLUDE_DIR}
         ${CMAKE_CURRENT_LIST_DIR}/include/
+        ${CESIUM_NATIVE_GSL_INCLUDE_DIR}
```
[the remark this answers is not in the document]
Samulus: Other targets using this as a dependency fail to link properly unless it's exposed here.

CesiumGltfReader/src/Reader.cpp
```diff
@@ -398,15 +399,20 @@ CesiumGltf::readImage(const gsl::span<const std::byte>& data) {
...
+        image.width * image.height * image.channels * image.bytesPerChannel;
+    // std::uint8_t is not implicitly convertible to std::byte, so we must use
+    // std::transform here to force the conversion.
+    std::transform(
         pImage,
-        pImage + image.width * image.height * image.channels *
-                     image.bytesPerChannel);
+        pImage + lastByte,
+        std::back_inserter(image.pixelData),
+        [](char c) { return std::byte(c); });
```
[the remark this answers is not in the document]
Samulus: Yes, reinterpret casting the underlying `std::byte*` to `std::uint8_t` is safe, updated to use std::copy and reinterpret_cast

modules/CodeCoverage.cmake
Samulus: https://github.com/bilke/cmake-modules

It's an amalgamation of unrelated cmake files, I didn't want to pollute our cmake namespace with a bunch of irrelevant cmake files so I did a direct copy.

extern/magic_enum/magic_enum.hpp
Samulus: Changed to submodule

extern/base64_encode/base64_encode.hpp
Samulus: Swapped base64_encode.hpp with modp_b64

CesiumGltfWriter/src/AnimationWriter.cpp
Samulus: Using `std::cerr` now. Clients might get annoyed having random stuff spewed on `stderror` though, let me know if you have a specific implementation in mind for error handling in an exceptionless land. (Although your only real options are logging stuff and returning error codes, so I guess I'm asking which is preferred).

CesiumGltfWriter/include/CesiumGltf/Writer.h
Samulus: Changed.

CesiumGltfWriter/include/CesiumGltf/WriteFlags.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+#include "WriterLibrary.h"
+namespace CesiumGltf {
+CESIUMGLTFWRITER_API enum WriteFlags {
+  GLB = 1,
+  GLTF = 2,
+  PrettyPrint = 4,
+  AutoConvertDataToBase64 = 8
+};
```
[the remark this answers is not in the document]
Samulus: Changed to use a struct that mimics `ReadModelOptions`

CesiumGltf/CMakeLists.txt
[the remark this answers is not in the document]
kring: Are you sure? GSL/include/CMakeLists.txt has:

```
    target_include_directories(GSL SYSTEM INTERFACE
        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
        $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
    )
```

So as long as GSL is a public `target_link_libraries` dependency, the include directories should be pulled in automatically. I'm ~98% sure this was working before, at least.

CesiumGltfWriter/src/AnimationWriter.cpp
kring: CesiumGltfReader returns a list of errors and warnings as part of its result. I imagine the same should work well here.

modules/CodeCoverage.cmake
kring: Why would the unused ones polute our namespace? Can't we `include()` only the one we need?

Samulus: nvm I forgot about `include`, yeah it's fine to use a submodule, i put it in `extern`

CesiumGltfWriter/src/AnimationWriter.cpp
Samulus: updated API to mimic CesiumGltfReader's `ModelReaderResult`

CesiumGltf/CMakeLists.txt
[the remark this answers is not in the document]
Samulus: I put together a minimal test case to verify that this links without this and it works now. No idea why it wasn't working before but I removed this line from `CesiumGltf/CMakeLists.txt`

PR #163, opened 2021-03-11
Support deserializing arbitrary extensions

Samulus:
Repasting my git commit message here:
```
Propagate `ReadModelOptions` to ALL JsonHandlers

This commit adds conditional support for propagating the
`ReadModelOptions` variable to the root JsonValue class for all of
its children. CesiumGltfReader was designed with the philosophy that
extensions should be well defined and that CesiumGltfReader should
take responsbility for decoding extensions. This is a fine idea, but we've
ran into a usecase where we need CesiumGltfReader to support decoding
new / developing / unknown / WIP extensions.

This commit adds the `deserializeExtensionsAsJsonValue` bool to ReadModelOptions
and for each ***JsonHandler, forwards it to the root via super() constructor invocations.
This opens the door for other  "while serializing" options to be stored in
ReadModelOptions
```

Basically this PR adds support for `CesiumGltfReader` to read arbitrary custom gltf extensions using the `JsonValue` interface. We have some use cases where we have custom extensions, and we need inversion of control so that we can read the JsonValues and work with them appropriately. Ideally we would have a common `GltfExtension` interface that clients are required to implement, then we could just use some kind of callback system so we don't completely toss type information but CesiumGltfReader doesn't need to be omnipotent about what glTF extensions exist. This approach is a stop gap for any client that wants support for arbitrary or non well defined extensions.

The inheritance structure of the `JsonHandlers` makes this PR large, the condition needs to be checked inside `ExtensibleObjectJsonHandler`, but inheritance chains require us to propagate the `ReadModelOptions` object basically everywhere. https://github.com/CesiumGS/cesium-native/pull/162 should be merged first.


kring:
@Samulus this looks good overall, but there are a few things that are less than ideal:

* It feels wrong to pass `ReadModelOptions` to every JsonHandler when most of it isn't and never will be used by any of them. It would be a little cleaner to have a purpose-built structure for this purpose, even if it means copying some of the options from one struct to another in `readModel`.
* Every handler holds `ReadModelOptions` by value, when a const reference would do just fine because the options instance is guaranteed to outlive the handlers.
* As currently written, we won't read any JSON extensions for any glTF objects that have a statically-typed extension (just `MeshPrimitive` so far).
* It's not possible to add statically-typed glTF extensions external to cesium-native.

I don't think it makes sense to ask you to address these things, given that I already have an idea of how I think this should work (especially the last one) and just need to execute it. So I've started the `gltf-extensions` branch, building on this one, and I'll PR it shortly so you can take a look.

Samulus:
Closing as https://github.com/CesiumGS/cesium-native/pull/208 deprecates this now.

PR #193, opened 2021-04-02
Don't always report zero-length images as errors.

kring:
`BingMapsRasterOverlay` purposely requests that the Bing servers return a zero-length image for non-existent tiles. That's what the `n=z` query parameter does. So we shouldn't report these zero-length images as a problem.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ### v0.1.1 - ?
 
 - Gave glTFs created from quantized-mesh terrain tiles a more sensible material with a `metallicFactor` of 0.0 and a `roughnessFactor` of 1.0. Previously the default glTF material was used, which has a `metallicFactor` of 1.0, leading to an undesirable appearance.
+- Reported zero-length images as non-errors as `BingMapsRasterOverlay` purposely requests that the Bing servers return a zero-length image for non-existent tiles.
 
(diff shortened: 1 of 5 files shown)


kring:
> remove incorrect move

If I'm not mistaken, those moves are valid because the lambda is marked mutable. So the lambda gets a copy, and then the copy is moved into the result.

baothientran:
I just googled it and it caught me surprised. If `const ref` is captured by lambda like this `[option]() mutable {}` , the option type is `const` too even the lambda is mutable. It was mentioned here https://stackoverflow.com/questions/31485041/lambda-capture-by-value-mutable-doesnt-work-with-const.

I tried it out and it is indeed const in the lambda too:
https://godbolt.org/z/dTzsKffGT

kring:
C++ is silly. 😆 
Ok so the problem is that the captured variable is const to begin with, so it gets captured as const even though it's also copied. We don't need to remove the `std::move` though, we just need to capture it differently. Like this: https://godbolt.org/z/P9ojG9ET6

Right?

baothientran:
Yup that's right lol

baothientran:
@kring can you please review the move again? Other than that, everything else looks good to me

kring:
Looks good to me, thanks @baothientran!

PR #197, opened 2021-04-06
Apply scale component of tile.transform to tile.geometricError

baothientran:
Report in the forum https://community.cesium.com/t/lod-problems-with-photogrammetry-model/12861. 

The spec currently says tile.transform should not apply to geometric error, but it seems to be a bug on the specs. It was mentioned [here](https://github.com/CesiumGS/3d-tiles/issues/367). Also cesiumJS already applies the transform to the geometric error as well


baothientran:
wait wrong branch

PR #200, opened 2021-04-07
Add gltfupaxis support

javagl:
Details discussed in https://github.com/CesiumGS/cesium-unreal/issues/315

### DIFF
--- a/Cesium3DTiles/include/Cesium3DTiles/Tileset.h
+++ b/Cesium3DTiles/include/Cesium3DTiles/Tileset.h
@@ -9,6 +9,7 @@
 #include "Cesium3DTiles/ViewUpdateResult.h"
 #include "CesiumAsync/AsyncSystem.h"
 #include "CesiumAsync/IAssetRequest.h"
+#include "CesiumGeometry/Axis.h"
 #include "CesiumGeometry/QuadtreeTileAvailability.h"
 #include <atomic>
 #include <memory>
@@ -426,7 +427,9 @@ public:
    *
    * @return The value representing the axis
    */
-  uint8_t getGltfUpAxis() const noexcept { return this->_gltfUpAxis; }
+  CesiumGeometry::Axis getGltfUpAxis() const noexcept {
+    return this->_gltfUpAxis;
+  }
 
 private:
   /**
@@ -727,15 +730,14 @@ private:
   /**
    * @brief The axis that was declared as the "up-axis" for glTF content.
    *
-   * The value indicates the axis, via 0=X, 1=Y, 2=Z.
-   *
    * The glTF specification mandates that the Y-axis is the "up"-axis, so the
-   * default value is 1. Older tilesets may contain a string property in the
-   * "assets" dictionary, named "gltfUpAxis", indicating a different up-axis.
-   * Although the "gltfUpAxis" property is no longer part of the 3D tiles
-   * specification, it is still considered for backward compatibility.
+   * default value is {@link Axis::Y}. Older tilesets may contain a string
+   * property in the "assets" dictionary, named "gltfUpAxis", indicating a
+   * different up-axis. Although the "gltfUpAxis" property is no longer part of
+   * the 3D tiles specification, it is still considered for backward
+   * compatibility.
    */
-  uint8_t _gltfUpAxis;
+  CesiumGeometry::Axis _gltfUpAxis;
 
   static void addTileToLoadQueue(
       std::vector<LoadRecord>& loadQueue,
--- a/Cesium3DTiles/include/Cesium3DTiles/Tileset.h
+++ b/Cesium3DTiles/include/Cesium3DTiles/Tileset.h
@@ -417,6 +417,17 @@ public:
     return this->_supportsRasterOverlays;
   }
 
+  /**
+   * @brief Returns the value indicating the glTF up-axis.
+   *
+   * This function is not supposed to be called by clients.
+   *
+   * The value indicates the axis, via 0=X, 1=Y, 2=Z.
+   *
+   * @return The value representing the axis
+   */
+  uint8_t getGltfUpAxis() const noexcept { return this->_gltfUpAxis; }
+
 private:
   /**
    * @brief The result of traversing one branch of the tile hierarchy.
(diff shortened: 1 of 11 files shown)


Cesium3DTiles/include/Cesium3DTiles/Tileset.h
```diff
@@ -417,6 +417,17 @@ class CESIUM3DTILES_API Tileset final {
...
+  /**
+   * @brief Returns the value indicating the glTF up-axis.
+   *
+   * This function is not supposed to be called by clients.
+   *
+   * The value indicates the axis, via 0=X, 1=Y, 2=Z.
+   *
+   * @return The value representing the axis
+   */
+  uint8_t getGltfUpAxis() const noexcept { return this->_gltfUpAxis; }
```
baothientran, line 429: maybe returning an enum is a better option since it makes the API more clear

Cesium3DTiles/src/Tile.cpp
```diff
@@ -238,6 +238,7 @@ void Tile::loadContent() {
       .thenInWorkerThread(
           [loadInput = std::move(loadInput),
            projections = std::move(projections),
+           ownerTileset = &tileset,
```
baothientran, line 241: it maybe dangerous to pass raw pointer to the lambda that will be passed to another thread since tileset can be destroyed before the lambda get called in the main thread (But that is a bug if it really does currently). I think it's better to capture the value of `getGltfUpAxis()` in the lambda so we don't have to manage object lifetime. Something like
```
uint8_t axisUp = tileset.getGltfUpAxis();
std::move(maybeRequestFuture.value()).thenInWorkerThread([axisUp]() { /*do something with it*/});
```

baothientran (commented):
@javagl This looks good to me. I only have some minor suggestions

Cesium3DTiles/src/Tile.cpp
javagl: Indeed that's much better. (I usually try to "pass on as little information as possible", but am not sure why I didn't do this here)

Cesium3DTiles/include/Cesium3DTiles/Tileset.h
javagl: Yes, I considered adding a full-fledged enum/class like https://github.com/CesiumGS/cesium/blob/master/Source/Scene/Axis.js , but wasn't sure where to put it (because it would be "public"). But ( also @kring ) it would probably make sense to add an `Axis` enum in `CesiumGeometry`, right?

kring: That sounds fine to me.

CesiumGeometry/include/CesiumGeometry/Axes.h
```diff
@@ -0,0 +1,50 @@
...
+
+#include "CesiumGeometry/Library.h"
+#include <glm/mat3x3.hpp>
+
+namespace CesiumGeometry {
+
+/**
+ * @brief Coordinate system conversion matrices
+ */
+struct CESIUMGEOMETRY_API Axes final {
```
baothientran, line 11: it just a small nick pick from me. I think maybe `AxisTransform` is a better name? cause Axes is kind of similar to `Axis.h` header

CesiumGeometry/include/CesiumGeometry/Axes.h
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * @brief A matrix to convert from y-up to z-up orientation,
+   * by rotating about PI/2 around the x-axis
+   */
+  static const glm::dmat4 Y_UP_TO_Z_UP;
```
baothientran, line 17: those matrices are good candidates for unit tests too. We can test it out by multiply a Y-vector with the matrix to see if it converts to Z-axis or something similar like that

CesiumGeometry/src/Axes.cpp
```diff
@@ -0,0 +1,65 @@
+#include "CesiumGeometry/Axes.h"
+#include <glm/mat3x3.hpp>
+
+namespace CesiumGeometry {
+
+namespace {
```
baothientran, line 6: I think unnamed namespace should be enough to make the function visible only in a local file. Static keyword maybe redundant

baothientran (commented):
@javagl this looks really good now. I only have some minor comments. After this, I think it can be ready to merge

CesiumGeometry/include/CesiumGeometry/Axes.h
javagl: Yeah, I often hesitate with stuff like that. 

On the one hand, it was supposed to "mimic" the https://github.com/CesiumGS/cesium/blob/master/Source/Scene/Axis.js . On the other hand, there are language-specific best practices. (E.g. in the Java world, a class that only contains static members/methods is indicated by the Plural, e.g. when there is a class `Person`, then `Persons` contains the static utility methods). 

I went for `AxisTransforms` now, roughly aligned with https://github.com/CesiumGS/cesium/blob/master/Source/Core/Transforms.js (but specialized to transforming the up-axis). 

An aside: We now have the enum `Axis.X`, but I could imagine that having some

    const glm::dvec4 X_AXIS{ 1.0, 0.0, 0.0, 0.0 };

somehere could also make sense sooner or later. (Or should it be a `dvec3` then? ... Well ....)

CesiumGeometry/include/CesiumGeometry/Axes.h
javagl: One problem is that "converting from one up-vector to another" can be done in different ways. Specifically (and maybe counterintuitively), 

    Y_AXIS * Y_UP_TO_Z_UP == -Z_AXIS 

Note that it's **MINUS** the `Z_AXIS`. This wouldn't necessarily be the case for other ways of "making the Z-axis point 'up'". 

That's why I *explicitly* added the *"...by rotating about PI/2 around the x-axis"* part in the comment - without that, the matrix could have many different forms, all achieving the result of Z pointing 'up'.

However, of course you're right regarding the unit test, and I just added them. Considering that the matrix columns are just the images of the unit vectors, these tests could be considered as a very elaborate form of saying 

    assert(X_TO_Z.elements == { 1, 0, 0, 0, 0, 0, 1, 0, 0, -1, 0, 0, 0, 0, 0, 1 })

but having that encoded as tests isn't wrong either.

CesiumGeometry/src/Axes.cpp
javagl: That's right, according to my (albeit a bit vague) understanding of https://stackoverflow.com/q/13396748 (that "internal vs. external linkage" should not be relevant here...)

baothientran:
thanks @javagl, I only reword a little bit on CHANGE.md. This can be merged after the CI pass

PR #202, opened 2021-04-12
First pass of clang tidy

jtorresfabra:
Mainly caching else after return, not using empty() instead of size() == 0, missing const references to avoid copies and removing redundant initializations. 

There are still missing several fixes that I left for following passes if we accept this first one, including:
* Incosistent case style for global constants
* Inconsistent case style for parameters/variables
* Differing parameters names between headers and sources
* Code that may throw exceptions in no-except declared functions
* Following rule of five if rule of zero can't be followed (Up to discussion on how to act on this)

Some of the changes explanation for reference:
https://clang.llvm.org/extra/clang-tidy/checks/performance-move-const-arg.html
https://clang.llvm.org/extra/clang-tidy/checks/performance-faster-string-find.html
https://clang.llvm.org/extra/clang-tidy/checks/performance-unnecessary-value-param.html
https://clang.llvm.org/extra/clang-tidy/checks/bugprone-branch-clone.html
https://clang.llvm.org/extra/clang-tidy/checks/readability-redundant-member-init.html
https://clang.llvm.org/extra/clang-tidy/checks/readability-simplify-boolean-expr.html

### DIFF
--- a/Cesium3DTiles/include/Cesium3DTiles/ExternalTilesetContent.h
+++ b/Cesium3DTiles/include/Cesium3DTiles/ExternalTilesetContent.h
@@ -43,7 +43,7 @@ private:
    * @return The {@link TileContentLoadResult}
    */
   static std::unique_ptr<TileContentLoadResult> load(
-      std::shared_ptr<spdlog::logger> pLogger,
+      const std::shared_ptr<spdlog::logger>& pLogger,
       const glm::dmat4& tileTransform,
       TileRefine tileRefine,
       const std::string& url,
--- a/Cesium3DTiles/include/Cesium3DTiles/GltfContent.h
+++ b/Cesium3DTiles/include/Cesium3DTiles/GltfContent.h
@@ -41,7 +41,7 @@ public:
    * @return The {@link TileContentLoadResult}
    */
   static std::unique_ptr<TileContentLoadResult> load(
-      std::shared_ptr<spdlog::logger> pLogger,
+      const std::shared_ptr<spdlog::logger>& pLogger,
       const std::string& url,
       const gsl::span<const std::byte>& data);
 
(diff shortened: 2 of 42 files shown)


shehzan10:
@jtorresfabra How much of this can be made part of clang-format?

baothientran:
@shehzan10 this is C++ linter clang-tidy which I think we should integrate it with cesium-native soon

jtorresfabra:
@shehzan10 as Bao said, it is a different tool, nothing of this can be made part of clang-format AFAIK.

shehzan10:
> @shehzan10 this is C++ linter clang-tidy which I think we should integrate it with cesium-native soon

Yeah I think this was my intention (I forgot that clang has a variety of tools). If we can make this process part of the CI, that would be fantastic.

kring:
Thanks @jtorresfabra! One thing I want to ask you about: member initialization. I noticed you removed a bunch of initializers, I guess because they're redundant. But then some were left in place that I thought also weren't needed, like `Tileset::_pRootTile` (which is a `std::unique_ptr`). My rule of thumb was always just to initialize every member in the order its declared, in order to avoid and surprises. Your rule is probably better, but can you articulate it? And maybe we should put it in the coding guidelines as well.

jtorresfabra:
Hi @kring

> But then some were left in place that I thought also weren't needed, like `Tileset::_pRootTile` (which is a `std::unique_ptr`)

This is only a first pass and clang-tidy is still complaining about a lot more of stuff. So probably it got buried in the big warning logs from clang-tidy I have. To be honest, my rule of thumb was the same as yours before clang-tidy checked for it, these days I just follow https://clang.llvm.org/extra/clang-tidy/checks/readability-redundant-member-init.html to ease readability. 

Say that, this check can be disabled if we don't find comfortable with it, so everything in this PR is up to discussion and we can tweak the clang-tidy rules if we are not happy with them.

PR #207, opened 2021-04-13
Watermask

nithinp7:
This is the cesium-native part of the watermask implementation. The cesium-unreal side to this is here: https://github.com/CesiumGS/cesium-unreal/pull/338.


Cesium3DTiles/src/QuantizedMeshContent.cpp
```diff
@@ -651,6 +677,92 @@ static std::vector<std::byte> generateNormals(
...
 
+// Create texture coordinates for tiles such that the texture fits exactly
+static void createFittedTextureCoordinatesForTile(
+    CesiumGltf::Model& model,
+    uint32_t positionsAccessorId,
+    uint32_t textureCoordinatesId,
+    const CesiumGeospatial::GlobeRectangle& rectangle) {
+  // TODO: How often exactly do these types of checks need to be made?
+  // If the accessor id is valid, then are the buffer view and buffer ids
+  // implicitly valid?
```
kring, line 688: No, a valid accessor doesn't guarantee the objects it references are valid. Easiest thing is to create an `AccessorView` and then check the `status`.

In fact, the code below won't work if the glTF has positions interleaved with other attributes. `AccessorView` takes care of all that.

kring: There's also an `AccessorWriter`.

Cesium3DTiles/src/Tileset.cpp
```diff
@@ -799,7 +799,7 @@ static BoundingVolume createDefaultLooseEarthBoundingVolume(
   context.requestHeaders.push_back(std::make_pair(
       "Accept",
       "application/vnd.quantized-mesh,application/octet-stream;q=0.9,*/"
-      "*;q=0.01"));
+      "*;q=0.01;extensions=watermask"));
```
kring, line 802: Should we make the watermask optional? It's a good amount of extra data to to download and won't be used at all in many applications.

nithinp7:
@baothientran I closed the other PR and just built the discussed changes on this branch. I think watermask is done now (both cesium-native and cesium-unreal parts). Can you have a look if you get the chance?

kring:
I made some small tweaks:
* Tweaked CHANGES.md.
* Made `projections.push_back(GeographicProjection());` happen only if the water mask is enabled, cause it potentially doubles the texture coordinate data and we shouldn't do that unnecessarily.
* Made it only request the watermask extension if the watermask is enabled. Cesium World Terrain ignores this anyway, but other terrain tilesets might honor it.

Looks great otherwise, merging!

PR #208, opened 2021-04-15
Improvements to glTF extension loading

kring:
Builds on #163 

* Added `CesiumJsonReader` library, which allows deserializing JSON into arbitrary statically-typed classes using the RapidJSON SAX API, without actually exposing RapidJSON at all. `CesiumGltf` is built on this, of course, but it can also be used for deserializing tileset.json, etc. in the future.
* Moved `JsonValue` into `CesiumUtility`, for lack of a better place. It's needed by both `CesiumGltf` and `CesiumJsonReader`, neither of which depend on each other.
* Renamed `CesiumGltf::Reader` to `CesiumGltf::GltfReader`, even though it's a little redundant.
* `GltfReader` is now meant to be instantiated. Previously it only had static methods. `readModel` and `readImage` are const and thread-safe (as long as you're not simultaneously adding extensions), however, so it's fine to have a single instance in most apps.
* Added `GltfReader::registerExtension`. It can be used to register a `JsonHandler` for an extension, so that new statically-typed extensions can be added without modifying cesium-native itself.
* Added `GltfReader::setExtensionState`. There are three possible states for an extension: enabled, disabled, or "json only". Disabled extensions are completely ignored. Enabled extensions are loaded into static classes if one is registered (`GltfReader::registerExtension` above) or into a `JsonValue` if one is not registered. "json only" extensions are loaded into a `JsonValue` even if there's a static class registered, which is useful if static support is added in a new version of cesium-native and you don't want to update your code yet.
* Generated `JsonHandlers` now expose a `ValueType` typedef indicating the type they deserialize to.
* Both statically-typed and non-statically-typed extensions can now co-exist on a single glTF object.

To do before merge:
* [x] Don't create a new `GltfReader` instance every time we want to load a glTF or image.
* [x] Create a cesium-unreal PR to update it based on these changes.

CC @Samulus


kring:
This PR is huge, but note when reviewing that most of the changes are to generated code.

.vscode/settings.json
```diff
@@ -99,7 +99,11 @@
         "codecvt": "cpp",
         "ranges": "cpp",
         "scoped_allocator": "cpp",
-        "typeindex": "cpp"
+        "typeindex": "cpp",
```
Samulus, line 102: Are these intentional or is this a vestigial change? I noticed that vscode likes to cram stuff here all the time no reason seemingly.

kring: That's VSCode just adding things itself, and me assuming it knows what it's doing while modifying its own settings file. Do you think it's a problem or otherwise better to avoid committing these changes?

kring:
I changed the base branch to `main` now that you closed the other PR, and now there are merge conflicts. I'll sort it out.

kring:
I'm going to merge this so we can get cesium-native and cesium-unreal working together again. Reviews after the fact are welcome!

.vscode/settings.json
Samulus: https://github.com/microsoft/vscode-cpptools/issues/722#issuecomment-301136398 I had no idea how this worked either.  I think we should just add this file to the gitignore, it just adds visual noise to PRs :man_shrugging:

baothientran (commented):
@kring there are still more files I need to review. But so far, everything looks good to me. I only have some comments below. I will review more in the mean time

.vscode/settings.json
kring: ok, that works for me. I can imagine we might want to actually have shared workspace settings at some point, but we don't have any meaningful ones at the moment so might as well cut down on the noise.

Cesium3DTiles/include/Cesium3DTiles/RasterOverlayTileProvider.h
```diff
@@ -400,5 +401,7 @@ class CESIUM3DTILES_API RasterOverlayTileProvider {
   int64_t _tileDataBytes;
   int32_t _totalTilesCurrentlyLoading;
   int32_t _throttledTilesCurrentlyLoading;
+
+  static CesiumGltf::GltfReader _gltfReader;
```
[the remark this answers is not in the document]
kring: Good idea. I wrote #221.

CesiumGltf/src/ExtensibleObject.cpp
```diff
@@ -0,0 +1,22 @@
+#include "CesiumGltf/ExtensibleObject.h"
+
+using namespace CesiumGltf;
+using namespace CesiumUtility;
+
+JsonValue* ExtensibleObject::getGenericExtension(
+    const std::string& extensionName) noexcept {
+  return const_cast<JsonValue*>(
+      const_cast<const ExtensibleObject*>(this)->getGenericExtension(
```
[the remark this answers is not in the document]
kring: I think the claim on StackOverflow that `const_cast` can't be used to add constness is incorrect, based on my reading of https://en.cppreference.com/w/cpp/language/const_cast.

> lvalue of any type T may be converted to a lvalue or rvalue reference to the same type T, more or less cv-qualified.

But an answer further down mentions `std::as_const` which seems like the most readable and correct thing to use (since C++17).

CesiumUtility/include/CesiumUtility/JsonValue.h
```diff
@@ -268,8 +275,53 @@ class CESIUMGLTF_API JsonValue final {
...
+   * aforementioned reasons.
+   * @remarks Compilation will fail if type 'To' is not an integral / float /
+   * double type.
+   */
+  template <
+      typename To,
+      typename std::enable_if<
+          std::is_integral<To>::value ||
+          std::is_floating_point<To>::value>::type* = nullptr>
+  [[nodiscard]] To getSafeNumericalValueOrDefaultForKey(
```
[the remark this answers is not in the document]
kring: It's not necessarily a problem to have methods that throw. They're perhaps preferred when exceptions are enabled, and they're valid even when exceptions disabled if you happen to _know_ the method won't be throwing in your particular case. A throw statement will still compile with exceptions disabled, it will just cause immediate program termination if it's executed.

The bigger problem we had was the use of a try/catch block. Under some compilers, at least (VS2017), that code won't even compile when exceptions are disabled. And even if it did, it presumably would not work as expected because the non-throwing contract of the method wouldn't have been able to work without the ability to throw and catch exceptions.

I think we may want to do another round of refactoring JsonValue in the future, but for now it's ok to have these throwing methods given that we also have non-throwing alternatives.

PR #214, opened 2021-04-16
Connection error messages

javagl:
This is acctually for https://github.com/CesiumGS/cesium-unreal/issues/314  - further details added there


kring:
Thanks @javagl!

Issue #211, opened 2021-04-15
Tileset that has ADD refinement isn't selected for rendering


baothientran:
We got a case where tileset with below structures failed to render if one of the children are failed to loaded (due to valid format but not supported by cesium-native currently):
```
                                             Root tile (Empty tile)
                                /                        |                          \
                               /                         |                           \
                 B3dm (Renderable)              I3dm(Unsupported)             pnts(Unsupported)
```

All of them have ADD refinement. The expected behavior is that the B3dm should be selected for rendering even though the other children and parent are not renderable. However, currently, all the children are kicked out of the rendering queue and the root get selected instead

PR #217, opened 2021-04-19
Fix add refinement not render when one of children has error

baothientran:
Fixes https://github.com/CesiumGS/cesium-native/issues/211 

The tile will fail to refine if one of the children is not renderable because that makes `allAreRenderable` to be false and none of them were rendered last frame. As a result, All of the children are kicked out of the rendering queue. 

If the tile's format can't be loaded because of invalid tile format, we simply return a nullptr content and treat it as an empty tile


kring:
Thanks @baothientran! I updated CHANGES.md.

javagl:
Just for the case that I'm not the only one who wondered by a "broken build" state was merged: The update of the `CHANGES.md` happened *immediately* before the merge (and the branch was deleted with the merge), but the change in `CHANGES.md` triggered a re-build on the (now deleted) branch.

Issue #213, opened 2021-04-16
Address misc GltfWriter technical debt.


Samulus:
https://github.com/CesiumGS/cesium-native/pull/125 was merged before @kring had a chance to point out a few issues.

### Cesium Unreal can't package plugin anymore
```
C:\Users\travis\build\CesiumGS\packages\CesiumForUnreal\HostProject\Plugins\CesiumForUnreal\Source\ThirdParty\include\CesiumUtility/JsonValue.h(323): error C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc
```
This step doesn't happen on Travis, and I don't do Unreal builds, so I was unable to catch this before it was merged. This is definitely the most imperative thing that needs to be fixed (compared to all the other issues below).

### Overflow Handling
* The library currently uses `gsl_narrow` to try to detect when large integers are added to the JSON
* Kevin mentioned that values > `2^53` cannot be represented in JavaScript anyway, but https://stackoverflow.com/a/13502497 suggests that they're valid numbers to store in JSON. So if we only care about glTF parsing in web browsers we can gut this complexity and keep everything as doubles. If we want to adhere to the JSON limits and not the JavaScript ones the narrowing can stay.

### Code Generation
* The GltfWriter classes were hand written with the assistance of editor macros to quickly generate each class. This is in contrast to the `CesiumGltfReader` library which uses a collection of JavaScript files to parse gltf schema files to generate C++ code. This was an intentional choice and my previous statements that "it was easier" / "it saved me time" were oversimplifications on my part:
    * The GltfWriter was created as unscoped work during an unrelated sprint, so the goal was to produce something as quickly as possible that would serve as a robust base for extending in `cesium-native`. The existing code generation makes heavy use of ES6 JavaScript templates to dynamically generate code. I thought it was more appropriate to use a template engine like HandlebarJS / Mako / Jinja  to do this. I wanted to refactor the code generation classes by reading dedicated template files and using a proper templating library to inject the necessary changes, but knew there wasn't enough time in the sprint to do this properly so I skipped it. This can be done retroactively, but isn't trivial and needs to be scoped properly.
    
### Misc.

See all the comments here: https://github.com/CesiumGS/cesium-native/pull/125#pullrequestreview-637326983 these are all small issues can be fixed in a single amalgamated PR.

kring:
[...]

PR #219, opened 2021-04-19
Address technical debt in CesiumGltfWriter

Samulus:
https://github.com/CesiumGS/cesium-native/pull/125 was merged before @kring had a chance to properly review it. This PR addresses his comments. There are more comments here as well: https://github.com/CesiumGS/cesium-native/issues/213

### Requested Changes 
* Using a bitmask flag to configure the gltf writing process has been changed to use a simple struct (like `CesiumGltfReader`)
* Doxygen comments have been fixed
* `magic_enum` is now included as submodule
* `base64_encode.hpp` is gone and `CesiumGltfWriter` now links to and uses `modp_b64`
* `std::transform + std::back_inserter` has been replaced with a preallocated vector, reinterpret_cast, and `std::copy`. I had the same `std::transform + std::back_inserter` algorithm in JsonWriter so I updated that as well.
* Antiquated variable naming has been corrected for some pointers
* `CodeCoverage.cmake` is now included as a submodule in extern
* `CesiumGltfWriter` does not throw exceptions anymore, it returns two vectors with warnings / errors like `CesiumGltfReader`
* `encodeAsBase64String` no longer performs an additional copy.
*  Removed extraneous `CESIUM_NATIVE_GSL_INCLUDE_DIR` from `CesiumGltf`'s public include directories. 

### Other Changes
* Animation / Buffer / Image Writer now all write the index of the array element that triggered an error (if one is encountered), should make easier for clients to diagnose why their glTF assets are malformed.


CesiumGltfWriter/src/EncodeBase64String.cpp
```diff
@@ -1,6 +1,17 @@
...
+  const std::size_t len = modp_b64_encode_len(data.size());
+  std::vector<char> destination(len, 0);
+  const char* asCharPointer = reinterpret_cast<const char*>(data.data());
+  modp_b64_encode(destination.data(), asCharPointer, data.size());
+
+  // TODO: Returning std:string(destination.begin(), destination.end())
+  //       adds *ALL* the padding bytes instead of just stopping at the first
+  //       null byte. It seems more correct to do that, however, if I do that
+  //       buffer.cesium.data is initalized to nothing when CesiumGltfReader
+  //       reads the glTF asset back in. Bug?
```
kring, line 15: `modp_b64_encode_len` returns the _max_ length of the encoded string, not the actual length. You should use the return value of `modp_b64_encode` as the length of the buffer. Actually it returns the length with the null terminator, so you should subtract one.

You can also read directly into a string, rather than reading into a vector and then copying into a string. Then resize the string to the actual encoded length at the end.

kring (commented):
Looks good, just one comment below. Can you also please update CHANGES.md? Most of this is under-the-hood and doesn't need to be mentioned, but the bitmask thing, at least, is an API change.

CesiumGltfWriter/src/EncodeBase64String.cpp
Samulus: Changed to remove the extra copy, but I'm not sure that minus one is correct:

```cpp
std::string
CesiumGltf::encodeAsBase64String(const std::vector<std::byte>& data) noexcept {
  const std::size_t paddedLength = modp_b64_encode_len(data.size());
  std::string result(paddedLength, '\0');
  const char* asCharPointer = reinterpret_cast<const char*>(data.data());
  const std::size_t actualLength = modp_b64_encode(result.data(), asCharPointer, data.size());
  result.erase(actualLength, std::string::npos);
  return result;
}
```

If `data` == `HelloWorld!`, then `result` should be set to `SGVsbG9Xb3JsZCE=0`, which is `17` characters [0-16]. `paddedLength` is set to `17` and `actualLength` is set to `16`. The docs for the function say it returns the length of the destination plus one for the null byte. As of C++11 [strings are required to be null terminated internally](https://stackoverflow.com/a/7554172). So resizing `result` to be `actualLength - 1` would result in `SGVsbG9Xb3JsZCE` which removes the '=' and generates an invalid base64 string as it would strip the padding. We can erase from `actualLength` to the end of the string to remove the extra bytes instead. I added extra tests for this.

Samulus:
Updated changes.md, if it's in the wrong section let me know, not sure what the versioning system for this project is.

CesiumGltfWriter/src/EncodeBase64String.cpp
kring: It seems the modp_b64 docs are wrong. The length returned by `modp_b64_encode` does _not_ include the null terminator. Looking at the source, the return value is:

```
    *p = '\0';
    return p - (uint8_t*)dest;
```

`dest` is the originall passed in pointer. So if we wrote one character, then incremented `p` and wrote the null character, then `p-dest` would give us 1, which is the number of characters written _without_ the null character. So yeah, the docs are definitely wrong.

So your code is correct, and `result.resize(actualLength)` would work well, too.

kring:
Thanks @Samulus! I made a couple more changes:

* Added `add_directory(magic_enum)` and then added it as a dependency of CesiumGltfWriter, rather than adding the include directory directly.
* Remove now unused INCLUDE_DIR vars for magic_enum and GSL.
* Replaced the erase with a resize in the base64 encoding because it should be a teensy bit more efficient and maybe clearer too?

I'll merge this as soon as CI is happy.

PR #223, opened 2021-04-20
Let tile be renderable if all rasters have ready tiles

### LINKED ISSUE #188: Tile::isRenderable reports true only when overlays are totally done loading, rather than when they're renderable
It checks that no mapped raster overlays have a "loading" tile. But once we have any "ready" tile, we can render.

There is a fix for this in https://github.com/CesiumGS/cesium-native/compare/tile-selection, but it is not well tested.

baothientran:
Fixes #188

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,15 @@
 # Change Log
 
+### v0.3.0 - ????-??-??
+
+##### Breaking Changes :mega:
+
+##### Additions :tada:
+
+##### Fixes :wrench:
+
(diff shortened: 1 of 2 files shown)


kring:
Thanks @baothientran. The logic looks good, but the implementation is _slightly_ less efficient than it used to be. Checking all the raster overlays is the slowest part of the process, so ideally we only do it if the other AND'd conditions pass.

baothientran:
ah you are right. I didn't notice that

baothientran:
@kring it's updated. It should be ready for another review

Issue #179, opened 2021-03-21
Report RasterOverlayTileProvider::loadTileImageFromUrl load warnings and errors more cleanly


kring:
Currently, `RasterOverlayTileProvider::loadTileImageFromUrl` takes pains to make sure the tile URL is included in a warning/error every time one is reported, because this is useful information for users to understand what has gone wrong. But really, it should be `RasterOverlayTileProvider::doLoad` that is responsible for providing this context.

The problem, though, is that `RasterOverlayTileProvider::doLoad` doesn't directly know the URL. Not all RasterOverlays are required to even load from URLs (e.g. they could be generated, like CesiumJS's TileCoordinatesImageryProvider).

So, to start with, `doLoad` should include the tile ID in its log of errors and warnings, i.e. `L10-X23-Y144`, because that's always available and useful. It would be nice to also include the URL _when there is one_. The simplest way to do that that I can think of is for `RasterOverlayTileProvider::loadTileImageFromUrl` to include the URL in the returned `LoadedRasterOverlayImage` as a separate and optional field. Then, `doLoad` prints the URL (in addition to the tile ID) if there is a URL, and prints just the tile ID if there isn't a URL.

There's some discussion about this in https://github.com/CesiumGS/cesium-native/pull/155.

kring:
Fixed in #226.

Issue #203, opened 2021-04-12
Add clang-tidy and cppcheck to CI process


baothientran:
clang-tidy and cppcheck are fantastic to prevent us from shooting ourselves in the foot because of C++

kring:
include-what-you-use, too!

kring:
There's also [C++ Core Check](https://docs.microsoft.com/en-us/cpp/code-quality/using-the-cpp-core-guidelines-checkers?view=msvc-160&viewFallbackFrom=vs-2019), which statically verifies (a subset of) the C++ Core Guidelines the form the basis of our coding standards. Probably not as mature or well maintained as clang-tidy, but maybe of value anyway.

PR #226, opened 2021-04-26
Log tile ids

javagl:
Addressing and maybe fixing https://github.com/CesiumGS/cesium-native/issues/179 : 

There now is a function to create a `std::string` from a `TileID`. The warning/error messages from `RasterOverlayTileProvider::doLoad` now contain these tile IDs. (**If** the underlying implementation uses `loadTileImageFromUrl`, then all warnings/errors will also contain the URL, but that was already done in another PR). 

The current structure is (depending on the implementation, in pseudocode) 

    void doLoad(Tile tile) {
        Result result = loadTileImage(tile.getId());
        if (result.error) {
            // Log something here...
        }
    }
    virtual Result  loadTileImage(TileID tileID) { // Depends on implementation
        String url = createUrlFrom(tileId);
        return loadTileImageFromUrl(url);
    }
    Result loadTileImageFromUrl(String url) {
        Response response = request(url);
        return readImageData(response.data);
    }
    Result readImageData(vector<byte> data) {
        return stb_load(data);
    }

So the question of "What should be logged where?" could be answered:

- Networking errors (404s) are reported in `loadTileImageFromUrl`, which does the request/response handling, including the URL
- Loading errors (image could not be read) are reported in `doLoad`, including the the tile ID
- In order to "transport" low-level errors from stb, the `loadedImage.errors` can be used. Since this is implemented locally in a lambda, only for the `loadTileImageFromUrl` case, the URL can be added there directly.

In any case: If there is a problem with a tile, it now prints

```
LogCesium: Warning: [2021-04-26 17:52:21.040] [warning] [RasterOverlayTileProvider.cpp:552] Warnings while loading image for tile L8-X463-Y199:
- ...
```

### DIFF
--- a/Cesium3DTiles/src/RasterOverlayTileProvider.cpp
+++ b/Cesium3DTiles/src/RasterOverlayTileProvider.cpp
@@ -2,6 +2,7 @@
 #include "Cesium3DTiles/IPrepareRendererResources.h"
 #include "Cesium3DTiles/RasterOverlay.h"
 #include "Cesium3DTiles/RasterOverlayTile.h"
+#include "Cesium3DTiles/TileID.h"
 #include "Cesium3DTiles/TilesetExternals.h"
 #include "Cesium3DTiles/spdlog-cesium.h"
 #include "CesiumAsync/IAssetResponse.h"
@@ -524,6 +525,58 @@ RasterOverlayTileProvider::loadTileImageFromUrl(
           });
 }
 
+RasterOverlayTileProvider::LoadResult
+RasterOverlayTileProvider::processLoadedImage(
+    const CesiumGeometry::QuadtreeTileID& tileId,
+    LoadedRasterOverlayImage&& loadedImage) {
+
+  const auto& pPrepareRendererResources = this->getPrepareRendererResources();
+  const auto& pLogger = this->getLogger();
+
+  if (!loadedImage.image.has_value()) {
+    std::string tileIdString =
+        Cesium3DTiles::TileIdUtilities::createTileIdString(tileId);
+    SPDLOG_LOGGER_ERROR(
+        pLogger,
+        "Failed to load image for tile " + tileIdString + ":\n- {}",
+        CesiumUtility::joinToString(loadedImage.errors, "\n- "));
+    LoadResult result;
+    result.state = RasterOverlayTile::LoadState::Failed;
+    return result;
+  }
+
+  if (!loadedImage.warnings.empty()) {
+    std::string tileIdString =
+        Cesium3DTiles::TileIdUtilities::createTileIdString(tileId);
+    SPDLOG_LOGGER_WARN(
+        pLogger,
+        "Warnings while loading image for tile " + tileIdString + ":\n- {}",
+        CesiumUtility::joinToString(loadedImage.warnings, "\n- "));
+  }
+
+  CesiumGltf::ImageCesium& image = loadedImage.image.value();
+
+  int32_t bytesPerPixel = image.channels * image.bytesPerChannel;
+  size_t requiredBytes =
+      static_cast<size_t>(image.width) * image.height * bytesPerPixel;
+  if (image.width > 0 && image.height > 0 &&
+      image.pixelData.size() >= requiredBytes) {
+    void* pRendererResources =
+        pPrepareRendererResources->prepareRasterInLoadThread(image);
+
+    LoadResult result;
+    result.state = RasterOverlayTile::LoadState::Loaded;
+    result.image = std::move(image);
+    result.credits = std::move(loadedImage.credits);
+    result.pRendererResources = pRendererResources;
+    return result;
+  }
+  LoadResult result;
+  result.pRendererResources = nullptr;
+  result.state = RasterOverlayTile::LoadState::Failed;
+  return result;
+}
+
 void RasterOverlayTileProvider::doLoad(
     RasterOverlayTile& tile,
     bool isThrottledLoad) {
@@ -537,61 +590,11 @@ void RasterOverlayTileProvider::doLoad(
 
   this->beginTileLoad(tile, isThrottledLoad);
 
-  struct LoadResult {
-    RasterOverlayTile::LoadState state = RasterOverlayTile::LoadState::Unloaded;
-    CesiumGltf::ImageCesium image = {};
-    std::vector<Credit> credits = {};
-    void* pRendererResources = nullptr;
-  };
-
   this->loadTileImage(tile.getID())
-      .thenInWorkerThread(
-          [tileRectangle =
-               this->getTilingScheme().tileToRectangle(tile.getID()),
-           projection = this->getProjection(),
-           pPrepareRendererResources = this->getPrepareRendererResources(),
-           pLogger =
-               this->getLogger()](LoadedRasterOverlayImage&& loadedImage) {
-            if (!loadedImage.image.has_value()) {
-              SPDLOG_LOGGER_ERROR(
-                  pLogger,
-                  "Failed to load image:\n- {}",
-                  CesiumUtility::joinToString(loadedImage.errors, "\n- "));
-              LoadResult result;
-              result.state = RasterOverlayTile::LoadState::Failed;
-              return result;
-            }
-
-            if (!loadedImage.warnings.empty()) {
-              SPDLOG_LOGGER_WARN(
-                  pLogger,
-                  "Warnings while loading image:\n- {}",
-                  CesiumUtility::joinToString(loadedImage.warnings, "\n- "));
-            }
-
-            CesiumGltf::ImageCesium& image = loadedImage.image.value();
-
-            int32_t bytesPerPixel = image.channels * image.bytesPerChannel;
-
-            if (image.width > 0 && image.height > 0 &&
-                image.pixelData.size() >=
-                    static_cast<size_t>(
-                        image.width * image.height * bytesPerPixel)) {
-              void* pRendererResources =
-                  pPrepareRendererResources->prepareRasterInLoadThread(image);
-
-              LoadResult result;
-              result.state = RasterOverlayTile::LoadState::Loaded;
-              result.image = std::move(image);
-              result.credits = std::move(loadedImage.credits);
-              result.pRendererResources = pRendererResources;
-              return result;
-            }
-            LoadResult result;
-            result.pRendererResources = nullptr;
-            result.state = RasterOverlayTile::LoadState::Failed;
-            return result;
-          })
+      .thenInWorkerThread([this, tileId = tile.getID()](
+                              LoadedRasterOverlayImage&& loadedImage) {
+        return processLoadedImage(tileId, std::move(loadedImage));
+      })
       .thenInMainThread([this, &tile, isThrottledLoad](LoadResult&& result) {
         tile._pRendererResources = result.pRendererResources;
         tile._image = std::move(result.image);
--- a/Cesium3DTiles/include/Cesium3DTiles/RasterOverlayTileProvider.h
+++ b/Cesium3DTiles/include/Cesium3DTiles/RasterOverlayTileProvider.h
@@ -357,7 +357,37 @@ protected:
       const LoadTileImageFromUrlOptions& options = {}) const;
 
 private:
+  struct LoadResult {
+    RasterOverlayTile::LoadState state = RasterOverlayTile::LoadState::Unloaded;
+    CesiumGltf::ImageCesium image = {};
+    std::vector<Credit> credits = {};
+    void* pRendererResources = nullptr;
+  };
+
   void doLoad(RasterOverlayTile& tile, bool isThrottledLoad);
+
+  /**
+   * @brief Processes the given `LoadedRasterOverlayImage`, producing a
+   * `LoadResult`.
+   *
+   * This function is intended to be called on the worker thread.
+   *
+   * If the given `loadedImage` contains no valid image data, then a
+   * `LoadResult` with the state `RasterOverlayTile::LoadState::Failed` will be
+   * returned.
+   *
+   * Otherwise, the image data will be passed to
+   * `IPrepareRendererResources::prepareRasterInLoadThread`, and the function
+   * will return a `LoadResult` with the image, the prepared renderer resources,
+   * and the state `RasterOverlayTile::LoadState::Loaded`.
+   *
+   * @param tileId The {@link TileID} - only used for logging
+   * @param loadedImage The `LoadedRasterOverlayImage`
+   * @return The `LoadResult`
+   */
+  LoadResult processLoadedImage(
+      const CesiumGeometry::QuadtreeTileID& tileId,
+      LoadedRasterOverlayImage&& loadedImage);
 
   /**
    * @brief Begins the process of loading of a tile.
--- a/Cesium3DTiles/include/Cesium3DTiles/TileID.h
+++ b/Cesium3DTiles/include/Cesium3DTiles/TileID.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "Cesium3DTiles/Library.h"
 #include "CesiumGeometry/OctreeTileID.h"
 #include "CesiumGeometry/QuadtreeTileID.h"
 #include <string>
@@ -35,5 +36,21 @@ typedef std::variant<
     CesiumGeometry::OctreeTileID,
     CesiumGeometry::UpsampledQuadtreeNode>
     TileID;
+
+struct CESIUM3DTILES_API TileIdUtilities {
+  /**
+   * @brief Creates an unspecified string representation of the given {@link
+   * TileID}.
+   *
+   * The returned string will contain information about the given tile ID,
+   * depending on its type. The exact format and contents of this string
+   * is not specified. This is mainly intended for printing informative
+   * log messages.
+   *
+   * @param tileId The tile ID
+   * @return The string
+   */
+  static std::string createTileIdString(const TileID& tileId);
+};
 
 } // namespace Cesium3DTiles
--- a/Cesium3DTiles/src/TileID.cpp
+++ b/Cesium3DTiles/src/TileID.cpp
@@ -1,0 +1,43 @@
+#include "Cesium3DTiles/TileID.h"
+
+#include <string>
+#include <variant>
+
+namespace Cesium3DTiles {
+
+/*static*/ std::string
+TileIdUtilities::createTileIdString(const TileID& tileId) {
+
+  struct Operation {
+
+    std::string operator()(const std::string& url) { return url; }
+
+    std::string
+    operator()(const CesiumGeometry::QuadtreeTileID& quadtreeTileId) {
+      // Strings of the form "L10-X23-Y144"
+      return "L" + std::to_string(quadtreeTileId.level) + "-" + "X" +
+             std::to_string(quadtreeTileId.x) + "-" + "Y" +
+             std::to_string(quadtreeTileId.y);
+    }
+
+    std::string operator()(const CesiumGeometry::OctreeTileID& octreeTileId) {
+      // Strings of the form "L10-X23-Y144-Z42"
+      return "L" + std::to_string(octreeTileId.level) + "-" + "X" +
+             std::to_string(octreeTileId.x) + "-" + "Y" +
+             std::to_string(octreeTileId.y) + "-" + "Z" +
+             std::to_string(octreeTileId.z);
+    }
+
+    std::string operator()(
+        const CesiumGeometry::UpsampledQuadtreeNode& upsampledQuadtreeNode) {
+      // Strings of the form "upsampled-L10-X23-Y144"
+      return "upsampled-L" +
+             std::to_string(upsampledQuadtreeNode.tileID.level) + "-" + "X" +
+             std::to_string(upsampledQuadtreeNode.tileID.x) + "-" + "Y" +
+             std::to_string(upsampledQuadtreeNode.tileID.y);
+    }
+  };
+  return std::visit(Operation{}, tileId);
+}
+
+} // namespace Cesium3DTiles
--- a/CesiumGeometry/include/CesiumGeometry/OctreeTileID.h
+++ b/CesiumGeometry/include/CesiumGeometry/OctreeTileID.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "CesiumGeometry/Library.h"
+#include <cstdint>
 
 namespace CesiumGeometry {
 


Cesium3DTiles/src/RasterOverlayTileProvider.cpp
```diff
@@ -524,6 +525,58 @@ RasterOverlayTileProvider::loadTileImageFromUrl(
           });
 }
 
+RasterOverlayTileProvider::LoadResult
+RasterOverlayTileProvider::processLoadedImage(
+    const CesiumGeometry::QuadtreeTileID& tileId,
+    LoadedRasterOverlayImage&& loadedImage) {
+
+  const auto& pPrepareRendererResources = this->getPrepareRendererResources();
+  const auto& pLogger = this->getLogger();
```
kring, line 534: We try to avoid passing pointers to `RasterOverlayTileProvider` and most other objects (including `this`) to a worker thread, because these objects are not thread safe (by design) and it's easy to accidentally misuse them from a thread.

In this case, it's probably fine (not a race condition). But by sticking to this rule we make any use of `this` or a pointer/reference to a non-thread-safe object a red flag, and that hopefully helps us avoid race conditions.

So this method should be static. At which point it could be a static global or private-namespaced method instead, which would have the additional benefit of letting us remove some implementation details from a public header.

Cesium3DTiles/src/RasterOverlayTileProvider.cpp
```diff
@@ -537,61 +590,11 @@ void RasterOverlayTileProvider::doLoad(
...
-              return result;
-            }
-            LoadResult result;
-            result.pRendererResources = nullptr;
-            result.state = RasterOverlayTile::LoadState::Failed;
-            return result;
-          })
+      .thenInWorkerThread([this, tileId = tile.getID()](
+                              LoadedRasterOverlayImage&& loadedImage) {
+        return processLoadedImage(tileId, std::move(loadedImage));
```
kring, line 596: IMO this is an example of an "extract method" refactoring make the code less clear and harder to follow. When looking at `doLoad`, it's less clear what processing a loaded image actually entails, and in particular how it relates to the work that happens next in the main thread. And while looking at `processLoadedImage`, we're missing the vital context that it's invoked as a continuation in a worker thread. So, for example, that use of `this` may look fine in code review when actually it's pretty sus.

Cesium3DTiles/src/RasterOverlayTileProvider.cpp
```diff
@@ -524,6 +525,58 @@ RasterOverlayTileProvider::loadTileImageFromUrl(
...
+
+  const auto& pPrepareRendererResources = this->getPrepareRendererResources();
+  const auto& pLogger = this->getLogger();
+
+  if (!loadedImage.image.has_value()) {
+    std::string tileIdString =
+        Cesium3DTiles::TileIdUtilities::createTileIdString(tileId);
+    SPDLOG_LOGGER_ERROR(
+        pLogger,
+        "Failed to load image for tile " + tileIdString + ":\n- {}",
```
kring, line 541: Minor point, but it's better to use `{}` for `tileIdString` rather than building the string with `+`. And while it's probably irrelevant for an error, it's also slightly better to put the call to `Cesium3DTiles::TileIdUtilities::createTileIdString(tileId)` right inside the `SPDLOG_LOGGER_ERROR` call rather than assigning it to a local variable beforehand. That way the call can be eliminated entirely if that log level is disabled with a compile-time flag.

kring (commented):
The tile ID changes look good, thanks!. I'm going to nitpick some of the other refactoring that happened here, though, for the sake of discussion.

javagl:
Regarding the use of `this` and the extracted method: You made a similar point in another PR. I'm aware of the difficulties of race conditions. The lack of clarity and documentation add another layer of difficulty (i.e. it is not clear which objects are "thread safe" in which way, or which methods may or may not be called on which thread). And as much as I'd like to avoid repeating myself, and trying to say that as diplomatically as possible: Throwing hundreds of lines of code into several nested layers of undocumented lambdas, capturing and passing on random parts of the surrounding environment (of which it is not clear whether *they* are thread-safe, FWIW) is **just not good**. We *have* to find a better solution for that.

Here, it's only two things that are required from `this`, namely the `IPrepareRendererResources`, and, as always, the logger. (Why had the `tileRectangle` and the `projection `also been captured in the lambda originally? Well \*shrugs\* ... why not?). So I think that converting this into a static method is OK.

Cesium3DTiles/src/RasterOverlayTileProvider.cpp
javagl: I'm aware of that. We seem to have different priorities here.

kring:
> The lack of clarity and documentation add another layer of difficulty (i.e. it is not clear which objects are "thread safe" in which way, or which methods may or may not be called on which thread).

You keep saying this, but it's not true. The rules are fairly clear and totally the norm across C++ code bases. Documentation should tell you when something is explicitly thread safe or explicitly _not_ thread safe, but when it doesn't, you may assume:

1. You can use different instances of a class simultaneously from different threads all you want.
2. You can share a single instance between threads safely so long as _none_ of the threads are modifying it. If even one thread is or may be modifying it, it is _not_ safe to even call const methods or access simple fields.
3. Instances that are potentially modified in one thread should not be used - and ideally shouldn't even be available to - any other threads.

If it helps to write this down somewhere, please do. But the above is what we should expect from _any_ C++ library (and most other languages as well); it's not anything special about cesium-native.

To be clear, `RasterOverlayTIleProvider` is firmly in camp 3.

kring:
> Why had the tileRectangle and the projection also been captured in the lambda originally? Well *shrugs* ... why not?

They were used by the old rectangle cutout system that we removed before 1.0, and we forgot to remove them. It's a shame none of the compilers flagged them as unused.

kring:
> Throwing hundreds of lines of code into several nested layers of undocumented lambdas, capturing and passing on random parts of the surrounding environment (of which it is not clear whether they are thread-safe, FWIW) is just not good.

I don't think I agree with anything in this statement.

* It's not hundreds of lines.
* They're not "undocumented lambdas", they're continuations; in other environments with an async/await type feature we could write this code completely linearly without an extra lambda.
* We're not capturing "random parts of the environment", we're capturing what we need to do the job; by reference for things that can be safely shared, by value for things that can't due to lifetime or thread safety.
* Any time we pass something to a lambda, we need to think about its lifetime to ensure that the lambda can't outlive it. If that lambda is to be executed in a worker thread, we also need to think about its thread safety. Hopefully my previous comment makes it clearer what thread safety assumptions we can make about the things we potentially want to use from a worker.

Cesium3DTiles/src/RasterOverlayTileProvider.cpp
kring: What priority did you have here that my suggestion conflicts with?

javagl:
> Documentation should tell you when something is explicitly thread safe or explicitly not thread safe, but when it doesn't, you may assume:

I know that. That's the standard. But it's subtle. For example, consider https://github.com/CesiumGS/cesium-native/pull/232 . The problem was: This lambda did *something* in the worker thread. Some other class did *something* in the main thread. And these "somethings" interfered, causing random crashes. 

The fix was was essentially a change from `thenInWorkerThread` to `thenInMainThread`. But maybe there is "something else" done there now. And and some other class does "something else" on the worker thread with the same object. And that may cause new random crashes. I did **not** verify that this was **not** the case, because it is not possible in practice. 

And that's a problem. 

More broadly: You cannot claim that we do *not* have an issue with concurrency and thread safety when we experience random crashes, flickering of tiles, or other (vague) ~"unexpected random behavior". 

---

> I don't think I agree with anything in this statement.

The severity of this problem varies. And I neither want to argue about whether it's "hundreds" or "dozens" of lines, or about whether we call it "continuation" or "lambda" (and I'll leave out anecdotes about all the conferences about parallel programming that I visited, and where I never saw even remotely similar constructs). But to re-phrase my statements in a form that will not allow you to vaguely disagree:

- There are *many* lines of code in lambdas, and some of them are even nested
- They are undocumented, period. There is *no documentation whatsoever* about what the lambdas are doing. In order to figure that out, one has to read them, line by line.
- They are capturing parts of the environment, and there is *no documentation whatsoever* about 1. what these captures are used for or 2. how the lambda may affect these captures. 

Of course, things like https://github.com/CesiumGS/cesium-native/blob/main/CesiumIonClient/src/Connection.cpp#L101 are extreme cases. Bringing that into a shape that is easier to read, maintain, and test could be a lot of effort (and have a bad effort-to-benefit ratio *in this particular case*, because it's unlikely (fingers crossed) that this code will ever have to be touched by anyone again). But for the "core" of cesium-native, maintainability, readability and testability should be priorized higher than they are right now.

kring:
> The problem was: This lambda did something in the worker thread. Some other class did something in the main thread. And these "somethings" interfered, causing random crashes.

Please don't make it sound so mysterious. The previous code in #232 was passing a pointer to a CreditSystem instance to a worker. CreditSystem was not documented to be thread safe. Therefore, sharing it with the worker thread was a violation of the rules. And it caused a bug. We could have fixed this by either making CreditSystem thread safe or by making sure it's not shared between threads. We initially thought we should go with the first, but then realized the cost would be too high and decided to go with the second instead.

Yes, this sort of thing is easy to miss. If you have ideas how to make it less error-prone, I'm all ears. The purpose of the patterns we're using (`AsyncSystem`, the policy of not capturing `this` in lambdas passed to `.thenInWorkerThread`) is to make it so that, to be confident our code is thread-safe, we only need to look at the list of things that are captured in the `.thenInWorkerThread` lambda and confirm that they're all either copied or thread safe. But if we don't do that, or if we miss something, we'll have a bug.

It's clearly more error prone than other concurrency models like JavaScript's share-nothing Web Workers or an Actor model. But it has some significant advantages (mostly performance) and doesn't seem totally out of control to me.

But if you know of a better model for our needs (from attending many conferences on parallel programming or otherwise), then pelase tell me.

kring:
> They are undocumented, period. There is no documentation whatsoever about what the lambdas are doing. In order to figure that out, one has to read them, line by line.

Frankly I'm a little exasperated here. The lambdas are part of a method that is documented to do a certain job asynchronously. Because some of that work needs to happen in a worker thread, and some needs to happen in the main thread, we need to use continuations. There is no reason those continuations need to be "documented", any more than a random dozen lines or whatever in a synchronous method need to be documented. If it's unclear what those dozen lines do, comments and method extraction can help explain. But there's no particular documentation needed just because we're using lambdas.

> They are capturing parts of the environment, and there is no documentation whatsoever about 1. what these captures are used for or 2. how the lambda may affect these captures.

I'm not sure what you mean by "how the lambda may affect these captures."

jtorresfabra:
Hey let me chime in (sorry for the intrusion).

From my experience lambdas can help readability or hurt readability. I agree with @javagl that if a lambda is long enough it really can hurt readability and can make you to lose the big picture pretty easy. For that matters sometimes I find more readable to use private (non-friend non-member) functions inside of an unnamed namespace. 

Say that, a private function certainly does not need to be documented for the user (as they won't see it, the same as lambdas) but definitely can have some dev comments here and there (and to be honest while scrubbing some parts of cesium native I miss some dev comments). I usually forget what I wrote two days after, so it's more a help for my future me than anything else :sweat_smile: .

Lambdas can equally have this kind of comments, and I would say extracting a lambda to a function is a common sense judgement. Also there are cases when lambdas cannot easily be changed to a private function, when capturing local variables for instance. 

Anyway I guess it's not a great deal to change those functions that are not very readable now  (either extracting or adding more comments).

All in all what I would love to see is more developer comments all over the place, but maybe is a personal preference as I tend to forget how things work very easily. Also when reading others people code, especially if there are some clever/smart tricks involved (template magic, ultramodern c++30 features, niche C++ features, or whatever).

Sorry for the disruption again, and keep up the good work!

javagl:
> The purpose of the patterns we're using [...] is to make it so that, to be confident our code is thread-safe, we only need to look at the list of things that are captured in the `thenInWorkerThread` lambda and confirm that they're all either copied or thread safe. 

The crucial point is that they are not used in a `thenInMainThread` anywhere else. And that's hard to figure out. 

> Frankly I'm a little exasperated here. The lambdas are part of a method that is documented to do a certain job asynchronously. Because some of that work needs to happen in a worker thread, and some needs to happen in the main thread, we need to use continuations.

The fact that these "parts of work" are entirely different and have to be strictly separated (in terms of thread rules) is a good reason to try and give **names** to these  parts of the work, and document their thread requirements:

    /** Supposed to be called on the main thread, for the CreditSystem */
    Result BingMapsRasterOverlay::processResponse(Response response, CreditSystem creditSystem) { 
        creditSystem.createCredit(...)
    }

    Result BingMapsRasterOverlay::createTileProvider(...) {
      return pAssetAccessor->requestAsset(asyncSystem, metadataUrl)
        // This an error that is somewhat easier to spot
        .thenInWorkerThread([....](....) { processResponse(response, creditSystem); })
        ...
    }

(For *private* methods, one could even go so far to call them `processResponseOnMainThread` or so...)

---

BTW: I do not know any coding guideline that does *not* explicitly state a limit on the "number of lines of code per method". Of course, "LOC" is not a sensible measure, but the ideas of "having a method do one thing (and only that)" and "limiting the complexity of a method" are clear. And "one method doing things on different threads" *is* a complexity that should be viewn with scrutiny

---

> I'm not sure what you mean by "how the lambda may affect these captures."

There is the function   

    void createTileProvider(...
      const std::shared_ptr<CreditSystem>& pCreditSystem)

Inside one of the implementations, there is 

    return pAssetAccessor->requestAsset(asyncSystem, metadataUrl)
      .thenInWorkerThread(
          [... pCreditSystem, ...]

and 120 lines further down the road, there is the critical call to `pCreditSystem->createCredit`. That's far easier to miss than a function that carries a `/** To be called on thread X, @param The thing that requires thread X */` comment. 

---

Related: I think that we should do a pass over our function signatures. My gut feeling is that many parameters could be changed from 

    const std::shared_ptr<SomeType>& pointer

to

    const std::shared_ptr<const SomeType>& pointer

e.g. for cases where no mutating (i.e. non-`const`) function is called on the object. This `const` *should* be added wherever it *can* be added. 

---

> Say that, a private function certainly does not need to be documented for the user

The "user" may well be a colleague who rightfully wants to *call* this function, and should have a clear idea about what the function does, and which caveats it might have - and this should preferably be possible **without** reading the method, line by line, and reading every method that is *called* from there, line by line. More generally: I'm *strongly* convinced that *writing comments* helps more in *writing better code* than many other high-level best-practice recommendations. When writing a comment to a class, and not being able to describe *what an instance of that class is*, then that's a red flag. And when writing a comment to a method, and the comment is long and contains lots of "if" and "and" and "when", then the method should be broken down.

kring:
Uh oh, if @javagl and @jtorresfabra mostly agree, and I'm on the other side, I must really be off my rocker! 😆

Thanks for the thoughts, though, @jtorresfabra. It all sounds reasonable to me, and I don't think I disagree with any of it.

I think maybe the thing that's missing when it comes to AsyncSystem is some higher-level documentation of what it's all about, how to use it, and what to watch out for. What should you capture in the lambdas and why? I don't think it's appropriate (or at least, I'm way too lazy) to try to document that at the (many) places it's used, but we should have it clearly laid out somewhere. I've been working under the assumption that the analogy to JS promises would make it inuitive to everyone without much documentation, and that the implications of thread safety on what you pass to the lambdas (and how) would be clear, but it's probably not true.

Meanwhile, I do stand by my original comment that started this discussion:

> IMO this is an example of an "extract method" refactoring make the code less clear and harder to follow. When looking at doLoad, it's less clear what processing a loaded image actually entails, and in particular how it relates to the work that happens next in the main thread. And while looking at processLoadedImage, we're missing the vital context that it's invoked as a continuation in a worker thread. So, for example, that use of this may look fine in code review when actually it's pretty sus.

Marco has since renamed `processLoadedImage` to `createLoadResultFromLoadedImage`, and the new name actually helps a lot. If the code were written this way originally, I might suggest putting the extracted method in an anonymous namespace to get it out of the header, but otherwise wouldn't complain. But I don't think it's necessary or even an improvement to go back and refactor all our code like this to immediately call an anonymous-namespace method from inside every `.thenX`.

The way I think about a method like this `doLoad` here is that it's a straightforward, linear process that just happens to be asynchronous. The asynchronicity unfortunately makes the code look a bit more awkward, but once you get used to the style and see it for the linear process it is, it's not hard to read. Habitually extracting methods for each lambda actually obfuscates the nature of this code and makes it a little bit more error prone.

I want to emphasize that I'm not arguing against method extraction in general. For example, I think what you did here @javagl is great:
https://github.com/CesiumGS/cesium-native/pull/232/files

If I can try to articulate the difference...  `collectCredits` in #232 is a clearly distinct part of the process of creating a BingMapsTileProvider, with succinct inputs and a single output. `createLoadResultFromLoadedImage` is the entire threadable part of the `doLoad`, it has a mess of inputs, and it has an output that is interesting only because it's what the non-threadable part of `doLoad` expects. Prior to the refactoring in this PR, `LoadResult` was a private struct of `doLoad`. In other words, the `collectCredits` refactoring improves the readability of the code by moving self-contained and less important details out of the main method. The `createLoadResultFromLoadedImage` refactoring, on the other hand, makes the code less cohesive, with little apparent benefit other than avoiding a kinda-long lambda.

I guess what I'm saying is that I'm 👍 with extracting methods from AsyncSystem continuations using the same heuristics you'd use to decide when to extract a method from any method: improve reuse, improve testability, to remove less important details, etc. I'm less ok with extracting an entire continuation into a separate function. Maybe I'm splitting hairs here. As I mentioned before, I wouldn't complain much if the code in this PR were written originally like this, rather than a refactoring of existing code. But Marco has been keen to refactor the code for readability, and if we're going to do that we should try to be clear about what it means. Hence all the bike shedding.

And as an aside, I really dislike what our current clang-format config does with lambdas. If we were actually able to format our C++ code the way Prettier formats javascript, the code would be less indented and would do a much better job of revealing the programmatic structure with the source structure. But I don't know how to fix it with clang-format's limitations.

kring:
> The crucial point is that they are not used in a thenInMainThread anywhere else. And that's hard to figure out.

No! It has almost nothing to do with `thenInMainThread`.

It's not safe to capture a pointer or reference for a `thenInWorkerThread` continuation unless the object pointed to or referenced is documented to be thread safe. End of story! Don't even look at how it's used. We shouldn't pass a non-thread-safe `CreditSystem` to a worker continuation _no matter how we use it_. Just like we should never pass `this` for a `RasterOverlayTileProvider` or anything else.

> and 120 lines further down the road, there is the critical call to pCreditSystem->createCredit. That's far easier to miss than a function that carries a /** To be called on thread X, @param The thing that requires thread X */ comment.

Just to reiterate, capturing a `CreditSystem` by pointer or reference in a `thenInWorkerThread` lambda should be a red flag, whether that CreditSystem is used on the next line, 120 lines later, or not at all. Because it means we're sharing (or risking sharing) a non-thread-safe object between threads.

> I do not know any coding guideline that does not explicitly state a limit on the "number of lines of code per method".

Hmm, well, I do.
The C++ Core Guidelines say nothing about it.
The LLVM Coding Standard says nothing about it.
The Google Coding Standard says to try to write short functions, but "We recognize that long functions are sometimes appropriate, so no hard limit is placed on functions length."

> Of course, "LOC" is not a sensible measure, but the ideas of "having a method do one thing (and only that)" and "limiting the complexity of a method" are clear. And "one method doing things on different threads" is a complexity that should be viewn with scrutiny

Of course having a method do one thing and limiting method complexity are good goals, no argument there. And of course doing things across threads is complexity. That complexity will unfortunately exist in `doLoad` whether we break its continuations into separate functions or not. The question is whether extracting each continuation into a separate lambda reduces complexity. I say no. The complexity is all still there, but now spread out into multiple places in such a way that it's harder to follow.

> e.g. for cases where no mutating (i.e. non-const) function is called on the object. This const should be added wherever it can be added.

Yeah that makes sense to me.

Keep in mind, though, that sharing an object across threads is safe only if _all_ threads (including the main one) treat it as const. It's not good enough for the worker thread to access it via a pointer to const if there's any possibility of another thread modifying it. Since this is usually very hard to determine, and even harder to ensure stays true as the code evolves in the future, we basically shouldn't ever do it.

javagl:
There are many levels on which the further discussion here could take place. I'll omit questions that could be considered as some sort of... handbrake-turn on the highway. (An example for such a question could be: Why do we only want *one* worker thread on a 6-core CPU?). So trying to narrow it down:

> I think maybe the thing that's missing when it comes to AsyncSystem is some higher-level documentation of what it's all about, how to use it, and what to watch out for. What should you capture in the lambdas and why?
> ...
> It's not safe to capture a pointer or reference for a thenInWorkerThread continuation unless the object pointed to or referenced is documented to be thread safe. End of story! Don't even look at how it's used.

It *would be* perfectly safe to use *anything* in the worker thread, *if* everything that is done with this object happened on the worker thread. End of story. The fact that there is ~"a collection of classes/objects that are *only supposed to be used (!) on the main thread*" is implicit knowledge. You "know" that because you created a mental model of the whole thing and implemented it. But deriving a certain mental model from code is hard or impossible. 

(Or to put it that way: It's not something that could sensibly be documented in the AsyncSystem, but only contextually, as a description of how exactly the AsyncSystem is supposed to be used in `cesium-native` in general. Even then, I don't have a clear idea about how to reduce the risk of accidentally using something on the wrong thread, be it technically or on the level of documentation). 

But my main point originally was not so much about the AsyncSystem itself, but more about the level of granularity and documentation in general. Looking at some examples, like the `Connection` one that I already linked to, **highly** critical (but far too complex) functions like `Tile::loadContent`, or something like https://github.com/CesiumGS/cesium-native/blob/main/Cesium3DTiles/src/TileMapServiceRasterOverlay.cpp#L121 , I'd strongly encourage you to try and take a step back, put yourself into the shoes of someone who has no clue what this is all about, and just has this one, simple question: *What is happening there?*. 

For the last example: Sure, it's called `createTileProvider`, so .... it will probably create a tile provider. But there are >100 LOC, and even though some of the ">" can be attributed to the clang formatting here, the only way of making an educated guess to answer that simple question is *to read the code*. Line by line. 

Iff this function was a bit closer to pseudocode like
```
LoadingContext loadingContext = { owner, accesor, whatnot };
struct TileDims { int width, height, minLevel, maxLevel };

return pAssetAccessor->requestAsset(asyncSystem, xmlUrl, this->_headers)
  .thenInWorkerThread([loadingContext](Request request) {
       JsonDocument document;
       if (!parseResponse(request->response, document)) return nullptr;
       
       TileDims tileDims = parseTileDims(document);
       Rectangle rectangle = parseRectangle(document, options);

       return std::make_unique<TileMapServiceTileProvider>(
           loadingContext, tileDims, rectangle);
  })
})
```
then answering the question of "What is happening there?" would involve reading 5 lines of code: It converts the response into a document, obtains the tile dimension and the coverage rectangle, and returns the tile provider. **If** it was necessary to know more details e.g. about what "computing the coverage rectangle" means, then one could read the doxygen documentation of `parseRectangle`, which would include the hint ~"to be called on worker thread".

 > The C++ Core Guidelines say nothing about it.

The C++ Guidelines and LLVM are targeting a *very* low level (and one reason of the difficulty of coming up with sensible coding standards, e.g. on a company level, is that "important" rules are buried under rules that dictate where to put a ` ` space or not...). 

Fortunately, these guidelines have already moved from the "non-sensical" LOC measure to others. But the first and most important points of the "Functions" category already try to encourage the right style, and https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-single still explicitly talks about "one-to-five-line functions" (which would be in line with the refactoring pseudocode that I sketched above). 

Some of the https://clang.llvm.org/extra/clang-tidy/checks/list.html try to establish measures for ["cognitive complexity"](https://clang.llvm.org/extra/clang-tidy/checks/readability-function-cognitive-complexity.html) or a ["function size"](https://clang.llvm.org/extra/clang-tidy/checks/readability-function-size.html) that *can* refer to LOC, but also to the number of statements (with the obvious caveat of some `SPDLOG_LOGGER_ERROR` macro that might well expand to hundreds or thousands of LOC/statements). 

(BTW: Now this starts to become related to https://github.com/CesiumGS/cesium-native/issues/203 - I'm not a CI expert and probably couldn't set this up without admin rights somewhere, but will read a bit more the CLang toos in general - the https://clangpowertools.com/ seem to be a good starter here).

Coming back to one point

> But I don't think it's necessary or even an improvement to go back and refactor all our code like this to immediately call an anonymous-namespace method

There's a certain amount of subjectivity involved. The C++ Guidelines stay vague here as well:

> * Flag functions that do not “fit on a screen.” How big is a screen? Try 60 lines by 140 characters; that’s roughly the maximum that’s comfortable for a book page.
> * Flag functions that are too complex. How complex is too complex? You could use cyclomatic complexity. Try “more than 10 logical path through.” Count a simple switch as one path.

Refactoring existing code to reduce the complexity of functions (particularly, functions that contain `then...` calls with large lambdas) is never "necessary", so to speak. I think that a general refactoring to break long functions and lambdas into shorter, documented ones would be worthwhile. But I know that there's always a trade off for that (roughly: it cannot be announced as a new feature - unless "reducing technical debt" is considered to be a "feature"...). But *if* this discussion leads to a state where

- ... people follow the "Boy Scout Rule" and break/document long functions ~"while they are working on that code part anyhow" and
- ... people don't *introduce* long functions in the first place

then the (my) main goal has been achieved, because I'm convinced that it **will** pay out, in better maintainability, testability, and fewer bugs.

kring:
I haven't processed all that, but one really quick correction... There isn't _one_ worker thread. Continuations registered with `.thenInWorkerThread` are dispatched with `ITaskProcessor::startTask` to "a background thread." Not necessarily the same one each time it's called. In Cesium for Unreal, they're dispatched to Unreal's task system, which uses a thread pool. I don't know offhand how many threads are in UE's pool by default, but it's definitely more than 1.

kring:
> The fact that there is ~"a collection of classes/objects that are only supposed to be used (!) on the main thread" is implicit knowledge. You "know" that because you created a mental model of the whole thing and implemented it. But deriving a certain mental model from code is hard or impossible.

I don't think there's any special knowledge required, beyond what could be easily documented in a "how to use AsyncSystem" tutorial. Any time you call `.thenInWorkerThread`, you are allowing - requesting, in fact! - that the continuation be invoked in some random other thread. Even if you're already in a worker thread, that continuation can and usually will be invoked in a _different_ worker thread. Therefore it is _never_ ok to give that lambda a pointer or reference to a non-thread-safe object. There's no implicit knowledge required to determine this, either. Look at the documentation for the object you're considering passing to the continuation. Does it explicitly say it's thread safe? No? Then you can't pass it by pointer or reference.

kring:
@javagl all this discussion is interesting and valuable, but let's get this PR merged. Unless you're opposed (or it doesn't work for some reason that isn't clear to me), please move `createLoadResultFromLoadedImage` and `LoadResult` out of the header and into the anonymous namespace in the .cpp. And then I'll merge this.

javagl:
Well, details are important: I assumed (and that assumption may have been naive, and it *is* definitely wrong, but ... not totally unreasonably, IMHO) that there was only *one*, dedicated worker thread - similar to the "HTTP Thread" or the "Game Thread" or the "IO Thread" or whatnot that exist in Unreal. So when `thenInWorkerThread` is rather `thenInAnyWorkerThread` (aka `throwThisIntoSomeThreadPool`), some of my statements make no sense. 

But again: I'm not dissecting the AsyncSystem here, or arguing about idioms in C++ vs. JS - the main points have been about keeping functions small, and lambdas even smaller, giving them *names*, and *documenting* them *and* their inputs (parameters/captures).

javagl:
> move createLoadResultFromLoadedImage and LoadResult out of the header and into the anonymous namespace

Done in the last commits. More generally, we could (and IMHO should) try to aim for some consistency here: There are many static functions and helper `struct`s right now that are *not* in an anonymous (or `impl`) namespace, although they could (and therefore IMHO *should*) be.

I already ... mentioned earlier that I ... already mentioned this earlier ... 😳 at the bottom of https://github.com/CesiumGS/cesium-native/pull/94#discussion_r547997889  - if you don't shout "STOP!", I'd open an issue for that later today.

kring:
Thanks @javagl!

Issue #215, opened 2021-04-17
Review the code for potential thread safety problems


javagl:
<sup>(Note: The actual contents of this issue is written "as if" it referred to a single function of a single class. It does not. There certainly are many places where a review with the appropriate scrutiny would uncover potential race conditions. I'm just using one class and one function as an example here. These "general considerations" will have to be broken down into individual issues that can be addressed via dedicated PRs)</sup>

There have been reports of [occasional crashes](https://github.com/CesiumGS/cesium-unreal/issues/348) in `cesium-unreal`, and the term "occasional" always means that there is some threading issue behind that. (Well, maybe not always. But occasionally...). As an example, some of the traces pointed to [`CreditSystem::createCredit`](https://github.com/CesiumGS/cesium-native/blob/d6587395aa3763799b09d25baa72782b838b9ed5/Cesium3DTiles/src/CreditSystem.cpp#L6). This is currently implemented as

    Credit CreditSystem::createCredit(const std::string& html) {
      // if this credit already exists, return a Credit handle to it
      for (size_t id = 0; id < _credits.size(); ++id) {
        if (_credits[id].html == html) {
          return Credit(id);
        }
      }

      // this is a new credit so add it to _credits
      _credits.push_back({html, -1});

      // return a Credit handle to the newly created entry
      return Credit(_credits.size() - 1);
    }
 
which iterates over a vector and may modify that same vector, so clearly is not thread safe. But looking at possible "call trees" leading to that function, one can see one call to that function in `Tileset.cpp` with:

    this->_externals.pAssetAccessor->requestAsset(this->_asyncSystem, ionUrl)
      .thenInMainThread([this](std::shared_ptr<IAssetRequest>&& pRequest) {
        this->_handleAssetResponse(std::move(pRequest));
      })

via

    void Tileset::_handleAssetResponse(std::shared_ptr<IAssetRequest>&& pRequest) {
        ...
          this->_tilesetCredits.push_back(
              this->_externals.pCreditSystem->createCredit(
              html->value.GetString()));

and one call from `BingMapsRasterOverlay.cpp` with

    Future<std::unique_ptr<RasterOverlayTileProvider>>
    BingMapsRasterOverlay::createTileProvider( 
      ...
      ) {
      ...
      return pAssetAccessor->requestAsset(asyncSystem, metadataUrl)
          .thenInWorkerThread(
          [ ... ](std::shared_ptr<IAssetRequest> pRequest)
[...]

PR #227, opened 2021-04-26
Added a mutex to CreditSystem

javagl:
As mentioned in https://github.com/CesiumGS/cesium-native/issues/215 

This (until now) only adds a mutex/lock to all methods of `CreditSystem`. And I think that's what has to be done to be really thread-safe. 

In a deeper analysis, one could remove the mutex/lock for some cases, and add disclaimers and say in a comment that it is *assumed* that, for example, `addCreditToFrame` and `startNextFrame` are always called from the same thread. 

Another step further, I'm not entirely sure about the conventions: `getCreditsToShowThisFrame` returns a reference to the vector. The caveats of having *one* thread operating on such a vector, and *another* thread modifying the vector (via any of the methods) are probably beyond what we can sensibly address...

BTW: I have added a ""test case"", but it looks crude: It just creates a pile of threads that all create credits like mad, and essentially expects the whole thing to *not* crash. *Without* the mutexes, this crashed pretty reliably (at different places). With the mutex, it never crashed. But there's probably some better, idiomatic way of testing this, so the text case is currently omitted via `#ifdef`. Any hint about how to test this sensibly is appreciated (otherwise, I'd try to find some infos on that, but that would probably not be in the scope of this PR)

### DIFF
--- a/Cesium3DTiles/src/CreditSystem.cpp
+++ b/Cesium3DTiles/src/CreditSystem.cpp
@@ -1,9 +1,13 @@
 #include "Cesium3DTiles/CreditSystem.h"
 #include <algorithm>
+#include <mutex>
 
 namespace Cesium3DTiles {
 
 Credit CreditSystem::createCredit(const std::string& html) {
+
+  std::lock_guard<std::mutex> lock(_mutex);
+
   // if this credit already exists, return a Credit handle to it
   for (size_t id = 0; id < _credits.size(); ++id) {
     if (_credits[id].html == html) {
@@ -19,6 +23,7 @@ Credit CreditSystem::createCredit(const std::string& html) {
 }
 
 const std::string& CreditSystem::getHtml(Credit credit) const {
+  std::lock_guard<std::mutex> lock(_mutex);
   if (credit.id < _credits.size()) {
     return _credits[credit.id].html;
   }
@@ -26,6 +31,7 @@ const std::string& CreditSystem::getHtml(Credit credit) const {
 }
 
 void CreditSystem::addCreditToFrame(Credit credit) {
+  std::lock_guard<std::mutex> lock(_mutex);
   // if this credit has already been added to the current frame, there's nothing
   // to do
   if (_credits[credit.id].lastFrameNumber == _currentFrameNumber) {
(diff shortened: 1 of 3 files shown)


Cesium3DTiles/src/CreditSystem.cpp
```diff
@@ -19,13 +23,15 @@ Credit CreditSystem::createCredit(const std::string& html) {
...
 const std::string& CreditSystem::getHtml(Credit credit) const {
+  std::lock_guard<std::mutex> lock(_mutex);
   if (credit.id < _credits.size()) {
     return _credits[credit.id].html;
   }
   return INVALID_CREDIT_MESSAGE;
 }
 
 void CreditSystem::addCreditToFrame(Credit credit) {
+  std::lock_guard<std::mutex> lock(_mutex);
```
kring, line 34: This lock is likely to be very expensive, as it can happen hundreds or even thousands of times per frame. I have to think about how we can avoid it...

kring:
@javagl I think I misled you in #215. It's fine to lock inside `createCredit`, because that method is called relatively infrequently and only in the course of a much more time consuming operating (like loading tiles). But what I didn't realize is that thread safety would necessitate locking in `addCreditToFrame`, too. And that method is called multiple times for every Bing raster overlay tile in every geometry tile in every frame. That's a _lot_ of locking. I've always been under the impression that locking a mutex is quite an expensive operation, even when it's uncontended but especially when it's contended. [This article](https://travisdowns.github.io/blog/2020/07/06/concurrency-costs.html) leads me to believe it's not quite as bad as I thought, but still wasteful if we have an alternative.

Do we have an alternative? Yes, I think so. All we need to do in this case is move the one `createCredit` call happening in a worker thread - in `BingMapsRasterOverlay::createTileProvider` - into the main thread. Creating a Bing Maps tile provider is a relatively cheap one-time operation, so there's really very little downside to this. And the upside is we don't need any locking in `CreditSystem` anymore, which should make the potentially thousands of calls to `addCreditToFrame` in every frame much cheaper. That's a great tradeoff.

javagl:
Indeed, I (already anticipated that you'd point that out, but) didn't have on the radar that the other functions are called **so** frequently. (If I had known that, I'd also hesitated to add this sledgehammer-locking there). 

So the conclusion/resolution is to explicitly declare `CreditSystem` as being **not** thread-safe, and making sure that it's only used from the main thread. This is done in https://github.com/CesiumGS/cesium-native/pull/232 , closing here.

PR #232, opened 2021-04-28
Moved credit creation into main thread

javagl:
As suggested in https://github.com/CesiumGS/cesium-native/pull/227#issuecomment-828201482 , the call in `BingMapsRasterOverlay::createTileProvider` that callled `CreditSystem::createCredit` has been moved from the worker thread to the main thread.

As a ... (hacky? or pragmatic?) ... well, as a "layer of validation", I added some 
```
std::string currentThreadString() {
    std::ostringstream ss;  
    ss << std::this_thread::get_id();
    return ss.str();
}
Credit CreditSystem::createCredit(const std::string& html) {

  SPDLOG_WARN("Called createCredit on {}", currentThreadString());
```
(and to all other functions of `CreditSystem`), and indeed, *most* functions had been called by the same thread *most* of the time, *except* for that few `createCredit` calls. After this change, **all** calls to the `CreditSystem` had been made on the same thread.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,12 +42,8 @@ jobs:
       # We're using Clang 10 instead of the default (Clang 7) because of std::variant related failures
       # on Clang 7, possibly related to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90397
       install:
-        - sudo apt-get check
-        - sudo apt-get clean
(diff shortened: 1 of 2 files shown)


kring:
Thanks @javagl!

Issue #196, opened 2021-04-06
Assume terrain tiles are quantized-mesh even if the Content-Type isn't right


kring:
Currently, terrain tiles aren't loaded any differently from other 3D Tiles content, which means that the only way we'll know we're dealing with quantized-mesh and use the right loader is if the `Content-Type` header is `application/vnd.quantized-mesh`. This is problem when trying to load local terrain from a `file://` URL because there's no way to way to set a response header on a file URL.

Two ways I can think of to deal with this:
1. Allow loaders to be registered for file extensions. These would be tried after the magic value and content type both fail to resolve.
2. When we start with an old-school terrain layer.json, assume the tile content is always terrain. Skip the usual loader process entirely.

leielyq:
What are the terrain formats currently supported?

leielyq:
The following file format will fail to load, and there will be a lot of blank areas. The http method is used.

```
{
    "attribution": "http://www.cesiumlab.com",
    "available": [
        [
            {
                "endX": 1,
                "endY": 0,
                "startX": 0,
                "startY": 0
            }
        ],
        [
            {
                "endX": 3,
                "endY": 1,
                "startX": 2,
                "startY": 0
            }
        ],
        [
            {
                "endX": 7,
                "endY": 3,
                "startX": 6,
                "startY": 2
            }
        ],
        [
            {
                "endX": 13,
                "endY": 5,
                "startX": 12,
                "startY": 4
            }
        ],
        [
            {
                "endX": 27,
                "endY": 11,
                "startX": 26,
                "startY": 10
            }
        ],
        [
            {
                "endX": 53,
                "endY": 21,
                "startX": 52,
                "startY": 20
            }
        ],
        [
            {
                "endX": 107,
                "endY": 43,
                "startX": 106,
                "startY": 42
            }
        ],
        [
            {
                "endX": 213,
                "endY": 87,
                "startX": 212,
                "startY": 84
            }
        ],
        [
            {
                "endX": 427,
                "endY": 173,
                "startX": 426,
                "startY": 170
            }
        ],
        [
            {
                "endX": 855,
                "endY": 345,
                "startX": 852,
                "startY": 342
            }
        ],
        [
            {
                "endX": 1711,
                "endY": 691,
                "startX": 1704,
                "startY": 686
            }
        ],
        [
            {
                "endX": 3421,
                "endY": 1383,
                "startX": 3410,
                "startY": 1374
            }
        ],
        [
            {
[...]

PR #233, opened 2021-04-29
Separate tileset options into streaming-options and content-options

nithinp7:
The old `TilesetOptions` only has streaming-related options, so I renamed it `TilesetStreamingOptions` and created a separate options struct called `TilesetContentOptions` (open to suggestions for a different name for this one). The `TilesetContentOptions` is also passed into the `Tileset` constructor, but unlike the streaming options, the content options gets passed (through `TileContext`) into `Tile` and then from there to `TileContentLoadInput`. There, it becomes available to the content loaders while they are parsing content as well as creating gltf models. In my current, very narrow use case, I'm only using it for the watermask, but I assumed this will be a useful infrastructure for the near future; to be able to configure content parsing / gltf creation from outside the cesium-native API.

Watermask can now trivially be enabled/disabled from the cesium-unreal side as a result of these changes.


nithinp7:
@kring Can you have a quick look at this when you get the chance? This is a pretty simple refactor PR. I made this separate so it would be easier to review compared to having this all be part of the watermask PR directly (since this is more general than just the watermask).

baothientran:
sorry for the delay, I'm looking through the PR currently

kring:
I'll let Bao review this properly, but I just took a quick look at a high level, and I want to throw out an idea. What do you think about keeping `TilesetOptions` with that name, and putting `TilesetContentOptions` as a field inside it? My reasoning is that there's at least one property (`credit`) that isn't exactly a streaming option, and there may be more in the future. And structuring it this way is a much less breaking change for existing code.

Cesium3DTiles/include/Cesium3DTiles/TileContentLoadInput.h
```diff
@@ -6,6 +6,8 @@
 #include "Cesium3DTiles/TileContext.h"
 #include "Cesium3DTiles/TileID.h"
 #include "Cesium3DTiles/TileRefine.h"
+#include "Cesium3DTiles/Tileset.h"
```
baothientran, line 9: I thinks it's better that TileContentLoadInput should require TilesetContentOption in its constructor, so that it doesn't depend on Tileset class.

baothientran (commented):
@nithinp7 For the code parts, it looks good to me. But for the architecture side, the Option design is a little awkward because 3D tile format doesn't have that concept of watermask (you probably knew this).

baothientran:
Let me think about this a little bit more. If this is done right, we might be able to tackle https://github.com/CesiumGS/cesium-native/issues/196 too

nithinp7:
After offline discussion, we decided it's best to hold off on this refactor until a more thorough refactor of the content loaders is ready. For now, TilesetContentOptions will go inside TilesetOptions and TilesetStreamingOptions doesn't need to be created (since it's a breaking change with no urgency). I am closing this since we now only need a small watermask-specific change that I will implement here: https://github.com/CesiumGS/cesium-native/pull/207.

Issue #218, opened 2021-04-19
`npm run format` can break doxygen `{@link foobar}` comments


Samulus:
I noticed this while fixing the documentation on my recently merged PR. You have to take special care to avoid clang-format from wrapping a `{@link foobar}` declaration to another line (this breaks the rendered doxygen html)

kring:
Oh geeze that's really annoying. Any ideas on how to fix it, short of disabling comment reflow entirely?

Samulus:
https://stackoverflow.com/a/51913607 I found this, I'll give it a try and make a trivial PR if it fixes it.

javagl:
I fixed the comments that had been wrong in https://github.com/CesiumGS/cesium-native/pull/234 

For avoiding that problem in the future, we considered different options. The fallback would have been to disable formatting for comments altogether. But the "least invasive" solution was (c.f. the link by @Samulus ) to use a `CommentPragmas` at https://github.com/CesiumGS/cesium-native/pull/234/files#diff-1026e0038b722990204a42bed8a6f7c0ec2302aa79e3fad1959d62ba968edfa2R37 

`CommentPragmas: '{.*}'`

this will cause `clang-format` to **NOT** break lines that contain `{@link SomeClass}` links.

An aside: I also checked whether there...

    /**
     * ...may be a problem with some line that is long `and where the part that would end up on the next line was backticked`
     */

and these are broken, as in

    /**
     * ...may be a problem with some line that is long `and where the part that would 
     * end up on the next line was backticked`
     */

**but** doxygen can cope with that, and the output still looks right.

javagl:
@Samulus The `CommentPragmas` now prevent `clang-format` from breaking lines that contain `{@link SomeClass}` links. It would be nicer if there was the option to *only* treat the `{@link SomeClass}` like "**one** word", but there doesn't seem to be such fine-grained control. 
I assume that this can be closed, and will do that soon.

PR #234, opened 2021-04-29
Comment fixes for clang formatting

javagl:
Fixes the doxygen errors that had been introduced by https://github.com/CesiumGS/cesium-native/issues/218

Added/fixed minor missing comments, and minor changes (e.g. from renamed parameters and such).

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,5 +1,5 @@
 
-# Based on the https://www.kernel.org/doc/Documentation/CodingStyle
+# Based on the LLVM coding style https://llvm.org/docs/CodingStandards.html
 BasedOnStyle: LLVM
 
 # Always add a line break after an opening bracket if the parameters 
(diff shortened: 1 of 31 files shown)


.clang-format
```diff
@@ -1,8 +1,37 @@
+
+# Based on the https://www.kernel.org/doc/Documentation/CodingStyle
```
kring, line 2: Thanks for adding the comments explaining the options. This link seems to be broken, though. Plus our style isn't really derived from the Linux kernel style at all, is it?

kring (commented):
Thanks @javagl, looks good other than one question.

.clang-format
javagl: Sorry, did that in a haste yesterday, not sure where that one came from. Updated to link to LLVM.

PR #237, opened 2021-04-30
Avoid unnecessarily restarting Bing Maps sessions

nithinp7:
There was some confusion in a previous attempt to implement this, but now I think it works.

### DIFF
--- a/Cesium3DTiles/src/BingMapsRasterOverlay.cpp
+++ b/Cesium3DTiles/src/BingMapsRasterOverlay.cpp
@@ -15,6 +15,7 @@
 #include <optional>
 #include <rapidjson/document.h>
 #include <rapidjson/pointer.h>
+#include <unordered_map>
 #include <utility>
 #include <vector>
 
@@ -30,12 +31,7 @@ struct CreditAndCoverageAreas {
   std::vector<CoverageArea> coverageAreas;
 };
 
-struct SessionCacheItem {
-  std::string metadataUrl;
-  std::vector<std::byte> responseData;
-};
-
-std::vector<SessionCacheItem> sessionCache;
+std::unordered_map<std::string, std::vector<std::byte>> sessionCache;
 } // namespace
 
 using namespace CesiumAsync;
@@ -242,7 +238,7 @@ namespace {
  * {
  *   "imageryProviders": [
  *     {
- *       "attribution": "� 2021 Microsoft Corporation",
+ *       "attribution": "© 2021 Microsoft Corporation",
  *       "coverageAreas": [
  *         { "bbox": [-90, -180, 90, 180], "zoomMax": 21, "zoomMin": 1 }
  *       ]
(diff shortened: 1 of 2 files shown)


Cesium3DTiles/src/BingMapsRasterOverlay.cpp
```diff
@@ -235,7 +242,7 @@ namespace {
  * {
  *   "imageryProviders": [
  *     {
- *       "attribution": "� 2021 Microsoft Corporation",
+ *       "attribution": "� 2021 Microsoft Corporation",
```
kring, line 245: Was this change intentional?

Cesium3DTiles/src/BingMapsRasterOverlay.cpp
```diff
@@ -352,73 +439,15 @@ BingMapsRasterOverlay::createTileProvider(
...
-            if (urlTemplate.empty()) {
-              SPDLOG_LOGGER_ERROR(
-                  pLogger,
-                  "Bing Maps tile imageUrl is missing or empty.");
-              return nullptr;
-            }
+            SessionCacheItem& item = sessionCache.emplace_back();
+            item.metadataUrl = metadataUrl;
+            item.responseData.resize(responseSize);
+            std::memcpy(item.responseData.data(), responseBuffer, responseSize);
```
kring, line 448: I think these two lines can be replaced with:
```
item.responseData = pResponse->data();
```

Worst case, this will definitely work:

```
item.responseData = std::vector<std::byte>(pResponse->data().begin(), pResponse->data().end());
```

Cesium3DTiles/src/BingMapsRasterOverlay.cpp
```diff
@@ -29,6 +29,13 @@ struct CreditAndCoverageAreas {
   Cesium3DTiles::Credit credit;
   std::vector<CoverageArea> coverageAreas;
 };
+
+struct SessionCacheItem {
+  std::string metadataUrl;
+  std::vector<std::byte> responseData;
+};
+
+std::vector<SessionCacheItem> sessionCache;
```
kring, line 38: Why not use `std::unordered_map<std::string, std::vector<std::byte>>` instead?

kring (commented):
Looks good, I just have some small suggestions to clean up the code a little.

Cesium3DTiles/src/BingMapsRasterOverlay.cpp
nithinp7: That's a good point. I remember being told it's sometimes good to be semantically clear by naming the inner types instead of parameterizing unnamed types as in: `<std::string, std::vector<std::byte>>`. But I wasn't sure when that principle is applicable and when it's unnecessary, I'll change this to an unordered_map!

Cesium3DTiles/src/BingMapsRasterOverlay.cpp
nithinp7: I changed it back to the copyright sign but it still marks the line as changed even though I can't tell what the difference is. I even tried copy-pasting the old version back in, but git doesn't detect that anything's changed. Do you have any idea what might be going on?

nithinp7:
@kring I applied your review suggestions I think this is ready now. The only thing is the extraneous change in the one comment with the copyright sign, I can't tell why it's still marked as changed after I copy-pasted in the old comment.

Cesium3DTiles/src/BingMapsRasterOverlay.cpp
kring: Yeah that's true. Using `std::vector<SessionCacheItem>` is definitely much better than using `std::vector<std::pair<std::string, std::vector<std::byte>>>` for example. But since sessionCache is truly a map, it's certainly ok to use a map type here. The little we lose by being unable to name the key and value are more than made up for by the conceptual clarity of using an actual map, IMO.

kring:
Thanks @nithinp7!

PR #238, opened 2021-05-03
Remove redundant `hasRootObject` from writeJsonValue

Samulus:
This parmaeter is ultimately unnecessary. The caller should be responsible
for ensuring that the CesiumGltf::JsonWriter state is balanced before calling
writeJsonValue. This should simplify our internal API.


kring:
Thanks @Samulus!

PR #242, opened 2021-05-06
Fix memory leak in GltfReader.cpp

Samulus:
`pImage` isn't being freed after being copied. This fixes that leak: We're running into out of memory crashes without it :grimacing:


kring:
Oh wow, thanks for noticing this one!

kring:
This definitely deserves a mention in CHANGES.md. Can someone please add it? Also it would be slightly better to use `stbi_image_free` instead of `free` (even though they're equivalent).

baothientran:
whoops sorry for miss this, I will update it right away

PR #243, opened 2021-05-07
Match draco indices to gltf if indices accessor doesn't match with draco

baothientran:
The asset to reproduce the error can be found in this forum post https://community.cesium.com/t/cesium-ion-data-crashes-in-unreal-engine/13523


kring:
Thanks @baothientran.

PR #245, opened 2021-05-12
Fix development package crash when shut down

baothientran:
Explicitly set `RasterOverlayProvider` owned by `RasterOverlay` to be `nullptr` instead of letting `unique_ptr` destructor get called implicitly. If we let `unique_ptr()` destructor called implicitly, `_pTileProvider` will get destructed first, but it will never set to be `nullptr`. So when `_pPlaceholder` is destroyed, its `_pPlaceholder` tile and _tiles member destructor will retrieve _pTileProvider instead of _pPlaceholder and it crashes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@
 
 ##### Fixes :wrench:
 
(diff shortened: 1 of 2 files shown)


kring:
Thanks @baothientran! I'm going to include this in the hotfix release.

PR #252, opened 2021-05-20
Clipping and Raster Overlay Refactor

nithinp7:
Partially solves: https://github.com/CesiumGS/cesium-unreal/issues/47

The cesium-unreal side to this change: https://github.com/CesiumGS/cesium-unreal/pull/446

Significant Changes:
* A `RasterizedPolygonsOverlay` with multiple `CartographicSelection`s can now be added to the tileset and it will be turned into rasterized textures on the tile gltfs. Currently only works for tiles with bounding regions. This is most directly applicable for visually culling out selected regions with an opacity mask, but it can be used for any type of masking-based shader behavior (e.g., custom water masks).
* There is now only one texture generated per geometry tile per overlay layer. To do this, all needed textures within a geometry tile are blitted together into a single texture.
* `RasterOverlay`s can be given names, so now each geometry tile receives a single named texture corresponding to each overlay.


Cesium3DTiles/src/Tileset.cpp
```diff
@@ -1143,6 +1147,127 @@ static void markTileAndChildrenNonRendered(
...
+ * @brief Returns whether the tile is completely clipped out by a clipping
+ * polygon.
+ *
+ * @param boundingVolume The bounding volume of the tile.
+ * @param cartographicSelections The list of clipping selections to check.
+ * @return Whether the tile is completely clipped out by a clipping polygon.
+ */
+static bool isCompletelyClipped(
+    const BoundingVolume& boundingVolume,
+    const std::vector<CartographicSelection>& cartographicSelections) {
```
kring, line 1160: This function is expensive to run for every tile every frame, especially with a complicated polygon. We should be able to optimize it by adding quick accept/reject checks. For example, if a tile's bounding region is entirely disjoint with the polygon's bounding region, there's no need to check every triangle in the polygon.

kring (commented):
This looks good @nithinp7! I have some small comments below, but also a bigger one that will need some discussion...

I think there's a _lot_ of overlap between the RasteOverlay system and this new CartographicSelection system, and we should figure out how to combine them if we can.

Fundamentally, the RasterOverlay system is a way to map textures onto tilesets. It does so in a level-of-detail aware fashion; it's not just one texture, but a pyramid of textures that are mapped onto the tileset in a sensible way.

Currently, RasterOverlay tiles are images downloaded from the network, but that need not be true. Overlay tiles can be generated on demand, and we already have examples of that in CesiumJS (cesium-native's RasterOverlays are very similar to CesiumJS's ImageryProviders).

Currently, RasterOverlay tiles are arranged in a quadtree in some projection, but that need not be true either. To improve the flexibility here, we only need to make `RasterOverlayTileProvider::mapRasterTilesToGeometryTile` virtual and move the current implementation into a `QuadtreeRasterOverlayTileProvider` subclass.

Currently, RasterOverlays are alpha blended on top of each other, but that also need not be true. A Material on the Unreal side could blend two raster overlays together while using a third as a water mask and a fourth as an opacity mask. At that point the name `RasterOverlay` might be misleading. Maybe these are `RasterLayers`?

In this hypothetical design, the "culling polygon" might be separate from the raster layers entirely, but supplied from the same source on the Unreal side.

What do you think?

kring:
Specifying clipping as a raster overlay would also allow us to easily provide different ways of specifying the clipping polygon. For example, it could be pre-rasterized tiles from the network. Or it could be a polygon in a GeoJSON file.

kring:
I'm going to try to flesh out the idea above a bit more...

A raster overlay in Cesium for Unreal is an ActorComponent that is attached to a Tileset. It effectively attaches (multiple?) color textures to each geometry tile. Our standard overlay material then alpha blends these together in a top-down manner.

The idea presented here is to extend this system to mapping arbitrary, named textures to geometry tiles, which can then be used as inputs to a Material to be used however the user wishes. Raster overlays are polymorphic (the Strategy pattern if you're into that sort of thing), allowing the actual texture data to come from anywhere you like. It can be downloaded from the network or rasterized from a polygon, for example.

To get to this point, where the raster overlay system can replace the `CartographicSelection` in this PR, we need to do a bunch of things:

1. Add a `name` property to `RasterOverlay`. This will be the name of the texture parameter given to the Tileset's Material.
2. Create a new `QuadtreeRasterOverlayTileProvider` derived from `RasterOverlayTileProvider`. Move the implementation of `mapRasterTilesToGeometryTile` that is currently in `RasterOverlayTileProvider` to `QuadtreeRasterOverlayTileProvider` and make that method pure virtual in `RasterOverlayTileProvider`. Move all the quadtree and tiling scheme related stuff to the quadtree class as well. Change all existing overlay classes to derived from `QuadtreeRasterOverlayTileProvider` instead.
3. Introduce new `RasterOverlay`- and `CesiumRasterOverlay`- derived classes representing a rasterized polygon. Maybe `RasterizedPolygonOverlay`. This replaces the CartographicSelection. The `mapRasterTilesToGeometryTile` implementation in `RasterizedPolygonOverlay` will simply create a new RasterOverlayTile by rendering the polygon into the provided GlobeRectangle of the geometry tile.
4. Deal with mismatched projections leading to multiple overlay tiles being mapped to a single geometry tile. More on this below.

Steps 1-3 are pretty straightforward. Step 4 is more complicated.

When raster overlay tiles have an inherent projection/tiling (like Bing overlays, and unlike `RasterizedPolygonOverlay`), there is no guarantee that the raster tiles are aligned with the geometry tiles. That is why a single Cesium World Terrain tile has up to three Bing tiles attached to it. This is a hassle to deal with in the material: lots of duplication of nodes, and how many textures do we need for a particular overlay anyway? and what do we name them? It also burns through texture units, which are a precious resource.

One way I can think of to deal with this is to force the `RasterOverlayTileProvider` to create exactly one texture per geometry tile. In the case of something like Bing, we could do this by creating a texture large enough to hold all the Bing pixels mapped to the geometry tile and then blitting the pixels from multiple Bing tiles into it. Such a texture would have the same width (west-east) as a single Bing tile, but would have up to double the height (south-north). If we want to stick to power of two texture sizes, it would have exactly double the height. In the worst-case scenario, this doubles the amount of texture memory required by overlays. In reality it won't be that bad, though. And it will use far fewer texture units. I'd like to claim rendering performance will be better, but I'm not certain that's true. I don't think it will be any worse, at least.

We should measure the impact on our texture memory usage before we get too far. If it's deemed unacceptable, the only other alternative I can think of is to simply expose the multiple overlay tiles as multiple named textures (e.g. Bing0, Bing1, Bing2) and force the Material to deal with it.

Speaking of Materials, though, I think we're getting to the point - whether we implement the plan described here to stick with the CartographicSelection system - where we need to decide what to do about Materials. The fundamental question, I think, is this: is it acceptable to require users to build a Material for their tileset themselves?

One sensible answer to that is that a simple unadorned tileset should work out of the box with a default material that we supply. But as soon as you add overlays, you need to create a material for it. We can use the Quick Add to help with this. For example, adding Cesium World Terrain + Bing would also add a Material suitable for rendering the Bing overlays on the tileset. And we can provide template Materials or even UI help to create a material for clipping, adding water, etc.

We should look into [Layered Materials](https://docs.unrealengine.com/4.26/en-US/RenderingAndGraphics/Materials/LayeredMaterials/) and [Material Layers](https://docs.unrealengine.com/4.26/en-US/RenderingAndGraphics/Materials/MaterialLayers/) (yes those are different things) to see if they can help us make this easier for our users.

I'm sure that doesn't answer all the questions, but hopefully it's a start. Let me know what doesn't make sense!

kring:
> The reason I changed it to a standalone actor was so that multiple tilesets could use the same polygon selection, which is super useful.

I have to admit it's not very obvious to me what that's useful for. Can you elaborate? There might be a way to work it in if necessary.

kring:
Depending on how common we expect this to be, a reasonable workaround might be to copy/paste the component from one Tileset to another.

baothientran:
@nithinp7 after the refactoring, can you also add unit tests for this feature? the reason is that I want cesium-native to move forward with 1.0, so if there are any refactors along the way, it's really helpful that we don't break the existing features

nithinp7:
@kring I did a solid amount of refactoring to attempt consolidating the cartographic selection system with the raster overlay system. It's still not done yet, it needs a new type of raster mapped to geometry type for rasterized polygons. Can you review the changes from the last 4 commits in particular? The rasterized polygons aren't working through the overlay system just yet, but I am paranoid I may have messed things up in the existing overlay system. Nothing seems to be broken but I would like a second pair of eyes on this to reaffirm my sanity here. 

All these changes seemed to be necessary for the consolidation but I could totally be convinced to backtrack and approach things differently. 

@baothientran If you have time to look at the last 4 commits as well, please let me know if you agree with the architectural changes.

Cesium3DTiles/include/Cesium3DTiles/CartographicSelection.h
```diff
@@ -0,0 +1,97 @@
...
+    return this->_targetTextureName;
+  }
+
+  /**
+   * @brief Returns the longitude-latitude vertices in WGS84 that define the
+   * perimeter of the selected polygon.
+   *
+   * @return The perimeter vertices in WGS84 longitude-latitude.
+   */
+  constexpr const std::vector<glm::dvec2>& getVertices() const {
```
baothientran, line 53: I maybe wrong on this one, but I think `constexpr` doesn't have effect for this class because up to C++17, std::string and std::vector can't be constructed at compile time. C++20 will allow them, but I haven't had any luck to find the compilers that can init the string at compile time (something like `constexpr std::string s = "Hello World"`, this can only be done with `std::string_view` at the moment). So `CartographicSelection` can't be constructed at compile time

baothientran:
@nithinp7 I haven't looked deep into the code yet, but I like where this is headed. I will review carefully tomorrow and will test it out to make sure nothing break

Cesium3DTiles/include/Cesium3DTiles/RasterOverlayTileProvider.h
```diff
@@ -198,75 +186,25 @@ class CESIUM3DTILES_API RasterOverlayTileProvider {
...
    * @brief Returns the {@link RasterOverlayTile} with the given ID, creating it
    * if necessary.
    *
-   * @param id The {@link CesiumGeometry::QuadtreeTileID} of the tile to obtain.
+   * @param id The {@link TileID} of the tile to obtain.
    * @return The tile.
    */
   CesiumUtility::IntrusivePointer<RasterOverlayTile>
-  getTile(const CesiumGeometry::QuadtreeTileID& id);
+  getTile(const TileID& id, const CesiumGeometry::Rectangle& imageryRectangle);
```
kring, line 197: I don't think this method is needed in the base class. This implementation should be moved to the `QuadtreeRasterOverlayTileProvider`, and `RasterizedPolygonsOverlay` can just create a new tile every time without the awkward use of an incremented int as an ID.

kring: Same with `getTileWithoutCreating` below.

Cesium3DTiles/src/QuadtreeRasterMappedTo3DTile.cpp
```diff
@@ -0,0 +1,154 @@
...
+namespace Cesium3DTiles {
+
+QuadtreeRasterMappedTo3DTile::QuadtreeRasterMappedTo3DTile(
+    const CesiumUtility::IntrusivePointer<RasterOverlayTile>& pRasterTile,
+    const CesiumGeometry::Rectangle& textureCoordinateRectangle)
+    : RasterMappedTo3DTile(pRasterTile, textureCoordinateRectangle) {}
+
+RasterMappedTo3DTile::MoreDetailAvailable
+QuadtreeRasterMappedTo3DTile::update(Tile& tile) noexcept {
+  if (this->getState() == AttachmentState::Attached) {
```
kring, line 17: This method appears closely tied to a quadtree at first glance, but really the only reason it uses the quadtree is to be able to find the raster tile's parent, grandparent, etc. What if `RasterOverlayTile` simply had a pointer to its parent, the way that `Tile` does? Could we get rid of `QuadtreeRasterMappedTo3DTile`, make `RasterMappedTo3DTile` non-polymorphic, and no longer allocate it on the heap?

Cesium3DTiles/include/Cesium3DTiles/RasterOverlayTileProvider.h
```diff
@@ -278,18 +216,26 @@ class CESIUM3DTILES_API RasterOverlayTileProvider {
...
 
   /** @copydoc mapRasterTilesToGeometryTile */
-  void mapRasterTilesToGeometryTile(
+  virtual void mapRasterTilesToGeometryTile(
       const CesiumGeometry::Rectangle& geometryRectangle,
       double targetGeometricError,
-      std::vector<RasterMappedTo3DTile>& outputRasterTiles,
-      std::optional<size_t> outputIndex = std::nullopt);
+      std::vector<std::unique_ptr<RasterMappedTo3DTile>>& outputRasterTiles,
+      std::optional<size_t> outputIndex = std::nullopt) = 0;
```
kring, line 230: I don't think this one needs to be part of the `RasterOverlayTileProvider` interface, because it's never used externally.

Cesium3DTiles/src/RasterizedPolygonsOverlay.cpp
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  virtual bool
+  hasMoreDetailsAvailable(const TileID& /*tileID*/) const override {
+    return false; // always true or always false??
+  }
+
+  virtual CesiumAsync::Future<LoadedRasterOverlayImage>
+  loadTileImage(const TileID& tileID) override {
+    CesiumUtility::IntrusivePointer<RasterOverlayTile> pTile =
```
kring, line 161: `loadTileImage` will be called from the main/game thread. So you probably should use `thenInWorkerThread` to rasterize the polygon in a worker.

Cesium3DTiles/src/Tile.cpp
```diff
@@ -104,21 +107,138 @@ bool Tile::isRenderable() const noexcept {
...
+            return rasterTile != nullptr &&
+                   rasterTile->getReadyTile() != nullptr;
           });
     }
   }
 
   return false;
 }
 
+static void applyCustomMasks(
```
kring, line 120: Is this part of the old implementation and is going away? If not, I don't understand what purpose it serves. Which may just mean I need to look at it closer.

kring (commented):
Yeah, I like where this is headed, too! I just have a few suggestions and questions below.

Cesium3DTiles/src/Tile.cpp
nithinp7: Yup this part will go away!

Cesium3DTiles/src/TileID.cpp
```diff
@@ -40,4 +42,58 @@ TileIdUtilities::createTileIdString(const TileID& tileId) {
...
+/*static*/ std::optional<TileID>
+TileIdUtilities::getParentTileID(const TileID& tileId) {
+  struct Operation {
+
+    std::optional<TileID> operator()(const uint32_t& /*id*/) {
+      return std::nullopt;
+    }
+
+    std::optional<TileID> operator()(const std::string& /*url*/) {
+      return std::nullopt;
```
baothientran, line 54: this feels somehow awkward as one would expect this method to return the url of the parent

baothientran (commented):
I still need to review more, but this first round, this looks good to me. The only problem which is outside of this PR is the way we communicate between Unreal and Cesium-Native through the gltf extra system, which can accidentally introduce breaking changes if we change the name of the extra attributes.

Cesium3DTiles/src/RasterizedPolygonsOverlay.cpp
```diff
@@ -0,0 +1,222 @@
...
+  // calculations should be normalized to the first vertex)
+  for (const CartographicSelection& selection : cartographicSelections) {
+
+    if (selection.getTargetTextureName() != textureTargetName) {
+      continue;
+    }
+
+    const std::vector<glm::dvec2>& vertices = selection.getVertices();
+    const std::vector<uint32_t>& indices = selection.getIndices();
+    for (size_t triangle = 0; triangle < indices.size() / 3; ++triangle) {
```
baothientran, line 45: this looks like it tries to rasterize all the triangles in the CartographicSelection; however, there is a good chance that majority of the triangles are outside the bound of the tile and needs not to be rasterized.

Cesium3DTiles/src/RasterizedPolygonsOverlay.cpp
```diff
@@ -0,0 +1,222 @@
...
+      const glm::dvec2& c = vertices[indices[3 * triangle + 2]];
+
+      glm::dvec2 ab = b - a;
+      glm::dvec2 ab_perp(-ab.y, ab.x);
+      glm::dvec2 bc = c - b;
+      glm::dvec2 bc_perp(-bc.y, bc.x);
+      glm::dvec2 ca = a - c;
+      glm::dvec2 ca_perp(-ca.y, ca.x);
+
+      for (size_t j = 0; j < 256; ++j) {
```
baothientran, line 57: I think it doesn't need to run 256x256 for most of the triangles. Another optimization I can think of is to calculate the intersection between the scan line and and the triangle, then determine which the range of the pixels that need to be rasterized. You can also calculate the vertical range of the pixels that the triangle covers to rasterize instead of rasterizing all row pixels
```
                                              O
                                           O    O
     rasterize this pixel ->  O         O <- to this pixel
                                       OOOOOOO
```

Cesium3DTiles/src/RasterizedPolygonsOverlay.cpp
baothientran: Another thing is if you can determine that the tile is completely bound by a triangle, it can return early with white image instead of doing any rasterization

Cesium3DTiles/src/RasterizedPolygonsOverlay.cpp
```diff
@@ -0,0 +1,222 @@
...
+#include <memory>
+#include <string>
+
+namespace Cesium3DTiles {
+namespace {
+void rasterizePolygons(
+    CesiumGltf::ImageCesium& image,
+    const CesiumGeospatial::GlobeRectangle& rectangle,
+    const std::string& textureTargetName,
+    const std::vector<CartographicSelection>& cartographicSelections) {
```
baothientran, line 19: Since cartographicSelections is just a 2D polygon, I think maybe rasterize whole polygon is more efficient than triangles, but I'm not sure on that. But in case you need to try, the algorithm for 2D polygon rasterization is here http://alienryderflex.com/polygon_fill/

Cesium3DTiles/src/RasterizedPolygonsOverlay.cpp
baothientran: the 2D line and triangle intersection hopefully should be fast since the line is also a parallel line

nithinp7:
@baothientran @kring Can you guys have another look? Particularly at `RastersMappedTo3DTile`, it's still a bit crashy currently due to multithreading-related issues. Please let me know any architectural concerns as well. The latest commit of: https://github.com/CesiumGS/cesium-unreal/pull/446 can be used with this.

kring:
I've done some work on this in the https://github.com/CesiumGS/cesium-native/tree/clipping-2.5d-kring branch (will PR it back here soon). It's not too crashy (though it's not extensively tested), and I think the architecture is inching closer, but there are quite a few outstanding issues:

* [ ] A combined texture my have a rectangle that is slightly bigger than the geometry tile's bounding region, in order to avoid resampling image pixels or changing their size. So we need to account for that when sampling the texture (e.g. with a small translation)
* [ ] Handling missing / failed quadtree tiles. Currently if any quadtree tile fails, it causes the entire geometry tile to fail. Instead, we should fill just that one quadtree tile with an ancestor image. Unless _all_ the quadtree tiles fail, that is.
* [ ] Caching of quadtree tiles is questionable. Previously, raster tile lifetime was controlled by counting references in geometry tiles, which worked pretty well. But now quadtree tiles will be unloaded the moment no geometry tiles are actively loading them. Quadtree tiles probably need their own LRU cache.
* [ ] I haven't really tried the actual _clipping_ functionality in awhile 😬
* probably lots more

javagl:
I did a short review, _solely on the code level_, based on the https://github.com/CesiumGS/cesium-native/tree/clipping-2.5d-kring branch (this is not part of this PR yet - otherwise, I'd have left some of these points as inlined comments).

- I'll keep pointing to 
  https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#i23-keep-the-number-of-function-arguments-low
  until I notice that PRs are making things *better* in that regard.

- Can any of the coordinates be greater than 3*PI here?
  https://github.com/CesiumGS/cesium-native/compare/clipping-2.5d-kring#diff-0b071d4189d9f4f18995dffd854354ceec20eb082346feb7147d3c49ba22c180R34
  (I don't know that, because none of the functions there has doxygen comments)
  If so, use `while` or `%`, or best: call some `clamp` function...
  Such a `clamp` function could probably also be used at
  https://github.com/CesiumGS/cesium-native/compare/clipping-2.5d-kring#diff-0b071d4189d9f4f18995dffd854354ceec20eb082346feb7147d3c49ba22c180R76
  https://github.com/CesiumGS/cesium-native/compare/clipping-2.5d-kring#diff-0b071d4189d9f4f18995dffd854354ceec20eb082346feb7147d3c49ba22c180R89
  (When the same `// One-line-comment` appears several times, then one should consider this...)

- The functions like `triangulatePolygon` etc. should also be in an anonymous namespace, IMHO

- The whole "normalization" of the polygon should be pulled out into a function, MAYBE it could then be re-used at 
  https://github.com/CesiumGS/cesium-native/compare/clipping-2.5d-kring#diff-6ef66d1406761bdc178e04328c6c5a6dc5823bad9f49bf0964e9cb8c5ad3af5aR71
  (but am not sure, just started reading)

- I'm *very* much **not** a fan of `protected` for variables (list of reasons can be given upon request)
  https://github.com/CesiumGS/cesium-native/compare/clipping-2.5d-kring#diff-8672c91315c97e190a4730b7bf9f2c9f39360c86cf6e1545f75cdd251eaea016R351

- When I see code like 
  https://github.com/CesiumGS/cesium-native/compare/clipping-2.5d-kring#diff-27f8e75f6d33fc51120654b848fb218a08c58b7d76f2db72d2cde3a4c66d9912R529
  I sometimes apply a "code-Blink-Comparator" (see https://en.wikipedia.org/wiki/Blink_comparator ) :
  BlinkComparator

- I'm pretty sure that the "subset" should not be part of the signature. 
  https://github.com/CesiumGS/cesium-native/compare/clipping-2.5d-kring#diff-27f8e75f6d33fc51120654b848fb218a08c58b7d76f2db72d2cde3a4c66d9912R516
  Imagine a function like 
  ```
  void updateValue(float value, bool reallyUpdate) {
    if (reallyUpdate) this->value = value;
  }
  ```
  This is a function for blitting images, it should NOT know anything related to ~"missing ancestor tiles" or whatnot. How should this be covered in a unit test? Computing the right source rectangle is in the responsibility of the caller.

- The fact that the `vector projections` was consistently changed to `unordered_set projections` is a bit surprising. I think it could make sense to let it be a set when *collecting* them, but in order to have a predictable order when they are processed, I think the container should have a predictable iteration order, because the order (e.g. for blitting and such) *might* be relevant under certain conditions. (Despearately waiting for `std::ranges::common_range` AKA `java.lang.Iterable`...)

- I'm pretty sure this
  https://github.com/CesiumGS/cesium-native/compare/clipping-2.5d-kring#diff-27f8e75f6d33fc51120654b848fb218a08c58b7d76f2db72d2cde3a4c66d9912R118
  should just be `intersection = intersect(geometryRectangle, imageryRectangle);` (Isn't there already an `intersect` function? If not, it should be created...)
  
A more high-level comment about asynchronicity will become a "discussion" issue....

kring:
Thanks Marco, I've addressed some of that in https://github.com/CesiumGS/cesium-native/commit/d689e3811d7a77c96366c4127074484347c18bed. Some of it I can't quite tell what you're referring to without the links working.

> I'm pretty sure that the "subset" should not be part of the signature.

Assuming you're referring to `blitImage`, there's no escaping having a subset parameter. `sourceRectangle` defines the extent of the entire `source`. `sourceSubset` defines the part of the source that should be copied to the target. Without both rectangles, we couldn't determine which pixels to copy.

javagl:
Yes, sorry. I thought that it was possible to directly link to changes in a comparison view, but obviously, this only refers to actual PRs.

The points that seem not to be covered in the cleanup commit:

- The static functions in https://github.com/CesiumGS/cesium-native/blob/d689e3811d7a77c96366c4127074484347c18bed/Cesium3DTiles/src/CartographicSelection.cpp should be in an anonymous namespace, and the comment `// check if the difference crosses the antipole` appears three times (that's where I thought that it should be solved with some sort of `clamp` or `wrap` function).

- The same file contains the part that does the [polygon normalization](https://github.com/CesiumGS/cesium-native/blob/d689e3811d7a77c96366c4127074484347c18bed/Cesium3DTiles/src/CartographicSelection.cpp#L25), and I think that what is done there could(!) be so similar to what is done at https://github.com/CesiumGS/cesium-native/blob/0c6c1a7a5ba016964514166ca3346a23b02d7e82/Cesium3DTiles/src/RasterizedPolygonsOverlay.cpp#L70 that an own `normalizePolygon` method may make sense

- Regarding the `subset`: Yes, I'm talking about the last parameter of [`QuadtreeRasterOverlayTileProvider::blitImage`](https://github.com/CesiumGS/cesium-native/blob/3620b4d835b29446616ecbe725cebe65f57611e7/Cesium3DTiles/src/QuadtreeRasterOverlayTileProvider.cpp#L560). 
 The function is called as
  `blitImage(target, targetRectangle, source, sourceRectangle, sourceSubset);`
  and the first thing that it does is to compute
  `Rectangle sourceToCopy = sourceSubset.value_or(sourceRectangle);`
  I don't see a reason why it should not be possible to call it as
  `blitImage(target, targetRectangle, source, sourceRectangle);`
  and omit the last parameter by setting `sourceRectangle = sourceSubset.value_or(sourceRectangle)` (and the same for `target`).    
  But alternatively, you can try to add a doxygen comment for `subset`, I'm curious what this would say...

- The last comment iirc referred to https://github.com/CesiumGS/cesium-native/blob/d689e3811d7a77c96366c4127074484347c18bed/Cesium3DTiles/src/QuadtreeRasterOverlayTileProvider.cpp#L118 : It *does* call `intersect`, but ... if there is none, it computes ... ~"some other sort of intersection(?)". This should have a name.

kring:
Thanks. I haven't looked closely at the `CartographicSelection` or `RasterizedPolygonsOverlay` code yet, but will definitely consider your feedback on those when I do.

> and omit the last parameter by setting sourceRectangle = sourceSubset.value_or(sourceRectangle) (and the same for target).

The sourceRectangle is also passed to `computePixelRectangle`. It needs to be the original one that defines the total source image rectangle, not the subset. What you're proposing wouldn't work at all.

> it does call intersect, but ... if there is none, it computes ... ~"some other sort of intersection(?)". This should have a name.

It's stretching the base layer imagery over the entire globe, as the comment above it says. This was ported from CesiumJS but is a little incomplete (and hence the code is admittedly a little weird) at the moment because we only support one raster overlay.

javagl:
> It needs to be the original one that defines the total source image rectangle, not the subset. 

Maybe I misunderstood the documentation of `computePixelRectangle` and its parameters. But it appears to do two things at once: 

1. Computing the "relative amount" ot the total rectangle that is covered with the part rectangle
2. Mapping that "relative amount" to the actual image pixel rectangle

I'd probably separate that, as in 

1. A function like `computeRelativePositionIn` that computes the relative position of one rectangle inside another, and returns a rectangle that contains values in [(0,0), (1,1)] (aka texture coordinates)
2. A function like `mapToPixels` that receives the size of the size of the image, and returns the pixels that are covered with the "texture coordinates" of a given rectangle

because these could be useful, generic building blocks (maybe in the `Rectangle` class), so that the implementation of `computePixelRectangle` could just be
```
PixelRectangle computePixelRectangle(
    const ImageCesium& image,
    const Rectangle& totalRectangle,
    const Rectangle& partRectangle) {
    return partRectangle.computeRelativePositionIn(totalRectangle).mapToPixels(image.size);
}
```
But that's probably not worth changing for now. If the current solution works, then this is fine, and there are certainly more important tasks.

nithinp7:
I was investigating the cause of [this issue(https://github.com/CesiumGS/cesium-unreal/pull/608#issuecomment-908409205) (pointed out by @javagl). The only lead I have so far is that, when deleting a raster overlay, the `~RasterOverlayTile()` destructor does not get called for any raster tiles except for a single placeholder tile that has no tile provider. For comparison, in main my breakpoint in the destructor gets hit plenty of times for valid tiles, and so the freeRaster() is called as expected.

nithinp7:
@kring My understanding is that these changes are included in #326 and/or #325. So I'll close this without merge, please reopen if this is incorrect.

Issue #205, opened 2021-04-13
Consider to not throw invalid_argument exceptions


javagl:
Right now, there are few places where an `invalid_argument` exception can be thrown:

- In the `Plane` constructor, when the given normal is not normalized ( https://github.com/CesiumGS/cesium-native/blob/8a2957d32f433438b6b91ad37059c1ca0c8665c5/CesiumGeometry/src/Plane.cpp#L13 )
- In the `Ray` constructor, when the ray direction is not normalized ( https://github.com/CesiumGS/cesium-native/blob/8a2957d32f433438b6b91ad37059c1ca0c8665c5/CesiumGeometry/src/Ray.cpp#L13 )
- In the `EllipsoidTangentPlane` class, when the origin is near the center ( https://github.com/CesiumGS/cesium-native/blob/b824699d3dd9503a5943007247f683b6adc132d4/CesiumGeospatial/src/EllipsoidTangentPlane.cpp#L51 )

As much as I'm a fan of early, clear error checks and responses, we should consider handling this differently: Recent crashes in Unreal ( https://github.com/CesiumGS/cesium-unreal/issues/290 and https://github.com/CesiumGS/cesium-unreal/issues/336 ) have been caused by this, and these crashes are hard, painful, and non-recoverable. 

I'm not sure about the best solution for this. 

- Simple solutions (like filling stuff with `NaN` or so) are certainly not appropriate, becauses these will have the effect of "Nothing is rendered", **without** any clue of what's happening at all.
- Enforcing to wrap the creation of these objects into `try-catch` will be hard (and inconvenient)
- More strict, invasive solutions could be (brainstorming) to make the constructors of these classes `private`, and only allow them to be created with factory methods that return an `std::optional<Plane>`. We could also consider separate `createThrowing`/`createUncheckedWithNaN`/`createOptional` factory methods. 

I just did a websearch about what e.g. GLM does for a `normalize` call on zero-length vectors, but... surprisingly there is no obvious information for that (have to try it out, I guess). The point is that we should consider "following conventions" for such cases ... if there *are* conventions. Otherwise, it largely depends on the desired and expected behavior, and the crucial question of "What should happen when the precondition does not hold?".

javagl:
That `tryConstruct` essentially the `createOptional` that I mentioned, and I think that could be reasonable solution. For *some* cases, having a `constructUnchecked` that does 1. assume the input to be valid or 2. does not care about NaN and such may also be OK, but can be added at any point in time.

[...]

Issue #212, opened 2021-04-16
Disable exceptions by default


kring:
Some of our downstream projects (including Cesium for Unreal) ship with exceptions disabled. But cesium-native itself has them enabled. This is _kind of_ ok as long as a) cesium-native itself is built with exceptions enabled, and b) the exceptions never leave cesium-native code. But it's easy to get this wrong. For example, a try/catch within a method is fine, unless that method happens to be inline or a template. In which case an exceptions-disabled project that tries to use it will fail to compile.

A safer approach, given that we expect cesium-native to be used in environments where exceptions aren't available, would be disable exceptions in cesium-native entirely. That way any attempts to use them would be immediately flagged.

There are some places where this would be pretty awkward, though, which is why we haven't done it already. We'll also likely need to change how errors are reported through AsyncSystem, and may even need to replace Async++.

javagl:
This is at least related to (or maybe a generalization of) https://github.com/CesiumGS/cesium-native/issues/205 . Also related: https://github.com/CesiumGS/cesium-native/issues/76 (disabling exceptions was mentioned there, at least). 

If completely disabling exceptions (on a level of compiler flags) would make it impossible to even *use* certain third-parts libs, one would at least have to carefully and diligently examine each call to a third-party lib and wrap that call into some `{ try { } catch () } noexcept` version. The question of "What should happen *instead* of throwing?" must be examined on a case-by-case basis, and probably include simple solutions like "Just 'log' the error and do nothing" or "Return some `optional`" to cases where it might be tricky to come up with a solution that fits all application cases, like the "I expect a vector to be normalized"-assumption in the linked issue.

xuelongmu:
It's not urgent by any means, but I'm curious if there are updates on this. For an internal project at Cesium that uses Cesium for Unreal, exceptions have been enabled in order to get the project to build (currently we are creating a Development build in our travis CI). However, as exceptions are always off in Shipping builds (re: https://github.com/CesiumGS/cesium-unreal/issues/387#issuecomment-832670363), it would be great to eventually resolve this in the `cesium-native` repo.

baothientran:
This one is currently not in the work at the moment. Unreal plugin travis CI is built with shipping build enabled I think. For the crashing in the shipping build that is related to Sqlite (https://github.com/CesiumGS/cesium-unreal/issues/245), I believe the fix is already in the main branch currently

PR #254, opened 2021-05-25
Do not throw exceptions from geometry classes

javagl:
Fixes https://github.com/CesiumGS/cesium-native/issues/205

As mentioned in the CHANGES.md:

> The `CesiumGeometry::Plane` and `CesiumGeometry::Ray` classes now offer factory
  methods to avoid throwing an `invalid_argument` exception in case of errors.
  They offer a `createUnchecked` method that assumes valid input, a `createOptional`
  method that returns an `std::optional` in case of errors, and a `createThrowing`
  method that throws an exception in case of errors. This has the effect of other
  classes also no longer throwing: The internal usage of these classes has been
  changed to generally use the `createUnchecked` method, because the preconditions
  for the exception to be thrown could not sensibly be propagated to the user.
  This means that the `CullingVolume` and `EllipsoidTangentPlane` do no longer
  throw for invalid inputs, but may return results that contain `NaN` values.

(Somewhat surprisingly, this does not seem to affect `cesium-unreal` on an API level: These classes had apparently not be used *directly*)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,17 @@
 
 ##### Breaking Changes :mega:
 
+- The `CesiumGeometry::Plane` and `CesiumGeometry::Ray` classes now offer factory
+  methods to avoid throwing an `invalid_argument` exception in case of errors.
+  They offer a `createUnchecked` method that assumes valid input, a `createOptional`
+  method that returns an `std::optional` in case of errors, and a `createThrowing`
+  method that throws an exception in case of errors. This has the effect of other
+  classes also no longer throwing: The internal usage of these classes has been
+  changed to generally use the `createUnchecked` method, because the preconditions
+  for the exception to be thrown could not sensibly be propagated to the user.
+  This means that the `CullingVolume` and `EllipsoidTangentPlane` do no longer
+  throw for invalid inputs, but may return results that contain `NaN` values.
+
 ##### Additions :tada:
 
 - Added `Cesium3DTiles::TileIdUtilities` with a `createTileIdString` function to create logging/debugging strings for `TileID` objects.
--- a/Cesium3DTiles/src/ViewState.cpp
+++ b/Cesium3DTiles/src/ViewState.cpp
@@ -43,7 +43,7 @@ ViewState::ViewState(
       _verticalFieldOfView(verticalFieldOfView),
       _sseDenominator(2.0 * glm::tan(0.5 * verticalFieldOfView)),
       _positionCartographic(positionCartographic),
-      _cullingVolume(createCullingVolume(
+      _cullingVolume(CullingVolume::createUnchecked(
           position,
           direction,
           up,
--- a/CesiumGeometry/include/CesiumGeometry/CullingVolume.h
+++ b/CesiumGeometry/include/CesiumGeometry/CullingVolume.h
@@ -35,32 +35,35 @@ struct CullingVolume final {
    *
    * Defaults to (0,0,1), with a distance of 0.
    */
-  CesiumGeometry::Plane leftPlane = CesiumGeometry::Plane::createUnchecked(glm::dvec3(0.0, 0.0, 1.0), 0.0);
+  CesiumGeometry::Plane leftPlane =
+      CesiumGeometry::Plane::createUnchecked(glm::dvec3(0.0, 0.0, 1.0), 0.0);
 
   /**
    * @brief The right plane of the culling volume.
    *
    * Defaults to (0,0,1), with a distance of 0.
    */
-  CesiumGeometry::Plane rightPlane = CesiumGeometry::Plane::createUnchecked(glm::dvec3(0.0, 0.0, 1.0), 0.0);
+  CesiumGeometry::Plane rightPlane =
+      CesiumGeometry::Plane::createUnchecked(glm::dvec3(0.0, 0.0, 1.0), 0.0);
 
   /**
    * @brief The top plane of the culling volume.
    *
    * Defaults to (0,0,1), with a distance of 0.
    */
-  CesiumGeometry::Plane topPlane = CesiumGeometry::Plane::createUnchecked(glm::dvec3(0.0, 0.0, 1.0), 0.0);
+  CesiumGeometry::Plane topPlane =
+      CesiumGeometry::Plane::createUnchecked(glm::dvec3(0.0, 0.0, 1.0), 0.0);
 
   /**
    * @brief The bottom plane of the culling volume.
    *
    * Defaults to (0,0,1), with a distance of 0.
    */
-  CesiumGeometry::Plane bottomPlane = CesiumGeometry::Plane::createUnchecked(glm::dvec3(0.0, 0.0, 1.0), 0.0);
+  CesiumGeometry::Plane bottomPlane =
+      CesiumGeometry::Plane::createUnchecked(glm::dvec3(0.0, 0.0, 1.0), 0.0);
 
 private:
   CullingVolume();
-
 };
 
 } // namespace Cesium3DTiles
--- a/CesiumGeometry/include/CesiumGeometry/CullingVolume.h
+++ b/CesiumGeometry/include/CesiumGeometry/CullingVolume.h
@@ -14,48 +14,53 @@ namespace Cesium3DTiles {
 struct CullingVolume final {
 
   /**
+   * @brief Creates a {@link CullingVolume} for a perspective frustum.
+   *
+   * @param position The eye position
+   * @param direction The viewing direction
+   * @param up The up-vector of the frustum
+   * @param fovx The horizontal Field-Of-View angle, in radians
+   * @param fovy The vertical Field-Of-View angle, in radians
+   * @return The {@link CullingVolume}
+   */
+  static CullingVolume createUnchecked(
+      const glm::dvec3& position,
+      const glm::dvec3& direction,
+      const glm::dvec3& up,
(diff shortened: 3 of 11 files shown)


javagl:
(An aside: Before merging, we should consider removing the `createOptional` and `createThrowing` methods: They are **not** used right now, and could be introduced later if they are required. The current state is mainly intended to show *how* they could be introduced later, but in a somewhat hopefully consistent manner)

kring:
Thanks @javagl! The approach of Unchecked/Optional/Throwing makes sense to me. But methods like those should be a last resort when we can't avoid the issue entirely by changing the interface slightly. So with that in mind, let's talk about the three thrown exceptions that you've eliminated in this PR:

## EllipsoidTangentPlane::computeEastNorthUpToFixedFrame

This method would throw when the position is too close to the center of the ellipsoid. You've eliminated it by silently returning an identity matrix. That seems very reasonable to me, and is a good example of what I'm talking about above. When there's something sensible we can do to make an exceptional case non-exceptional, we should do it. The only thing I might suggest we change is to make the origin the position translation part of the matrix. The rotation/scale would still be identity.

## Plane constructor

It would previously throw if given a non-normalized normal. The old code looks like this:

```cpp
  //>>includeStart('debug', pragmas.debug);
  if (!Math::equalsEpsilon(glm::length(normal), 1.0, Math::EPSILON6)) {
    throw std::invalid_argument("normal must be normalized.");
  }
  //>>includeEnd('debug');
```

Those weird comments came from CesiumJS, and are the markers for code that gets removed in release builds. Therefore, a natural translation of that CesiumJS code to C++ is to use an assertion:

```cpp
assert(Math::equalsEpsilon(glm::length(normal), 1.0, Math::EPSILON6));
```

With that, we don't need exceptions or unchecked/optional/throwing methods. It's simply part of the contract that the normal must be normalized, and in debug builds we verify that this is true.

Why didn't I use an assertion to begin with? Uhh honestly this was some of the first code I wrote in cesium-native and I hadn't written C++ code in a little while so I had sort of forgotten that assertions exist. 😆

## Ray constructor

This case is exactly the same as the Plane constructor. It should be an assertion.

javagl:
I just noticed that I didn't update the unit tests, but in hindsight, that's a good thing - I could have burnt some more time with that, considering that the suggested solution for the issue that caused this PR is now "just use `assert` instead of `throw`". Using `assert` *might* be some sort of "Best Practice" in C++, but I think it raises questions ...

- Do we really want to _always_ crash in these cases in Debug mode? 
- Do we really want to _never_ be informed about these cases in Release mode?
- What about the timing?

The latter refers to the problem that we don't exactly know where these cases appear. From ~"subjecive evidence", these crashes usually happened in some `Actor::Tick` call, that was apparently called at a point in time where some `dmat4` member had not (yet) been initialized properly (because some other function, `OnConstruction` or whatnot, had not been called yet). The complex interaction of things like `bCanEverTick`  and (particularly) `ShouldTickIfViewportsOnly` make it hard to analyze things like this *deterministically*. 

This is related to the difficulty of "propagating" statements about error conditions (aka preconditions) up the call hierarchy. This may be doable for specific, low-level cases like `computeEastNorthUpToFixedFrame`: *"Throws if position is too close to the center of the ellipsoid"*. How close? Where is this method called, and how will this condition be phrased at the call site?

    /* throws if ... some position is somewhere ... we don't know for sure ... better don't call this */
    void MyHighLevelBlueprintClass::update() {
        EllipsoidTangentPlane::computeEastNorthUpToFixedFrame(this->computePosition(), this->ellipsoid);
    }    

The general approach of "ignoring" invalid values may *appear* to be an easy solution: Sure, it will not crash any more. But if it then just does not render anything, and one has to figure out the reason for that, *without any clue what the reason might be* (only to find out that this is caused by some `NaN`, deeply hidden in some property), can literally involve *hours* of debug-stepping through Unreal Engine code. Having some `logVeryVerbose("Hey, the length should be x but is 0.0!")` could help to prevent that.

---

However: If this is *explicitly* reqested, I can create another PR, just replacing these checks with `assert`.

kring:
> Do we really want to always crash in these cases in Debug mode?

Yes. Calling these methods with a normal that is not normalized is improper use of the methods. We should fail loudly. If you're in a situation where you're not sure whether a given normal will be normalized or not, the proper thing to do is to check before calling the method.

> Do we really want to never be informed about these cases in Release mode?

More like, by the time we're in a release build, we're assuming the library is being used properly and we don't need to spend cycles at runtime verifying this. Is it a perfect assumption? Of course not. I believe it's possible to configure a build to check assertions in a release build, if necessary, but it's rarely necessary.

> Having some logVeryVerbose("Hey, the length should be x but is 0.0!") could help to prevent that.

This is exactly the use-case that assertions are intended for: `assert(length == x)`. They're good for verifying preconditions that _should_ always be true but that the compiler / type system can't guarantee to always be true. Liberal use of assertions helps flag problems (violations of assumptions) early, rather than, say, letting a NaN propagate around until it causes an obvious problem later, by which point it's hard to tell where it came from.

> However: If this is explicitly reqested, I can create another PR, just replacing these checks with assert.

Just to be clear, this PR currently ignores these conditions in Ray and Plane entirely. What I'm proposing is that we instead assert the expected condition. I think what you've done with `computeEastNorthUpToFixedFrame` in this PR is correct. I wouldn't assert there.

javagl:
I'm wondering about the "flow of events" for this. We did not encounter the reported crashes at development time (so we didn't trigger the exception, and thus, wouldn't have triggered the `assert`). In the wild, users either did something unexpected, or something unexpected happened for reasons beyond our knowledge (timing reasons or whatnot). In these cases, the exception was thrown. One problem was that it crashed the whole application, while some `log("Whoopsie in plane!")` would have been sufficient. But thanks to the stack trace, we knew **what** was wrong, exactly. **Iff** this had been an `assert` only, then ... people would have reported things like: "Hey, my stuff is not rendered" (or whatever happens when a the `NaN` spead through the UE). We wouldn't know what's wrong. We could only *hope* to be able to reproduce the problem in a debug build. That sounds like a potential debugging-time-sinkhole.

Long story short: I don't know the "silver bullet" here, but will doubtfully replace the `throw`s with `assert` in another PR.

kring:
I don't disagree with you @javagl. But surely an assertion is better than the `createUnchecked` that was originally in this PR, right?

FWIW, I think the biggest problem was that the exception was causing immediate program termination in release builds due to the lack of exception support. The NaNs, while silent, may or may not be a serious problem, depending on how transient the condition that led to them is.

Anyway, exceptions are not an option in UE release builds. Even logging is questionable. So our options are extremely limited for this scenario where something goes wrong in a release build and we end up passing an unnormalized normal into this code.

javagl:
> But surely an assertion is better than the `createUnchecked`

The way that I'm thinking about certain things revolves about the ability to _emulate_ one approach with the other. It's easier (and less of a "breaking change") to add an `assert` in a `createUnchecked` method. But having an `assert` in a constructor basides raising the questions that I mentioned) is a binding contract that can never be changed, and hardly even be modified, e.g. with adding a log statement. 

---

<sup>(I considered to move the following part to https://github.com/CesiumGS/cesium-native/issues/212 , but it focusses on `CesiumGeometry` in some way - there number of libraries in this repo is becoming too large...)</sup>

I tried to figure out some "best practices here", but ... that seems to be hard. I was surprised to see https://github.com/g-truc/glm/blob/master/manual.md#-710-why-some-glm-functions-can-crash-because-of-division-by-zero , for example: From my tests, it does *not* crash, but just create NaN. (Maybe whether or not it throws can be defined with some `#define`, but if it is, it is not properly documented). 

Other examples of ~"linear algebra"/~"geometry" libraries and their error handling could be 
http://eigen.tuxfamily.org/dox/TopicAssertions.html or https://doc.cgal.org/latest/Manual/devman_checks.html , but deriving a silver bullet from that may be hard as well: The latter plainly forbids `assert`, and they generally define own macros for checks with different granularities. Although macros are ... yuck... a relict from the 1970's ... , we might still consider this, if we make sure that they are cleanly defined, only used internally at a few critical places.

kring:
> It's easier (and less of a "breaking change") to add an assert in a createUnchecked method. But having an assert in a constructor (besides raising the questions that I mentioned) is a binding contract that can never be changed, and hardly even be modified, e.g. with adding a log statement.

I don't think I follow you here. We're not concerned with breaking changes in cesium-native yet (other than we should try to document them). But even if we were, why is changing a constructor inherently more breaking than changing some other method?

javagl:
> We're not concerned with breaking changes in cesium-native yet

Every time something changes, the calling code has to be reviewed, adapted, and tested extensively, in order to (only) decrease the likelihood of introducing a bug - even if it is "only" used in `cesium-native` itself, or in `cesium-unreal`. The "less breaking" refers to the fact that whatever the constructor does, it will always happen. If we want to offer anything that behaves differently (like a `createThrowing` method), then I think this will be much harder. (But maybe it's only a "technical inconvenience": it's tremendously hard to figure out where a constructor is called... except for ~"making it `private` and look through the compile errors" - instead of doing a full-text search for `Plane::createUnchecked`).

But I think that the general questions that are discussed here in this PR should be discussed in a broader sense (and led to a conclusion, as in a paradigm that is followed in `cesium-native` and preferably *all* other Cesium C++ libraries), so this should probably be done in https://github.com/CesiumGS/cesium-native/issues/212 (and in any case, this PR is obsolete, so I'll close it for now).

Issue #251, opened 2021-05-20
Build failure x64/MSVC: Warning C4267 size_t to Uint  httplib


tfiner:
`C:\Work\lab\cesium-native\extern\cpp-httplib\httplib.h(2555): warning C4267: '=': conversion from 'size_t' to 'uInt', possible loss of data [C:\Work\lab\cesium-native\build\CesiumIonClient\CesiumIonClient.vcxproj]`

Followed by 
`C:\Work\lab\cesium-native\extern\cpp-httplib\httplib.h(2555): error C2220: warning treated as error - no 'object' file generated [C:\Work\lab\cesium-native\build\CesiumIonClient\CesiumIonClient.vcxproj]`

The solution is simple for x64, static_cast size_t to uInt.  Although, if anyone ever compresses more than 4gb, you'll probably want to revisit this by either using a safe_cast that throws on overflow, or breaking the data up into chunks.

I recommend turning warnings as errors off until the code is more stable and you have all your supported compilers / platforms being used by multiple people to catch problems like this.

kring:
@tfiner VS2017 on Windows x64 is our primary compilation target for cesium-native, and we haven't seen that warning. Are you building cesium-native has part of another project that overrides the compilation settings perhaps? Generally, third-party headers like this should be considered "system headers" and get less strict warning treatment.

kring:
Here's the log of our CI system building the latest main (as of this writing) on Windows x64 with VS2017:
https://travis-ci.com/github/CesiumGS/cesium-native/jobs/506616305

It's also built with VS2019 and for Clang and GCC on Linux:
https://github.com/CesiumGS/cesium-native/runs/2629831808

tfiner:
I'm actually building cesium-native by itself with a fairly up to date VS 2019.  A couple of caveats:
1. That warning is legitimate: if you assign a 32 bit unsigned int with a size_t on x64, it is truncating the integer.  Now, is it a problem really?  Probably not.  
2. I didn't upgrade the toolset. My understanding is that 2019 uses 2017 under the hood when you don't upgrade the toolset.  But.. this is probably the reason for my seeing the warning that you guys aren't.

I agree with you about third party headers, they should be looser and should probably have warnings as errors turned off.  

Otherwise, you'll run into more 3rd party libraries that prevent builds like this one did.  

[...]

PR #255, opened 2021-05-27
Cmake fix for cpp-httplib

nithinp7:
Possibly fixes: #251. 

Tells `cpp-httplib` not to look for zlib or openssl. Quoting @baothientran for what the issue may have been:
>hm good chance is your dev environment have zlib and openssl installed and httplib link against them when building. But those libraries are not copied in to unreal third party and it didn't know how to link against those symbol 

I'm not going to pretend I understand cmake or what exactly httplib, zlib, and openssl do, but this change finally solved a very painful issue I was stuck on for the last two days :). The issue does not happen in everybody's dev environment, so I think Bao's idea makes a lot of sense. 

@jtorresfabra @kring

### DIFF
--- a/extern/CMakeLists.txt
+++ b/extern/CMakeLists.txt
@@ -33,6 +33,8 @@ endif()
 add_subdirectory(modp_b64)
 
 set(HTTPLIB_USE_BROTLI_IF_AVAILABLE OFF CACHE BOOL "Don't use Brotli")
+set(HTTPLIB_USE_ZLIB_IF_AVAILABLE OFF CACHE BOOL "Don't use Zlib") 
+set(HTTPLIB_USE_OPENSSL_IF_AVAILABLE OFF CACHE BOOL "Don't use OpenSSL")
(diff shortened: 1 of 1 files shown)


shehzan10:
@nithinp7 I had run into the OpenSSL one with Android. We added the same option to the toolchain file. This looks like the right fix to me

https://github.com/CesiumGS/cesium-unreal/blob/android-platform/extern/unreal-android-toolchain.cmake#L15.

kring:
We're using cpp-httplib as a tiny web server we can run on localhost as part of the OAuth handshake with Cesium ion. Because we're only running on localhost, we don't need https support, so we don't need OpenSSL. We also don't particularly care whether the responses returned from our server are compressed, so we don't need zlib either.

If we ever start using cpp-httplib's HTTP _client_ support, both of these will be much more important. And we'll probably need to include zlib and openssl properly. This is pretty unlikely, though, considering that cpp-httplib requests are blocking.

Long story short, this looks good!

PR #256, opened 2021-05-28
Add an option to prefix all sqlite3 symbols with `cesium_`

kring:
This helps us avoid a conflict with other sqlite implementations, notably the limited one built into Unreal Engine.

The way it works is by, at configure time, reading sqlite3.c and sqlite3.h, doing a search/replace of `sqlite3` with `cesium_sqlite3`, and writing the output to the build directory. The rewritten files are then added as the sqlite3 target's source and include files. They're written using `file(GENERATE OUTPUT ...)` so they won't be rewritten (and thus recompiled) unless they actually change.

Now in our code, we wrap all use of SQLite symbols in a macro, `CESIUM_SQLITE`, defined in `cesium-sqlite3.h`, which prefixes the symbol with `cesium_` only if the build is configured to rename the sqlite symbols. The end result is a configure-time selection of whether to privatize the sqlite3 symbols or not.

I also refactored `SqlCache` slightly to use the pimpl idiom to avoid exposing sqlite types in a public header file. This way we don't expose the renaming (or lack of renaming) to users of cesium-native.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@
 
 - Added `Cesium3DTiles::TileIdUtilities` with a `createTileIdString` function to create logging/debugging strings for `TileID` objects.
 - Accessing the same Bing Maps layer multiple times in a single application run now reuses the same Bing Maps session instead of starting a new one each time.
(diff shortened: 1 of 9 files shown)


kring:
I'm going to merge this once CI is happy, because (I just realized) cesium-unreal already uses it as of when I merged CesiumGS/cesium-unreal#443.

PR #262, opened 2021-06-09
feat: CesiumGeometry::clipTriangleAtAxisAlignedThreshold add unit test

zy6p:
Add catch test for CesiumGeometry::clipTriangleAtAxisAlignedThreshold. And test cases refer to [https://github.com/CesiumGS/cesium/blob/master/Specs/Core/Intersections2DSpec.js](https://github.com/CesiumGS/cesium/blob/master/Specs/Core/Intersections2DSpec.js).

 - [x] clang format.
 - [x] catch test pass.


kring:
Love it, thanks @zy6p!

PR #267, opened 2021-06-22
Load tracing and performance optimizations

kring:
## Tracing Framework

In `CesiumUtility/Tracing.h`. This originally came from asset-pipeline, was ported to cesium-native by Bao, and then I extended it to support tracing async processes across threads. To enable it, add `-DCESIUM_TRACING_ENABLED=ON` to your CMake configure CLI. You also need to uncomment the line that adds `CESIUM_TRACING_ENABLED` to `PublicDefinitions` in `CesiumRuntime.build.cs` in Cesium for Unreal.

Once you've done all that, run e.g. Cesium for Unreal, shut it down, and look for `cesium-trace-[number].json` in `C:\Program Files\Epic Games\UE_4.26\Engine\Binaries\Win64` or equivalent. Visit `chrome://tracing` in Chrome and drag/drop that JSON file onto it. It looks something like this:

image

You zoom and pan with the W/A/S/D keys. Choose the "Selection" mode (it looks like an arrow) and drag a box around a part of the trace you're interested in. It will show all the events in that time period. Click the Occurrences column twice to show the most frequently occurring events at the top.

image

Then you can click on one, click the "Duration" radio button, and see a nice chart of the distribution of times taken by this event:

image

## AsyncSystem improvements

Separated the giant AsyncSystem.h into separate source files. Most of the code is "private" and ideally wouldn't be in a public header, but it needs to be because of how templates work.

Added `Future<T>::thenImmediately` to run a continuation immediately when the future resolves, in whatever thread does the resolving. For really quick tasks that you don't care where they run.

Changed `Future<T>::thenInWorkerThread` and `Future<T>::thenInMainThread` to run immediately in the _same thread_ if the previous continuation finishes in the correct thread (worker or main). This saves some overhead of queuing and task dispatching.

Add `AsyncSystem::createThreadPool` and `Future::thenInThreadPool` to dispatch continuations to a user-managed thread pool rather than main thread or the ITaskProcessor interface / UE task system.

## Created a dedicated thread for the SQLite cache

`CachingAssetAccessor` now uses the `createThreadPool` and `threadInThreadPool` methods described above to do all cache database access in a single designated thread (it's a thread pool with one thread). This is a significant optimization because when multiple threads share a single SQLite connection (as we were doing), all access gets serialized using a mutex. So we'd end up dispatching work to UE tasks only for them to sit there blocked while another thread read from or wrote to the disk cache. Having one thread dedicated to reading/writing the cache DB frees up UE tasks threads for more important work than waiting around for the disk.

## To do

* [x] I've seen crashes at shutdown, seemingly related to the lifetime of the queue that holds continuations for the main thread. Need to investigate.


Cesium3DTiles/src/RasterOverlayTileProvider.cpp
```diff
@@ -614,7 +626,8 @@ void RasterOverlayTileProvider::doLoad(
   // Don't let this tile be destroyed while it's loading.
   tile.setState(RasterOverlayTile::LoadState::Loading);
 
-  this->beginTileLoad(tile, isThrottledLoad);
+  [[maybe_unused]] int64_t loadID = this->beginTileLoad(tile, isThrottledLoad);
+  CESIUM_TRACE_ASYNC_ENLIST(loadID);
```
baothientran, line 630: I'm curious what is the purpose of this macro? It looks like it's not needed since we don't use CESIUM_TRACE_BEGIN and CESIUM_TRACE_END in this class

baothientran (commented):
@kring For the tracing part, it looks good to me. I tried it on the Unreal side and everything is rendered faster than before. I have some questions about the tracing system and some suggested changes below. I will look more into the new async system tomorrow.

Cesium3DTiles/src/RasterOverlayTileProvider.cpp
```diff
@@ -649,16 +662,54 @@ void RasterOverlayTileProvider::doLoad(
...
+  int64_t loaderID = -1;
+
   ++this->_totalTilesCurrentlyLoading;
   if (isThrottledLoad) {
+    size_t maxLoads = size_t(std::max(
+        this->getOwner().getOptions().maximumSimultaneousTileLoads,
+        0));
+    if (this->_tilesBeingLoaded.size() != maxLoads) {
+      this->_tilesBeingLoaded.resize(maxLoads, nullptr);
+      this->_loadingIDs.resize(maxLoads, 0);
```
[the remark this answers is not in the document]
kring: Yeah this is messy, let me clean it up.

Cesium3DTiles/src/Tileset.cpp
```diff
@@ -96,7 +106,8 @@ Tileset::Tileset(
             ionAssetID,
             e.what());
         this->notifyTileDoneLoading(nullptr);
-      });
+      })
+      .thenImmediately([]() { CESIUM_TRACE_END("Tileset from ion startup"); });
```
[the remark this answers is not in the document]
kring: I think the `thenImmediately` should be fairly cheap, but you're right, we might as well eliminate it.

I was thinking of that `thenImmediately` as a sort of `finally`. Technically, the continuation in `thenInMainThread` and the one in `catchInMainThread` can _both_ be executed, and I was kinda sorta worried that if that happened the tracing would get corrupted if we put the END in both places. We can't really ever execute both ENDs, though, if the END is the last line in each. Unless the END itself throws, but then we have bigger problems.

Cesium3DTiles/src/RasterOverlayTileProvider.cpp
kring: Does the documentation on it help at all? It says "this scope in this thread is part of an async process with this ID". That async process ID is then used for any non-ID TRACE calls, as well inherited by any work dispatched through the AsyncSystem.

CesiumAsync/include/CesiumAsync/Impl/AsyncSystemSchedulers.h
```diff
@@ -0,0 +1,78 @@
...
+namespace Impl {
+// Begin omitting doxgen warnings for Impl namespace
+//! @cond Doxygen_Suppress
+
+class AsyncSystemSchedulers;
+
+class ImmediatelyInMainThreadScheduler {
+public:
+  ImmediatelyInMainThreadScheduler(AsyncSystemSchedulers* pSchedulers)
+      : _pSchedulers(pSchedulers) {}
```
baothientran, line 18: since `ImmediatelyInMainThreadScheduler` uses `async::fifo_scheduler` in its method, maybe we should have this class own the `async::fifo_scheduler` directly and remove `AsyncSystemSchedulers` from it. That way, we won't have dependency between it and `AsyncSystemSchedulers`

CesiumAsync/include/CesiumAsync/Impl/AsyncSystemSchedulers.h
```diff
@@ -0,0 +1,78 @@
...
+  WorkerThreadScheduler(AsyncSystemSchedulers* pSchedulers)
+      : _pSchedulers(pSchedulers) {}
+
+  void schedule(async::task_run_handle t);
+
+private:
+  AsyncSystemSchedulers* _pSchedulers;
+};
+
+class ImmediatelyInWorkerThreadScheduler {
```
baothientran, line 37: Since `ImmediatelyInWorkerThreadScheduler` uses `WorkerThreadScheduler`, which in turn uses `ITaskProcessor` in its implementation, I think we can consolidate them together in one class. Maybe `ImmediatelyInWorkerThreadScheduler` will own `ITaskProcessor` directly and use the processor in its `schedule()` method. That way, we can remove the cyclic dependency between `AsyncSystemSchedulers` and  `ImmediatelyInWorkerThreadScheduler`. Also the WorkerThreadScheduler may not be needed.

With the change, `AsyncSystemScheduler` will still own `ImmediatelyInWorkerThreadsScheduler` and `ImmediatelyInMainThreadsScheduler`, but it won't own `ITaskProcessor` and `async::fifo_scheduler`

CesiumAsync/include/CesiumAsync/Impl/ContinuationFutureType.h
```diff
@@ -0,0 +1,26 @@
...
+template <typename T> class Future;
+
+namespace Impl {
+// Begin omitting doxgen warnings for Impl namespace
+//! @cond Doxygen_Suppress
+
+template <typename Func, typename T> struct ContinuationFutureType {
+  typedef Future<typename RemoveFuture<
+      typename ContinuationReturnType<Func, T>::type>::type>
+      type;
```
baothientran, line 17: slightly nickpick from me, but `using` keyword is more readable in this case, and since we use it a little below, it will be consistent

baothientran (commented):
I still need to look more into the async system in the next round, but overall everything looks good to me in this second round. I have some suggestions below.

Cesium3DTiles/src/Tileset.cpp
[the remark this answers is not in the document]
kring: Oh I see why I did this. The `thenInMainThread` in this case returns another Future. We don't want to do the END until _that_ Future resolves as well. The `thenImmediately` is the cleanest way to do that that I can see.

Cesium3DTiles/src/RasterOverlayTileProvider.cpp
```diff
@@ -649,16 +662,54 @@ void RasterOverlayTileProvider::doLoad(
...
+
+      for (size_t i = 0; i < this->_loadingIDs.size(); ++i) {
+        int64_t id = CESIUM_TRACE_ALLOCATE_ASYNC_ID();
+        this->_loadingIDs[i] = id;
+        CESIUM_TRACE_BEGIN_ID(
+            ("Overlay Loading Slot " + std::to_string(id)).c_str(),
+            id);
+      }
```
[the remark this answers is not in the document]
kring: Good idea. It ended up being a lot more work than I expected, but it's done now.

CesiumAsync/src/ThreadPool.cpp
```diff
@@ -0,0 +1,25 @@
+#include "CesiumAsync/ThreadPool.h"
+
+using namespace CesiumAsync;
+
+/*static*/ thread_local std::vector<ThreadPool::Scheduler*>
+    ThreadPool::schedulersInThreadPoolThread;
```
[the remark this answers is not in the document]
kring: You're right that the code to add the scheduler to the vector was missing! Fixed now.

I don't think your proposed solution would have worked. If I'm understanding you right, that would check if we're in the thread that created the thread pool. We need to check if we're in a thread created _by_ the thread pool.

CesiumAsync/include/CesiumAsync/Impl/AsyncSystemSchedulers.h
```diff
@@ -0,0 +1,78 @@
...
+  void startInWorkerThread();
+  void endInWorkerThread();
+
+  // If an AsyncSystemSchedulers instance is found in these thread-local
+  // vectors, then the current thread is currently inside the "main thread" or a
+  // "worker thread" for that scheduler.
+  static thread_local std::vector<AsyncSystemSchedulers*>
+      schedulersInMainThread;
+  static thread_local std::vector<AsyncSystemSchedulers*>
+      schedulersInWorkerThread;
```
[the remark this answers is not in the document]
kring: Same problem. Also there isn't just one thread ID which is the right one. Thread pools and the UE task system use multiple, interchangeable threads.

CesiumAsync/include/CesiumAsync/Impl/AsyncSystemSchedulers.h
kring: I'm not quite sure what the advantage of that would be. `ImmediatelyInMainThreadScheduler` is an embedded member of AsyncSystemSchedulers, so the lifetime dependency still exists. And the current approach is more symmetric with the other schedulers. Also, originally, `mainThreadScheduler` was meant to be usable directly, e.g. if you _don't_ want to execute immediately even if you're in the main thread. I eventually got rid of that because it wasn't immediately useful. But that's how JS promises work (by design) so it's easy to imagine we might want to support that behavior in the future.

CesiumAsync/include/CesiumAsync/Impl/AsyncSystemSchedulers.h
kring: Same reason as above. WorkerThreadScheduler was originally intended to be used as an independent scheduler, and making it easy to do so in the future (or at least not changing the code so it's harder!) is worthwhile.

CesiumAsync/src/ThreadPool.cpp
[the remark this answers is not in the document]
baothientran: Ahh that's right. Sorry I misread the code

CesiumAsync/include/CesiumAsync/Impl/AsyncSystemSchedulers.h
baothientran: My initial thought is that with the proposed change, the `AsyncSystemSchedulers` still depends on the `ImmediatelyInMainThreadScheduler`, but if we can make the `ImmediatelyInMainThreadScheduler` not depend on the `AsyncSystemSchedulers` (even though it's harmless), I think we should

CesiumAsync/include/CesiumAsync/Impl/AsyncSystemSchedulers.h
baothientran: Similar to the one above, the proposed change mostly is to make `ImmediatelyInWorkerThreadsScheduler` not to depend on `AsyncSystemSchedulers`. I think we can keep `WorkerThreadScheduler` but let `ImmediatelyInWorkerThreadsScheduler` owns it directly and remove `AsyncSystemSchedulers` from `ImmediatelyInWorkerThreadsScheduler`

kring: ok so if `ImmediatelyInWorkerThreadScheduler` owns a `WorkerThreadScheduler` directly, then in order to schedule directly with the `WorkerThreadScheduler` we'd have to "reach into" the `ImmediatelyInWorkerThreadScheduler` to grab its embedded `WorkerThreadScheduler`. Or have two `WorkerThreadScheduler` instances maybe. Either way that doesn't sound like an improvement to me.

I guess the way I think about it is that these various schedulers would ideally just be methods on `AsyncSystemSchedulers`. There would be a `scheduleInWorkerThread`, `scheduleImmediatelyInWorkerThread`, etc. But to use them with Async++, we need separate instances that all have a method called `schedule`. So `AsyncSystemSchedulers` plus all the individual scheduler types are conceptually a single class. Maybe the scheduler types should be nested to make that more clear.

baothientran: Admittedly, it's not an improvement in the implementation of the `AsyncSystem`. But from the reusable point of view, I still think it is, since if someone needs to use `ImmediatelyInWorkerThreadScheduler` outside of `AsyncSystem` context, they won't need to bring in the `AsyncSystemSchedulers` which has the `ImmediatelyInMainThreadScheduler` that they don't need. But that is an exacerbation on my part as the component won't be used outside of the `AsyncSystem`, so I guess the dependency change is not worth it. Also the scheduler is simple enough to reason about the object lifetime, and it probably won't get more complicated in the future to be much of the concern.

kring: I'm still not really sold on the reusability angle. Or rather, given that this is an entirely private API, I think it would be wiser to refactor for reuse if and when we're actually going to expose it publicly or reuse it internally, rather than speculatively investing that time now. But I had a go at implementing some approximation of what you were describing, and I do agree it's an improvement. Let me know if this is close to what you had in mind.

baothientran: It’s better than what I had in mind actually. I agree on the reusable aspect. I realized that it’s used only in the AsyncSystem internally, so it’s unreasonable of me to ask to invest the time in it

baothientran (commented):
I looks through async system, and things looks good to me. I only have some minor suggestion and a question below

Cesium3DTiles/src/Tileset.cpp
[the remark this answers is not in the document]
baothientran: oh okay I know what you mean now

CesiumAsync/include/CesiumAsync/Future.h
```diff
@@ -0,0 +1,256 @@
...
+    // dispatching of the work.
+    auto task = this->_task.then(
+        async::inline_scheduler(),
+        [tracingName, CESIUM_TRACE_LAMBDA_CAPTURE_TRACK()](T&& value) mutable {
+          CESIUM_TRACE_USE_CAPTURED_TRACK();
+          if (tracingName) {
+            CESIUM_TRACE_BEGIN_IN_TRACK(tracingName);
+          }
+          return std::move(value);
+        });
```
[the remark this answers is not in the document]
kring: Yes, exactly.

CesiumAsync/include/CesiumAsync/Impl/CatchFunction.h
```diff
@@ -0,0 +1,53 @@
...
+namespace CesiumAsync {
+namespace Impl {
+// Begin omitting doxgen warnings for Impl namespace
+//! @cond Doxygen_Suppress
+
+template <typename Func, typename T, typename Scheduler> struct CatchFunction {
+  Scheduler& scheduler;
+  Func f;
+
+  decltype(auto) operator()(async::task<T>&& t) {
```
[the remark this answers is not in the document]
kring: Yeah good point. While I'm admittedly a little fuzzy on difference, I originally saw this as a forwarding function, and so I wanted to forward the exact return type. But in reality this function will never return anything other than `async::task<T>` anyway, so I've changed it to say so explicitly.

baothientran:
thanks @kring, the PR looks good to me

PR #268, opened 2021-06-26
Use uint32_t instand of int32_t as count of tiles and use const T& not const T

zy6p:
There are two small problems in `Cesium3DTiles/include/Cesium3DTiles/RasterOverlayTileProvider.h`:
1. `Cesium3DTiles::RasterOverlayTileProvider::_throttledTilesCurrentlyLoading` is int32_t but its get() function return uint32_t, so I change 3 associated variables.
2. `Function Cesium3DTiles::RasterOverlayTileProvider::getCredit()` return a `const std::optional<Credit>` and this cause clang-tidy reports ['const' at top level, which may reduce code readability without improving const correctness](https://stackoverflow.com/questions/63972019/understanding-const-at-top-level-which-may-reduce-code-readability-without-i). Just need to add '&'.

### DIFF
--- a/Cesium3DTiles/include/Cesium3DTiles/RasterOverlay.h
+++ b/Cesium3DTiles/include/Cesium3DTiles/RasterOverlay.h
@@ -21,7 +21,7 @@ struct CESIUM3DTILES_API RasterOverlayOptions {
    * @brief The maximum number of overlay tiles that may simultaneously be in
    * the process of loading.
    */
-  int32_t maximumSimultaneousTileLoads = 20;
+  uint32_t maximumSimultaneousTileLoads = 20;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


zy6p:
> Thanks @zy6p! Can you please move the properties to use `int32_t` consistently rather than `uint32_t`? Consistent with the advice in the C++ Core Guidelines, we should avoid the use of unsigned integer types:
> [isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es106-dont-try-to-avoid-negative-values-by-using-unsigned](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es106-dont-try-to-avoid-negative-values-by-using-unsigned)

Oh, sorry! I will move the properties to use int32_t consistently rather than uint32_t soon!

kring:
Thanks @zy6p!

PR #272, opened 2021-07-01
Add .editorconfig and .gitattributes

kring:
Adds a `.editorconfig` file so that editors know that our source files:

* Use spaces, not tabs.
* Are indented with two spaces.
* Have a final newline at the end.
* Don't have trailing whitespace.

Visual Studio should pick this up automatically. For Visual Studio Code, install the EditorConfig plugin:
image

For C++ source files, clang-format will take care of most of this anyway, but having an editorconfig avoids some obvious mismatches as you're writing code in the first place, which is nice. Also the editorconfig applies in some places clang-format doesn't, like Markdown files.

Also added `.gitattributes` indicating that all files should be checked out with platform-specific line endings, but stored in the repo with Unix/Mac line endings.

The remaining changes in this PR fix files that were committed to the repo with Windows line endings.


lilleyse:
CC @IanLilleyT @jtorresfabra @Samulus - might be good to add to tilers too

baothientran:
thanks @kring, looks good to me

PR #275, opened 2021-07-04
Use angle brackets to #include catch2.

kring:
We were using quotes instead of angle brackets to #include catch2. So this trivial PR just changes it to angle brackets, consistent with all the other third party libraries.


baothientran:
This looks straightforward. I'm going to review the AsyncSystem test PR and merge into main after that

kring:
I'd recommend merging the PRs closest to main first (so, this one). After you merge this PR, GitHub will redirect PRs that target this branch to instead target main. So that way you'll see a minimal number of changes in each review.

baothientran:
I didn't know that. Let merge it then

PR #277, opened 2021-07-06
Add some tests for AsyncSystem and Future

kring:
Certainly not exhaustive, but exercises a lot of the main functionality.

Also, from the changelog:
* Added `AsyncSystem::dispatchZeroOrOneMainThreadTask` to dispatch a single task, rather than all the tasks that are waiting.
* Fixed a bug that prevented `.then` functions from being used on a `Future<void>` when CESIUM_TRACING_ENABLED was ON.


kring:
This isn't quite ready yet. The test failures on GCC and Clang reveal a deeper problem in how exceptions are handled.

kring:
Ready now!

CesiumAsync/include/CesiumAsync/Impl/WithTracing.h
```diff
@@ -9,7 +9,23 @@ namespace Impl {
...
+    return
+        [tracingName, CESIUM_TRACE_LAMBDA_CAPTURE_TRACK()](T&& result) mutable {
+          CESIUM_TRACE_USE_CAPTURED_TRACK();
+          if (tracingName) {
+            CESIUM_TRACE_BEGIN_IN_TRACK(tracingName);
+          }
+          return std::move(result);
+        };
+#else
+    return Impl::unwrapFuture<Func, T>(std::forward<Func>(f));
```
baothientran, line 24: I maybe wrong here, but I think the type `f` of this function is rvalue reference and not universal reference, so returning std::move(f) is appropriate in this case instead of std::forward

CesiumAsync/include/CesiumAsync/AsyncSystem.h
```diff
@@ -15,6 +15,91 @@ class ITaskProcessor;
...
 template <typename T> class Future;
 
+class AsyncSystem;
+
+/**
+ * @brief A promise that can be resolved or rejected by an asynchronous task.
+ *
+ * @tparam T The type of the object that the promise will be resolved with.
+ */
+template <typename T> class Promise {
```
baothientran, line 25: Nitpick from me, should we separate this out into another file, similar to Future?

baothientran (commented):
Thanks @kring! This looks good to me. I only have a couple questions and some small suggestion below

CesiumAsync/include/CesiumAsync/AsyncSystem.h
```diff
@@ -213,6 +290,17 @@ class CESIUMASYNC_API AsyncSystem final {
...
+  /**
+   * @brief Runs a single waiting task that is currently queued for the main
+   * thread, if there is one.
+   *
+   * The task is run in the calling thread.
+   *
+   * @return true A single task was executed.
+   * @return false No task was executed because none are waiting.
+   */
+  bool dispatchZeroOrOneMainThreadTask();
```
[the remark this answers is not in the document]
kring: No, it has little to do with how many tasks are in the queue. `dispatchMainThreadTasks` keeps running tasks until the queue is empty, then returns. `dispatchZeroOrOneMainThreadTask` runs just one task and then returns. If there are zero tasks in the queue, it returns immediately without running any tasks.

This is currently only useful to us in the tests, but it's a useful capability in general. Async++ has the same thing, named `try_run_one_task`.

I've renamed it to `dispatchOneMainThreadTask` and improved the docs to hopefully make this clearer.

CesiumAsync/include/CesiumAsync/Impl/WithTracing.h
kring: Ah good catch! Though, we actually do want perfect forwarding here, so I tweaked the templates so that `f` will be a universal reference. I made a similar mistake in `unwrapFuture` so I fixed it there too.

CesiumAsync/include/CesiumAsync/AsyncSystem.h
kring: Good idea, done.

kring:
Thanks @baothientran. I've addressed all the feedback and this is ready for another look.

baothientran:
@kring the intermittent failed test maybe related to the lifetime of the `TaskProcessor` that is captured in the lambda of its schedule() function [here](https://github.com/CesiumGS/cesium-native/blob/a033ebaa60a4d3f5f36482d87f98bc9d40972474/CesiumAsync/src/TaskScheduler.cpp#L29). 

This is just my educated guess at the moment, but I think when `async::task_run_handle` finish running at this [line](https://github.com/CesiumGS/cesium-native/blob/async-system-tests/CesiumAsync/src/TaskScheduler.cpp#L27), the waiting future (`Future` that calls its `wait()` method) is resolved immediately in the main thread, which makes `TaskProcessor` finish its lifetime in the main thread and release all of its schedulers. However, the `markEnd()` [here](https://github.com/CesiumGS/cesium-native/blob/a033ebaa60a4d3f5f36482d87f98bc9d40972474/CesiumAsync/src/TaskScheduler.cpp#L29) is not finished yet, and so we have a dangling pointer in this case where the lambda refers to an already-deleted scheduler. A solution I can think of maybe to use `std::enable_shared_from_this` for `TaskProcessor`. So instead of letting the lambda capture `this` raw pointer in this [line](https://github.com/CesiumGS/cesium-native/blob/a033ebaa60a4d3f5f36482d87f98bc9d40972474/CesiumAsync/src/TaskScheduler.cpp#L23), the lambda needs to capture `std::shared_from_this()` to keep the `TaskProcessor` alive as long as the lambda is finished executing. 

Also I can replicate the bug more frequently when I have a heavy-computing task running on the background of my laptop (for example, compiling o3de engine while debugging the bug 😆)

baothientran:
This is kind of dodgy but I can replicate the issue with this test below to confirm it:
- I modify the TaskScheduler's schedule() method like this:
```C++
void TaskScheduler::schedule(async::task_run_handle t) {
  // std::function must be copyable, so we can't put a move-only
  // task_run_handle in the capture list of a lambda we want to use with it.
  // So, we wrap it with a copyable type (shared_ptr).
  // https://riptutorial.com/cplusplus/example/1950/generalized-capture has
  // a good explanation of this problem.

  struct Receiver {
    async::task_run_handle taskHandle;
  };

  std::shared_ptr<Receiver> pReceiver = std::make_shared<Receiver>();
  pReceiver->taskHandle = std::move(t);

  this->_pTaskProcessor->startTask([this, pReceiver]() mutable {
    this->immediate.markBegin();

    try {
      pReceiver->taskHandle.run();
    } catch (...) {
      this->immediate.markEnd();
      throw;
    }

    // let worker thread sleep shorter than main thread so that markEnd() will get triggered when the below test is executed
    // You can see before the sleep below, the future is finished waiting in the main thread
    using namespace std::chrono_literals;
    std::this_thread::sleep_for(7000ms);
    this->immediate.markEnd();
  });
}
```

- Then run this test below alone:
```C++
  SECTION("worker continuations following a worker run immediately") {
    bool executed1 = false;
    bool executed2 = false;

    {
      // let this taskprocessor go out of scope
      std::shared_ptr<MockTaskProcessor> pTaskProcessor =
          std::make_shared<MockTaskProcessor>();
      AsyncSystem asyncSystem(pTaskProcessor);

      Promise<void> promise = asyncSystem.createPromise<void>();
      Future<void> trigger = promise.getFuture();

      auto future =
          std::move(trigger)
              .thenInWorkerThread([&executed1]() { executed1 = true; })
              .thenInWorkerThread([&executed2]() { executed2 = true; });

      // Now that both continuations are attached, set the chain in motion.
      promise.resolve();
      future.wait();

      CHECK(pTaskProcessor->tasksStarted == 1);
    }

    CHECK(executed1);
    CHECK(executed2);

    using namespace std::chrono_literals;
    std::this_thread::sleep_for(100000ms);
  }
```

kring:
Thanks for the detective work @baothientran. This should be fixed. I moved to RAII for marking the start/end of scheduling a thread to do some work, which is much cleaner for the TaskScheduler and the QueuedScheduler, but still a little awkward for the ThreadPool because Async++ only exposes prerun/postrun hooks and has no ability to pass a scope object between the two. So instead we store it in another thread_local.

It's still possible for `schedulersCurrentlyDispatching` to briefly contain a pointer to an object that has been deleted. This doesn't cause a problem because the pointer is never dereferenced or otherwise used. It's only a "marker" that the current thread is doing some work for that scheduler (whether or not that scheduler still exists). The problem we had before was that _getting_ that pointer involved differencing a potentially dangling pointer. But no longer: the scheduler pointer is now stored in the RAII object, which is guaranteed to live long enough.

baothientran (approved):
Thanks @kring I can confirm that fix the issue. I only has a small comment below regarding for Future::wait() that throws exception. Do you think if it's a quick fix to add another interface that doesn't throw? If it's not, the suggestion below doesn't need to hold up the PR though

CesiumAsync/include/CesiumAsync/Future.h
```diff
@@ -175,18 +174,10 @@ template <typename T> class Future final {
...
-  Impl::FutureWaitResult_t<T> wait() {
-    try {
-      return Impl::FutureWaitResult<T>::getFromTask(this->_task);
-    } catch (std::exception& e) {
-      return e;
-    } catch (...) {
-      return std::runtime_error("Unknown exception.");
-    }
-  }
+  T wait() { return this->_task.get(); }
```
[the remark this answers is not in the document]
kring: I had such an interface originally. It would return `std::variant<T, std::exception>`. But that didn't work at all, because copying the exception into the variant would slice the exception (i.e. call the std::exception copy/move constructor on an instance that is actually of a derived type). Once I realized that, I tried making it `std::variant<T, std::exception_ptr>` instead, but that doesn't really help because the only way to get an exception from an exception_ptr is by throwing, so it doesn't help clients with exceptions disabled. It's almost like the C++ language committee doesn't want you to turn off exceptions! 😆 

The only solution that I can think of is to make the user provide two callbacks to `wait`, one to receive a value and one to receive an exception. At least then we could pass the complete, derived-class exception object to the exception callback. But clients would probably still need a way to figure out what type of exception was thrown in order to do anything useful with it. dynamic_cast? 😬 Plus that interface (two callbacks) is far less convenient.

So apps with exceptions disabled really just need to avoid rejected exceptions getting all the way to a wait. That's relatively easy to do just by adding a `catchImmediately` or `catchInMainThread` that turns a rejection into success with an error object.

In the future, if we stop wrapping Async++ and implement this stuff ourselves, we can probably do better. Not using exceptions to represent rejected future results, for example. But that's way outside the scope of this PR.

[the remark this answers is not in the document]
baothientran: yeah the alternative doesn't look good to me too.

baothientran:
thanks @kring! Once the CI finishes, I can merge this

PR #279, opened 2021-07-07
Put all generated glTF JSON handlers into a single C++ source file, rather than one file per class

kring:
They're all generated together anyway, and compiling one source file rather than dozens saves a bunch of time.


baothientran:
@kring I haven't ran the build yet, but how fast is the compilation process after the combination? I have a mixed feeling about this as I still prefer separate files for easy to maintain and debugging even though the compilation is longer. But cesium-native has fast compilation time at the moment, so the trade off maybe not significant to combine them all?

baothientran:
Another thing I worry is that if the file is getting bigger in the future, some IDE and code editor may have a trouble to read and parse them (e.g sqlite3.c file). But that's probably not much of a concern compared to the maintenance above as the combined file is only around thousand lines

kring:
@baothientran in general I agree with you regarding maintenance, but in this case it's generated that code you're literally not supposed to touch. If you're maintaining it, you're doing it wrong. :) I haven't measured the time, but ~30 fewer source files to compile will be significant.

baothientran:
That's true about not touching the generated files, but I'm still uncertain about this due to maybe being unfamiliar with it. I may need some more  opinions from others too to decide on this. @IanLilleyT, @lilleyse, and @jtorresfabra, since our inner project may need the gltf, do you have any opinions on combining the generated json handlers into one file?

baothientran:
For the maintenance and debugging point, I guess it's still doable with one combined file, but still want to have more some advice before going forward

kring:
I don't see any downside, but 🤷.

Timing wise, I did this:

```powershell
rm -recurse -force build/CesiumGltfReader
Measure-Command { ."C:\Program Files\CMake\bin\cmake.EXE" --build c:/Dev/cesium-unreal-samples/Plugins/cesium-unreal/extern/cesium-native/build --config Debug --target ALL_BUILD -j 14 -- }
```

So, just measuring the time to build CesiumGltfReader. It also triggers a CMake configure, which takes a few seconds.

In this branch, it reports about 16 seconds. In main, it reports about 32 seconds.

jtorresfabra:
From my point of view, and as far as the documentation is clear on how to use it (examples, tests, README.md, you name it...) and we don't need to dive into the class I'm pretty ok with the change. :+1:

CesiumGltfReader/generated/GeneratedJsonHandlers.cpp
```diff
@@ -0,0 +1,1612 @@
...
+
+  return this->readObjectKeyNamedObject(objectType, str, *this->_pObject);
+}
+// This file was generated by generate-gltf-classes.
+// DO NOT EDIT THIS FILE!
+#include "CesiumGltf/Mesh.h"
+#include "MeshJsonHandler.h"
+
+#include <cassert>
+#include <string>
```
baothientran, line 385: Ideally the headers should be at the top of the file. But that shouldn't hold the PR back though

baothientran:
if everyone is okay with the change, then I will go ahead with the PR

PR #280, opened 2021-07-07
Add support for compiling with C++14.

kring:
This PR makes cesium-native compile successfully with C++14. Previously it required C++17. To use C++14, add `-DCESIUM_CXX_STANDARD=14` when you CMake configure.

I had to disable the tests in TestReader.cpp and TestTilesetSelectionAlgorithm.cpp because they use std::filesystem, which isn't available. Also, `CesiumGltfWriter` is not included in C++14 builds because it uses magic_enum, which is explicitly a C++17 library.

The reason we need this is because Unreal Engine doesn't support C++17 on iOS in any 4.26.x or 4.27.x version. It does support C++17 on iOS in ue5-main, but not in any early access build yet.

Meanwhile on Android, UE supports C++17 in 4.26.2, but not any earlier version. Because Epic's Marketplace build process uses 4.26.0 instead of 4.26.2, it's not possible to publish an Android-supporting plugin that uses C++17. And like iOS, UE5 does support C++17 on Android, but only in the ue5-main branch on github, not in a released early access build yet.

To do:
* [x] Polyfill std::filesystem for C++14, or replace it. Then remove the `#if __cplusplus >= 201703L` around two test files.
* [x] Add one or more C++14 CI jobs.


lilleyse:
Does Epic have plans to update their build process to 4.26.2 soon?
What are the C++ requirements for other engines, e.g. O3DE? Is C++ 17 widely used at this point?

Should we just embrace C++ 14 and not try to support both 14 and 17 simultaneously?

shehzan10:
> Does Epic have plans to update their build process to 4.26.2 soon?

@lilleyse Epic builds separate packages for each minor version for Unreal Engine - so 4.26 (and in the future, 4.27, and then 5).

For each minor release, the plugin is built separately using the `X.Y.0` patch version of Unreal Engine. So we can't specifically target 4.26.2 for compilation. We need the plugin to be compilation compatible with all versions of 4.26 by building on 4.26.0.

The way it looks right now is that as long as we want to support 4.26, we will have to support 4.26.0 compilation.

@kring Can correct me if I wrote anything incorrect or incomplete.

baothientran:
O3DE currently targets C++17, which makes this a little easier to decide

lilleyse:
@shehzan10 thanks that clarifies some things. When is 4.27 expected and would we set that to the plugin's minimum requirements? Or do we plan on supporting 4.26 for a while? How fast are Unreal users to upgrade?

shehzan10:
@lilleyse I expect that we will continue to support 4.26 for a while. Because UE is not shy to break APIs between minor releases, it does create a tail where developers don't immediately upgrade.

That said, 4.27 adds a decent number of georeference features which may convince users with a better experience.

The flip side to that coin is that there may be a portion of developers who may not see the value in upgrading to 4.26 and believe that they will want to invest those cycles in going to Unreal Engine 5 at some point.

TLDR: I expect that we will support 4.26 on the Unreal Marketplace for the foreseeable future.

baothientran:
There are a couple things that concerns me about the PR, and I'm not quite sure how to go about it:
- We are defining alias for third party libraries `optional`, `variant`, `any`, and `byte` in the `std` namespace when compiling C++14. Though I understand the reason of that is to polyfill `std` type and enable easy transitions to C++17, it's still a bad practice which @kring already know. It's undefined behavior to extend namespace `std` with any declarations and definitions with only very few  exceptions (https://en.cppreference.com/w/cpp/language/extending_std).
- Though the change is mostly for Unreal iOS and Android, I don't think we will be able to use full C++17 features as long as we want to support 4.26. For example, we can't use `if constexpr`, or `fold expression`, or any new language features (not new std library features) in this link https://en.cppreference.com/w/cpp/17. So instead of pretending to be a C++17 library, I think it maybe best to constraint ourselves to C++14 to be consistent in the style. If we go with C++14, I'm still favor to define all the new std library feature in an independent namespace, instead of defining the alias the `std` currently

kring:
> It's undefined behavior to extend namespace std with any declarations and definitions with only very few exceptions

Technically true, yes. The reason the C++ standard has this rule is so that the standard committee and implementors don't need to worry about a) their introducing new types in `std` breaking people's code, and b) dodgy stuff added to the `std` namespace causing the standard-provided functionality to break in subtle ways. By declaring the whole mess undefined, they're providing cover for any complaints. If you add something to std, and your program stops working or misbehaves in an obvious or subtle way, well, too bad, that's what undefined behavior means.

But in our case, we're only adding types to the std namespace that C++17 has already added itself, and those types have the same semantics as the actual C++17 types (to within the limitations, if any, imposed by the C++14 compiler). So while it may be _technically_ undefined, for all practical purposes the behavior will be well understood and reliable.

It's an ugly hack, no doubt, but "the standard says it's UB!" is not a sufficient argument against it in this context, given that we have no better alternatives.

> I don't think we will be able to use full C++17 features as long as we want to support 4.26.

That's true. But our ability to benefit from C++17 will still be far greater than if we throw the baby out with the bath water and go with pure C++14:

1. We still benefit from automatic C++17 compiler improvements, such as copy ellision.
2. We can use C++17 features like `[[nodiscard]]` via a #define, and benefit from that useful compile-time check when compiling with C++17.
3. We use the actual C++ standard types, like `std::variant` and `std::string_view` in our public API, which improves usability because people are familiar with them. 
4. We benefit from better-tested, possibly more performant implementations of the new C++17 standard library on platforms that can use C++17, rather than using a far less widely used third-party implementation everywhere.
5. We make it easier to use cesium-native in other projects that want to use C++17. We have two examples of this already.
6. We can use C++17 unhindered in libraries that aren't needed in Cesium for Unreal on Android and iOS. For example, CesiumGltfWriter is still C++17-only in this PR.
7. Our path to dropping C++14 support in the future is clearer and requires less work. I estimate we'll be able to do that in about a year.

kring:
We no longer need this because Epic accepted CesiumGS/cesium-unreal#544. 😅

Issue #126, opened 2021-02-05
Discussion about the traversal process


javagl:
The current traversal procedure of tiles is - virtually by definition - the most simple traversal process that will ever exist. 

In the future, there will be ...

- additional culling criteria. 
- additional visibility checks. 
- additional flags that can be switched on or off. 
- additional tile types and providers. 
- refinements for the refinements. 
- extensions for the prioriztations of tile loading. 
- additional information (statistics, performance measures) that have to be collected during the traversal. 
- additional operations that are supposed to be done during the traversal. 
- additional strategies for the order in which the children of a tile will be visited. 

Each of these could offer great options for new features and improvements. But it also means that the complexity of the traversal **can** only increase (and I'd say it's bound to increase *dramatically*). If there is no clear separation of concerns, no abstraction, and no clear, high-level outline for the process itself and the state transitions, the state space will face a combinatorial explosion. 

But even though the current state is the most simple state that will ever exist, it's already ... complicated. 

---

An aside: It may sound like an inappropriate and/or bold claim, but I'm pretty sure that we're already at a point where nobody knows exactly what happens during the traversal. One example for what caused my concerns here is that the traversal starts with this method:

    Tileset::TraversalDetails Tileset::_visitTileIfNeeded(
        const FrameState& frameState,
        uint32_t depth,
        bool ancestorMeetsSse,
        Tile& tile,
        ViewUpdateResult& result

I tried to figure out the meaning of the `ancestorMeetsSse` flag. And I think this is essentially not possible. (Spoiler alert: It certainly does not say whether an ancestor meets the SSE). And when it's impossible to clearly state the conditions under which this flag, which is passed into the top-level-traversal call (!) and passed along (and changed occasionally...) is `true` or `false`, then that's not good: Any unit-test that is supposed to do something that is even indirectly affected by this flag could just capture the current state, but not say whether the current state is "right" or "wrong". 

---

From an idealistic perspective, I wondered whether it could be possible to move towards a state where the traversal process looks more like this pseudocode...:

    ViewUpdateResult updateView(ViewState viewState) {
        TraversalResult traversalResult = traverse(root, TraversalState(viewState, frameNumber));

        addTilesToLoadQueues(traversalResult.tilesToLoad);
        passTilesToRenderer(traversalResult.tilesToRender);
        removeTilesFromRenderer(traversalResult.tilesToRemove);
        
        return ViewUpdateResult(traversalResult);
    }
[...]

PR #283, opened 2021-07-13
Support multiple frustums

nithinp7:
Allows arbitrary list of view frustums to be passed in to the refinement engine.


baothientran:
@nithinp7 can you separate the orthographic frustum commit in a separate PR?

javagl (commented):
Two nitpicks inline, will inline further thoughts as comments, and still have to test it a bit

javagl:
The current solution, from just reading it, seems to be a very pragmatic solution to the immediate goal.

---

Beyond that: The traversal is one of my pet peeves, with some elaborate discussion at https://github.com/CesiumGS/cesium-native/issues/126 . Some of the "shortcomings" that I mentioned there are not supposed to be affected by this PR. I think that some of the changes are not really improvements in that regard, but that might not be considered to be important, in view of the need to achieve a certain goal quickly (and that's what it does). 

But one thing that I wondered about: There have been methods like `bool isVisibleInFog(double distance, double fogDensity)` and `bool _meetsSse(const ViewState& viewState, const Tile& tile, double distance, bool culled)`, but also `addTileToLoadQueue` (I already talked about [the _meetsSse method before](https://github.com/CesiumGS/cesium-native/issues/126#issuecomment-774504271), and the load queues in general...). Now, these methods receive _vectors_ of `distances` and `frustums` (which actually are `viewStates`, but ... I'm not gonna talk about that one for now). When there already is a method 

    static bool isVisibleInFog(double distance, double fogDensity) {
        ...
    }

and then "that thing" has to be checked for multiple distances and densities, then I wonder whether there is a reason to pass in the vectors, when the given method already does this one, specific thing, and is documented and such. I'd then rather tend to *call* this method in another one:
```
static bool isVisibleInFog(
    const std::vector<ViewState>& frustums,
    const std::vector<double>& distances,
    const std::vector<FogDensityAtHeight>& fogDensityTable) {
    for (...) {
        double density = computeDensity(fogDensityTable, frustums[i]);
        if (!isVisibleInFog(distances[i], density) {
            return false;
        }
    }
    return true;
}
```
similar to how it was done for `isVisibleFromCamera`.  

The same applies to the `addTileToLoadQueue` method, but I see that it's a bit more difficult there: I already added a comment about some `priorityFor(tile, viewState, distance)`, because the statement: "In order to add a loading task to a queue, you need a distance" sounds odd, and more generally, I think there should not be three load queues (because that makes no sense), but ... I'll try to avoid further bragging about all the predictions that I made in https://github.com/CesiumGS/cesium-native/issues/126 coming true (e.g. "if we had a `TraversalState`, then this whole change could have been much, **much** simpler"), and will do some tests now instead.

---

**tl;dr**: If the tests don't bring up any issues, I have no further comments for now.

baothientran (commented):
This looks really good @nithinp7! I only has some small comments below

Cesium3DTiles/src/Tileset.cpp
```diff
@@ -1201,12 +1197,28 @@ static bool isVisibleFromCamera(
...
  */
-static bool isVisibleInFog(double distance, double fogDensity) {
-  if (fogDensity <= 0.0) {
-    return true;
+static bool isVisibleInFog(
+    const std::vector<ViewState>& frustums,
+    const std::vector<double>& distances,
+    const std::vector<FogDensityAtHeight>& fogDensityTable) {
+
+  for (size_t i = 0; i < frustums.size() && i < distances.size(); ++i) {
```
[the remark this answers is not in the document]
baothientran: I agree this should be using assert as it's developer error at this point

baothientran:
@nithinp7 can you also please add tests for multiple frustums?

Cesium3DTiles/src/Tileset.cpp
```diff
@@ -1201,12 +1197,28 @@ static bool isVisibleFromCamera(
...
+    const ViewState& frustum = frustums[i];
+    const double& distance = distances[i];
+
+    double fogDensity = computeFogDensity(fogDensityTable, frustum);
```
baothientran, line 1209: this may re-compute fog density for every tile even though fog-density only change between frustums

Cesium3DTiles/src/Tileset.cpp
```diff
@@ -1201,12 +1197,28 @@ static bool isVisibleFromCamera(
...
+
+    double fogScalar = distance * fogDensity;
+    if (glm::exp(-(fogScalar * fogScalar)) > 0.0) {
+      return true;
+    }
   }
-  double fogScalar = distance * fogDensity;
-  return glm::exp(-(fogScalar * fogScalar)) > 0.0;
+
+  return false;
```
baothientran, line 1221: should we fog-cull a tile if it is fog-culled by all the frustums? This looks like it culls a tile if the tile is fog-culled by one of the frustum

Cesium3DTiles/src/Tileset.cpp
```diff
@@ -1868,24 +1911,34 @@ static bool anyRasterOverlaysNeedLoading(const Tile& tile) {
...
+
+      glm::dvec3 tileDirection = boundingVolumeCenter - frustum.getPosition();
+      double magnitude = glm::length(tileDirection);
+
+      if (magnitude >= CesiumUtility::Math::EPSILON5) {
+        tileDirection /= magnitude;
+        double loadPriority =
+            (1.0 - glm::dot(tileDirection, frustum.getDirection())) * distance;
+        if (loadPriority > maxLoadPriority) {
+          maxLoadPriority = loadPriority;
```
kring, line 1936: I think you want the minimum load priority. Lower priority values load sooner.

Cesium3DTiles/src/Tileset.cpp
```diff
@@ -1224,18 +1236,30 @@ Tileset::TraversalDetails Tileset::_visitTileIfNeeded(
...
-          boundingVolume,
-          this->_options.renderTilesUnderCamera)) {
+  const std::vector<ViewState>& frustums = frameState.frustums;
+
+  if (std::none_of(
+          frustums.begin(),
+          frustums.end(),
+          [boundingVolume,
+           renderTilesUnderCamera = this->_options.renderTilesUnderCamera](
+              const ViewState& frustum) {
```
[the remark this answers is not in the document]
kring: fwiw I quite like this functional-ish style.

Cesium3DTiles/src/Tileset.cpp
kring: The logic looks correct to me. It returns true (visible) if its not in fog in any of the frustums, and returns false (culled) only if its in fog for all the frustums.

Cesium3DTiles/src/Tileset.cpp
javagl: I also didn't spot that, although, as you might remember, I once criticized that "low `priority`" means "high priority". Things like this cause bugs, period.

Cesium3DTiles/src/Tileset.cpp
baothientran: ah that's right! Sorry for the false alarm

Cesium3DTiles/src/Tileset.cpp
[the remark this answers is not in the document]
baothientran: I don't have strong opinion on this, so if everyone is good with it, then it's okay for me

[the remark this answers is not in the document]
javagl: Subjetively, I also like this functional style (and it has several advantages - there's a compelling reason for nearly all major programming languages having moved into that direction since 2005 - "The free lunch is over", roughly related to "internal iteration" vs "external iteration" and such...)

**But** ... I don't like large, anonymous lambdas. This specific block would IMHO be *much* better if it was just a line 
```
if (std::none_of(frustums.begin(), frustums.end(), tileIsInFrustum)) {
    ...
}
```
with a proper `auto tileIsInFrustum = ...`  lambda.

nithinp7:
@baothientran @kring I added a few tests, I definitely need to write more tests for it but maybe that can be a separate PR for the sake of time? Otherwise, I think this works as expected. I'll note though that the frustum/fog culling is a bit awkward and maybe could be better optimized for cases where the different frustums are on different sides of the world. It would also be awkward to add a hardware occlusion query stage to this the way it is currently structured, so we may want to slightly refactor that code or at least create an issue for doing so.

Cesium3DTiles/src/Tileset.cpp
```diff
@@ -1201,10 +1216,11 @@ static bool isVisibleFromCamera(
  * @param fogDensity The fog density
  * @return Whether the tile is visible in the fog
  */
-static bool isVisibleInFog(double distance, double fogDensity) {
+static bool isVisibleInFog(const double& distance, const double& fogDensity) {
```
kring, line 1219: Why the change here? It's generally better to pass doubles by value.

PR #285, opened 2021-07-15
Add `AsyncSystem::all` and `Future<T>::catchImmediately`

kring:
Builds on #277 so merge that first.

`all` is needed for #252, and is also likely to be used for i3dm support. `catchImmediately` is useful for the tests.


baothientran:

PR #286, opened 2021-07-16
Add SharedFuture

kring:
Mostly just a wrapper around Async++'s `shared_task`: https://github.com/Amanieu/asyncplusplus/wiki/Tasks#shared-tasks

This is needed for the raster overlay changes, and is a useful capability in general.

Builds on #285 so merge that first.


CesiumAsync/include/CesiumAsync/SharedFuture.h
```diff
@@ -0,0 +1,254 @@
...
+    // It would be nice if tracingName were a template parameter instead of a
+    // function parameter, but that triggers a bug in VS2017. It was previously
+    // a bug in VS2019, too, but has been fixed there:
+    // https://developercommunity.visualstudio.com/t/internal-compiler-error-when-compiling-a-template-1/534210
+#if CESIUM_TRACING_ENABLED
+    // When tracing is enabled, we measure the time between scheduling and
+    // dispatching of the work.
+    auto task = this->_task.then(
+        async::inline_scheduler(),
+        Impl::WithTracing<T>::begin(tracingName, std::forward<Func>(f)));
```
baothientran, line 216: this code doesn't seem to compile when enabling tracing. I think it's because `Impl::WithTracing<T>::begin()` returns a function with rvalue reference, but `shared_task` seems to only accepts value or const lvalue&

baothientran:
Thanks @kring! this looks good to me overall. There is a compiling error when tracing enables. Once it's fixes, I think we can merge this

kring:
@baothientran I fixed the compile errors. Tracing is not going to work very well with shared futures, though, because the multiple continuations will all run in the same "track" and `chrome://tracing` will get confused. Solving this will be tricky, and I don't think it needs to hold up this PR.

baothientran:
@kring is that happening even with regular `Future` without using another track (e.g when we call `CESIUM_TRACE_USE_TRACK_SET(loadingSlot)`)? Below is the test case when using two tracks ( `CESIUM_TRACE_USE_TRACK_SET` is used twice for two different futures)
```C++
class MockTaskProcessor : public ITaskProcessor {
public:
  std::atomic<int32_t> tasksStarted = 0;

  virtual void startTask(std::function<void()> f) {
    ++tasksStarted;
    std::thread(f).detach();
  }
};

TEST_CASE("AsyncSystem") {
  std::shared_ptr<MockTaskProcessor> pTaskProcessor =
      std::make_shared<MockTaskProcessor>();
  AsyncSystem asyncSystem(pTaskProcessor);

  SECTION("conversion to SharedFuture") {
    CESIUM_TRACE_INIT("C:\\Users\\bao\\Documents\\trace.json");
    {
      CESIUM_TRACE_DECLARE_TRACK_SET(loadingSlots, "Future");
      CESIUM_TRACE_USE_TRACK_SET(loadingSlots);
      auto firstPromise = asyncSystem.createPromise<int>();
      auto firstFuture = firstPromise.getFuture();

      bool executed1 = false;
      auto one = std::move(firstFuture)
                     .thenInWorkerThread([&executed1](int value) {
                       CESIUM_TRACE_BEGIN_IN_TRACK("Future One")
                       CHECK(value == 1);
                       CHECK(!executed1);
                       return 2;
                     })
                     .thenImmediately([](int value) {
                       CESIUM_TRACE_END_IN_TRACK("Future One");
                       return value;
                     })
                     .thenInWorkerThread([](int value) {
                       CESIUM_TRACE_BEGIN_IN_TRACK("Future One 2nd")
                       return value;
                     })
                     .thenImmediately([](int) {
                       CESIUM_TRACE_END_IN_TRACK("Future One 2nd")
                     });

      CESIUM_TRACE_USE_TRACK_SET(loadingSlots);
      bool executed2 = false;
      auto secondPromise = asyncSystem.createPromise<int>();
      auto secondFuture = secondPromise.getFuture();
      auto two = std::move(secondFuture)
                     .thenInWorkerThread([&executed2](int value) {
                       CESIUM_TRACE_BEGIN_IN_TRACK("Future Two")
                       CHECK(value == 1);
                       CHECK(!executed2);
                       return 2;
                     })
                     .thenImmediately([](int value) {
                       CESIUM_TRACE_END_IN_TRACK("Future Two");
                       return value;
                     })
                     .thenInWorkerThread([](int value) {
                       CESIUM_TRACE_BEGIN_IN_TRACK("Future Two 2nd")
                       return value;
                     })
                     .thenImmediately([](int) {
                       CESIUM_TRACE_END_IN_TRACK("Future Two 2nd")
                     });
      firstPromise.resolve(1);
      secondPromise.resolve(1);

      one.wait();
      two.wait();
    }

    CESIUM_TRACE_SHUTDOWN();
  }
}
```
With this test, I got the tracer to produce two different tracks like below ([trace.json.txt](https://github.com/CesiumGS/cesium-native/files/6936038/trace.json.txt)):
TwoTracks

Then, I try to use one track only by comment out the second use of `CESIUM_TRACE_USE_TRACK_SET` in the test below, and chrome produces a different input which seems to make less sense compared to the one above ([trace.json.txt](https://github.com/CesiumGS/cesium-native/files/6936105/trace.json.txt)):

OneTrack

For the second one, I would expect the order of: `waiting for worker thread...Future One...waiting for worker thread...Future One Two` for the first continuation. The second continuation is similar. But in the second one, the order is a little mixed. Not sure if I miss something

I modify the test above to use the share future and give the same results as well

kring:
Right, you have two threads of work happening in one track. A track is a single sequential flow, like a thread (but it's allowed to cross threads). Chrome will try to interpret the events in this light, leading to something between weirdness and total brokenness. SharedFuture just makes it a little easier to accidentally end up in this situation.

Issue #270, opened 2021-06-30
Extend Gltf generator tool into generic Json schema generator tool


baothientran:
I just stumble `quicktype` library [here](https://github.com/quicktype/quicktype) that can generate C++ classes from json schema along with serializer and deserializer. You can try it online here https://app.quicktype.io/. However, the code it generates is quite poor, for example: a `double` property is turned into `std::shared_ptr<double>`, or it uses `regex` header. Given that our gltf generator tool may almost work with any json file, I think we should extend it to generate C++ code with a generic json schema and add support for both serialization and deserialization too. Bonus if we can generate validator from the schema as well

kring:
Yeah the glTF generator was definitely intended to be generic, with 3D Tiles in mind in particular, though obviously that vision isn't fully realized. It does require the particular JSON Schema version that the glTF schema uses, but I think 3D Tiles uses the same one so that should be a good fit.

kring:
The same code is now used to generate classes for 3D Tiles, too.

PR #291, opened 2021-07-22
3d tiles serialization

krupkad:
#### Summary
- moved the non-glTF specific `CesiumJson` writer components to `CesiumJsonWriter` (new) so it can be shared without mixing in things named `glTF`
- renamed `Cesium3DTiles` to `Cesium3DTilesPipeline` to make room for the 3D Tiles reader/writer (forgot the original name we mentioned, let me know if this should be changed)
- added the autogenerated 3D tiles writer - PR for updated generator coming shortly and separately (unless it should go here)

First two changes are large but all renames/moves - they're isolated to their own commits.

#### More info on current state and next steps
Basically the glTF autogenerated readers are a little coupled to hand-written extension handling in `CesiumGltf` (`ExtensionsJsonHandler::readObjectKey` invokes  `GltfReader::createExtensionHandler`), so it's complicated to do that factoring-out in clean chunks (re: #270). Therefore doing the 3D Tiles (de)serialization completely separately, with an eye to merging them in the future.

Had to make some autogenerator changes to handle the glTF schema and produce writers, which still needs a few more changes to be compatible with the existing glTF output. For first-pass the generated structs a) have no `final` or `Spec`, b) no non-trivial extension awareness.

Follow-up plan is:
- finish up the autogenerator changes
- add readers
- add extension handling

The writer has unit test coverage - basically creates a `Tileset` from structs by hand, serializes it, then parses it with rapidjson and checks that all expected fields exist.

Opening in draft for discussion of style/approach.

Particular questions:
- the reader/writer/struct libraries look like they'll only be a single cpp/h pair each, do we still want these all separate, or would it make more sense to combine them into e.g. `Cesium3DTilesSerialization`


lilleyse:
@krupkad and I talked offline about splitting this into multiple PRs, possible one per commit since the commits are segmented out nicely already. @kring @baothientran what do you think?

baothientran:
yup I agree. It's easier to review with smaller PR

krupkad:
Closing in favor of separate PRs:
- factoring out JSON writing from CesiumGltfWriter
- renaming Cesium3DTiles -> Cesium3DTilesPipeline
- adding Cesium3DTilesWriter

PR #293, opened 2021-07-23
CMake option for code coverage

IanLilleyT:
This PR adds a cmake option to enable or disable code coverage. I also removed the "cesium-native-is-submodule" check because it didn't seem to be working, plus there's no need for it now that you can enabled or disable manually

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ project(cesium-native
 
 option(PRIVATE_CESIUM_SQLITE "ON to rename SQLite symbols to cesium_sqlite3_* so they won't conflict with other SQLite implemenentations" OFF)
 option(CESIUM_TRACING_ENABLED "Whether to enable the Cesium performance tracing framework (CESIUM_TRACE_* macros)." OFF)
-option(CESIUM_COVERAGE_ENABLED "Whether to enable code coverage" ON)
+option(CESIUM_COVERAGE_ENABLED "Whether to enable code coverage" OFF)
 
 if (CESIUM_TRACING_ENABLED)
     add_compile_definitions(CESIUM_TRACING_ENABLED=1)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,7 @@ project(cesium-native
 
 option(PRIVATE_CESIUM_SQLITE "ON to rename SQLite symbols to cesium_sqlite3_* so they won't conflict with other SQLite implemenentations" OFF)
 option(CESIUM_TRACING_ENABLED "Whether to enable the Cesium performance tracing framework (CESIUM_TRACE_* macros)." OFF)
+option(CESIUM_COVERAGE_ENABLED "Whether to enable code coverage" ON)
 
 if (CESIUM_TRACING_ENABLED)
     add_compile_definitions(CESIUM_TRACING_ENABLED=1)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -16,23 +17,23 @@ endif()
...
-        DEPENDENCIES cesium-native-tests
-    )
+if (CESIUM_COVERAGE_ENABLED)
+    # Add a coverage target iff:
+    # - gcovr is available
+    # - g++ is the active compiler.
+    # This branch must occur before adding other targets, or gcovr
+    # won't be able to inject the necessary coverage information.
+    find_program(_gcovr "gcovr")
+    if(CMAKE_COMPILER_IS_GNUCXX AND _gcovr)
```
Samulus, line 27: Should we raise a `FATAL_ERROR` during configuration if `CESIUM_COVERAGE_ENABLED` is true but `gcovr` is not in the system path?

Samulus (commented):
I like using an option here, it's less fragile and users of the native library can turn it off explicitly.

CMakeLists.txt
```diff
@@ -16,23 +17,23 @@ endif()
...
-        EXCLUDE "${PROJECT_SOURCE_DIR}/extern/*" "${PROJECT_BINARY_DIR}"
-        DEPENDENCIES cesium-native-tests
-    )
+if (CESIUM_COVERAGE_ENABLED)
+    # Add a coverage target iff:
+    # - gcovr is available
+    # - g++ is the active compiler.
+    # This branch must occur before adding other targets, or gcovr
+    # won't be able to inject the necessary coverage information.
+    find_program(_gcovr "gcovr")
```
baothientran, line 26: I think CodeCoverage.cmake has logic to find `gcovr` too. Maybe we shouldn't do it here

CMakeLists.txt
baothientran, line 27: I think AppleClang work with code coverage too. Maybe we shouldn't check any condition at all here and let the error from `CodeCoverage.cmake` print out?

CMakeLists.txt
baothientran: this maybe just an extreme, but I kind of prefer the build system to just explicitly error out if the condition is not met. It looks like CodeCoverage.cmake has quite a lot of those error messages, so I think we can just let it handle those conditions

CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ project(cesium-native
 
 option(PRIVATE_CESIUM_SQLITE "ON to rename SQLite symbols to cesium_sqlite3_* so they won't conflict with other SQLite implemenentations" OFF)
 option(CESIUM_TRACING_ENABLED "Whether to enable the Cesium performance tracing framework (CESIUM_TRACE_* macros)." OFF)
+option(CESIUM_COVERAGE_ENABLED "Whether to enable code coverage" ON)
```
baothientran, line 11: the options maybe need to turn off since some compiler flags are added in the release build

IanLilleyT: ok, I changed it to OFF and that will also fix CI since the script doesn't install gcovr

IanLilleyT:
I removed some of the extra options and it works perfectly on linux and mac as long as `gcovr` is installed. The `CodeCoverage.cmake` fails on MSVC because `gcov` is not found - understandably. So I disable code coverage on Windows

baothientran:
@IanLilleyT can you also update the CHANGE.md? After that, I think it should be ready to merge

CMakeLists.txt
IanLilleyT: The `CodeCoverage.cmake` does it for us already:

```
cmake .. -DCMAKE_BUILD_TYPE=Debug -DCESIUM_COVERAGE_ENABLED=ON
-- Performing Test HAVE_fprofile_abs_path
-- Performing Test HAVE_fprofile_abs_path - Success
-- Appending code coverage compiler flags: -g -fprofile-arcs -ftest-coverage -fprofile-abs-path
CMake Error at extern/cmake-modules/CodeCoverage.cmake:476 (message):
  gcovr not found! Aborting...
```

CMakeLists.txt
IanLilleyT: Yep, `CodeCoverage.cmake` detects missing libs

CMakeLists.txt
IanLilleyT: Mac worked out of the box after removing this if statement

Issue #228, opened 2021-04-27
Generate code for glTF writers, just as we do with the glTF readers


kring:
Currently the writer functions are hand-written. We should generate from the glTF JSON Schema, just as we do with the readers.

kring:
This was done in #416.

PR #294, opened 2021-07-23
Extract JSON writing to non-glTF library

krupkad:
Related to #228, #270, and 3D Tiles serialization. Extracting JSON writing functionality into a non-glTF associated library. Everything here is just renames/moves to a different namespace (`CesiumJsonWriter`).

Split from #291.


CesiumJsonWriter/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+        ${CMAKE_CURRENT_LIST_DIR}/include
+    PRIVATE
+        ${CMAKE_CURRENT_LIST_DIR}/src
+        ${CMAKE_CURRENT_LIST_DIR}/generated
+)
+
+target_link_libraries(CesiumJsonWriter
+    PUBLIC
+        CesiumUtility
+)
```
baothientran, line 46: I just noticed this `cmake` doesn't have install() script like other ones

baothientran:
@krupkad this looks straight forward and nicely decoupling json writer from gltf, which is reasonable since we are going to generate tileset writer soon

Also I just notice that we have two writers class PrettyJsonWriter and JsonWriter which are very identical. The only difference is that they use different rapidjson::Writer<> which is an interface class on its own. Maybe we should remove PrettyJsonWriter? and replace it by having `CesiumJsonWriter::JsonWriter` accepts `rapidjson::Writer` pointer, so that we can pass a compact `rapidjson::Writer` or `rapidjson::PrettyWriter` depends on the use case. A disadvantage of this approach is that rapidjson::Writer will be leaked into the public interface, but I would argue that `rapidjson::Writer` header is already in the header of our `CesiumJsonWriter::JsonWriter`, so people who use our writer is still needs rapidjson itself. This doesn't need to be in this PR though

CesiumJsonWriter/CMakeLists.txt
baothientran: Once this is updated, I think we can merge

kring:
This looks good to me. Just a couple of comments...

> Maybe we should remove PrettyJsonWriter? and replace it by having CesiumJsonWriter::JsonWriter accepts rapidjson::Writer pointer, so that we can pass a compact rapidjson::Writer or rapidjson::PrettyWriter depends on the use case.

I don't think that will work, because `Writer` doesn't use virtual methods. `CesiumJsonWriter::JsonWriter` would have to become a template.

> rapidjson::Writer will be leaked into the public interface, but I would argue that rapidjson::Writer header is already in the header of our CesiumJsonWriter::JsonWriter, so people who use our writer is still needs rapidjson itself.

Because of the nature of RapidJSON, we can't completely avoid exposing it to cesium-native's users. But we can minimize it as much as possible. For example, `PrettyJsonWriter.h` should only `#include <rapidjson/fwd.h>` instead of `#include <rapidjson/prettywriter.h>` and `#include <rapidjson/stringbuffer.h>`. Ideally, we should be able to swap out RapidJSON in the future without anyone noticing. And if they do notice, it should be more along the lines of "oh I was using RapidJSON too, and previously you were including it for me, but now I need to include it myself". It shouldn't be part of the public API without a very good reason.

baothientran:
yeah I just check again, and it's indeed not gonna work without virtual

PR #295, opened 2021-07-24
Metadata tweaks

kring:
After merging #281, I found some small things to fix:

* Don't interpret `extras` or `extensions` as property names, because they're special. Without this change, Cesium for Unreal was throwing assertion failures and log messages when loading Cesium OSM Buildings.
* Add performance tracing to `upgradeBatchTableToFeatureMetadata` (the time spent in there is very reasonable).
* Updated CHANGES.md.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@
 
 ##### Additions :tada:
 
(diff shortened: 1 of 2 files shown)


baothientran:
thanks @kring! I can confirm it fixed the issue

PR #298, opened 2021-07-26
Update draco to fix gcc 11 build

lilleyse:
Updates Draco to the latest commit so that Cesium Native can compile in gcc 11. This brings in fixes from https://github.com/google/draco/pull/636 and https://github.com/google/draco/pull/708. The second fix hasn't been included in a tagged release yet which is why we just grabbed the latest commit from `master`.

In the latest version of draco the target is called `draco_static` on linux and `draco` on windows. This is what most of the cmake changes are for.

The only problem now is compile warnings. I haven't had any luck suppressing these warnings in gcc.

```
In file included from /home/slilley/Code/cesium-native/extern/draco/src/draco/attributes/attribute_octahedron_transform.cc:19:
/home/slilley/Code/cesium-native/extern/draco/src/draco/compression/attributes/normal_compression_utils.h:321:27: warning: using integer absolute value function 'abs' when argument is of floating point type [-Wabsolute-value]
    const float x = 1.f - abs(y) - abs(z);
                          ^
/home/slilley/Code/cesium-native/extern/draco/src/draco/compression/attributes/normal_compression_utils.h:321:27: note: use function 'std::abs' instead
    const float x = 1.f - abs(y) - abs(z);
                          ^~~
                          std::abs
/home/slilley/Code/cesium-native/extern/draco/src/draco/compression/attributes/normal_compression_utils.h:321:36: warning: using integer absolute value function 'abs' when argument is of floating point type [-Wabsolute-value]
    const float x = 1.f - abs(y) - abs(z);
                                   ^
/home/slilley/Code/cesium-native/extern/draco/src/draco/compression/attributes/normal_compression_utils.h:321:36: note: use function 'std::abs' instead
    const float x = 1.f - abs(y) - abs(z);
                                   ^~~
                                   std::abs
2 warnings generated.
```

In `decodeDraco.cpp` I tried:

```
#ifdef _MSC_VER
#pragma warning(push)
#pragma warning(disable : 4127 4018 4804)
#endif
#ifdef __GNUC__
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wabsolute-value"
#endif

#include <draco/compression/decode.h>
#include <draco/core/decoder_buffer.h>

#ifdef _MSC_VER
#pragma warning(pop)
#endif
#ifdef __GNUC__
#pragma GCC diagnostic pop
#endif
```


kring:
Thanks @lilleyse. We don't need to worry about compiler warnings while building third-party libraries. If we're getting warnings in our CPP files from including third-party headers, we do have to deal with that, though.

PR #299, opened 2021-07-27
Extract extension handling

krupkad:
Related to #270 and 3D Tiles serialization. Factoring out the extension handling machinery in `CesiumGltfReader` for more general use (e.g. 3D tiles).

These are mostly copies of the same-named files in `CesiumGltf`/`CesiumGltfReader` up to some renaming. The exceptions are:
- Addition of `ExtensionContext.cpp/.h`, which is the subset of functionality in `GltfReader` that does extension handling.
- Removal of `ReaderContext`, which was an (unnecessary?) coupling of `GltfReader` to the extension machinery

The glTF versions can eventually (or now?) be removed and replaced with this common version.


kring:
This new library, `CesiumJsonExtensions`, has two types of things in it:

1. `ExtensibleObject`, which has virtually no dependencies except for `JsonValue`, which is in `CesiumUtility`.
2. `ExtensionContext`, `ExtensibleObjectJsonHandler`, `ExtensionsJsonHandler`, and `IExtensionJsonHandler`, all of which are specific to _reading_ and could go in `CesiumJsonReader`.

So my feeling is that all the (2) stuff should go in `CesiumJsonReader`, while `ExtensibleObject` goes into `CesiumUtility`. I can also see some wisdom in introducing a `CesiumJson` library, and moving `ExtensibleObject` plus `JsonValue.h` and `JsonHelpers.h` from CesiumUtility into there instead.

It might also make sense to rename `ExtensionContext` so that its name captures the fact that it is specific to reading extensions.

I don't think we need a library specific to extensions.

krupkad:
@kring agreed on the reader-only parts and `ExtensibleObject` - moved those as appropriate. My concern on `ExtensionContext` is that the 3D Tiles writer PR is going to be adding writer registration to it too - unsure where it would belong at that point.

kring:
> My concern on ExtensionContext is that the 3D Tiles writer PR is going to be adding writer registration to it too - unsure where it would belong at that point.

I think I would just have a separate `ExtensionReaderContext` and `ExtensionWriterContext`. I don't think there's much overlap between them.

kring:
Thanks @krupkad, this looks good. Just two requests:

* Remove the CesiumGltf versions of this code.
* Update CHANGES.md to mention the breaking change of moving these types to their new home.

Unless there's a reason you wanted to do these steps in a separate PR?

krupkad:
@kring thanks,
> Unless there's a reason you wanted to do these steps in a separate PR?

Yes, the existing glTF reader classes inherit from `CesiumGltfReader::ExtensibleObjectJsonHandler` whose constructor takes `CesiumGltfReader::ReaderContext` which is involved in the `GltfReader` coupling there. So the changes to do that migration at minimum require touching all of the reader classes, and some light refactoring of the glue code wrapping them.

baothientran:
This looks good to me!

PR #300, opened 2021-07-28
Default constructors for views, IsMetadataInteger/Floating.

kring:
Some small changes to the metadata system to support CesiumGS/cesium-unreal#545:

* Add default constructors for `MetadataPropertyView` and `MetadataArrayView`, to reduce the need for std::monostate. You can think of these as views on invalid properties and empty arrays, respectively.
* Added `IsMetadataInteger<T>` and `IsMetadataFloat<T>` to distinguish between the two types of numeric properties.
* Make the `PropertyType` enum not a bitmask, because it's not used as one and as a bitmask it won't fit into a uint8, which is required by Unreal Blueprints enums.


baothientran:

PR #304, opened 2021-08-11
Optionally generate smooth normals when they are missing

nithinp7:
Fixes: https://github.com/CesiumGS/cesium-unreal/issues/510

Also related to: https://github.com/CesiumGS/cesium-unreal/issues/511


javagl:
IIRC the relevant classes and methods are going to be moved to `CesiumGltf`. So in the meantime, just a comment from quickly looking over it: There was some repetition of the `switch (primitive.mode)` part, which I *think* could be avoided with something like https://github.com/CesiumGS/cesium-native/commit/ca467bb2802902ab107f4370b5e4e3e451b04cdc (just a draft, thus, no OR). Of course, this approach of extracting the "index access" into a `std::function` could be taken further, eventually allowing to omit the template parameter of that function, and there are certainly other ways of tackling that, but ... I just wondered whether the "structurally equal" code for both `switch` blocks could somehow be combined. 

(BTW: I'd offer [lambdas like this one](https://github.com/CesiumGS/cesium-native/blob/1711d90b31f4aea5bcf03d94450626eb5013d021/Cesium3DTilesSelection/src/GltfContent.cpp#L483) (or more generally, the lambdas that are passed to `forEachPrimitiveInScene`) as a dedicated functions, because I think it could make sense to have these as "building blocks" that can be applied to a specific primitive, but .,, this may just be my tendency of trying to avoid "long lambdas"...)

javagl:
Doesn't that `static glm::dmat4 gltfToCesiumAxes` cause that infamous "internal compiler error" for you? (Maybe it could be replaced by one of the https://github.com/CesiumGS/cesium-native/blob/8b0041c45204fd1c59735c72ed55eae3474a75e6/CesiumGeometry/src/AxisTransforms.cpp , but I'd have to take a closer look at the dependencies...)

nithinp7:
For the sake of moving flat normal computation to cesium-native, I started implementing a `flattenIndexedAttributes` functionality in full generality that duplicates vertex attributes and modifies gltfs in cesium-native. It quickly introduced several complications, only some of which I could justify spending the time on to overcome:
* Indices can have different types and vertex attributes can have different types and component types.
* Vertex attributes may be interleaved, one-after-another in the same buffer, or separate buffers altogether.

There will be no concrete advantage to moving flat normals to cesium-unreal, although eventually that may be a nice-to-have improvement in cesium-native. Smooth normals on the other hand were straight forward to implement in cesium-native, and provided a concrete benefit to both repos. So I will roll back the flat normal functionality for now and restore the previous client-side implementation in cesium-unreal.

CesiumGltf/src/Model.cpp
```diff
@@ -208,3 +211,366 @@ void Model::merge(Model&& rhs) {
     this->scene = int32_t(this->scenes.size() - 1);
   }
 }
+
+namespace {
+void forEachPrimitiveInMeshObject(
```
baothientran, line 216: Those functions are quite useful. I can imagine you can replace the Unreal gltf loadNode code and the like with those foreach

CesiumGltf/src/Model.cpp
```diff
@@ -208,3 +211,366 @@ void Model::merge(Model&& rhs) {
...
+    const glm::dmat4x4& transform,
+    const Model& model,
+    const Node& node,
+    const Mesh& mesh,
+    std::function<Model::ForEachPrimitiveInSceneConstCallback>& callback) {
```
baothientran, line 221: nitpick from me but I wonder if replacing std::function with a template maybe a little faster than std::function. Std::function may allocate if the lambda is large enough

baothientran (commented):
I still need to review more, but so far so good. Please make sure to add tests

CesiumGltf/src/Model.cpp
baothientran: cc@kring for more opinion

CesiumGltf/src/Model.cpp
```diff
@@ -208,3 +211,366 @@ void Model::merge(Model&& rhs) {
...
+    }
+    break;
+
+  default:
+    return;
+  }
+
+  // normalizes the accumulated vertex normals
+  for (size_t i = 0; i < count; ++i) {
+    normals[i] = glm::normalize(normals[i]);
```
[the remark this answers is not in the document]
javagl: `NaN` would be the best case. Crashes _might_ be the result. It's tremendously hard to find reliable information there. E.g. https://github.com/g-truc/glm/blob/master/manual.md#section7_10 says that it ~"might crash", and **explicitly** mentions the case of `normalize` for a vector with length 0.0. **But** we didn't see crashes for that case until now (but only `NaN`). All this might well depend on some compiler settings or so...

CesiumGltf/include/CesiumGltf/Model.h
```diff
@@ -19,6 +22,61 @@ struct CESIUMGLTF_API Model : public ModelSpec {
...
+
+  /** @copydoc Gltf::forEachPrimitiveInScene() */
+  void forEachPrimitiveInScene(
+      int sceneID,
+      std::function<ForEachPrimitiveInSceneConstCallback>&& callback) const;
+
+  /**
+   * @brief Fills in smooth normals for any primitives with missing normals.
+   */
+  void generateMissingNormalsSmooth();
```
baothientran, line 78: I just notice that this generates normal for every gltf primitives in the model. It maybe problematic here since some primitives may requires tangent and bitangent too, which will require flat normals in those cases. For example, imagine a scene has two primitives with both missing normals and tangents too. However, one primitive has material with no-normal mapping. In this case, it doesn't requires tangent and bitangent and therefore, generating smooth normal should be okay. However, another primitive has normal mapping in the material. It will require tangent to be generated and flat normal needs to be used here, instead of a smooth normal

CesiumGltf/src/Model.cpp
kring: Yeah I agree, this would be better as a template. There's no good reason I wrote it as a std::function originally.

CesiumGltf/include/CesiumGltf/Model.h
kring: > some primitives may requires tangent and bitangent too, which will require flat normals in those cases.

I could be missing something obvious, but why do tangents and bitangents necessitate flat normals?

CesiumGltf/src/Model.cpp
nithinp7: @baothientran Just so I understand, should the template callback type only be used for the local static functions, or should it be used for the public `forEachPrimitiveInScene` functions as well?

nithinp7:
@baothientran I added tests and addressed the other review comments I think, can you have another look and let me know if anything else is missing?

javagl:
I think that tests like

    REQUIRE(glm::abs(glm::dot(vertex0Normal, expectedNormal) - 1.0f) < 1e-6f);

([e.g. here](https://github.com/CesiumGS/cesium-native/pull/304/files#diff-ec10cab6946129e6cacb0c19ae978220342535ddb96b1b3a300c1b368f92eed6R404)) do not make sooo much sense. While it ~"may work" in this case, the dot product could be 1.0 for totally different vectors as well. I think that a _component-wise_ comparison (with a certain `#define`d epsilon) would make more sense, e.g. as in https://github.com/CesiumGS/cesium-native/commit/8ddb163f6d0cf1167ea96101b3e153c0d7846229

nithinp7:
@javagl That makes a lot of sense, thank you for the suggestion! I merged your commit into this branch.

javagl:
Coming back to the [comment above](https://github.com/CesiumGS/cesium-native/pull/304#issuecomment-904036718), I just stumbled over `Math::equalsEpsilon` at https://github.com/CesiumGS/cesium-native/blob/raster-overlay-refactoring/CesiumUtility/include/CesiumUtility/Math.h#L141 - but I think that using the built-in `glm` function is fine here, so will just leave that as a pointer.

kring:
I think all or almost all the feedback has been addressed and there's nothing blocking us from merging this, so I'm going to do so. Please write any remaining concerns or suggestions as separate issues.

CesiumGltf/src/Model.cpp
```diff
@@ -208,3 +212,422 @@ void Model::merge(Model&& rhs) {
...
+  size_t normalBufferStride = sizeof(glm::vec3);
+  size_t normalBufferSize = count * normalBufferStride;
+
+  std::vector<std::byte> normalByteBuffer(normalBufferSize);
+  gsl::span<glm::vec3> normals(
+      reinterpret_cast<glm::vec3*>(normalByteBuffer.data()),
+      count);
+
+  int64_t numIndices;
+  std::function<TIndex(int64_t)> getIndex;
```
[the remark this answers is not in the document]
javagl: <sup>Yeah, you're right</sup>

Generating the normals of https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/MetalRoughSpheres (which are ~256k normals, for whatever reason), for 1000 times, takes on average ~8.9 seconds with the template version, and ~10.2 seconds with the `std::function` version (in `RelWithDebInfo`, similar with `Release`). 

An extreme case, for sure (although the cross product is by far the biggest chunk in both, of course), but also a bit disappointing. I would have expected the compiler/inlining to be smarter here :-/

PR #306, opened 2021-08-11
3d tiles writer

krupkad:
Adding a writer for 3D Tiles.

- Added `ExtensionWriterContext` (similar to #299) and tests for it under `CesiumJsonWriter`
- Added the writers themselves under `Cesium3DTilesWriter`
  - All autogenerated
  - Includes tests


kring:
Thanks @krupkad! What's doing the code generation? The comments say it's generate-gltf-classes, but I don't see any changes to that in this PR.

krupkad:
@kring Just updated the description, there's a branch/PR in #307 with that which I separated because it's _extensive_.

Cesium3DTiles/include/Cesium3DTiles/Cesium3DTiles.h
```diff
@@ -0,0 +1,199 @@
...
+    static inline constexpr const char* TypeName = "Property";
+
+    std::optional<BinaryBodyReference> v0;
+    std::optional<std::vector<double>> v1;
+    std::optional<double> v2;
+  };
+  struct CODEGEN_API GlobalPropertyScalar {
+    static inline constexpr const char* TypeName = "GlobalPropertyScalar";
+
+    struct CODEGEN_API Variant0 : public CesiumUtility::ExtensibleObject {
```
baothientran, line 127: I kinda know why we use Variant0 naming here since schema doesn't define the name for it. But it still not a good word for API alone

Cesium3DTiles/include/Cesium3DTiles/Cesium3DTiles.h
```diff
@@ -0,0 +1,199 @@
...
+      static inline constexpr const char* TypeName = "Variant0";
+
+      double byteOffset;
+    };
+
+    std::optional<Variant0> v0;
+    std::optional<std::vector<double>> v1;
+    std::optional<double> v2;
+  };
+  struct CODEGEN_API GlobalPropertyCartesian3 {
```
baothientran, line 137: Nitpick from me, but we should have space between structs

baothientran (commented):
Code generation looks really nice! I just have a couple of comments below

Cesium3DTiles/include/Cesium3DTiles/Cesium3DTiles.h
```diff
@@ -0,0 +1,199 @@
...
+
+    struct CODEGEN_API Variant0 : public CesiumUtility::ExtensibleObject {
+      static inline constexpr const char* TypeName = "Variant0";
+
+      double byteOffset;
+    };
+
+    std::optional<Variant0> v0;
+    std::optional<std::vector<double>> v1;
+    std::optional<double> v2;
```
baothientran, line 135: This may have the same problem with the bounding volume above. The schema specifies GlobalPropertyScalar can only accept one of the properties above. Should using `std::variant` be a better choice in this case? cc @kring for more opinions

baothientran (commented):
That's all I have for the generated code in this round. The only comment I have is that some of the construct in the 3D tiles spec requires one of the several properties to appear at most once (e.g Bounding Volume requires to be a bounding sphere or Box, etc), should we use std::variant in those cases?

Cesium3DTiles/include/Cesium3DTiles/Cesium3DTiles.h
krupkad: Did consider this, my concern was that the reader becomes much more complicated since it has to "read every alternative at the same time", so
- some template/STL gymnastics is required to get each type instantiated as a read destination
- there's no obvious choice when multiple alternatives are valid other than "first in the specified order", which also requires some STL mess to accomplish

Not opposed to doing it that way, just will take a little more work and be a little STL-heavy.

The "multiple `optional`s` approach also covers the semantics of both `oneOf` and `anyOf`, granted that's not relevant for glTF/3DTiles

Cesium3DTiles/include/Cesium3DTiles/Cesium3DTiles.h
```diff
@@ -0,0 +1,199 @@
...
+      UNSIGNED_BYTE,
+      SHORT,
+      UNSIGNED_SHORT,
+      INT,
+      UNSIGNED_INT,
+      FLOAT,
+      DOUBLE
+    };
+
+    double byteOffset;
```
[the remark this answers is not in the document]
krupkad: Issue is that the 3D Tiles schema has `number` for this, which does properly correspond to a `double`. Options are
- change the schema to `integer` like glTF, which is technically a non-standard type (but the generator does handle)
- add logic for handling this special case, which feels "incorrect"

Cesium3DTiles/include/Cesium3DTiles/Cesium3DTiles.h
```diff
@@ -0,0 +1,199 @@
...
+};
+
+struct CODEGEN_API BoundingVolume : public CesiumUtility::ExtensibleObject {
+  static inline constexpr const char* TypeName = "BoundingVolume";
+
+  std::optional<std::vector<double>> box;
+
+  std::optional<std::vector<double>> region;
+
+  std::optional<std::vector<double>> sphere;
```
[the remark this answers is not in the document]
krupkad: Should be possible, but I do think the generated code should be as schema-faithful as possible with minimum "customization", with any custom logic done through other means e.g. inheritance, non-member non-friend utility functions. So maybe in this case we provide e.g. a non-member non-friend `BoundingVolume createBox(const glm::dvec3& center, const glm::dvec3& halfSize);` in `include/Utils.h` and `src/Utils.cpp`.

Cesium3DTiles/include/Cesium3DTiles/Cesium3DTiles.h
krupkad: Agreed but unsure how best to address this - looks like it would be fairly not-easy to override this in generator config

Cesium3DTiles/include/Cesium3DTiles/Cesium3DTiles.h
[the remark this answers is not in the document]
kring: We should change the spec. `integer` is standard in the version of JSON Schema 3D Tiles and glTF use: https://datatracker.ietf.org/doc/html/draft-zyp-json-schema-04

Cesium3DTiles/include/Cesium3DTiles/Cesium3DTiles.h
kring: Without necessarily understanding the full implications, std::variant sounds right to me. It more closely matches the schema semantics and will be more memory efficient as well.

Cesium3DTiles/include/Cesium3DTiles/Cesium3DTiles.h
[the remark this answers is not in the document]
krupkad: Opened CesiumGS/3d-tiles#476.

krupkad:
@baothientran  @kring the schema has been changed so the `byteOffset`s are ints, and the writer now uses `std::variant`.

cesium-concierge:
Thanks again for your contribution @krupkad! 

Cesium3DTiles/include/Cesium3DTiles/Cesium3DTiles.h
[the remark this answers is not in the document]
lilleyse: Related discussion in https://github.com/CesiumGS/cesium-native/pull/362

PR #310, opened 2021-08-13
Added missing include of AsyncSystem in TilesetExternals

javagl:
I _think_ the `AsyncSystem.h` has to be included in the `TilesetExternals`, because it is a (non-pointer, non-reference) member.


baothientran:
Thanks @javagl!

PR #313, opened 2021-08-17
Fixed property name of normal texture info

javagl:
According to [the specification](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#reference-material), the property name for the normal texture should be `normalTexture` (its *type* is called `normalTextureInfo`, but not the property).

This caused a validation error 

    UNEXPECTED_PROPERTY | Unexpected property. | /materials/0/normalTextureInfo

for assets that contained a normal texture.

This PR just fixes that property name.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@
 
 - Added `Future<T>::share`, which returns a `SharedFuture<T>` and allows multiple continuations to be attached.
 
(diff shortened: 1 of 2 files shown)


kring:
The writers have been overhauled in #307, but this looks like a good chance in the meantime. Thanks!

PR #301, opened 2021-07-30
Rename Cesium3DTiles -> Cesium3DTilesSelection

krupkad:
Renaming `Cesium3DTiles` to `Cesium3DTilesSelection` to make room for 3D Tiles serialization under the current naming scheme (`Cesium3DTiles`, `Cesium3DTilesReader` and `Cesium3DTilesWriter`)

Issue #311, opened 2021-08-17
The glTF content cannot be written with glTF writer


javagl:
<sup>Edited based on Kevins comment</sup>

The [`GltfContent`](https://github.com/CesiumGS/cesium-native/blob/2848ea62d406f21e70d32c0d75696df1b5602ad2/Cesium3DTilesSelection/src/GltfContent.cpp#L45) class and the [`QuantizedMeshContent`](https://github.com/CesiumGS/cesium-native/blob/2848ea62d406f21e70d32c0d75696df1b5602ad2/Cesium3DTilesSelection/src/QuantizedMeshContent.cpp#L1167) extend the glTF model with an `extras` entry that contains a string value:

    loadedModel.model.value().extras["Cesium3DTilesSelection_TileUrl"] = url;

The current [`Writer`](https://github.com/CesiumGS/cesium-native/blob/2848ea62d406f21e70d32c0d75696df1b5602ad2/CesiumGltfWriter/src/Writer.cpp#L114) tries to write these extras, it crashes (although the code in [`JsonObjectWriter`](https://github.com/CesiumGS/cesium-native/blob/2848ea62d406f21e70d32c0d75696df1b5602ad2/CesiumJsonWriter/src/JsonObjectWriter.cpp#L90) **looks** like this case should be handled...).

kring:
> As best practice, extras should be an Object rather than a primitive value for best portability.

I read that as "the entire `extras` property value should be an object" rather than "every property inside `extras` should have an object for a value." So by my reading, `"extras": { "Cesium3DTilesSelection_TileUrl": "whatever" }` is fine. Certainly agree it shouldn't crash either way, though.

javagl:
@kring You're right, and what I wrote there initially was remarkably nonsensical. Updated the post accordingly.

lilleyse:
Fixed in https://github.com/CesiumGS/cesium-native/pull/416

PR #315, opened 2021-08-19
Fix tile URL "extra" in glTFs.

kring:
We add an extra to loaded glTFs indicating the URL they were loaded from, `Cesium3DTiles_TileUrl`, which is really handy for debugging. But in #301 it was accidentally renamed to `Cesium3DTilesSelection_TileUrl`. This PR changes it back, so double-clicking tiles in Unreal Engine tells you their URL again.

### DIFF
--- a/Cesium3DTilesSelection/src/GltfContent.cpp
+++ b/Cesium3DTilesSelection/src/GltfContent.cpp
@@ -42,7 +42,7 @@ GltfContent::load(const TileContentLoadInput& input) {
   }
 
   if (loadedModel.model) {
-    loadedModel.model.value().extras["Cesium3DTilesSelection_TileUrl"] = url;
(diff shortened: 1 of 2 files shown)


javagl:
Oddly related: https://github.com/CesiumGS/cesium-native/issues/311 

(I haven't looked yet where this `extra` is used, but maybe it could be changed to an `object` at some point?)

baothientran:
Should this change be on Unreal side?

kring:
It was never changed on the Unreal side when #301 was merged, which is why Unreal lost its ability to tell us which tile was clicked. After this PR native and unreal will be back in sync.

kring:
If you're asking whether we should leave this `Cesium3DTilesSelection_TileURL` and change the Unreal side to match, I say no. It was never meant to be a namespace name.

baothientran:
thanks @kring! it looks good to me

PR #318, opened 2021-08-23
Add Rectangle::computeUnion method and rename intersect to match

kring:
This is a small PR broken out of #252.

* Added `Rectangle::computeUnion` (we can't call it `union` because that's a C++ reserved word).
* Renamed `Rectangle::intersect` and `GlobeRectangle::intersect` to `computeIntersection` (to match).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,9 +2,14 @@
 
 ### Next Release - ?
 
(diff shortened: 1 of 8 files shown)


baothientran:
Thanks @kring! look good to me

PR #319, opened 2021-08-23
Add `Future::isReady`.

kring:
More changes from #252.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 ##### Additions :tada:
 
+- Added `Future<T>::isReady`.
 - Added `Future<T>::share`, which returns a `SharedFuture<T>` and allows multiple continuations to be attached.
 
 ### v0.6.0 - 2021-08-02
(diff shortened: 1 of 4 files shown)


javagl:
Most of this is just exposing the `.ready` method from async++.

But I stumbled over the change [`here`](https://github.com/CesiumGS/cesium-native/pull/319/files#diff-f9a98f0642af6c7e5643f048b31ab70d016988e19b5cd5a5ac5faae925d8e5a3R196), and wondered whether this **might** be a breaking change in some odd corner case. The only thing that I could come up with where this *might* be relevant is something like

    struct TestStruct { int x; };
    void foo(TestStruct& t) { t.x = 123; }
    ...
    auto promise = asyncSystem.createPromise<TestStruct>();
    auto future = promise.getFuture().share();
    promise.resolve(TestStruct());
    foo(future.wait());

where the `foo` call now issues an error about the `const` being dropped, whereas it originally only issued a warning For VS, that was warning was treated as an error because of compilation settings, but I'm not sure about GCC or other compilers...

In any case: I **assume** that this is not an issue, and that this change was intentional and a general improvement, so unless there is a comment/commit that includes the word "Whooopsie" until tomorrow, I'd hit "Merge" soon.

baothientran:
@javagl for that case, the object must be copied first before passing to foo()

baothientran:
I probably go ahead and merge the PR since this looks good to me

Issue #221, opened 2021-04-20
Move image loading into a separate library or class


kring:
Currently, we use `CesiumGltf::GltfReader` to load jpegs, pngs, etc. That's a little odd.

javagl:
<sup>(Triggered by this (and the source, https://github.com/CesiumGS/cesium-native/pull/208#discussion_r615892972 ) being open in some abandoned Browser Tabs for nearly a month now: )</sup>

The solutions for this range from 

- A `CesiumUtilities::ImageLoader` class with a single static `loatImage` function that essentially wraps `stb-image` away. That would be good: It would make dependency management easier and clearer, and would allow to quickly try out other PNG decoders to see whether they are faster (the image deconding is a severe bottleneck right now). But some questions to think about: 1. Would `CesiumGltf*` have a dependency to `CesiumUtilites` then? (probably yes, is this desired?). And 2. we all know that this will not be sufficient in the medium- or long term, so how far will we go towards the second option:
- A `CesiumImages` library. This would allow to define a *clearer* dependency. (I mean, what are `CesiumUtilities` after all?). The API of this library could be relatively small, and *mainly* wrap `stb_image` in the first shot, **but** should already be laid out keeping in mind that this library will/should also support things like KTX2 in the near future.

People who have already looked into KTX2 might give some hints for the second option. KTX2 regularly discussed in the 3dformats calls, but I'm not familiar with it on a technical level. For example, I assume that medium-term, a single, static `ImageLoader::loadThat(rawData)` will **not** be sufficient. There will likely be some sort of configuration object like

    struct ImageReaderParameters { 
        bool decompressThisAndThat; 
        int compressSoManyComponents;
        bool retainThisAndThatInformation; 
        ...
    }

The result may not be a single `CesiumImage` with `uint32_t rgba` pixels, but something that can transport other forms of image data, and maybe even some sort of Mipmap levels or metadata or whatnot.

And ... this issue is talking about image *loading*. Image *writing* will also be a topic (although with much, much lower priority, of course - at least for `cesium-unreal`. The tilers folks might want some `writeKtx(pixels)` function rather sooner than later...),

javagl:
If this was implemented, it would also fix https://github.com/CesiumGS/cesium-native/issues/150

j9liu:
At the very least we should update this issue with a better description, though, so I'll do that now

azrogers:
[...]

PR #321, opened 2021-08-23
Add ImageManipulation class with blitting operations

kring:
Another from #252.

It probably makes sense for this to move to a CesiumImage library or somesuch at some point, but I think it's ok here for the time being.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Additions :tada:
 
 - Added `Future<T>::share`, which returns a `SharedFuture<T>` and allows multiple continuations to be attached.
+- Added `ImageManipulation` class to `CesiumGltfReader`.
 
 ### v0.6.0 - 2021-08-02
 
--- a/CesiumGltfReader/include/CesiumGltf/ImageManipulation.h
+++ b/CesiumGltfReader/include/CesiumGltf/ImageManipulation.h
@@ -1,0 +1,88 @@
+#pragma once
+
+#include "CesiumGltf/ReaderLibrary.h"
+#include <cstddef>
+#include <cstdint>
+
+namespace CesiumGltf {
+
+struct ImageCesium;
+
+/**
+ * @brief Specifies a rectangle pixels in an image.
+ */
+struct PixelRectangle {
+  /**
+   * @brief The X coordinate of the top-left corner of the rectangle.
+   */
+  int32_t x;
+
+  /**
+   * @brief The Y coordinate of the top-left corner of the rectangle.
+   */
+  int32_t y;
+
+  /**
+   * @brief The total number of pixels in the horizontal direction.
+   */
+  int32_t width;
+
+  /**
+   * @brief The total number of pixels in the vertical direction.
+   */
+  int32_t height;
+};
+
+class CESIUMGLTFREADER_API ImageManipulation {
+public:
+  /**
+   * @brief Directly copies pixels from a source to a target, without validating
+   * the provided pointers or ranges.
+   *
+   * @param pTarget The pointer at which to start writing pixels.
+   * @param targetRowStride The number of bytes between rows in the target
+   * image.
+   * @param pSource The pointer at which to start reading pixels.
+   * @param sourceRowStride The number of bytes between rows in the source
+   * image.
+   * @param sourceWidth The number of pixels to copy in the horizontal
+   * direction.
+   * @param sourceHeight The number of pixels to copy in the vertical direction.
+   * @param bytesPerPixel The number of bytes used to represent each pixel.
+   */
+  static void unsafeBlitImage(
+      std::byte* pTarget,
+      size_t targetRowStride,
+      const std::byte* pSource,
+      size_t sourceRowStride,
+      size_t sourceWidth,
+      size_t sourceHeight,
+      size_t bytesPerPixel);
+
+  /**
+   * @brief Copies pixels from a source image to a target image.
+   *
+   * If the source and target dimensions are not the same, the image is resized.
+   * Otherwise, the source pixels are exactly copied into the target. Resizing
+   * is only supported when bytesPerChannel is 1.
+   *
+   * The images must have the same number of channels and bytes per channel, and
+   * the provided rectangles must be valid for the images. If any of these
+   * constraints are violated, this function will return false and do nothing.
+   *
+   * @param target The image in which to write pixels.
+   * @param targetPixels The pixels to write in the target.
+   * @param source The image from which to read pixels.
+   * @param sourcePixels The pixels to read from the target.
+   * @returns True if the source image was blitted successfully into the target,
+   * or false if the blit could not be completed due to invalid ranges or
+   * incompatible formats.
+   */
+  static bool blitImage(
+      ImageCesium& target,
+      const PixelRectangle& targetPixels,
+      const ImageCesium& source,
+      const PixelRectangle& sourcePixels);
+};
+
+} // namespace CesiumGltf
--- a/CesiumGltfReader/include/CesiumGltf/ImageManipulation.h
+++ b/CesiumGltfReader/include/CesiumGltf/ImageManipulation.h
@@ -9,7 +9,7 @@ namespace CesiumGltf {
 struct ImageCesium;
 
 /**
- * @brief Specifies a rectangle pixels in an image.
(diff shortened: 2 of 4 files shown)


javagl:
As you mentioned, the class does not really belong there, but this is, to some extent, already tracked via https://github.com/CesiumGS/cesium-native/issues/221 . The class is currently not used, but is probably supposed to be used in the raster overlay area. If this is right, then one could make a case for moving it to `Cesium3DTilesSelection`, where it could indeed be a _private_ class, without the `CESIUMGLTFREADER_API` specififier. Right now, moving it into an image library would be a breaking change. 

---

The implementation of the `unsafeBlitImage` looks OK. 

---

For the `blitImage`, there are 5 different `return false` cases. If I was supposed to use this function, I'd **really** appreciate better error handling. One part of this would be to move the validity checks into an own function, so that `blitImage` would **roughly** look like this:
```
bool blitImage(
    Image target, Rectangle targetPixels, 
    Image source, Rectangle sourcePixels) {
    if (!isValid(target, targetPixels, source, sourcePixels)) {
        return false;
    }
    blitImageKnowningEverythingIsValid(....);
}
```
But more importantly: More fine-grained checks and logging messages (!) like
```
  if (sourcePixels.x < 0 || sourcePixels.y < 0 || sourcePixels.width < 0 ||  sourcePixels.height < 0) {
     LOG("Negative input in source: %d %d %d %d", ...);
     return false;
  }
```
could safe **hours** of debugging. This is what makes the difference between "It did not show the image, and I tried to figure out whether it was a wrong `fileName.png` or a bug in the Unreal engine", and "It did not show the image, and I immediately knew that it was because of computing `source.x = min-max` instead of `source.x = max-min`".

---

I looked at the tests. I _think_ that it does not cover cases where `rowStride != width*bytesPerPixel` (but maybe that is intentionally not considered?)

---

I tried to understand the [`verifySuccessfulCopy`](https://github.com/CesiumGS/cesium-native/pull/321/files#diff-39cea7779bb531a79c9be10e4b092063ced7453b8094f07972a826bbebac2aacR169), and think that a `PixelRectangle::contains` method could be handy, and allow this to be implemented a bit simpler, **roughly** (drafted!) as in
```
  // Use size_t or int32_t here ----------------------v ? The latter, probably...
  auto contains = [](const PixelRectangle& rectangle, size_t x, size_t y) {
    if (x < rectangle.x) {
      return false;
    }
    if (y < rectangle.y) {
      return false;
    }
    if (x >= rectangle.x + rectangle.width) {
      return false;
    }
    if (y >= rectangle.y + rectangle.height) {
      return false;
    }
    return true;
  };

  auto verifySuccessfulCopy = [&target, &targetRect, &contains]() {
    size_t targetBytesPerPixel(target.bytesPerChannel * target.channels);
    for (size_t j = 0; j < size_t(target.height); ++j) {
      for (size_t i = 0; i < size_t(target.width); ++i) {

        size_t index(j * target.width + i);
        size_t offset(index * targetBytesPerPixel);
        std::byte expected = contains(targetRect, i, j) ? std::byte(2) : std::byte(1);
        for (size_t k = 0; k < targetBytesPerPixel; ++k) {
          CHECK(target.pixelData[offset + k] == expected);
        }
      }
    }
  };
```
but this (and the role of a stride != `width*bytes`) would have to be reviewed more closely as well.

kring:
Thanks Marco.

> The class is currently not used, but is probably supposed to be used in the raster overlay area.

Yep. Originally it was just private where it was needed. I broke it out and made it public because I was going to use it to create power-of-two sized textures on the Unreal side, too. I ended up deciding I don't need to do that in the short term, though.

>  If this is right, then one could make a case for moving it to Cesium3DTilesSelection, where it could indeed be a private class, without the CESIUMGLTFREADER_API specififier. Right now, moving it into an image library would be a breaking change.

I don't think it matters and I'd rather avoid the churn. If I'm going to move this functionality around, I'd rather just move it into CesiumImage or wherever we think it will go in the end. But mostly I'd rather than move it around right now. We're not too concerned about breaking changes in cesium-native at the moment.

> I'd really appreciate better error handling

I hear you, but this is way too low-level for logging. Assertions could be an option. In another context exceptions would be an obvious choice. A more detailed error code rather than success/failure might be nice. But given the low-level nature and the fact that any occurrence of "false" is basically a developer error (API misuse, precondition violation), I think simply returning false and allowing the caller to assert/throw is sensible enough. If a user sees the function return false and doesn't know why, hopefully it isn't hours of debugging, but just minutes to set a breakpoint on each of the `return false` lines and see which one is being hit.

> I think that it does not cover cases where rowStride != width*bytesPerPixel

Maybe I don't know what you mean, but I think `succeeds for non-scaled blit` covers this. The copied subset is smaller than either the source or target image, so `width * bytesPerPixel` won't equal either the source or target row stride.

If you're referring to the `unsafeBlitImage`, then `ImageManipulation::unsafeBlitImage subset of source` should have the same property.

> I tried to understand the verifySuccessfulCopy, and think that a PixelRectangle::contains method could be handy

Ok, sure, your code is a little easier to understand (even if it's twice as long), so I've changed it.

I'm not sure if you were suggesting `PixelRectangle::contains` should be public, rather than the lambda you've written... but I don't see much value there. Outside of test code, testing individual pixels for containment in a rectangle is a horribly inefficient way to go about solving any problem I can think of. So I've left it a lambda.

javagl:
> logging [...] Assertions [...] exceptions [...] error code

Yeah, these are the usual ones, and the "best" choice depends on many questions that are both obvious and hard to answer. I agree that logging might be an overkill _in this particular case_. A "clean" solution would be to `throw std::std::invalid_argument("A message telling you EXACTLY what is wrong (not just 'Something is wrong')")`, but ... maybe we won't even have exceptions, eventually...

However, some responsibility **can** be moved to the caller (and this is the case for exceptions as well). So when I find some 

    if (!blitImage(...)) {
        return; // Just ignore that....
    }

in the code that uses this class, then I'll scream, loudly: INSERT SOME LOG MESSAGE THERE!!!111 😉 

---

> Maybe I don't know what you mean, but I think succeeds for non-scaled blit covers this. 

Maybe that was a bit overthinking, but in some contexts, there are images where each *row* has certain alignment requirements (e.g. each starting address of a row must be a multiple of 64 or so - and this is not directly related to (but still similar to) the usual power-of-two issue). In these cases, there may be an image that is 53 pixels wide, but has a "row stride" of 64 pixels. 

But now, I think that this may not even matter here, because the `CesiumImage` only has a `width` field, but no `rowStride` field - so this cannot be represented anyhow.  (Maybe it should, at some point, in a `CesiumImageLibrary`, but ... that's a different story).

---

> I'm not sure if you were suggesting PixelRectangle::contains should be public,

To some extent, yes. More generally, the current `PixelRectangle` could be an `IntRectangle` and have largely the same structure and methods as [`CesiumGeometry::Rectangle`](https://github.com/CesiumGS/cesium-native/blob/main/CesiumGeometry/src/Rectangle.cpp), including a `contains(ivec2)` method. One could even go so far and create a templated `Rectangle<T>`, with specializations where necessary. 

But for now, the `PixelRectangle` is only supposed to be used for that single class, and the `contains` is used only in the test, so keeping the scope of that as small as possible is fine. (It could be a function in an anonymous namespace, but that doesn't matter so much).

---

Speaking of "places where things should be used":

I wondered whether `PixelRectangle` would have to be declared `CESIUMGLTFREADER_API` to be used externally. But apparently, this is not the case. But what's confusing for me right now: Even when **removing** the `CESIUMGLTFREADER_API` specifier from the `ImageManipulation` class, and then using that class in `cesium-unreal`, there are no "Unresolved external..." linker errors - and I don't exactly understand why this works. I think there *should* be errors... 

BTW: When trying to use this branch _directly_ in `cesium-unreal`, there are some linker errors about duplicated symbols, due to the `STB_IMAGE_RESIZE_IMPLEMENTATION` at https://github.com/CesiumGS/cesium-unreal/blob/7bb30ff8cba5315dccc41e7505e5744f2e81f02b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp#L44 **and** in the new class. But I guess that will be sorted out in the branch where this class will be used...

javagl:
(BTW: All the previous comments are not a reason to not merge this one, and I'd do this not later than tomorrow)

PR #328, opened 2021-08-31
Raster overlay refactoring fix compiler error

javagl:
As mentioned in https://github.com/CesiumGS/cesium-native/commit/acd276388a4bda5426ddb72d2fd17c18241c7ab9#r55708457  (and earlier, but apparently, the comment was overlooked)

### DIFF
--- a/Cesium3DTilesSelection/src/GltfContent.cpp
+++ b/Cesium3DTilesSelection/src/GltfContent.cpp
@@ -1,5 +1,6 @@
 #include "Cesium3DTilesSelection/GltfContent.h"
 #include "Cesium3DTilesSelection/spdlog-cesium.h"
+#include "CesiumGeometry/AxisTransforms.h"
 #include "CesiumGltf/AccessorView.h"
 #include "CesiumGltf/AccessorWriter.h"
 #include "CesiumGltf/Model.h"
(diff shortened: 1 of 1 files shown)


nithinp7:
Oh that's right I will merge this right away. For my benefit, how is that internal compiler error reproduced?

Also is there a reason this targets `raster-overlay-refactor` instead of main? I believe I introduced this code during the "generate smooth normals" PR. Is it just to avoid more merges from main in the `raster-overlay-refactor` branch?

nithinp7:
I know it may not be necessary to support out-of-date compiler versions, but I think this is more concise and less redundant anyhow. Thanks @javagl!

javagl:
It appears in VS Version 16.7.7, which is not terribly outdated.... (but it's probably hard to reproduce if you have a newer version where it does not appear any more). Regardless of that, I think that using the existing constant (with its name saying what the transform is) is reasonable.

In the normals branch, I had usually fixed this locally, and mentioned it at https://github.com/CesiumGS/cesium-native/pull/304#issuecomment-900428677 , but I might not have checked the state after the normals branch as merged to main. 
Now I fixed this in the `raster-overlay-refactoring` branch, because there it became apparent for me (it's merged in the meantime - otherwise, I could have done that based on `main`, but the overlay branch is supposed to be merged soon anyhow...)

Issue #314, opened 2021-08-19
External tilesets don't work right with enableFrustumCulling=false and enforceCulledScreenSpaceError=false


kring:
Originally reported here with Cesium OSM Buildings, which uses external tilesets extensively:
https://community.cesium.com/t/runtime-created-cesium-tileset-with-material-osm-buildings-for-example-doesnt-render/14834/2

In the tile selection algorithm, we treat external tilesets as if they have a really huge geometric error. This guarantees that we never select them, but instead always refine past them. At least, it usually does.

A problem comes up when enableFrustumCulling is set to false. In this scenario, tiles that would otherwise be culled will be included in the render list. When enforceCulledScreenSpaceError is also false, we also will not (ever!) refine such tiles. And this is true even for external tilesets with their huge geometric error. Thus, we'll end up selecting an external tileset for rendering, even though it's not a thing we can render.

Now this alone wouldn't be a _huge_ problem. It would just be a tile in the render list without any renderable content and we could just ignore it. But because `_meetsSse` is returning true for these culled external tileset tiles (and this is the _only_ situation in which it will), we add it to the render list with `_renderInnerTile`. And `_renderInnerTile` will note that the tile's `isRenderable` method returned false (which is fair enough, it's not renderable). Because this tile isn't renderable, and never has been, the next level up in the selection traversal will kick this tile _and its siblings_ out of the render list, waiting for them all to become renderable, presumably by further loading. But an external tileset will never become renderable no matter how long we wait for it to load.

End result: we don't render this whole branch of the tile tree with these settings.

javagl:
It may not strictly be necessary, but the keywords reminded me of the discussion around https://github.com/CesiumGS/cesium-native/issues/126#issuecomment-774352476 , so having this cross-ref here might be helpful in the future.

PR #335, opened 2021-09-03
Fixed external tilesets not loading when frustum culling was disabled

nithinp7:
Fixes: #314

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Fixed a bug introduced in v0.7.0 where Bing credits were not being collected.
+- Fixed a bug where disabling frustum culling caused external tilesets to not load.
 
 ### v0.7.0 - 2021-09-01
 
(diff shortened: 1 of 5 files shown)


kring:
This looks good, thanks @nithinp7! One idea to run past you, though: what do you think about using `std::numeric_limits<double>::infinity()` as the geometric error to mark a tile for unconditional refinement, rather than the bool? `glm::isinf` can be used to test for it. The advantage is less data in the Tile, which means (a little bit) less CPU cache thrashing during traversal. The disadvantage is perhaps it's a little less clear, but the `get/setUnconditionallyRefine` functions go a long way there.

nithinp7:
@kring That's a good idea! I hadn't know that using infinity was idiomatic and safe, but it makes a ton of sense.

nithinp7:
And by that last part, you mean keep the getter/setter there but change it so that it packs and unpacks infinity from the geometric error right?

kring:
Yeah that's what I had in mind.

kring:
I guess `setUnconditionallyRefine(false)` doesn't really make sense anymore, but I don't think it's needed, either. Just change it to `setToUnconditionallyRefine()` (parameterless) or somesuch.

kring:
Thanks @nithinp7!

PR #336, opened 2021-09-03
Fix for "Forbid Holes" not working

nithinp7:
There was an issue (may or may not be related to [this issue](https://github.com/CesiumGS/cesium-unreal/issues/440)) where enabling "Forbid Holes" completely broke the loading of CWT and probably any other tileset. To reproduce simply load up CWT, check "Forbid Holes", and refresh the tileset.

Why I think it was happening:
Children are initially not considered "renderable" so the parent tile refuses to refine.  I've seen the children have placeholder loading raster tiles (not mapped yet) and I've seen the children have loaded raster tiles that weren't yet put into the ready slot. In both cases they're considered not renderable so the parent does not refine. While the parent is waiting, `Tile::loadContent` gets called on the children, but there is never a `Tile::update` call and so sometimes the raster to geometry mapping is missed (when the tile provider was a placeholder during `Tile::loadContent`?) and `RasterMappedTo3DTile::update` is never called (so even loaded raster tiles are never set to ready).

So since the children were not being considered renderable and they were not making any progress towards being renderable, "Forbid Holes" forced the parents to wait indefinitely for the children.

I'm only describing the problem in this detail so someone can verify that this is a reasonable fix 🙂.

Edit: I also added a fix here for the issue of forbid holes not working with external tilesets.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Fixed a bug introduced in v0.7.0 where Bing credits were not being collected.
+- Fixed the "forbidHoles" option not working when frustum culling was enabled and when external tilesets were used.
 
 ### v0.7.0 - 2021-09-01
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@
 ##### Fixes :wrench:
 
 - Fixed a bug introduced in v0.7.0 where Bing credits were not being collected.
-- Fixed the "forbidHoles" option not working when frustum culling was enabled and when external tilesets were used.
+- Fixed a bug where the "forbidHoles" option was not working with raster overlays and external tilesets.
 
 ### v0.7.0 - 2021-09-01
 
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tile.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tile.h
@@ -465,6 +465,11 @@ public:
   bool isRenderable() const noexcept;
 
   /**
+   * @brief Determines if this tile is has external tileset content.
+   */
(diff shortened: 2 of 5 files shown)


kring:
This makes sense to me. I think it would be reasonable to unconditionally call `update` just before the `_markTileVisited`, just as we do when we `_visitTileIfNeeded` in the normal mode. Or did you see a problem doing that?

nithinp7:
@kring I believe I tried that and it _seemed_ to work but it caused one of the tests to fail:

test-fail

Basically the child tile should fail temporarily when a network request fails. But for some reason (maybe a race condition), if I call update before the child is loaded, the child tile seems to fail permanently. I don't know if I just buried the symptoms with the conditional call to update or if I actually fixed the issue though.

baothientran:
Nithin for the forbid hole, it fixed the CWT, but it's still a problem for the regular tileset. You can try the Boston nearmap asset ID 354759. I think that can be a separate PR though. Doesn't need to block this one

kring:
I guess `setUnconditionallyRefine(false)` doesn't really make sense anymore, but I don't think it's needed, either. Just change it to `setToUnconditionallyRefine()` (parameterless) or somesuch.

kring:
> if I call update before the child is loaded, the child tile seems to fail permanently.

I think this is a feature not a bug. `update` is supposed to transition from `FailedTemporarily` to `Failed` when there's no `failedTileCallback` to handle it otherwise. So it didn't make sense for the children to remain in the FailedTemporarily state, and test was asserting the presence of an incorrect result. We just need to change the test.

nithinp7:
@baothientran I was able to fix the issue for external tilesets as well just now. Feel free to try it again if you have time to review.

kring:
All looks good to me now. @baothientran I'll let you merge when you're happy with it.

PR #337, opened 2021-09-07
Refactored MaximumCachedBytes

shikharvashistha:
Greetings of the day @kring,

Reference : https://github.com/CesiumGS/cesium-unreal/pull/635

### DIFF
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetOptions.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetOptions.h
@@ -157,7 +157,7 @@ struct CESIUM3DTILESSELECTION_API TilesetOptions {
    * unloaded until the total is under this number or until only required tiles
    * remain, whichever comes first.
    */
-  int64_t maximumCachedBytes = 512 * 1024 * 1024;
+  int64_t MaximumCachedBytes = 256 * 1024 * 1024;
 
   /**
    * @brief A table that maps the camera height above the ellipsoid to a fog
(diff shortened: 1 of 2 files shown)


Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetOptions.h
```diff
@@ -157,7 +157,7 @@ struct CESIUM3DTILESSELECTION_API TilesetOptions {
    * unloaded until the total is under this number or until only required tiles
    * remain, whichever comes first.
    */
-  int64_t maximumCachedBytes = 512 * 1024 * 1024;
+  int64_t MaximumCachedBytes = 256 * 1024 * 1024;
```
kring, line 160: This should remain lowercase, consistent with the cesium-native coding standards.

shikharvashistha: Done !

shikharvashistha:
Greetings of the day @kring,

I think [this](https://github.com/CesiumGS/cesium-native/pull/337) needs to be merged first then a rerun of CI would be good ?

Thanks & Regards 
Shikhar Vashistha

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetOptions.h
baothientran: should we still keep the default value 512 mb in cache since this property is now exposed in the Unreal side, so that user can adjust their cache depend on their use case? Changing it lower may worsen the performance for some users

baothientran:
@kring since the PR on Unreal seems to be enough to let user customize their cache size, should we close this one?

kring:
Yes, I think so. THanks for the PR @shikharvashistha but I think we'll leave the cesium-native default as it is for now.

PR #338, opened 2021-09-07
Make tile content loading asynchronous

nithinp7:
Changes:
- Made the `TileContentLoader` interface future-based so that the tile loading can happen asynchronously. 
- Added the `AsyncSystem`, `IAssetAccessor`, and `IAssetRequest` to the `TileContentLoadInput` struct so that tile content loading can be asynchronous, and so further network requests can be made during the tile content loading (ion or otherwise).


nithinp7:
@kring Sorry for the delay, I was worried I introduced #341 here, but I ended up being able to reproduce that in main. So this should be ready for review.

Cesium3DTilesSelection/src/CompositeContent.cpp
```diff
@@ -97,69 +106,79 @@ CompositeContent::load(const TileContentLoadInput& input) {
...
   uint32_t pos = sizeof(CmptHeader);
 
-  for (uint32_t i = 0; i < pHeader->tilesLength; ++i) {
+  for (uint32_t i = 0; i < pHeader->tilesLength && pos < pHeader->byteLength;
+       ++i) {
     if (pos + sizeof(InnerHeader) > pHeader->byteLength) {
       SPDLOG_LOGGER_WARN(
           pLogger,
           "Composite tile ends before all embedded tiles could be read.");
+      pos = static_cast<uint32_t>(pHeader->byteLength);
```
kring, line 123: I might just be missing it, but I can't see what purpose this line serves?

Cesium3DTilesSelection/src/CompositeContent.cpp
```diff
@@ -97,69 +106,79 @@ CompositeContent::load(const TileContentLoadInput& input) {
...
+    innerTiles.push_back(TileContentFactory::createContent(
+        asyncSystem,
+        pAssetAccessor,
+        requestHeaders,
+        derive(input, innerData)));
   }
 
-  return pResult;
+  return asyncSystem.all(std::move(innerTiles))
+      .thenInMainThread(
```
kring, line 147: This doesn't need to be done in the main thread, does it? `thenInWorkerThread` would be better, I think.

Cesium3DTilesSelection/src/Tile.cpp
```diff
@@ -259,14 +261,43 @@ void Tile::loadContent() {
...
       .thenInWorkerThread(
-          [loadInput = std::move(loadInput),
+          [handleLoadResult,
+           handleLoadError,
+           loadInput = std::move(loadInput),
            projections = std::move(projections),
            gltfUpAxis,
            pPrepareRendererResources =
                tileset.getExternals().pPrepareRendererResources,
+           &asyncSystem = tileset.getAsyncSystem(),
```
kring, line 298: Better to pass AsyncSystem by value.

Cesium3DTilesSelection/src/Tile.cpp
```diff
@@ -306,74 +337,80 @@ void Tile::loadContent() {
...
-              }
-            }
-
-            LoadResult result;
-            result.state = LoadState::ContentLoaded;
-            result.pContent = std::move(pContent);
-            result.pRendererResources = pRendererResources;
 
-            return result;
+            TileContentFactory::createContent(
```
kring, line 341: You should return the Future returned by createContent. That way the outer Future doesn't resolve until the one created by createContent does. I think that will greatly simplify this method (i.e. you don't need the two levels of handling result/errors).

kring (commented):
Looks good @nithinp7, just a few suggestions.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileContentFactory.h
```diff
@@ -92,13 +93,20 @@ class CESIUM3DTILESSELECTION_API TileContentFactory final {
...
    * no loader registered for the magic header of the given
    * input, and no loader for the content type of the input.
    */
-  static std::unique_ptr<TileContentLoadResult>
-  createContent(const TileContentLoadInput& input);
+  static CesiumAsync::Future<std::unique_ptr<TileContentLoadResult>>
+  createContent(
+      const CesiumAsync::AsyncSystem& asyncSystem,
+      const std::shared_ptr<CesiumAsync::IAssetAccessor>& pAssetAccessor,
+      const std::vector<std::pair<std::string, std::string>>& requestHeaders,
```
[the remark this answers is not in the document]
nithinp7: I was thinking about that too, but I was a little worried about needless copies and the overhead of the atomic reference count increment with the shared_ptrs. For instance: 
- Should the AsyncSystem be wrapped with a shared_ptr in TilesetExternals, or should AsyncSystem be kept as a reference in TileContentLoadInput?
- I assume the pAssetAccessor can be a shared ptr without issue.
- How would I avoid making a copy of the request headers? Could I even move it into TileContentLoadInput? I assumed not since the Tile may fail at loading and try again, possibly needing the request headers again.

Cesium3DTilesSelection/src/Tile.cpp
nithinp7: Wouldn't that end up becoming a "future of a future"? Or will that automatically be unwrapped twice in the result-handler?

kring: It'll get unwrapped. See https://github.com/Amanieu/asyncplusplus/wiki/Tasks#task-unwraping and replace "task" with "future". :)

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileContentFactory.h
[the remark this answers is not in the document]
kring: > Should the AsyncSystem be wrapped with a shared_ptr in TilesetExternals, or should AsyncSystem be kept as a reference in TileContentLoadInput?

AsyncSystem should be stored by value in TileContentLoadInput.

> I assume the pAssetAccessor can be a shared ptr without issue.

Yep

> How would I avoid making a copy of the request headers? Could I even move it into TileContentLoadInput? I assumed not since the Tile may fail at loading and try again, possibly needing the request headers again.

The `data` is already passed by reference (via a span) to TileContentLoadInput, so if that is ok then it should be ok to do the same with the headers. But actually that's a little worrying. I don't think anything is currently guaranteeing that the lifetime of that data (or headers) is extended to the end of the Future returned by the loader. So if the loader continues to access the data/headers after the initial load function returns and the `std::shared_ptr<IAssetRequest>` passed to the continuation goes out of scope, then it'll be undefined behavior.

An easy and efficient - but conceptually messy - solution here is to std::move the entire `std::shared_ptr<IAssetRequest>` into the `TileContentLoadInput`. Then we can remove `data`, `contentType`, and `url`, and we don't need to add `headers`. That'll save some allocating/copying. It's messy because ideally content loading wouldn't assume that the content always came from an asset request. Given that IAssetRequest is polymorphic, though, it's not totally disastrous. For example, we could easily write an `IAssetRequest` that represents an entirely in-memory asset. So I think that might be the best solution, barring a better idea that we haven't thought of yet. :)

nithinp7:
To get this PR caught up with the current issues that were discussed offline:
- Currently, this lambda mysteriously cannot be declared mutable which forces us to take a weird work-around for both passing along the status code as well as making shared_ptr copies. https://github.com/CesiumGS/cesium-native/blob/84a6e708b1206f4840b0f8306ed6e42eadc353b3/Cesium3DTilesSelection/src/Tile.cpp#L267-L272
- A Tileset traversal test which mocks failed tiles by setting their status code to 404 is causing a segfault in this branch.

kring (commented):
Not a full review, but a couple quick things I noticed. There's a chance that the capture-by-reference is the cause of the access violation in the tests.

kring:
Actually the test segfault is more mundane than that. Looks like the Future returned by `TileContentFactory::createContent` is resolving to a nullptr. So then `pContent->httpStatusCode = statusCode;` dereferences a null pointer and crashes.

Cesium3DTilesSelection/src/Tile.cpp
```diff
@@ -257,62 +259,88 @@ void Tile::loadContent() {
...
   std::move(maybeRequestFuture.value())
-      .thenInWorkerThread(
-          [loadInput = std::move(loadInput),
+      // TODO: for some reason, this first lambda completely breaks if we make
+      // it mutable
+      .thenInWorkerThread([this,
+                           &asyncSystem = tileset.getAsyncSystem(),
+                           &pLogger_ = tileset.getExternals().pLogger,
+                           &pAssetAccessor_ =
+                               tileset.getExternals().pAssetAccessor](
```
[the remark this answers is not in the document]
nithinp7: Since the lambda can be mutable now I can pass the shared_ptr by value now. But as for the "this" pointer, I thought a Tile was guaranteed to not be deleted when its content was loading? Or is that not always the case?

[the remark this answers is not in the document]
kring: That's true, the Tile won't be deleted while the content is loading. But passing `this` is still dangerous because the Tile is not thread safe and so _using_ the `this` is unsafe. And it's especially easy to do that accidentally because `this` is implicit. So best thing is to avoid passing non-thread-safe objects to the worker thread lambda in general, but if you need to you should at least avoid naming them `this`. e.g. `[pTile = this, ...]` is an improvement.

nithinp7:
@kring I just finished addressing all review comments so far, I think this is ready for a final review when you have time.

kring (commented):
Looking really close now!

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileContentLoadInput.h
```diff
@@ -89,37 +97,30 @@ struct CESIUM3DTILESSELECTION_API TileContentLoadInput {
...
    */
-  std::string contentType;
+  std::shared_ptr<CesiumAsync::IAssetRequest> pRequest;
 
   /**
-   * @brief The source URL.
+   * @brief The content data to use.
    */
-  std::string url;
+  std::optional<gsl::span<const std::byte>> data;
```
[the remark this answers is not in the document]
nithinp7: Ah good idea that sounds like a much cleaner way to do it!

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileContentLoadInput.h
```diff
@@ -29,46 +32,39 @@ struct CESIUM3DTILESSELECTION_API TileContentLoadInput {
...
+   * @param asyncSystem The async system to use for tile content loading.
    * @param pLogger The logger that will be used
-   * @param data The actual data that the tile content will be created from
-   * @param contentType The content type, if the data was received via a
-   * network response
-   * @param url The URL that the data was loaded from
-   * @param tile The {@link Tile} that the content belongs to
+   * @param pAssetAccessor The asset accessor to make further requests with.
+   * @param pRequest The original tile request and its response.
+   * @param data The content data to use. The response data will be used
```
kring, line 50: This parameter doesn't exist anymore.

Cesium3DTilesSelection/src/CompositeContent.cpp
```diff
@@ -30,6 +34,64 @@ static_assert(sizeof(InnerHeader) == 12);
...
+      const IAssetResponse* pOriginalResponse_,
+      const gsl::span<const std::byte> derivedData_)
+      : pOriginalResponse(pOriginalResponse_), derivedData(derivedData_) {}
+
+  uint16_t statusCode() const override {
+    return pOriginalResponse->statusCode();
+  }
+
+  std::string contentType() const override {
+    return pOriginalResponse->contentType();
```
kring, line 54: In the original implementation, the `contentType` for the nested content was set to empty string. I'm not sure if that was actually important or not, though.

Cesium3DTilesSelection/src/Tile.cpp
```diff
@@ -297,21 +293,46 @@ void Tile::loadContent() {
...
+              return asyncSystem.createResolvedFuture(std::move(pLoadResult));
             }
 
-            loadInput.data = pResponse->data();
-            loadInput.contentType = pResponse->contentType();
-            loadInput.url = pRequest->url();
-            std::unique_ptr<TileContentLoadResult> pContent =
-                TileContentFactory::createContent(loadInput);
-
+            loadInput.asyncSystem = asyncSystem;
```
kring, line 299: It's reasonable to `std::move` the asyncSystem, too.

Cesium3DTilesSelection/src/Tile.cpp
```diff
@@ -297,21 +293,46 @@ void Tile::loadContent() {
...
+                        pContent->httpStatusCode = statusCode;
+                      }
+
+                      return std::move(pContent);
+                    });
+          })
+      .thenInWorkerThread(
+          [gltfUpAxis,
+           &boundingVolume = this->_boundingVolume,
+           &transform = this->_transform,
```
kring, line 319: These references are a bit unsafe. I mean, it's probably mostly fine as the code is currently written, but it's kind of a policy violation that might lead to (subtle, hard to debug) bugs in the future.

The violated policy is: never share not-thread-safe objects between threads. The bounding volume and transform are members of the Tile, and Tile is not thread safe. So sharing a reference to them with the worker thread means multiple threads can end up accessing them simultaneously. Now that's technically fine as long as neither thread is _modifying_ either field, and I think the code as currently written will not modify a Tile's transform or bounding volume while that Tile is in the process of being loaded. So it's not an immediate problem.

But this is fragile, because it means in all our modifications in the future, we have to continue to make sure that a Tile's bounding volume and transform are not modified while an async Load is in progress. And if we fail to do so, it'll probably mostly continue to work just fine, unless we just happen to have some really bad timing and then things will start failing in super mysterious ways (e.g. totally wrong texture coordinates in rare circumstances because we used a half-updated transform to generate them).

We have a policy against _ever_ sharing not-thread-safe objects between threads in order to avoid the need to ever think "will this be modified?" That's a hard enough question to answer for the code base today, and it's even harder to ensure our answer doesn't change in the future.

The only exception should be when the cost of copying is too high, and we can't see another way to go about it. I don't think we currently have any cases like that.

I don't think a bounding volume or transform are so expensive that we need to stress too much about copying them. One way we could avoid it, though, (which may be worth considering), is by making the second `thenInWorkerThread` here inside of the other one, as a continuation of the `thenImmediately` above. That would allow the `TileContentLoadInput` to be moved in, making the code overall much more similar to how it was in main. But at the cost of some extra indentation.

kring (commented):
So close now, just a couple more small things!

Cesium3DTilesSelection/src/Tile.cpp
nithinp7: @kring I combined the second `thenInWorkerThread` with the nested `thenImmediately` and it looks cleaner now, is this what you were thinking? It looks like `loadInput` still needs to be captured-by-reference to avoid a copy, but I think this is much safer than capturing the Tile members directly by reference. The `loadInput` can't be moved into the lambda capture since it will still be used by the `createContent` invocation, but by the time the `thenInWorkerThread` runs, `createContent` is clearly done, so the `loadInput` reference will be safe to use. Mostly just expounding on this to make sure it doesn't violate some policy like before.

kring: Capturing the loadInput by reference is not safe. That object is owned by the outer lambda, so you're relying on the lifetime of that outer lambda extending until after the inner lambda completes. I'm actually somewhat surprised that's working. I thought Async++ destroys the function immediately after it returns. Even if it doesn't, that's an implementation detail we shouldn't rely on.

I should be safe and reasonable to move loadInput into the inner lambda. Yes, it also gets passed to `TileContentFactory::createContent` by reference, but that happens first, so it should be fine to move the instance into the lambda afterward. `TileContentFactory::createContent` isn't allowed to hold on to and use that reference past the point that it returns, consistent with normal rules for reference parameters, so that shouldn't stop it. A `createContent` lambda capturing that loadInput parameter by reference would be unsafe in the same way that it's unsafe for the lsambda here to do so. If `createContent` needs some or all of the TileLoadInput later, it needs to make a copy.

Cesium3DTilesSelection/src/Tile.cpp
```diff
@@ -297,65 +302,81 @@ void Tile::loadContent() {
...
+
+                          Tile::generateTextureCoordinates(
+                              model,
+                              loadInput.tileTransform,
+                              // Would it be better to use the content bounding
+                              // volume, if it exists?
+                              // What about
+                              // TileContentLoadResult::updatedBoundingVolume?
+                              loadInput.tileBoundingVolume,
+                              projections);
```
nithinp7, line 356: @kring I think everything dubious has been fixed now! What are your thoughts on this last thing, does it make a difference which bounding volume is used for the texture coordinates generation?

kring: That's a really interesting question!

I think we need to use the original tile bounding volume for texture coordinate generation, not the content bounding volume. That's because we sometimes expect to be able to use parent tile raster overlay textures when rendering child tiles. And we can't do that reliably if we use a bounding volume that (pretty much by definition) doesn't cover parts of the child tiles. At the very least, the raster overlay tile selection algorithm would get more complicated. Currently if all the overlay tiles that are attached to a child tile fail to load, we use the parent texture. But if the parent texture coordinate space doesn't cover the child, we couldn't do that anymore. We'd have to load additional parent-level raster overlay tiles for use with the child. So I think we should keep this simple until we identify a tileset or use-case where the extra complexity provides a compelling benefit.

As for the `updatedBoundingVolume`, it's unfortunately not totally clear (either in my head or in the doc) whether this is meant to bound the content or the content plus all child content. In the case where it's used - quantized mesh terrain - the two are the same so the question didn't really come up. But `Tile::update` uses it to update the tile bounding volume (not the content bounding volume), so I guess that's something like an authoritative answer. So, since that field is an "improved" version of the tile bounding volume, I'd say that implies we should use it to generate texture coordinates if it's available.

kring:
:shipit: 🎉

Issue #258, opened 2021-05-30
Replace glTF enum properties with primitives (int/string), and provide a class with known enum values


javagl:
The issue that was reported in https://community.cesium.com/t/problems-with-loading-local-3dtiles-models/12869/20 is caused by the JSON data of the GLB of the B3DM containing images that have the `"mimeType":"image/webp"`.

After drilling into the code for a few hours to find the reason for the bug (ouch) I tried to figure out a solution, but I think that the deeper problem might be that the MIME type is modeled as an `enum` in the first place. It should just be a string. 

Depending on how the decision about the output is made in the `generate-gltf-classes`, this may require a special handling for the image MIME type: The specification declares it as a string, https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/image.schema.json#L24 , and *usually*, an `anyOf` with `string` should still be modeled as an `enum`, but not in this case. 

(Somewhat related: There are considerations to formalize the process of extending the "set of valid MIME types", see `https://github.com/KhronosGroup/glTF/issues/1966#issuecomment-822571685`)

---

EDIT: More generally, unknown enum values should not cause "unspecified parse errors", but rather say .... "unspecified enum value X found in Y" or so. Depending on the efforts, this issue might either be generalized, or a new issue may be opened for that.

kring:
Based on the issue described here, and Marco's demonstration of the problem in #259, I think we should avoid using an enumeration as the type of _any_ glTF property. Instead, we should:

* Use the appropriate primitive type from the JSON Schema as the property type. e.g. `image.mimeType` should be declared as `std::string`.
* Provide a class with static constant properties for the known enumeration values when they are defined in the JSON Schema.
* In the generated Doxygen for the property, add a link to the generated enumeration value class.

We can't predict which enum properties will need custom values via an extension in the future, so better to just provide the flexibility everywhere from the start. We lose a bit of static typing, but I think it's a necessary cost.

Our glTF classes are meant to be able to represent every valid glTF, but make no attempt to guarantee that every valid instance of a `Model` is a valid glTF. That's the job of a validator.

javagl:
I think that this might be related: https://github.com/KhronosGroup/glTF/pull/2031 - There are considerations to replace `enum` in the schema with `const`. Maybe I'm misinterpreting something there, and I left a comment, hoping to better understand the reasoning behind that, but it might be necessary to update the code generator accordingly if this change is applied.

kring:
This was implemented in #350.

PR #259, opened 2021-05-30
Added test for unknown MIME type

javagl:
Just a draft, showing the issue https://github.com/CesiumGS/cesium-native/issues/258

PR #350, opened 2021-09-23
glTF generator enum type fixes

javagl:
This is supposed to address https://github.com/CesiumGS/cesium-native/issues/258 , 

# but ...

... how this relates to https://github.com/CesiumGS/cesium-native/pull/307 has to be sorted out. There are several loose ends to consider, summarized quickly from the tip of my head:

- In the linked PR, the `generate-gltf-classes` project has been refactored significantly
- This refactored state is only available as a *new* project (`schema-gen`)
- How this new project relates to the original project (or which refactoring steps had been undertaken exactly) is not clear
- The new project was supposed to not only generate the _structures_ and the _reader_ classes, but also the _writer_ classes (in the current `main` branch, the writer classes are written manually) 
- The new project (as of the state of the PR) cannot directly re-generate the `CesiumGltf` classes as they are. It *might* be possible to solve this quickly, but not without diving deeper into the code (and thus, it would be _not_ "quickly"...). 
- Both the old and the new project are not documented. I mean, not _at all_. \*sigh\*

---

This PR is, roughly speaking, the attempt to do the "smallest modifications" to the original code generator that allow us to write code where the `enum`s use their underlying type. The results, summarized with some examples here:

### Old integer enums:
```
struct CESIUMGLTF_API Sampler final : public NamedObject {
  static inline constexpr const char* TypeName = "Sampler";

  /**
   * @brief Magnification filter.
   *
   * Valid values correspond to WebGL enums: `9728` (NEAREST) and `9729`
   * (LINEAR).
   */
  enum class MagFilter {
    NEAREST = 9728,

    LINEAR = 9729
  };

...
  /**
   * @brief Magnification filter.
   *
   * Valid values correspond to WebGL enums: `9728` (NEAREST) and `9729`
   * (LINEAR).
   */
  std::optional<MagFilter> magFilter;

  ...

  /**
   * @brief s wrapping mode.
   *
   * S (U) wrapping mode.  All valid values correspond to WebGL enums.
   */
  WrapS wrapS = WrapS(10497);
```

### New integer enums:
```
  /**
   * @brief Known values for Magnification filter.
   */
  class MagFilter {
  public:
    static const int32_t NEAREST = 9728;

    static const int32_t LINEAR = 9729;
  };

  ...

  /**
   * @brief Magnification filter.
   *
   * Known values are defined in {@link MagFilter}.
   *
   *
   * Valid values correspond to WebGL enums: `9728` (NEAREST) and `9729`
   * (LINEAR).
   */
  std::optional<int32_t> magFilter;

  ...
  
  /**
   * @brief s wrapping mode.
   *
   * Known values are defined in {@link WrapS}.
   *
   *
   * S (U) wrapping mode.  All valid values correspond to WebGL enums.
   */
  int32_t wrapS = WrapS::REPEAT;

```

### Old string enums:

```
struct CESIUMGLTF_API AccessorSpec : public NamedObject {
  static inline constexpr const char* TypeName = "Accessor";
...
  /**
   * @brief Specifies if the attribute is a scalar, vector, or matrix.
   */
  enum class Type {
    SCALAR,

    VEC2,

    VEC3,

    VEC4,

    MAT2,

    MAT3,

    MAT4
  };

  /**
   * @brief Specifies if the attribute is a scalar, vector, or matrix.
   */
  Type type = Type();
}
```

### New string enums:
```
struct CESIUMGLTF_API AccessorSpec : public NamedObject {
  static inline constexpr const char* TypeName = "Accessor";
  ...
  
  /**
   * @brief Known values for Specifies if the attribute is a scalar, vector, or
   * matrix.
   */
  class Type {
  public:
    inline static const std::string SCALAR = "SCALAR";

    inline static const std::string VEC2 = "VEC2";

    inline static const std::string VEC3 = "VEC3";

    inline static const std::string VEC4 = "VEC4";

    inline static const std::string MAT2 = "MAT2";

    inline static const std::string MAT3 = "MAT3";

    inline static const std::string MAT4 = "MAT4";
  };

   ...
   /**
   * @brief Specifies if the attribute is a scalar, vector, or matrix.
   *
   * Known values are defined in {@link Type}.
   *
   */
  std::string type = std::string();
```

---

The code for the generated **Readers** is updated accordingly. Some places in the not-auto-generated classes will have to be updated manually - mostly `switch` statements, but a few others. Unfortunately, the code for the current **Writers** is also not auto-generated, and this may require a bunch of tedious updates that are supposed to be ... overwritten by the auto-generated writers soon. But if this is desired (depending on the priorities), I can tackle that. 

--- 

Finally... I *have* to link to [Lava Flow](https://en.wikipedia.org/wiki/Lava_flow_(programming)) here.


javagl:
I have updated the generator to use `struct` and `constexpr`, as suggested. Based on the resulting state of the generator, the following commits are

- Update of the resulting auto-generated classes
- Updating the non-auto-generated classes accordingly
- Removal of the `magic_enum` dependency (it was only used in `CesiumGltfWriter` for the enums, and is no longer required)
- Fixes for failing tests...

--- 

The update of the non-auto-generated classes was less of a hassle than I thought. The `mimeTypeToMimeString` function could be removed, and I hope/think that the [update for `mimeTypeToExtensionString`](https://github.com/CesiumGS/cesium-native/pull/350/commits/486eb57a93044827841c25a34fbc01fd199b18f5#diff-ef0bf21d3f67205425f2b5652b6bc46dec376bfb91c9e675eee9cd728b0f5393R12) makes sense. 

---

There's one question about error handling: 

Files like `Accessor.h` contained functions like

    int8_t computeNumberOfComponents(CesiumGltf::Accessor::Type type);

which are now a

    static int8_t computeNumberOfComponents(const std::string& type);

which means that the error handling becomes more important. These should at least print an error message (as marked with a few `TODO Print a warning here` markers). 

Right now, `spdlog` is a dependency of `Cesium3DTilesSelection`, with the [`spdlog-cesium.h`](https://github.com/CesiumGS/cesium-native/blob/main/Cesium3DTilesSelection/include/Cesium3DTilesSelection/spdlog-cesium.h) convenience wrapper. 

In order to use `spdlog` in `CesiumGltf` as well, one could 

- add `spdlog` as a dependency to `CesiumGltf`, and add a similar convenience header there
- move that header to `CesiumUtility`, and use it from there (most other projects, particularly `Cesium3DTilesSelection` and `CesiumGltf`, already depend on `CesiumUtility`)

I'm slightly leaning towards the latter. One concern is that it makes `spdlog` an _implicit_ part of the _public(!)_ API of `CesiumUtility`. But ... what is the "public API of `CesiumUtility`" anyhow? So this might not matter so much here. If there are any preferences, just let me know...

---

Initially, the tests failed, because there are some assumptions about the initial values for the enum members. For example, the `AccessorSpec::type` was initialized to be an empty string (because it is optional (because it does not have a default value)). But the tests assumed that it had the value `Type::SCALAR` if it was not initialized. This was fixed in the subsequent commits: Now each enum member is initialized with its default value (if that was given), or with "the first" enum value otherwise.

---

I was a bit confused about the role of the files in the `generated` directory. There had been some `.cpp` files that apparently didn't belong there, and they have been removed in https://github.com/CesiumGS/cesium-native/pull/350/commits/0f6645f34ba3d9db769b796df3870545f63382c8 (if this is an issue, I'll have another look what happened there...).

lilleyse:
I opened a follow up PR that brings us up-to-date with the latest glTF schema: https://github.com/CesiumGS/cesium-native/pull/351


> I was a bit confused about the role of the files in the `generated` directory. There had been some `.cpp` files that apparently didn't belong there, and they have been removed in https://github.com/CesiumGS/cesium-native/pull/350/commits/0f6645f34ba3d9db769b796df3870545f63382c8 (if this is an issue, I'll have another look what happened there...).

I was also confused by this.

kring:
I don't think CesiumGltf has any business logging, especially not in such a low-level function. Documenting that it returns 0 for an unknown type is sufficient. Then the caller can assert, throw an exception, or log as appropriate.

kring:
This looks good to me, let's get it merged for our release tomorrow.

javagl:
Before this becomes part of a release, it should be tested more thoroughly.

I have re-introduced the unit test that I had created in https://github.com/CesiumGS/cesium-native/pull/259 to illustrate the issue. This one passes now. 

**But**... this only means that it no longer bails out with an "Unspecified parse error".

Further tests will be necessary. Particularly: **If** there is a B3DM with a glTF with a `image/webp` MIME type, then it has to be decided where and how to handle the fact that we currently cannot decode WebP images. (Some relevant discussion on a low technical level is in https://github.com/KhronosGroup/glTF/issues/1966#issuecomment-845702479 , but how to actually detect and present that error has to be decided for our implementation).

Unfortunately, we don't have a tileset that uses WebP images. The closest that we have was a single B3DM that was provided in the forum thread https://community.cesium.com/t/problems-with-loading-local-3dtiles-models/12869/20 where this issue came up. As a _quick_ test, I just took https://github.com/CesiumGS/cesium/tree/main/Specs/Data/Cesium3DTiles/Batched/BatchedTextured , changed the `jpeg` to `webp` with a hex editor, and loaded that in CesiumForUnreal. It does no longer print the parse error, but ... it does not print an error at all: Apparently, the `mimeType` is ignored in the loader part, underlying `stb` image loader detects the actual image data (which is still JPEG) and processes it. (Well, it does not display anything ... but it also doesn't display anything for the unmodified test data, so sorting this out is a different story...). 

In order to have a real-world test, we need a tileset that actually contains a B3DM with a glTF with a WebP image. And preferably, with the same complexity as the `BatchedTextured` test, and *not* as some random, non-open, 200MB tileset that some user provided. I can try to create such tileset during the weekend, add it to the https://github.com/CesiumGS/cesium-native/tree/main/CesiumGltfReader/test/data , and create an actual test. (That test would pass iff it does not crash despite the WebP data (but prints a warning)).

And... sure, I'd fix the merge conflicts...

kring:
I tried the `Modern Office (WebP)` tileset in the cesiumlab ion account. It prints messages like this to the console:

```
LogCesium: Warning: Texture source index must be non-negative and less than 1, but is -1
```

But it otherwise loads and displays fine (no textures of course):

image

Before this PR, we got messages like this:

```
LogCesium: Error: [2021-10-01 12:18:35.240] [error] [GltfContent.cpp:44] Failed to load binary glTF from https://assets.cesium.com/17743/0material54_0.b3dm?v=4:
- JSON parsing error at byte offset 1491: Parsing was terminated.
```

And the model didn't display at all. So we could definitely do better on the warning message, but this still a major improvement.

PR #351, opened 2021-09-26
Update generated files with newer glTF schema

lilleyse:
There have been a series of language revisions to the glTF spec recently, e.g. https://github.com/KhronosGroup/glTF/pull/1997 and https://github.com/KhronosGroup/glTF/pull/2028. This PR updates the generated glTF reader files with those changes. There are no functional changes as far as I can tell, just documentation.

The `npm run generate` command now uses `CesiumGS/glTF/3d-tiles-next` instead of `kring/glTF/metadata-schema-cleanup` since https://github.com/CesiumGS/glTF/pull/10 was merged and `3d-tiles-next` branch is now synced up with KhronosGroup `main`. Also the subfolders for `EXT_feature_metadata` were removed so `1.0.0` was removed from `glTF.json`.

I built locally and spot-checked the generated Doxygen documentation, no issues.

Just as a warning, we may need to support both draft4 schemas and 2020-12 schemas once https://github.com/KhronosGroup/glTF/pull/2034 is merged since glTF will be using 2020-12 exclusively but glTF extensions and 3D Tiles will be on draft4 for at least a little bit longer.

### DIFF
--- a/CesiumGltf/include/CesiumGltf/AccessorSparse.h
+++ b/CesiumGltf/include/CesiumGltf/AccessorSparse.h
@@ -10,31 +10,27 @@
 
 namespace CesiumGltf {
 /**
- * @brief Sparse storage of attributes that deviate from their initialization
- * value.
+ * @brief Sparse storage of accessor values that deviate from their
+ * initialization value.
  */
 struct CESIUMGLTF_API AccessorSparse final : public ExtensibleObject {
   static inline constexpr const char* TypeName = "AccessorSparse";
 
   /**
-   * @brief Number of entries stored in the sparse array.
-   *
-   * The number of attributes encoded in this sparse accessor.
+   * @brief Number of deviating accessor values stored in the sparse array.
    */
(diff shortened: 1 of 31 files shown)


lilleyse:
Updated to support JSON schema 2020-12 now that https://github.com/KhronosGroup/glTF/pull/2034 is merged. The only real difference is that glTF now uses `const` instead of a single element `enum` arrays for enums. `const` was introduced in JSON schema draft6.

Screenshot from 2021-10-03 14-29-12

The code still works with draft4 since we'll need it for glTF extensions and 3D Tiles. To test that out point the code generator to `3d-tiles-next-json-schema-4`. The actual structs should be identical though some descriptions changed between draft4 and 2020-12.

The main code change was to move a lot of the enum checking into two functions `getEnumValue` and `isEnum` which handle both the draft4 and 2020-12 style of describing enums. The rest of the changes are the result of running prettier on the code.

kring:
Thanks @lilleyse!

PR #352, opened 2021-09-28
Don't ignore RTC_CENTER when it has an integer coordinate value.

kring:
Fixes problems with the "Washington State" tileset, ion asset ID 55416.

Much to my surprise, RapidJSON's `IsDouble` returns false for integers, even though `GetDouble` works and performs the expected conversion, and `IsInt64` returns true for a 32-bit integer. So in the problem tileset, one of the RTC_CENTER coordinate values happened to have an integer value, and so `IsDouble` was returning false, and so we were assuming the RTC_CENTER value was invalid and ignoring it, and so the tileset was rendered in the wrong place.

This PR checks for _any_ number and then converts it to a double.

CC @argallegos @nithinp7

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@
 
 - Made tile content loading asynchronous and allowed it to make further network requests, to Cesium ion or otherwise.
 
+##### Fixes :wrench:
+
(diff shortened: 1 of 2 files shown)


argallegos (approved):
Looks good, thank you!

PR #354, opened 2021-10-01
Improved error handling for unhandled image types like WebP

javagl:
This is a follow-up for https://github.com/CesiumGS/cesium-native/pull/350

I wanted to add tests to cover the error handling for glTF data, according to [a comment in the linked PR](https://github.com/CesiumGS/cesium-native/pull/350#issuecomment-931447988). As part of that, I wanted to load a tileset in `cesium-unreal` that contained only a simple glTF asset, once with PNG textures, and once with WebP textures. After figuring out why `cesium-unreal` did not display anything even for the PNG case ([it's subtle...](https://github.com/CesiumGS/cesium-native/commit/1eb4caad97be9c59aa89eddba6d37c3f9537343d)), I could do the comparison based on a tileset that was created from the glTF test data that is added with this PR (and which, in turn, is only a WebP version of the [BoxTextured glTF sample](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/BoxTextured) - it uses the Cesium logo! Isn't that great?!)

With the PNG version, everything looks fine:

CesiumUnreal Simple B3DM

With the WebP version, it now prints an error message:

CesiumUnreal Simple B3DM with WebP

(and importantly: It does **not** crash or so...)

The error message is of the form

```
LogCesium: Error: [2021-10-01 20:42:16.672] [error] [GltfContent.cpp:44] Failed to load binary glTF from file://c:/Develop/J3DTiles/j3dtiles/j3dtiles-main/data/tilesets/BoxTexturedWebp/content.b3dm:
- Image not of any known type, or corrupt
- Declared image MIME Type: image/webp
```

More generally, the *image* reader in `cesium-native` did not forward any error messages to the *model* reader, so all the messages that had been created by `stb` had been swallowed silently.

(Of course, users who try to load a tileset that contains thousands of tiles with WebP data will be flooded with these messages. But... that may, in fact, be the right thing to do here...)

---

An aside, @kring : In the linked PR, you mentioned a warning

> LogCesium: Warning: Texture source index must be non-negative and less than 1, but is -1 

I did not see this warning at any point in time. Are you sure that this is related to the WebP content, or could that be another issue with that model?

---

For the case that someone wants to run the test in `cesium-unreal`, here are are the test tilesets that I used:

[WebP glTF and Tilesets test data.zip](https://github.com/CesiumGS/cesium-native/files/7269508/WebP.glTF.and.Tilesets.test.data.zip)

(This does not really belong here, but rather into `cesium-unreal`, but we don't have a dedicated setup/environment for this sort of "test models" there...)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
+- Errors and warnings that occur while loading glTF textures are now include in the model load errors and warnings.
+
 ### v0.8.0 - 2021-10-01
 
 ##### Breaking Changes :mega:


javagl:
The output of the failed build (which is *only*  Windows + VS2019) says

> Installing visualstudio2019buildtools...
> 
> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Looks like a glitch to me. Not being sure what may have caused this, I'd <strike>switch it off and on again</strike> re-trigger the build, unless someone has a better idea...

Issue #353, opened 2021-09-28
Add support for 3DTiles_implicit_tiling extension


kring:
https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next/extensions/3DTILES_implicit_tiling

I think this extension should integrate pretty nicely with the implicit tiling support already in cesium-native. In short:

* When parsing tileset.json, we should look for an `3DTILES_implicit_tiling` extension on the tile. Such a tile is known as an "implicit root".
* When we see one, we should create a new `ImplicitTilingContext` for it and arrange for the tile's content URI to be the root subtree URI. `ImplicitTilingContext` needs to be extended to capture more of the extension. In particular, it needs to capture the subtree URI, `subtreeLevels`, and `maximumLevel`.
* When loadContent is invoked on the implicit root, we load the subtree URI and populate the `availableTileRectangles` of the `TileContentLoadResult`. Or - more likely - we'll need to define an alternative way to describe availability if the extension's availability can't be easily presented as `availableTileRectangles`. The availability in `TileContentLoadResult` gets copied into the `ImplicitTilingContext` in `Tile::update`.
* Any time we want to traverse past a tile with no children, we check if the children are available implicitly and create them if so. Actually, we already do this, but we need to extend the existing logic to also load a new subtree file when needed. The easiest way to do that will be by creating a new tile with the subtree as its content.

That's a really quick and dirty summary, but hopefully enough to get you started @nithinp7 and I'll try to flesh it out a bit more tomorrow.

lilleyse:
See https://github.com/CesiumGS/3d-tiles/pull/487 for a recent addition to the spec to specify geometric error at per-tile granularity.

Also check out https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next/specification/Metadata/Semantics - there are ways to override the default implicit tiling bounding volumes using per-tile metadata.

PR #355, opened 2021-10-04
Support 3DTILES_implicit_tiling

nithinp7:
Adds support for the [3DTILES_implicit_tiling extionsion](https://github.com/CesiumGS/3d-tiles/tree/main/extensions/3DTILES_implicit_tiling) in Cesium Native. 

Resolves: #353


kring (commented):
I have a few comments below, but it's not based on a thorough review by any means. Overall it looks good. I think my most important high-level observation is that it seems kind of awkward for the Tile to be responsible for availability loading. I think this should be primarily the responsibility of the TileContext / ImplicitTilingContext, with help from the Tileset itself as necessary. For example...

During Tile traversal, the Tileset / ImplicitTilingContext could note that required availability information is missing, so it can't traverse deeper in the Tile tree. So it queues up a subtree load and stops traversal there. This might even be a good use for a polymorphic ImplicitTilingContext, so we can easily have different fields and implementations for quantized mesh implicit tiling versus 3DTILES_implicit_tiling. The tiling context will need to keep track of which subtrees are already loading so it won't generate duplicate requests.

That's a bit handy-wavy, but let me know what you think.

Cesium3DTilesSelection/src/Tile.cpp
```diff
@@ -320,10 +322,9 @@ void Tile::loadContent() {
 
   Tileset& tileset = *this->getTileset();
 
-  std::optional<Future<std::shared_ptr<IAssetRequest>>> maybeRequestFuture =
-      tileset.requestTileContent(*this);
-
-  if (!maybeRequestFuture) {
+  // TODO: rethink some of this logic, in implicit tiling, a tile may be
+  // available but have no content. It may still have children.
+  if (!maybeContentRequest) {
```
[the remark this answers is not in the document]
kring: More generally, I don't quite understand why `loadContent` needed to be changed to receive the request as a parameter, rather than starting the request itself when it deems it to be necessary. Aside from the case above (which is a bug), it looks to be totally equivalent.

Cesium3DTilesSelection/src/Tile.cpp
```diff
@@ -766,111 +863,12 @@ void Tile::update(
...
-      createImplicitTile(implicitContext, *this, this->_children[2], nwID, nw);
-      createImplicitTile(implicitContext, *this, this->_children[3], neID, ne);
-    }
+  std::optional<CesiumGeospatial::GlobeRectangle> rectangle =
+      estimateGlobeRectangle(this->getBoundingVolume());
+  if (!rectangle) {
+    return;
   }
 
-  // TODO: if there's no model, we can actually free any existing overlays.
```
kring, line 873: This TODO is still relevant, I think. Though it should probably become an issue rather than a TODO.

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -469,6 +487,13 @@ Tileset::requestTileContent(Tile& tile) {
     return std::nullopt;
   }
 
+  // TODO: should we be checking for content availability or tile availability
+  // here? For implicit tiles, check if they are available.
```
kring, line 491: This TODO looks important. Or has it been addressed elsewhere?

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -1275,18 +1515,35 @@ static bool isVisibleInFog(double distance, double fogDensity) noexcept {
...
-  this->_markTileVisited(tile);
 
-  const Tileset* pTileset = tile.getTileset();
-  if (!pTileset) {
-    return TraversalDetails();
+  // If the tile is ContentLoaded, complete processing it and push it to Done
+  // unless we are waiting on a subtree load. We have to wait for the subtree
+  // to load so that we don't create fake (upsampled) children before knowing
+  // whether or not the tile has actual implicit children by looking at the
+  // subtree.
```
kring, line 1528: Not the end of the world, but it's a shame we have to avoid finishing/rendering a tile that is ready to go just to wait for a subtree that will only be relevant for refinement. I'm guessing if you could see any easy way to avoid this, though, you would have already implemented it. Is that right?

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -2056,10 +2410,142 @@ void Tileset::processQueue(
 
   for (LoadRecord& record : queue) {
     CESIUM_TRACE_USE_TRACK_SET(this->_loadingSlots);
-    record.pTile->loadContent();
+    // TODO: for implicit tilesets only create content request if content is
+    // available.
+    if (record.pTile->getState() == Tile::LoadState::Unloaded) {
+      record.pTile->loadContent(this->requestTileContent(*record.pTile));
+    } else {
+      record.pTile->loadContent();
```
kring, line 2418: If `loadContent` doesn't go back to requesting its own content... then `loadContent` does two totally different things, depending on whether the `Tile` is in the `Unloaded` state to start. Now that those two cases are separated at the call site, it'd be a lot cleaner to make them two separate methods.

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -2056,10 +2410,142 @@ void Tileset::processQueue(
...
+  }
+
+  std::sort(this->_subtreeLoadQueue.begin(), this->_subtreeLoadQueue.end());
+
+  for (SubtreeLoadRecord& record : this->_subtreeLoadQueue) {
+    // TODO: tracing code here
+    loadSubtree(std::move(record));
+    if (this->_subtreeLoadsInProgress >=
+        this->_options.maximumSimultaneousSubtreeLoads) {
+      break;
```
kring, line 2546: Very possible I've missed something here, but if I'm reading this right... A subtree will only be added to `_subtreeLoadQueue` by `addSubtreeToLoadQueue` if the implicit info's `pCurrentNode` is null. And `addSubtreeToLoadQueue` _also_ sets `pCurrentNode` to no longer be null. So if we hit this break here, because we're loading too many subtrees, then the subtree will _never_ load. Right?

kring (commented):
There's a lot to work through here, this isn't a complete review yet. But I thought it would be useful to submit these intial comments. Overall it looks very good!

Cesium3DTilesSelection/src/Tileset.cpp
nithinp7: Oof great catch here, `pCurrentNode` should be set to no longer null in `loadSubtree` instead, only once we're sure we're going to load it. This will mean that after `addSubtreeToLoadQueue`, `pCurrentNode` should still be null. Hopefully, that change works nicely with the rest of the implicit traversal state-space logic.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
```diff
@@ -409,6 +428,7 @@ class CESIUM3DTILESSELECTION_API Tileset final {
 
   TraversalDetails _renderLeaf(
       const FrameState& frameState,
+      ImplicitTraversalInfo&& implicitInfo,
```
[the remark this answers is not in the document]
nithinp7: IIRC, it can't be a const reference because `pParentNode`  needs to be mutable when adding a new subtree node during subtree load. It made sense to me to move the `implicitInfo` into the `SubtreeLoadRecord` so that `pParentNode`, `usingImplicitQuadtreeTiling`, `usingImplicitOctreeTiling`, were all available to be potentially used in `loadSubtreeContent` without a copy. That being said I definitely failed to create a move constructor for it, which I will do right away unless you still find move semantics unnecessary here.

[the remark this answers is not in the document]
nithinp7: Hm okay, I think I now see what you must have actually meant by "...doesn't benefit from move semantics anyway." `ImplicitTraversalInfo` is pretty much a POD struct since it is not responsible for managing the memory of the pointers. There is no point to move semantics since all of the members are primitive types which will be copied anyways. And in general the struct is not heavy enough to warrant wrapping it in a unique_ptr since it would need to be heap allocated in that case. Hopefully I'm understanding that right, I'm still trying to wrap my head around move semantics.

Cesium3DTilesSelection/src/Tile.cpp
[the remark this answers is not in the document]
nithinp7: I wasn't sure if this was the best change, but I did it to set up a way to _not_ request content for implicit tiles that we know not to have content. To decide that, we need the `ImplicitTraversalInfo` which I was considering putting in the `LoadRecord`, so it would be available in `processLoadQueue`. Since it didn't make much sense to further pass the `ImplicitTraversalInfo` into `Tile::loadContent`, I figured it would be better to create a content request in `processLoadQueue` (potentially conditioned on implicit tiling content availability). Actually implementing that seems like a huge pain though, since each callsite of `addTileToLoadQueue` potentially needs the `ImplicitTraversalInfo`. 

The biggest blocker in deciding to go through fully with that change was confusion about the function of `loadContent`. Like you said in the other comment below, `loadContent` does two different things depending on whether or not there is a content request to make. Splitting those into separate functions like you suggest below will help us be more clear on what we're actually doing when there is no content to request. Maybe at that point, the naming of the Tile load states (ContentLoading / ContentLoaded) need to be revisited.

Cesium3DTilesSelection/src/Tileset.cpp
nithinp7: It does seem hard to work around, but actually the issue gets even more complicated once we actually avoid making content requests for implicit tiles that are marked as content unavailable (referencing above conversations): 

So firstly, the only tiles that have to wait like this are going to each be the root tile of some unloaded or loading subtree. The fact that this subtree root tile exists at this point, implies that the parent subtree had marked this subtree as available; note also that the spec states the root tile of each available subtree must be available. That is why this Tile object has already been created, _however_, we actually don't know the _content_ availability of the tile. So until the subtree is loaded, we can't commit to treating the tile's content as being unavailable, and we can't (or shouldn't) blindly request the content from the template uri, and of course we can't create implicit children since we don't know their availabilities either before the subtree is loaded. 

In the current state though, since we are blindly requesting content on every TILE_AVAILABLE implicit tile anyways, this blocking only serves to avoid accidentally creating fake, upsampled children prematurely.

Cesium3DTilesSelection/src/Tileset.cpp
nithinp7: I agree with this, I'll have to sort out what exactly happens in those two different cases, but I think that separation is worth making.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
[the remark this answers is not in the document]
kring: > it can't be a const reference because pParentNode needs to be mutable when adding a new subtree node during subtree load.

You mean in `addSubtreeToLoadQueue`? You just need to copy it and modify the copy. It's a very inexpensive operation for this struct.

> Hopefully I'm understanding that right

Yep, exactly right.

Cesium3DTilesSelection/src/Tileset.cpp
kring: Ok, I think I understand. And in a way that makes this simpler. We can't load this tile's content _at all_ until we load its subtree, because the subtree tells us whether the tile has content. So we do have to wait for the subtree before we can render this tile, but it's essential waiting required by the spec, not accidental waiting as a result of our implementation.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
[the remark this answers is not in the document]
nithinp7: I've changed it so it's just doing the copy there (it's a non-const reference now). But if the `ImplicitTraversalInfo` is a const reference instead, then even after copying you could only access a const `info.pParentNode` pointer right? I think we need a non-const `pParentNode` pointer so that we can add a new child subtree node onto it when we begin loading. 

Just so I'm not making a C++ mistake here, if you had this type of situation:

```
struct A {
  Obj* ptr
};

const A a;
```
What is the type of `a.ptr`, is it `const Obj*` or `Obj const*`?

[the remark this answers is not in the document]
kring: Somewhat surprisingly, `a.ptr` will be `Obj* const`, meaning you can't change the value of the pointer, but you _can_ change the thing pointed to by it. See https://godbolt.org/z/WW1qh6rq3

(`Obj const*` and `const Obj*` are identical I believe, the former is just a less commonly used way of expressing it)

Cesium3DTilesSelection/src/Tileset.cpp
nithinp7: This logic ended up going away, because tiles should not even end up in the ContentLoaded state if their content availability is not known yet (the subtree hasn't loaded yet). `addTileToLoadQueue` will now refuse to move a tile out of `Unloaded` if the availability is still loading.

nithinp7:
@kring I think this is ready for another review when you get the chance! I think I'm happy with the resolutions to the last round of review comments.

kring (commented):
Your recent changes look great! A few picky things below that I happened to notice. Also I think it'd be worth doing a pass for any (new) TODOs. Remove the ones that are no longer relevant, and turn the rest into a checklist in this PR (removing them from the code). Any that don't urgently need to be addressed can then become github issues after we merge this.

Cesium3DTilesSelection/src/Tile.cpp
```diff
@@ -307,42 +308,38 @@ void* processNewTileContent(
...
+  const UpsampledQuadtreeNode* pSubdivided =
+      std::get_if<UpsampledQuadtreeNode>(&this->getTileID());
+  if (pSubdivided) {
+    // We can't upsample this tile until its parent tile is done loading.
+    if (this->getParent()) {
+      if (this->getParent()->getState() == LoadState::Done) {
         std::vector<Projection> projections = mapOverlaysToTile(*this);
         this->upsampleParent(std::move(projections));
-      } else {
+      } else if (this->getParent()->getState() != LoadState::Unloaded) {
```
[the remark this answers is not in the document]
nithinp7: Hm that's right, I'm trying to think why we thought that was necessary yesterday. It's the case now that a tile _must_ have loadable content for `loadContent` to be called, but it turns out it only makes sense to ever call `continueLoadingContent` when a tile has loadable content as well. Otherwise, there will be no model and therefore no need to update rasters that don't exist.

kring:
Ok I've now looked at everything, reasonably thoroughly. It's looking fantastic, and I'm going to merge it as soon as CI passes and I do one more check to make sure I didn't break anything with some small changes I made while reviewing (https://github.com/CesiumGS/cesium-native/pull/355/commits/c43bfcb3a7caff5261c3fc7e6d802d0c621dac62). My changes were:
* Updated the doc on `loadContent` to reflect the latest iteration.
* Added some length checks to the availability parsing so that bad data won't lead to a crash.
* Slightly reordered some code in `AvailabilitySubtreeContent` to avoid a move.
* Removed an erroneous check for a valid result from `estimateGlobeRectangle` in Tile.cpp, even though that result was never used. I _think_ this was just a bad merge, but let me know if there's some deeper reason for it that I missed.

I'm still a little concerned about all the TODOs, and I'd still like to see an inventory and triage of them. But it doesn't have to hold up merging this. Best to get this in main now so we've been using it daily and are confident in it by the time the December release rolls around.

This was a huge piece of work in a very complicated area of the code, great job!

kring:
I see problems when upsamping for raster overlays... tiles just disappearing. It sounds similar to the problem you mentioned in the S2 branch, so we can sort it out later.

PR #363, opened 2021-10-11
Add Cesium3DTilesReader project

lilleyse:
Merge https://github.com/CesiumGS/cesium-native/pull/362 first.

Adds a tileset JSON reader with generated files from https://github.com/CesiumGS/cesium-native/pull/361.

The changes in `ArrayJsonHandler.h` are for supporting recursive types like in `TileJsonHandler.h`. @krupkad did all of the work here in https://github.com/CesiumGS/cesium-native/pull/308 which I'm thankful for.

The biggest question is how much code duplication is acceptable? `IExtensionJsonHandler`, `TilesetReader`, `ExtensibleObjectJsonHandler`, and `ExtensionsJsonHandler` are essentially copy-pasted from `GltfReader`.

The scaffolding is set up for reader options and extensions handling but those aren't being used currently. That is likely to change in a future PR.


CesiumNativeTests/CMakeLists.txt
```diff
@@ -4,9 +4,8 @@ configure_cesium_library(cesium-native-tests)
 # Add tests here, ensure they define the TEST_SOURCES / TEST_HEADERS
 # properties.
 set(cesium_native_targets
+    Cesium3DTilesReader
     Cesium3DTilesSelection
-    CesiumAsync
-    CesiumGeometry
```
kring, line 9: I'm guessing these were removed accidentally. I'll add them back.

lilleyse: Yeah that was a bad merge on my part, thanks.

lilleyse:
Thanks for helping out with this PR as well @kring

PR #369, opened 2021-10-18
Regenerate classes for EXT_feature_metadata

lilleyse:
After cleaning up the `EXT_feature_metadata` schema in https://github.com/CesiumGS/glTF/pull/29 there were a few good changes that the code generator picked up - `Type`, `ComponentType`, and `OffsetType` are now "real" types instead of `CesiumUtility::JsonValue`. Updated the code to use these types instead of hardcoded strings.

### DIFF
--- a/CesiumGltf/include/CesiumGltf/MetadataFeatureTableView.h
+++ b/CesiumGltf/include/CesiumGltf/MetadataFeatureTableView.h
@@ -93,9 +93,9 @@ public:
 
     PropertyType type = convertStringToPropertyType(pClassProperty->type);
     PropertyType componentType = PropertyType::None;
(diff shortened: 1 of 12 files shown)


CesiumGltf/include/CesiumGltf/MetadataFeatureTableView.h
```diff
@@ -108,7 +108,7 @@ class MetadataFeatureTableView {
       getArrayPropertyViewImpl(
           propertyName,
           *pClassProperty,
-          type,
+          componentType,
```
lilleyse, line 111: I'm pretty sure this used to be a bug.

PR #371, opened 2021-10-21
Support external buffers/images in glTFs

nithinp7:
Support external gltf buffers and images to enable the [implicit tiling PR](https://github.com/CesiumGS/cesium-native/pull/371) to work with CDB to 3D Tiles Next tilesets.

EDIT: Closing this and opening #391, which is the same branch but rebased onto main.


nithinp7:
@kring I split this out into a separate PR from #355, can you have a look? There are a few TODOs that can be searched in the diff for decisions I felt were questionable:
- There are two types we currently have for HTTP headers, `CesiumAsync::IAssetAccessor::THeader` and `CesiumAsync::HttpHeaders`. I needed to convert from one to another, and it seemed to require a copy. Should we consolidate those two types?
- There is a very suspicious lambda captured pointer around L407 in `GltfReader.cpp`, but in this specific case I _think_ it's thread-safe. But I would like to get someone to confirm that, or someone to discourage me from capturing that just out of principle.

CesiumGltfReader/src/GltfReader.cpp
```diff
@@ -357,8 +361,106 @@ ModelReaderResult GltfReader::readModel(
...
+    if (buffer.uri) {
+      resolvedBuffers.push_back(
+          pAssetAccessor
+              ->requestAsset(
+                  asyncSystem,
+                  Uri::resolve(baseUrl, *buffer.uri),
+                  tHeaders)
+              .thenInWorkerThread(
+                  // TODO: is this safe? result should be alive until
+                  // the final continuation is invoked.
```
kring, line 408: Yeah I'd say this is unsafe. It took me a minute to even realize why it works. `result` is passed into this method by reference. Sure it's an r-value reference, which tells us the caller isn't planning to use it after the function call returns, but it's still a reference. So as soon as `resolveExternalData` returns, we can't count on the referred-to object continuing to exist. When this method is called from GltfContent, the reference is to an object in the `GltfContent::load` function's local scope, so it will become invalid as soon as that function returns. The `result` captured in the final lambda is a totally different instance with a totally different address.

So why doesn't it crash? Well, because when the result is captured by the final lambda by move, all of the vectors inside of the Model are moved. So while pointers to the result and its embedded Model become invalid, pointers to objects inside the vectors just happen to remain valid. I'm not entirely sure where this falls along the spectrum between "spec-guaranteed behavior that may not be obvious" and "unspecified behavior that just happens to work out for us, and probably will continue to do so in most situations."

But either way it's probably worthwhile to make it more obviously correct. The easiest solution is to move the result into a heap-allocated instance, i.e. `auto pResult = std::make_unique<ModelReaderResult>(std::move(result))`, and then move that to the final lambda. That will guarantee that the lifetime of that entire pResult is extended until after all the buffer image and image lambdas complete, and there's no concern by pointers becoming invalid.

CesiumGltfReader/src/GltfReader.cpp
```diff
@@ -357,8 +361,106 @@ ModelReaderResult GltfReader::readModel(
...
+  // TODO: Can we avoid this copy conversion?
+  std::vector<IAssetAccessor::THeader> tHeaders(headers.begin(), headers.end());
+
+  if (!result.model) {
+    return asyncSystem.createResolvedFuture(std::move(result));
+  }
+
+  size_t externalBufferCount = 0;
+  for (const Buffer& buffer : result.model->buffers) {
+    if (buffer.uri) {
```
kring, line 380: I think we should explicitly exclude data URIs. With default settings, they'll already have been loaded and cleared before we get here. But if the user disabled the "clear data URIs" option, we'll end up loading them twice.

Cesium3DTilesSelection/src/GltfContent.cpp
```diff
@@ -16,19 +16,28 @@
...
+/*static*/
+Future<std::unique_ptr<TileContentLoadResult>> GltfContent::load(
+    const AsyncSystem& asyncSystem,
     const std::shared_ptr<spdlog::logger>& pLogger,
     const std::string& url,
+    const HttpHeaders& headers,
+    const std::shared_ptr<IAssetAccessor>& pAssetAccessor,
     const gsl::span<const std::byte>& data) {
   CESIUM_TRACE("Cesium3DTilesSelection::GltfContent::load");
   std::unique_ptr<TileContentLoadResult> pResult =
```
baothientran, line 43: this ptr is recreated in the final lambda. I think we can remove it

PR #376, opened 2021-10-29
Fixes handling of reserved c++ keywords in gltf reader generator

nithinp7:
Properties with names that are reserved C++ keywords are appended with "Property" so that they do not cause issues in the C++ GltfReader code. But when actually parsing a gltf we should check the JSON for the original string, even if it is a C++ keyword.


kring:
Thanks @nithinp7!

PR #327, opened 2021-08-27
Raster Overlays on 3d Tiles

kring:
This builds on #326 and adds the ability to use raster overlaps (including for clipping) on 3D Tiles instead of just on terrain.

PR #378, opened 2021-10-31
Fix use of potentially invalidated iterator.

kring:
I've occassionally seen crashes when using Cesium for Unreal, showing a call stack that indicates a problem with raster overlays. I think the cause is the use of an iterator after it has been invalidated, and it's fixed by this PR.

I originally made this change in #327 and didn't see the crash anymore afterward. That PR isn't quite going to make it into our November release, which is the reason for this separate PR.


kring:
I'm self-merging this because it's a clear improvement and should go into today's release.

PR #380, opened 2021-11-01
Added stricter GLM defines

IanLilleyT:
The [GLM manual](https://github.com/g-truc/glm/blob/master/manual.md#section2) has some defines we can set to make GLM more strict and hopefully avoid bugs:

```
GLM_FORCE_XYZW_ONLY # Disable .rgba and .stpq to make it easier to view values from debugger
GLM_FORCE_EXPLICIT_CTOR # Disallow implicit conversions between dvec3 <-> dvec4, dvec3 <-> fvec3, etc
GLM_FORCE_SIZE_T_LENGTH # Make vec.length() and vec[idx] use size_t instead of int
```

I was also thinking about adding `GLM_FORCE_INLINE` but there was a compilation failure in one of the unit tests where the GLM function could not be inlined. I haven't checked if `GLM_FORCE_INLINE` has any runtime improvement so it could be worth looking into some other time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,10 +92,12 @@ function(configure_cesium_library targetName)
         CXX_EXTENSIONS NO
     )
 
-    target_compile_definitions( ${targetName} PUBLIC
-        GLM_FORCE_XYZW_ONLY # Disable .rgba and .stpq to make it easier to view values from debugger
-        GLM_FORCE_EXPLICIT_CTOR # Disallow implicit conversions between dvec3 <-> dvec4, dvec3 <-> fvec3, etc
-        GLM_FORCE_SIZE_T_LENGTH # Make vec.length() and vec[idx] use size_t instead of int
+    target_compile_definitions(
+        ${targetName}
+        PUBLIC
+            GLM_FORCE_XYZW_ONLY # Disable .rgba and .stpq to make it easier to view values from debugger
+            GLM_FORCE_EXPLICIT_CTOR # Disallow implicit conversions between dvec3 <-> dvec4, dvec3 <-> fvec3, etc
+            GLM_FORCE_SIZE_T_LENGTH # Make vec.length() and vec[idx] use size_t instead of int
     )
 
     if (BUILD_SHARED_LIBS)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,7 +92,7 @@ function(configure_cesium_library targetName)
         CXX_EXTENSIONS NO
     )
 
-    target_compile_definitions( ${targetName} PRIVATE
+    target_compile_definitions( ${targetName} PUBLIC
         GLM_FORCE_XYZW_ONLY # Disable .rgba and .stpq to make it easier to view values from debugger
         GLM_FORCE_EXPLICIT_CTOR # Disallow implicit conversions between dvec3 <-> dvec4, dvec3 <-> fvec3, etc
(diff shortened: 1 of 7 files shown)


javagl:
+1 for `GLM_FORCE_EXPLICIT_CTOR` alone. It kinda baffled me that Travis builds keep failing because of obscure rules that compilers establish for conversions between integer types, but GLM silently converted between structurally and semnatically totally unrelated types like "a 3D point" and "a 4D point"...

CMakeLists.txt
```diff
@@ -92,6 +92,12 @@ function(configure_cesium_library targetName)
         CXX_EXTENSIONS NO
     )
 
+    target_compile_definitions( ${targetName} PRIVATE
```
kring, line 95: I think we want `PUBLIC` rather than `PRIVATE`, so that other libraries that depend on cesium-native via cmake will use the same definitions.

CMakeLists.txt
```diff
@@ -92,6 +92,12 @@ function(configure_cesium_library targetName)
...
+        GLM_FORCE_XYZW_ONLY # Disable .rgba and .stpq to make it easier to view values from debugger
+        GLM_FORCE_EXPLICIT_CTOR # Disallow implicit conversions between dvec3 <-> dvec4, dvec3 <-> fvec3, etc
+        GLM_FORCE_SIZE_T_LENGTH # Make vec.length() and vec[idx] use size_t instead of int
```
kring, line 98: I don't think cesium-native uses `.length()` anywhere, so I'm not too fussed about this. But I am curious: what problem does this one solve for you?

kring (commented):
Thanks @IanLilleyT! Just one suggested change and a question below.

Big 👍 from me too on `GLM_FORCE_EXPLICIT_CTOR`. BTW @javagl the reason for those conversion warnings/errors on GCC and Clang is because we've enabled a high warning level that flags them. By default those compilers wouldn't care, same as VC++.

CMakeLists.txt
IanLilleyT: `.length` doesn't matter for me that much, but bracket operator does. It's to do `vec[idx]` instead of `vec[int(idx)]` for size_t types.

IanLilleyT: Since cesium-native's use of GLM carries over to libraries that use cesium-native, I'm ok with keeping `int` as the default to minimize disruption. But it is annoying to cast every time when accessing with brackets

CMakeLists.txt
IanLilleyT: Do you know how to make these values propagate? Doesn't look like it's happening automagically

kring: I was under the impression that they'll propagate automatically if they're declared PUBLIC, but I've never really tested it.

CMakeLists.txt
kring: I really don't have strong feelings on this either way, I was just curious what the motivation was. That makes sense if you frequently have size_t indices.

CMakeLists.txt
IanLilleyT: I think the problem is I'm using ExternalProject_Add for cesium-native so it doesn't have a real target to work with. But I changed the PR to be `PUBLIC` so that it works with `add_subdirectory` or similar

IanLilleyT:
@kring ready for another review. The only difference is it uses `PUBLIC` instead of `PRIVATE`

kring:
Thanks @IanLilleyT!

Issue #364, opened 2021-10-11
Add support for the 3DTILES_bounding_volume_S2 extension


kring:
https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next/extensions/3DTILES_bounding_volume_S2

This builds on #355.

We'll need:
* A new `S2BoundingVolume`, and to include it in the `BoundingVolume` variant.
* A new `S2TileID` in the TileID variant.
* The implicit tiling code will need to know how to subdivide S2TileIDs in addition to quadtrees and octrees.
* For tile selection, we'll need to be able to cull the new bounding volume against the frustum and to compute the distance to it from the camera. Look to the CesiumJS implementation and to https://github.com/google/s2geometry for inspiration. Worst case scenario, we can compute an `OrientedBoundingBox` from the S2 volume and use that for culling and distance checks, similar to what we do for `BoundingRegion`.

nithinp7:
Finished after #387 and #390.

PR #387, opened 2021-11-05
Add support for 3DTILES_bounding_volume_S2

kring:
This does _not_ cover the implicit tiling side of it (that's a separate PR). Just support for the extension and new bounding volume type.

Still needs a bit more test and doc work.

This covers part of #364.


kring:
I'm reasonably happy with the doc and test situation now, so I think this is ready.

nithinp7:
@kring I cherry picked some small changes made on `implicit-s2` that belong here, including the addition of `fromQuadtreeTileID(...)`, `getFace()`, the `HilberOrder` class, and formatting / gcc + clang warning fixes.

PR #390, opened 2021-11-09
Integrate S2 Bounding Volume With Implicit Tiling

nithinp7:
This is a PR into: https://github.com/CesiumGS/cesium-native/pull/355 to integrate the S2 bounding volume (implemented by @kring [here](https://github.com/CesiumGS/cesium-native/pull/387)) with implicit tiling support.


nithinp7:
@kring There is an issue in this PR, or maybe related to S2 in general where adding overlays to an S2 tileset causes a scary-looking crash in Unreal (in this case an implicit S2 tileset but I don't know if that matters). The crash seems to happen when we attempt to build physics meshes in cesium-unreal. Inspecting the vertex buffers there immediately before creating the physics mesh, it appears that the UV coords and possibly the normals have denormalized, invalid double values.

kring:
Changes look good (surprisingly simple!). Once CI passes I'm going to merge this, ignoring the raster overlay crash for now. Then I'll merge it into #399, try it there, and fix raster-overlays-on-S2 in that branch if they're still broken.

PR #393, opened 2021-11-12
Add a macOS build to TravisCI

kring:


kring:
In addition to activating CI for macOS, this PR fixes a macOS compile problem with s2geometry and changes one test slightly so its output is slightly better when it fails. The tests are not passing on macOS (see #394), but they've been temporarily disabled so we still have a green tick here.

kring:
I'm merging this because the s2geometry problem fixed in this PR is preventing successfully Cesium for Unreal builds against main at the moment.

PR #396, opened 2021-11-16
Add DebugColorizeTilesRasterOverlay

kring:
This is a simple raster overlay that adds a random 50% opaque color to every tile, which is useful for visualizing how a tileset is broken up into tiles. Here's Melbourne:

image

There are lots of ways to do this, but the nice part about doing it with a raster overlay (a 1x1 texture per tile) is that it doesn't require any changes to existing code, just a new raster overlay.


nithinp7:
This looks super useful, thanks @kring!

PR #397, opened 2021-11-16
Fix compiler error when tracing shared future

baothientran:


kring:
Thanks @baothientran!

PR #398, opened 2021-11-18
Fix macOS test failures.

### LINKED ISSUE #394: Tests fail on macOS (xcode12)
This has probably been true for a long time, but we didn't have a build until #393.

A few failures seem to be caused by serializing the contents of a `vector<bool>` to JSON, reading it back and getting integers instead of booleans. I really don't understand why.

Another failure is a segfault in the tests of the `CachingAssetAccessor`. Again, I don't understand why.

See a complete log here:
https://app.travis-ci.com/github/CesiumGS/cesium-native/jobs/547862152

kring:
Fixes #394 

I think the problem is that `operator[]` for `std::vector<bool>` doesn't return an actual bool on this platform, but rather something that can be implicitly converted to bool. When we pass that thing to RapidJSON's PushBack, it still doesn't get implicitly converted to a bool. So the JSON ends up incorrect. Fixed by explicitly casting return return value of `operator[]` to the right type.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,6 @@ jobs:
       osx_image: xcode12
       script:
         - cmake --build . --config Debug
-        - # Tests fail on macOS, see https://github.com/CesiumGS/cesium-native/issues/394
-        - ctest -V || echo "There are test failures, but we're not failing the build. See https://github.com/CesiumGS/cesium-native/issues/394"
+        - ctest -V
         - # Doc build doesn't work on Mac (missing Doxygen). Let's just skip it.
         - #cmake --build . --config Debug --target cesium-native-docs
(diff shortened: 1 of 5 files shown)


kring:
The caching accessor tests are still segfaulting on macOS, needs investigation.

kring:
Ok this is ready now! The segfaults on macOS were caused by a conflicting definition of `MockTaskProcessor` (and possibly other types). `MockTaskProcessor` was defined in two source files, `TestCacheAssetAccessor.cpp` and `TestAsyncSystem.cpp`. In neither case were these in the unnamed namespace, so they were the "same" class. Yet the definitions were slightly different. I guess it wasn't a multiply-defined linker error because all of the method implementations were also inlined. So I put both definitions into the unnamed namespace and the segfault went away.

It took ages to track this down. What eventually gave it away was when I tried to add some debug code to `MockTaskProcessor` and it had no effect!

nithinp7:
Good work hunting this one down @kring, thanks!

PR #399, opened 2021-11-19
Partial fix for broken upsampling with raster overlays.

kring:
This fixes the problem mentioned in https://github.com/CesiumGS/cesium-native/pull/355#issuecomment-972805713 where geometry upsampled for raster overlays would just disappear. That was caused by failing to account for the bufferView's byteOffset when reading positions for upsampling.

But even with that fix, texture coordinates generated for raster overlays are sometimes broken. The problem appears to be that the bound region for upsampled tiles is (sometimes) very wrong. This PR fixes the symptom, but I need to spend some more time understanding what is really going on here, which is why this PR is a draft for the moment.

### DIFF
--- a/Cesium3DTilesSelection/src/BoundingVolume.cpp
+++ b/Cesium3DTilesSelection/src/BoundingVolume.cpp
@@ -200,13 +200,13 @@ estimateGlobeRectangle(const BoundingVolume& boundingVolume) {
 const CesiumGeospatial::BoundingRegion*
 getBoundingRegionFromBoundingVolume(const BoundingVolume& boundingVolume) {
   const BoundingRegion* pResult = std::get_if<BoundingRegion>(&boundingVolume);
-  if (!pResult) {
-    const BoundingRegionWithLooseFittingHeights* pLoose =
-        std::get_if<BoundingRegionWithLooseFittingHeights>(&boundingVolume);
-    if (pLoose) {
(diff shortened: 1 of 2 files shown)


kring:
Ok this is ready now. It... grew. See the changelog for all that changed.

The two big bugs were:
1) we weren't accounting for the glTF bufferView.byteOffset when accessing positions for upsampling (so the upsampled geometry was totally wrong if the byteOffset wasn't zero).
2) There was a typo (copy/paste error probably) in `createQuadtreeSubdividedChildren` that caused incorrect bounding volumes for some of the children in the non-implicit case. (see https://github.com/CesiumGS/cesium-native/blob/f3b2dc7d545657f27012dbb6920afc78d854eb5e/Cesium3DTilesSelection/src/Tile.cpp#L815, `se` repeated where we should have `ne` instead).

But in the course of finding and fixing those two problems, I noticed that our bounding region and texture coordinate generation code was not great. It mostly worked, most of the time, but large tiles were a mess. Some of the low-detail Cesium World terrain tiles would end up with pretty wrong bounding regions, for example. So I ended up reworking it, and it should be much more reliable now.

kring:
I tried adding a raster overlay to the S2 globe in this branch, and it worked fine, but there were a bunch of errors printed to the console because of primitives with no vertices. I added some code to eliminate these empty primitives when upsampling and everything looks good now.

Issue #400, opened 2021-11-21
Access Violation In copyDecodedAttribute function


lyhkop:
in this code:
`
const auto doCopy = [pMesh, pAttribute, numberOfComponents](auto pOut) {
    for (draco::PointIndex i(0); i < pMesh->num_points(); ++i) {
      const draco::AttributeValueIndex valueIndex = pAttribute->mapped_index(i);
      pAttribute->ConvertValue(valueIndex, numberOfComponents, pOut);
      pOut += pAttribute->num_components();
    }
};
`
pMesh->num_points() maybe greater than pOut size which reference buffer.cesium.data，that will causes 'memory access violation

'

kring:
Yes I think you're right @lyhkop. Here's the specific code:
https://github.com/CesiumGS/cesium-native/blob/e8cac1748afb0495dd127712774f1444483f4fef/CesiumGltfReader/src/decodeDraco.cpp#L234

A simple solution is to change:

```
for (draco::PointIndex i(0); i < pMesh->num_points(); ++i) {
```

To:

```
for (int64_t i = 0; i < pAccessor->count; ++i) {
```

Because pAccessor->count is guaranteed to be `<= num_points()` by a check earlier in the code.

PR #401, opened 2021-11-21
draco mesh vertices number maybe greater than attribute accessor count

lyhkop:
it will fix the issuses "https://github.com/CesiumGS/cesium-native/issues/400"

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,10 @@
 - Added `GltfContent::applyRtcCenter` and `applyGltfUpAxisTransform`.
 - Clipping polygon edges now remain sharp even when zooming in past the available geometry detail.
 
+##### Fixes :wrench:
+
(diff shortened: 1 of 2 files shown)


kring:
Thanks for the PR @lyhkop! I originally thought we should allow the accessor to have a `count` that is _less_ than the number of points in the Draco mesh. But re-reading the `KHR_draco_mesh_compression` spec, it's not really clear that this is the case. So I think this change is ok (and certainly it's better than memory corruption!).

Before I can merge this, though, can you please sign the Contributor License Agreement? https://github.com/CesiumGS/cesium/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla

This ensures that we and other cesium-native users are allowed to use your changes. Thanks again!

lyhkop:
@kring Hi kring  I've signed the CLA. You can merge this pr, when you have time.

kring:
Thanks @lyhkop!

PR #402, opened 2021-11-22
Support KTX2 Texture Compression

### LINKED ISSUE #196: Assume terrain tiles are quantized-mesh even if the Content-Type isn't right

nithinp7:
Supports the [KHR_texture_basisu extension](https://github.com/CesiumGS/glTF/tree/main/extensions/2.0/Khronos/KHR_texture_basisu). 

Uses the [KTX-Software library](https://github.com/KhronosGroup/KTX-Software) to transcode KTX v2 textures into the ideal, platform-specific gpu pixel compression format. Clients can report which formats are supported on the current platform and Cesium Native will choose the ideal one for each particular case. Optionally, the KTX v2 texture can be fully decompressed to raw pixels as well.

Main Changes:
- `TilesetOptions` and `RasterOverlayOptions` each have a new option to report which formats are supported on the client platform. Ideal formats amongst the available ones are picked by roughly following the logic [specified here](https://github.com/KhronosGroup/3D-Formats-Guidelines/blob/main/KTXDeveloperGuide.md).
- The image API can now convey which gpu pixel compression format (if any) is used. This will inform what to expect in the image's pixel buffer.
- The image API can now also contain pre-computed mipmaps, if they exist. In that case, all the mips will be in the pixel buffer and the delineation between each mip will be described in `ImageCesium::mipPositions`. 
- Added the auto-generated classes for KHR_texture_basisu.
- Tilesets can now have content with the following extensions: ".gltf", ".glb", and ".terrain" (should fix #196 as well).


CesiumGltf/src/Ktx2TranscodeTargets.cpp
```diff
@@ -0,0 +1,120 @@
...
+    this->ETC1S_RGB = GpuCompressedPixelFormat::BC1_RGB;
+  } else if (supportedFormats.PVRTC1_4_RGB) {
+    this->ETC1S_RGB = GpuCompressedPixelFormat::PVRTC1_4_RGB;
+  }
+
+  // Find the transcode target for ETC1S Red-Green
+  if (supportedFormats.ETC2_EAC_RG11) {
+    this->ETC1S_RG = GpuCompressedPixelFormat::ETC2_EAC_RG11;
+  } else if (supportedFormats.ETC1_RGB) {
+    this->ETC1S_RG = GpuCompressedPixelFormat::ETC1_RGB;
```
kring, line 43: The chart says to transcode to two ETC1 RGB textures here. I don't totally understand why, but it seems fairly clear:

> On platforms that support only ETC1 RGB, applications could transcode each ETC1S slice to its own ETC1 RGB texture and use two texture samplers simultaneously.

I'm fine with skipping the hassle of creating two textures for now and just moving on to the next option (even if that means uncompressed), but we probably shouldn't pack the two channels into one ETC1_RGB texture. We should also write an issue for remember to do the texture splitting in the future.

CesiumGltf/src/Ktx2TranscodeTargets.cpp
```diff
@@ -0,0 +1,120 @@
...
+  } else if (supportedFormats.PVRTC1_4_RGB) {
+    this->ETC1S_R = GpuCompressedPixelFormat::PVRTC1_4_RGB;
+  }
+
+  // Find the transcode target for UASTC RGBA
+  if (supportedFormats.ASTC_4x4_RGBA) {
+    this->UASTC_RGBA = GpuCompressedPixelFormat::ASTC_4x4_RGBA;
+  } else if (supportedFormats.BC7_RGBA) {
+    this->UASTC_RGBA = GpuCompressedPixelFormat::BC7_RGBA;
+    // } else if (HighQuality) { // decode to pixels
```
kring, line 64: "PreserveHighQuality" could just be another flag in `SupportedGpuCompressedPixelFormats` I think, with appropriate documentation. Or just write an issue for the future.

CesiumGltf/src/Ktx2TranscodeTargets.cpp
```diff
@@ -0,0 +1,120 @@
...
+    this->UASTC_RG = GpuCompressedPixelFormat::BC7_RGBA;
+    // } else if (HighQuality) { // decode to pixels}
+  } else if (supportedFormats.ETC2_EAC_RG11) {
+    this->UASTC_RG = GpuCompressedPixelFormat::ETC2_EAC_RG11;
+  } else if (supportedFormats.BC5_RG) {
+    this->UASTC_RG = GpuCompressedPixelFormat::BC5_RG;
+  }
+  // TODO: else { decode to RG8 or RGR565 }
+
+  // Find the transcode target for UASTC RG
```
kring, line 105:
```suggestion
  // Find the transcode target for UASTC R
```

CesiumGltfReader/src/GltfReader.cpp
```diff
@@ -220,20 +221,21 @@ void postprocess(
...
 
   if (options.decodeDataUrls) {
-    decodeDataUrls(reader, readGltf, options.clearDecodedDataUrls);
+    decodeDataUrls(reader, readGltf, options);
   }
 
   if (options.decodeEmbeddedImages) {
     CESIUM_TRACE("CesiumGltf::decodeEmbeddedImages");
-    for (Image& image : model.images) {
+    for (CesiumGltf::Image& image : model.images) {
```
kring, line 229: There's a `using namespace CesiumGltf;` declaration at the top, so qualifying with `CesiumGltf::` shouldn't be necessary. Just for clarity?

CesiumGltfReader/src/GltfReader.cpp
```diff
@@ -422,16 +428,184 @@ Future<GltfReaderResult> GltfReader::resolveExternalData(
...
+  CESIUM_TRACE("CesiumGltf::readImage");
 
   ImageReaderResult result;
 
   result.image.emplace();
   CesiumGltf::ImageCesium& image = result.image.value();
 
+  if (isKtx(data)) {
+    // TODO: better error handling; make sure KTX-Software doesn't throw
+    // exceptions
```
kring, line 456: Can it? If so, wrapping in a try/catch is worthwhile so the exception doesn't escape cesium-native.

CesiumGltfReader/test/TestGltfReader.cpp
```diff
@@ -400,3 +400,15 @@ TEST_CASE("Can parse doubles with no fractions as integers") {
       options);
   CHECK(!result.warnings.empty());
 }
+
+TEST_CASE("KTX2_hacky_test") {
```
kring, line 404: Could use a better test, or at least a better name. :)

extern/CMakeLists.txt
```diff
@@ -4,6 +4,17 @@ add_subdirectory(Catch2)
...
+option(KTX_FEATURE_STATIC_LIBRARY "" on)
+option(KTX_FEATURE_TESTS "" off)
+add_subdirectory(KTX-Software)
+
+#get_target_property(ktx_read_defs ktx_read COMPILE_DEFINITIONS)
+#message("********************${ktx_read_defs}")
+#list(REMOVE_AT ktx_read_defs 0)
+#STRING(REPLACE "$<$<CONFIG:Debug>:_DEBUG;DEBUG>;" "" ktx_read_defs "${ktx_read_defs}")
+#message("${ktx_read_defs}")
+#set_target_properties(ktx_read PROPERTIES COMPILE_DEFINITIONS "${ktx_read_defs}")
```
kring, line 16: Remove this?

kring (commented):
Looks awesome @nithinp7! Just a few small things below.

CesiumGltf/include/CesiumGltf/Ktx2TranscodeTargets.h
```diff
@@ -0,0 +1,127 @@
...
+ * {@link GpuCompressedPixelFormat}s on a client platform.
+ *
+ * Clients can use this struct to convey which gpu compressed pixel formats are
+ * supported. This can be used to construct a {@link Ktx2TranscodeTargets}.
+ *
+ * Each entry in the struct is a bool that represents whether the gpu
+ * compressed pixel format with the corresponding name is supported.
+ */
+struct CESIUMGLTF_API SupportedGpuCompressedPixelFormats {
+  bool ETC1_RGB;
```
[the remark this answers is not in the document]
nithinp7: These would default to false, no? Or is it better for clarity to explicitly initialize to false?

Cesium3DTilesSelection/src/TileContentFactory.cpp
```diff
@@ -35,6 +35,27 @@ void TileContentFactory::registerContentType(
 
 CesiumAsync::Future<std::unique_ptr<TileContentLoadResult>>
 TileContentFactory::createContent(const TileContentLoadInput& input) {
+  // This is a hack to support gltf/glb content directly. Is there a more
+  // elegant way to do this instead of check the extension here?
```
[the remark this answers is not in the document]
nithinp7: Yeah this was totally a hack so I could put together KTX2 tilesets from KTX2 glTFs. I was thinking maybe we should support (EXT_content_gltf)[https://github.com/CesiumGS/3d-tiles/tree/main/extensions/3DTILES_content_gltf) instead. We probably shouldn't encourage invalid tilesets by being lenient on the content type like this. I will add support for that extension today.

Cesium3DTilesSelection/src/TileContentFactory.cpp
```diff
@@ -35,6 +35,27 @@ void TileContentFactory::registerContentType(
...
+  const std::string& url = input.pRequest->url();
+  size_t extensionPos = url.rfind(".");
```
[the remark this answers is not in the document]
nithinp7: It should now be removing the query parameters. Is there anything else I should look for and cut off on the rhs of the url before I look for the extension? Are hash parameters  possible in this situation? I assumed not.

Cesium3DTilesSelection/src/TileContentFactory.cpp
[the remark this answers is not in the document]
nithinp7: I just added a file extension to loader registration for when the magic and content type fail to determine a loader. It now supports .gltf, .glb, and .terrain.

CesiumGltf/include/CesiumGltf/Ktx2TranscodeTargets.h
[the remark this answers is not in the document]
kring: No, I don't think they're guaranteed to be initialized to false. In particular if you use it like this:

```
SupportedGpuCompressedPixelFormats supported;
...
```

Then fields of primitive type will have an undefined value.

But I'm not 100% sure because [Initialization in C++ is Bonkers](https://accu.org/journals/overload/25/139/brand_2379/). Best advice is to always initialize. You can do so here just by adding `{}` after each field:

```
bool ETC1_RGB{};
```

CesiumGltf/include/CesiumGltf/Ktx2TranscodeTargets.h
```diff
@@ -0,0 +1,129 @@
...
+   */
+  GpuCompressedPixelFormat UASTC_RGB = GpuCompressedPixelFormat::NONE;
+
+  /**
+   * @brief The gpu pixel compression format to transcode RGBA UASTC textures
+   * into. If NONE, it will be decompressed into raw pixels.
+   */
+  GpuCompressedPixelFormat UASTC_RGBA = GpuCompressedPixelFormat::NONE;
+
+  Ktx2TranscodeTargets() {}
```
kring, line 113:
```suggestion
  Ktx2TranscodeTargets() = default;
```

CesiumGltfReader/test/TestGltfReader.cpp
kring: The test doesn't seem too hacky to me (maybe the tested model is), so how about just giving it a better name?

Cesium3DTilesSelection/src/TileContentFactory.cpp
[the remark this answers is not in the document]
kring: No I think this is fine. Ideally we'd use `uriparser` instead of a manual URI manipulation, but it's ok for now. Really we need a better URI library that isn't such a hassle to use (mostly cause it's 'C' and very low-level).

nithinp7:
@kring I think this should be done now, I addressed all your remaining comments!

kring:
Looks great, thanks @nithinp7! Please update CHANGES.md in main.

PR #404, opened 2021-11-24
set indices accessor offset value to 0 after decode draco mesh

lyhkop:
i think accessor offset value should set to 0, because we create a new bufferview and buffer.


kring:
Thanks @lyhkop, I agree. From [the spec](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_draco_mesh_compression/README.md#conformance):

> When loading each accessor, you must ignore the bufferView and byteOffset of the accessor and go to the previously decoded Draco geometry in the primitive to get the data of indices and attributes.

Thanks for the PR!

PR #408, opened 2021-12-01
Support textures in EXT_feature_metadata 

nithinp7:
Expands EXT_feature_metadata support by adding convenient "views" for feature textures and feature id textures. This helps facilitate CPU-based picking implementations when combined with a front-end implementation.

In particular, this PR adds:
- Utility to query the feature ID at a given texture coordinate in a feature ID texture.
- Utility to query the property value at a given texture coordinate of a feature texture property.
- Several miscellaneous improvements to the metadata API to better support rendering engine integrations (this has been integrated successfully here: https://github.com/CesiumGS/cesium-unreal/pull/698).


CesiumGltf/include/CesiumGltf/FeatureTexturePropertyView.h
```diff
@@ -0,0 +1,208 @@
...
+  InvalidTextureIndex,
+  InvalidTextureSamplerIndex,
+  InvalidTextureSourceIndex,
+  InvalidChannelsString
+};
+
+enum class FeatureTexturePropertyComponentType {
+  Uint8
+  // TODO: add more types. Currently this is the only one outputted by stb,
+  // so change stb call to output more of the original types.
```
kring, line 32: Even if STB only outputs images in certain formats, a user could construct a glTF with textures in alternate formats, right? We should support them, unless it creates a lot of complexity.

CesiumGltf/include/CesiumGltf/FeatureTexturePropertyView.h
```diff
@@ -0,0 +1,208 @@
...
+    if (textureAccessor.channels.length() > 4 ||
+        textureAccessor.channels.length() > this->_pImage->channels ||
+        textureAccessor.channels.length() !=
+            static_cast<size_t>(this->_componentCount)) {
+      this->_status = FeatureTexturePropertyViewStatus::InvalidChannelsString;
+      return;
+    }
+
+    for (int i = 0; i < textureAccessor.channels.length(); ++i) {
+      // TODO: should this be case-sensitive?
```
kring, line 109: Another question for the 3D Tiles team I think.

CesiumGltf/include/CesiumGltf/FeatureTexturePropertyView.h
```diff
@@ -0,0 +1,208 @@
...
+        _pImage(nullptr),
+        _pClassProperty(&classProperty),
+        _textureCoordinateIndex(textureAccessor.texture.texCoord),
+        _status(FeatureTexturePropertyViewStatus::InvalidUninitialized),
+        _channelOffsets(),
+        _type(FeatureTexturePropertyComponentType::Uint8),
+        _componentCount(0),
+        _normalized(false) {
+
+    if (textureAccessor.texture.index < 0 ||
```
kring, line 71: This implementation should be in a CPP file too.

CesiumGltf/src/MetadataFeatureTableView.cpp
```diff
@@ -124,9 +124,13 @@ MetadataPropertyViewStatus MetadataFeatureTableView::getBufferSafe(
     return MetadataPropertyViewStatus::InvalidValueBufferIndex;
   }
 
+  // TODO: investigate why this causes lots of metadata gltfs to fail, are they
+  // all incorrectly built?
```
kring, line 128: We should check with the 3D Tiles team on this, too, if you haven't already.

kring (commented):
Looks good @nithinp7! A few comments below.

CesiumGltf/include/CesiumGltf/FeatureIDTextureView.h
```diff
@@ -0,0 +1,150 @@
...
+
+    // Ignore sampler, we will always use nearest pixel sampling.
+    if (texture.source < 0 || texture.source >= model.images.size()) {
+      this->_status = FeatureIDTextureViewStatus::InvalidImageIndex;
+      return;
+    }
+
+    this->_pImage = &model.images[texture.source].cesium;
+
+    // TODO: is this right?
```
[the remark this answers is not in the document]
javagl: That's an interesting one. I also stumbled over this, c.f. https://github.com/CesiumGS/3d-tiles-samples/pull/40#issuecomment-1019315404 . Actually, the spec does not say much about this, so I looked into the glTF spec and ... this does not say much about this either: It talks about red, green, blue and alpha channels, but does not _explicitly_ say anything about how many channels there "are" (or might be), or the order of these channels (particularly for the case that there are fewer than 4 channels). 

Vaguely related statements from https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#metallic-roughness-material might be...

> The base color texture MUST contain 8-bit values encoded with the sRGB opto-electronic transfer function ..

(even though that does not say, for example, whether an alpha channel is present), and

> For example, assume an 8-bit RGBA value of [64, 124, 231, 255] is sampled from baseColorTexture  ...

which indicates that RGBA may be stored there, in this order, but that's only an example, and not sufficient for answering the underlying question.

Surprisingly, the part at https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_texture_basisu/README.md#ktx-v2-images-with-basis-universal-supercompression seems to be more specific in terms of matching the different textures with their types, which channels are present, and the order of these channels. The answer to the question from the code...

>     // For instance, if the image has two channels, will those be referred to
>     // as r and g or r and a? 

might be that it is "r and g" based on that, but that's just a gut feeling for now.

(It might even be that this question is worth being "escalated" it to an issue in the glTF repo...)

CesiumGltf/src/MetadataFeatureTableView.cpp
javagl: Even at the risk of this being the result of confusion on my side (in terms of `EXT_feature_metadata` and `EXT_mesh_features`, and "Property Table" vs. "Feature Table"): The spec says at https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_mesh_features#property-tables 

> > Each buffer view byteOffset must be aligned to a multiple of its component size.
>
> **Implementation note**: Authoring tools may choose to align all buffer views to 8-byte boundaries for consistency, but  client implementations should only depend on 8-byte alignment for buffer views containing 64-bit component types.

It **might** well be that some of the glTFs from https://github.com/CesiumGS/3d-tiles-samples/pull/40 are "invalid" in that way, because ... from a glTF perspective, a `bufferView` does not have "components". It's just raw data. The components are defined via the _accessor_, and here, we don't have an accessor ( c.f. https://github.com/CesiumGS/glTF/pull/44 ). 

From a quick glance, I think that the examples generally only have one buffer view with an offset of 0, so it would be good to know which examples fail exactly. 

Beyond that: The fact that a (core glTF) `bufferView` may have a "component type" _implicitly_ defined via the metadata that refers to it raises some questions. I'd consider to review this more thoroughly on a spec level,

nithinp7: @javagl For EXT_feature_metadata (which is what this PR is aiming to support currently), the spec seems to say:
> Each buffer view byteOffset must be aligned to a multiple of 8 bytes. If the buffer view's buffer is the GLB-stored BIN chunk the byte offset is measured relative to the beginning of the GLB. Otherwise it is measured relative to the beginning of the buffer.

javagl: So indeed seems to be something that changed in a subtle way between EXT_feature_metadata and EXT_mesh_features. I wonder how this implementation will eventually be "lifted up" to the EXT_mesh_features state. I've seen quite some `if`-statements for the "legacy" path in the CesiumJS code, wondered whether this will also be necessary here, and particularly, how such subtle changes like a changed/relaxed alignment requirement will be covered. But fingers crossed, many of the necessary changes might be derived from unit tests.

kring:
@nithinp7 is this one ready for another review? It looks good for the most part but I think there are is a small comment or two not addressed (even by saying we're not going to address it right now).

kring:
Sorry for the delay on this one @nithinp7, it all looks good! Please just update CHANGES.md in main.

PR #409, opened 2021-12-02
Miscellaneous cleanup and code generator changes

lilleyse:
This PR is mix of cleanup and code generator changes that have been happening behind the scenes on the gltf-writer branch that are ready to go into main before the rest of the code.

* **Namespace changes**: namespaces are now based on the project name rather than conceptual groups. Previously CesiumGltfReader, CesiumGltfWriter, and CesiumGltf shared the CesiumGltf namespace. Now they are separate namespaces. This is a breaking API change.
* **Folder structure changes**: split up generated code from non-generated code, e.g. Accessor and AccessorSpec are no longer in the same folder. Generated code now always goes in a `generated` folder. The directory structure for each project looks like:

```
	generated
		include
		src
	include
	src
```

* ... where public headers go in `include` and private headers and cpp files go in `src`, like before. Generated files can refer to non-generated files in a different location so for simplicity all include paths are now scoped with the project name. This has no effect on the public API.
* **Autogenerated extension registration**: extensions registration is no longer hand-coded in GltfReader.cpp and TilesetReader.cpp. They now call an autogenerated function. See `registerExtensions`. That way if you add an extension you no longer need to remember to update the code.
* Updated some third party libraries due to build problems in other repos that use cesium-native: httplib and rapidjson

CC @IanLilleyT who was involved in a lot of this work as well


kring:
Thanks @lilleyse! This looks good for the most part. Why was it necessary, though, to add library-name subdirectories of `src` and `generated/src` and include private headers as `#include "CesiumGltf/Whatever.h"` rather than `#include "Whatever.h"`? Can't we just add both `src` and `generated/src` as private include paths?

lilleyse:
@kring That was for consistency so that all include paths were project-scoped, but no practical reason. I changed the folder structure back based on your suggestion.

Cesium3DTiles/generated/include/Cesium3DTiles/Asset.h
```diff
@@ -2,7 +2,7 @@
 // DO NOT EDIT THIS FILE!
 #pragma once
 
-#include "Library.h"
+#include "Cesium3DTiles/Library.h"
```
kring, line 5: Ok so on the one hand, this is totally not-that-important. But on the other hand, we recently merged two PRs that made this all more consistent, so we should make sure we're not about to merge a regression...

So the rule is supposed to be that:
1) Public headers should always include other public headers _in the same library_ using quotes (rather than angle brackets).
2) #includes that use quotes should be paths relative to the current file.

This change is violating (2), because the new path isn't resolvable as a relative path.

But it's awkward, right? The correct relative path is `../../../include/Cesium3DTiles/Library.h`. And in addition to being a mouthful, that's not actually going to be the correct relative path after we install.

I want to say that the correct approach here is to consider `include/Cesium3DTiles` and `generated/include/Cesium3DTiles` to be _the same path_, because after install they will be, and just use `#include "Library.h"`. But I can't really see a good way to make that work. We'd have to add `include/Cesium3DTiles` and `generated/include/Cesium3DTiles` to the library's public header path. It can't even be private, because other cesium-native libraries need to be able to resolve that `#include "Library.h"`. So that's really painful.

So the most consistent approach I can think of (short of going back to not having a `generated` directory for public headers) is to just always include public headers using fully-qualified angle brackets. So in this case `#include <Cesium3DTiles/Library.h>`. And we'd do the same everywhere, including in cpp files. The only place we'd use quotes is for private headers in the same directory as their including cpp file.

If that's the plan, it can be a separate PR. Does it make sense though? Any better ideas?

kring:
This looks good other than that one bike-sheddy question, so I'm merging it!

Cesium3DTiles/generated/include/Cesium3DTiles/Asset.h
lilleyse: I'm on board with that, though one annoying side effect of moving to fully qualified angle brackets is that clang-format no longer sorts the main header first. I'm not sure what changes would need to be made to `.clang-format` or if this is just some inherent limitation.

Previously `Model.cpp` would look like:

```
#include "CesiumGltf/Model.h"

#include "CesiumGltf/AccessorView.h"
#include "CesiumGltf/ExtensionKhrDracoMeshCompression.h"
```

Now it looks like

```
#include <CesiumGltf/AccessorView.h>
#include <CesiumGltf/ExtensionKhrDracoMeshCompression.h>
#include <CesiumGltf/Model.h>
```

PR #410, opened 2021-12-02
Use 2 space indentation for glTF writer

lilleyse:
Thoughts on changing the default indentation from 4 spaces to 2 spaces for the glTF writer?


kring:
Sounds good to me, merging.

PR #411, opened 2021-12-06
Refactor tileset loading code

kring:
There are no functional changes here. This PR just moves some of the loading-related code out of Tileset.cpp and into new .cpp files specific to the load operation. Probably not going to win any design awards here, but it removes over a thousand lines of code from Tileset.cpp, which hopefully makes things a little easier to understand in pieces. The new home for the removed code is in nested classes of Tileset, which are implicity friends of Tileset, so these new classes can still manipulate Tileset itself as necessary (yep definitely not going to win that design award).

A fun twist is that  these "implementation detail" loading classes themselves have a private portion in a further-nested class (named `Private`). This way methods used internally as part of the loading process don't need to be visible in the main load operation header, which must be included in Tileset.cpp in order to be callable. Mostly to reduce the number of includes required. Welcome to cesium-native, where our implementation details have implementation details. 😎 

The bigger picture here is that we need to make some changes to the load processes, in order to provide more detailed error reporting back to cesium-unreal and other clients (i.e. not just logging), in order to respond better to things like invalid/expired Cesium ion tokens. So this is just a bit of refactoring ahead of that.


cesium-concierge:
Thanks again for your contribution @kring! 

nithinp7:
@kring This looks really good, I like the encapsulation of this pattern and I'm glad to see `Tileset.cpp` slim down! 

We should do something similar to decouple the traversal implementation from the Tileset as well. I think the only Tileset state-info that the traversal needs is `TilesetOptions` and its output is the collection of load queues. The Tileset state is essentially unmutated after the traversal (since the load queues are emptied every frame, they are pretty much "output" even though on an implementation-level they are a state change). Anyways, so I think it would be pretty easy to pull apart in a similar pattern, but of course it doesn't have to be part of this PR, I'll write an issue for it.

Once we pull apart the traversal as well, we'll be pretty close (I think) to being able to use the Cesium3DTilesReader types. The Cesium3DTilesReader integration may also allow simple drop-in replacements for `TilesetLoadTileFromJson` and `TilesetLoadTilesetDotJson` now that they are cleanly pulled apart from the Tileset.

The code changes look good to me, I'll test it out with Unreal for a bit, merge master back in, and then I'll merge it. Thanks Kevin!

PR #412, opened 2021-12-07
3D Tiles Writer

lilleyse:
Adds a new project `Cesium3DTilesWriter` that serializes a `Tileset` struct to a tileset JSON.

Much of this work is built on top of  https://github.com/CesiumGS/cesium-native/pull/306 and https://github.com/CesiumGS/cesium-native/pull/307 with similar generated code and improved handling of extensions, extras, and default values.

@krupkad's original summary still mostly applies

> Writer generation:
> - Writers are primarily implemented as overloads of `writeJson` in an anonymous namespace in cpp files
> - They are exposed as classes (e.g. `TilesetWriter`) with a single static method `write`, which just calls `writeJson`
> - The cpp files contain `writeJson` implementations for primitive types (double, string, integer, bool), composites (string->X dictionary, array), and `optional`
> 

The code generation is split between `generate.js` and `generateCombinedWriter.js`. `generate.js` creates the pieces for each writer and `generateCombinedWriter.js` combines them into a single header and cpp file. There's special handling for default values. We don't write a property if

* The property is optional and the value is the default value. This is important for both keeping the files small and preventing invalid files from being generated. For example, in glTF `matrix` and `translation/rotation/scale` are mutually exclusive and we shouldn't write out default values for both.
* The property is an optional glTF id with value -1
* The property is a vector or map type that's empty
* The property has the std::optional type and it doesn't have a value

The code has been tested with both 3D Tiles (this PR) and glTF (follow up PR) and seems to produce valid results for both, though there could be lingering issues with the default value checks.


kring:
Thanks @lilleyse!

PR #413, opened 2021-12-07
Support file:// under Windows

137734949:


kring:
Thanks for the PR @137734949. I don't think we can merge this in its current form, though. It looks like you're trying to address #196, but this isn't the right way to do it. We can't assume that any local file is a quantized mesh tile. And there's also nothing inherent to Windows about this that justifies the ifdefs. The original issue (#196) has some suggestions about how the underlying problem might be addressed, but they'll admittedly be a bit of work.

I'm going to close this PR, but feel free to re-open if you make further improvements and want us to take another look.

PR #416, opened 2021-12-17
Update GltfWriter to use autogenerated files

lilleyse:
This is a follow up to https://github.com/CesiumGS/cesium-native/pull/412 that replaces the current GltfWriter with an autogenerated GltfWriter.

The new writer takes a very hands-off approach. It serializes the CesiumGltf struct as-is without any special handling for external files or data uris. It's expected that the user would manage this on their own before calling `writeGltf` or `writeGlb`. This was to avoid injecting custom code in the autogenerated code or copying or modifying the `CesiumGltf` struct that's passed in. The full doc is below.

```
  /**
   * @brief Serializes the provided model into a glTF JSON byte vector.
   *
   * @details Ignores internal data such as {@link CesiumGltf::BufferCesium}
   * and {@link CesiumGltf::ImageCesium} when serializing the glTF. Internal
   * data must either be converted to data uris or saved as external files. The
   * buffer.uri and image.uri fields must be set accordingly prior to calling
   * this function.
   *
   * @param model The model.
   * @return The result of writing the glTF.
   */
  GltfWriterResult writeGltf(const CesiumGltf::Model& model) const;

  /**
   * @brief Serializes the provided model into a glb byte vector.
   *
   * @details The first buffer object implicitly refers to the GLB binary chunk
   * and should not have a uri. Ignores internal data such as
   * {@link CesiumGltf::BufferCesium} and {@link CesiumGltf::ImageCesium}.
   *
   * @param model The model.
   * @param bufferData The buffer data to store in the GLB binary chunk.
   * @return The result of writing the glb.
   */
  GltfWriterResult writeGlb(
      const CesiumGltf::Model& model,
      const std::vector<std::byte>& bufferData) const;
```


CesiumGltf/generated/include/CesiumGltf/Material.h
```diff
@@ -12,6 +12,9 @@
 #include <optional>
 #include <vector>
 
+// OPAQUE is defined in wingdi.h
+#undef OPAQUE
```
baothientran, line 16: this turns out to be quite cumbersome to use the glTF library in O3DE too. I think we should rename the enum to something else (Maybe add prefix gltf to enum)

baothientran: or maybe use `Opaque` as a name with the first letter capitalized (Not entirely fool-proofed for what window does, but seems to be good enough)

lilleyse: Yeah it's unfortunate. I'm not happy with `undef`'ing things in Cesium Native so maybe renaming is a better choice. We already do that already for properties that conflict with reserved cpp words, like "class" becomes "clasProperty". Maybe "OPAQUE" could become "OPAQUE_ENUM".

lilleyse:
Pushed another commit https://github.com/CesiumGS/cesium-native/pull/416/commits/108c90b8679f437cf8f96c9507df8e593d4316da so the enums in EXT_meshopt_compression are processed correctly ([here](https://github.com/KhronosGroup/glTF/blob/24ae1bb305f4be06ce675dfbfcbe8b9b2f0ef671/extensions/2.0/Vendor/EXT_meshopt_compression/schema/bufferView.EXT_meshopt_compression.schema.json#L34-L44)). It should also support the older style enums in AGI_articulations ([here](https://github.com/KhronosGroup/glTF/blob/24ae1bb305f4be06ce675dfbfcbe8b9b2f0ef671/extensions/2.0/Vendor/AGI_articulations/schema/articulation.stage.schema.json#L16-L60)) and KHR_lights_punctual ([here](https://github.com/KhronosGroup/glTF/blob/24ae1bb305f4be06ce675dfbfcbe8b9b2f0ef671/extensions/2.0/Khronos/KHR_lights_punctual/schema/light.schema.json#L30-L50)) though I haven't tried those.

CesiumGltf/generated/include/CesiumGltf/Material.h
lilleyse: Updated the code

lilleyse:
All the code generator and cleanup changes are now in https://github.com/CesiumGS/cesium-native/pull/418 so make sure to review that one first.

PR #420, opened 2022-01-13
Implicit Tiling: Fix endianness bug in child subtree index calculation

nithinp7:
Thanks @javagl for generating the thorough [sample implicit tileset here](https://github.com/CesiumGS/3d-tiles-samples/tree/main/next/SparseImplicitQuadtree) and for identifying this Cesium Native bug exposed by the tileset.


kring:
The change looks good to me @nithinp7. But can you please add a note to CHANGES.md mentioning the fix?

javagl:
As a totally unrelated side-task, I had a look at how https://github.com/CesiumGS/cesium-native/blob/implicit-fix/CesiumGeometry/src/OctreeAvailability.cpp#L226 could be addressed, and tried it out in https://github.com/CesiumGS/cesium-native/compare/remove-redundant-redundancy , but that's a quick stab (i.e. there are no doxygen comments yet), it only covers the `Quadtree...` one, and may not be considered to be important right now, but ... I'll leave it here as comment.

nithinp7:
@javagl Just had a quick glance, that looks a lot more elegant. I can finish building that branch out with doxygen comments and expand it for Octrees when I get the chance, thanks for the start there it looks a lot nicer!

javagl:
Note that these changes have been largely "syntactical": I just tried to spot equal parts of the code, and pull them out into methods, without trying to deeply _understand_ the code. I'm _reasonably_ sure that the changes do **not** affect the behavior in any way, but that has to be verified. (I don't know how far the behavior is covered with sorts of unit-tests). 

---

An aside: Beyond reducing the redundancy purely "syntactically", further refactorings might be possible by someone who more deeply understands what the code is supposed to do. For example, I was a bit confused by the `computeAvailability` function that computes all availabilities, and combines them in a single `uint8_t` (together with something that was called `REACHABLE`...). In some local ("experimental") implementations, I tried to keep these things separate, and to hide the "constant-vs-bufferView" distinction behind an interface, as in
```
interface ImplicitTreeInfo<T>
{
    AvailabilityInfo<T> getTileAvailabilityInfo();
    AvailabilityInfo<T> getContentAvailabilityInfo();
}
interface AvailabilityInfo<T>
{
    boolean isAvailable(T index);
}
```
But maybe that does not make sense for the use case of `cesium-native`. Maybe `cesium-native` does need the "child subtree availability" (which is deliberately not contained in the example above). And maybe there is no way to replace the "large-scale" redundancy of `QuadtreeAvailabilty` and `OctreeAvailability` ....

Cesium Availability

... with a single `Availability<T>`. All these thoughts might be a form of "idealism", that far too often stand in the way of _getting things done_.

PR #421, opened 2022-01-18
Notify tile done loading when ion response can't be parsed

baothientran:
We forget to notify tile done loading when ion response can't be parsed, which makes the tileset unable to be destructed


nithinp7:
Thanks @baothientran!

PR #425, opened 2022-01-20
Report Tileset and Raster Overlay load errors programmatically

kring:
This builds on #424 so merge that first.

So that we can handle them, rather than just logging them.


nithinp7:
This looks good to me and works well with https://github.com/CesiumGS/cesium-unreal/pull/731. Thanks @kring!

PR #429, opened 2022-01-21
enable IntrusivePointer to be created from a derived class

### LINKED ISSUE #276: Intrusive_ptr of derived class can't be assigned to Intrusive_ptr of base class
Example:
```C++
class Base {
public:
    virtual ~Base() = default;
    // implement releaseReference() and addReference() interface for Intrusive_ptr
}

class Derived : public Base {}

int main() {
    Intrusive_ptr<Derived> pDerived = new Derived();
    Intrusive_ptr<Base> pBase = pDerived; // Compiler error
    return 0;
}
```

Solution is to implement the constructor for the case:
```C++
template <typename T>
class Intrusive_ptr {
public:
   template<typename U>
   Intrusive_ptr(const Intrusive_ptr<U>& rhs, std::enable_if_t<std::is_convertible_v<T*, U*>, int> = 0);
}
```

We should do the same for assignment operator and equal operator.

Disclaimer I haven't tested the solution though

joseph-kaile:
Fixes #276

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,11 +17,11 @@
 ##### Additions :tada:
 
 - Added `Cesium3DTilesWriter` library.
+- Enable `IntrusivePointer<T>` to be converted to `IntrusivePointer<U>` if U is a base class of T.
 
 ##### Fixes :wrench:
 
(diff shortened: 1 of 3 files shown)


kring:
Thanks @joseph-kaile, this looks great! Can you please just update CHANGES.md to mention that this has been changed? Probably in the "Additions" section would be best.

joseph-kaile:
Thanks, I added a test which made me realize I was missing an include.

PR #433, opened 2022-01-27
Refine value should be case sensitive

### LINKED ISSUE #345: Should enum string value comparisons be case-sensitive?
Again, sparked by a [forum thread](https://community.cesium.com/t/add-3dtiles-gltf1-0/15171) where [this check in `Tileset.cpp`](https://github.com/CesiumGS/cesium-native/blob/fc3f11dd2bfd670a5c4fba1bcfa71f263c47742a/Cesium3DTilesSelection/src/Tileset.cpp#L809) causes the error message

> Tile contained an unknown refine value: replace

I'd say: **Yes**, the comparison should be case-*sensitive*. A tileset with `"refine" : "replace"` should be considered to be invalid, but I wanted to bring it up.

---

EDIT:

Just for reference: [In CesiumJS, `"replace"` is allowed](https://github.com/CesiumGS/cesium/blob/1837859300e61ae2ea1bc0e3832923f87e752fed/Source/Scene/Cesium3DTile.js#L148), but causes a warning. Surprisingly, something like `"RePlace"` would **not** print a warning, but silently be accepted... (why?)

joseph-kaile:
Fixes #345


Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -1051,10 +1051,28 @@ static void parseImplicitTileset(
...
+      std::string refineUpper = refine;
+      std::transform(
+          refineUpper.begin(),
+          refineUpper.end(),
+          refineUpper.begin(),
+          [](unsigned char c) -> unsigned char {
+            return static_cast<unsigned char>(std::toupper(c));
+          });
+      if (refineUpper == "REPLACE" || refineUpper == "ADD") {
+        SPDLOG_LOGGER_ERROR(
```
kring, line 1063: This should be `SPDLOG_LOGGER_WARN` rather than ERROR.

PR #434, opened 2022-01-27
Write GLTF extras

### LINKED ISSUE #311: The glTF content cannot be written with glTF writer

joseph-kaile:
Fixes #311

### DIFF
--- a/CesiumGltfWriter/src/Writer.cpp
+++ b/CesiumGltfWriter/src/Writer.cpp
@@ -115,6 +115,7 @@ CesiumGltfWriter::WriteModelResult writeModel(
   CesiumGltfWriter::writeExtensions(model.extensions, *writer);
 
   if (!model.extras.empty()) {
+    writer->Key("extras");
     CesiumJsonWriter::writeJsonValue(model.extras, *writer);
   }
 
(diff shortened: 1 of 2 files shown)


lilleyse:
I think the issue this PR is addressing was fixed in https://github.com/CesiumGS/cesium-native/pull/416 when we overhauled the glTF writer. E.g. `Writer.cpp` and `TestWriter.cpp` no longer exist in main.

joseph-kaile:
Thanks, I see the code now!

PR #435, opened 2022-01-28
Accept double as integer if fractional part is zero

### LINKED ISSUE #357: The integer JSON handler should accept (zero) fractional values
Properties that have the type `integer` in the schema may sometimes be written with a fractional part. For example, one has to anticipate to encounter something like
``` 
  "bufferViews" : [ {
    "buffer" : 0.0,
    ...
  }, {
```

The [section about JSON Encoding in the glTF specification](https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/Specification.adoc#27-json-encoding) explicitly says this as well:

> Some of glTF properties are defined as integers in the schema. Such values MAY be stored as decimals with a zero fractional part or by using exponent notation. Regardless of encoding, such properties MUST NOT contain any non-zero fractional value.

(Some related discussion: https://github.com/KhronosGroup/glTF/pull/1574 - the **tl;dr** is: It's difficult, and one has very limited control over what "JSON" is in the wild...)

---

Specifically (to my understanding), the `IntegerJsonHandler` should add overrides for `double` as well, at https://github.com/CesiumGS/cesium-native/blob/main/CesiumJsonReader/include/CesiumJsonReader/IntegerJsonHandler.h#L21 . 

One may consider printing a warning, maybe even configurable with a flag. And if there is a non-zero fractional part, it should *definitely* print a warning or issue an error. But for strings with a zero fractional part, it should not just *ignore* the value.

joseph-kaile:
Fixes #357

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,7 @@
 - Added `SchemaWriter` and `SchemaReader` for serializing and deserializing schemas in [EXT_mesh_features](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_mesh_features) and [3DTILES_metadata](https://github.com/CesiumGS/3d-tiles/tree/main/extensions/3DTILES_metadata).
 - Added `hasExtension` to `ExtensibleObject`.
 - Added `CESIUM_TESTS_ENABLED` option to the build system.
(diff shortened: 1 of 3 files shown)


kring:
Thanks @joseph-kaile, the code looks good. Can you please add a test for this and also update CHANGES.md?

PR #437, opened 2022-02-02
Ensure all elements in JSON array of double/floats are double/floats before parsing

joseph-kaile:
Fixes parsing the 3D Tiles with no texture from [here.](https://www.geospatial.jp/ckan/dataset/plateau-tokyo23ku-3dtiles-2020)


kring:
Good catch @joseph-kaile! I think anywhere that section of code sets any of the `isX` flags to anything other than false, it should be `&=` rather than `=`. So I see at least two more places it should be changed. Also, please update CHANGES.md.

PR #441, opened 2022-02-10
Load tiles with missing geometric error

### LINKED ISSUE #440: Allow missing geometricError property in leaf tiles
The 3D Tiles spec clearly requires a `geometricError` property in every tile, but at least one tool in the wild (FME) generates tilesets with this property missing in the leaf tiles. CesiumJS tolerates this, with a warning. cesium-native should do the same.

The code that handles it in CesiumJS is here, and we should take the same approach:
https://github.com/CesiumGS/cesium/blob/2de96feae0fd59abb70af6b9f10479455c642f35/Source/Scene/Cesium3DTile.js#L134

joseph-kaile:
Fixes #440
Before:
HighresScreenshot00009
After:
HighresScreenshot00010


kring:
Thanks @joseph-kaile!

PR #442, opened 2022-02-14
Cache ion endpoints

### LINKED ISSUE #382: Cache Cesium ion asset endpoint responses in memory to avoid repeated requests
cesium-native should avoid repeatedly requesting the some ion asset endpoint in a single process, like CesiumJS does within a single web page:
https://github.com/CesiumGS/cesium/blob/3831cb9ff1c5f63a2e656223e798eb0a4d52575e/Source/Scene/IonImageryProvider.js#L166

This accomplishes two things:
1. It's a small performance win for the user because they don't need to wait for the ion request/response more than once per asset per process.
2. It avoids using extra Bing sessions unnecessarily.

Mentioned on the forum here:
https://community.cesium.com/t/reducing-my-number-of-bing-maps-sessions-with-cesium-for-unreal/15894

joseph-kaile:
Fixes #382


Cesium3DTilesSelection/src/TilesetLoadIonAssetEndpoint.cpp
```diff
@@ -60,6 +60,9 @@ struct Tileset::LoadIonAssetEndpoint::Private {
           assert(false);
         });
   }
+
+  static std::unordered_map<std::string, std::shared_ptr<IAssetRequest>>
+      endpointCache;
```
kring, line 65: We shouldn't hold onto the IAssetRequest instance like this. The problem is that IAssetRequest is implemented by cesium-native's client (e.g. Cesium for Unreal or O3DE or whatever) and may (probably does) hold some of the client's resources hostage. For example, maybe there's a limit on the number of requests that can exist at once, and these requests will count against that limit. Or maybe it's really important that request resources be released before, say, the HTTP module shuts down, but we can't do that if a static field is keeping them alive.

So it may be possible to just cache the parts of the request/response that are needed. Or, if necessary, you can even create another class that implements IAssetRequest and store instances of that class in the cache instead. We do something similar in CachingAssetAccessor.cpp.

Cesium3DTilesSelection/src/TilesetLoadIonAssetEndpoint.cpp
```diff
@@ -149,12 +210,11 @@ Tileset::LoadIonAssetEndpoint::Private::mainThreadHandleResponse(
...
            attributionsIt->value.GetArray()) {
-
+        AssetEndpointAttribution endpointAttribution;
         const auto html = attribution.FindMember("html");
         if (html != attribution.MemberEnd() && html->value.IsString()) {
-          tileset._tilesetCredits.push_back(
-              tileset.getExternals().pCreditSystem->createCredit(
-                  html->value.GetString()));
+          endpointAttribution.html = html->value.GetString();
+          endpoint.attributions.push_back(endpointAttribution);
```
kring, line 217: Slightly cleaner / more performant to do:

```
AssetEndpointAttribution& endpointAttribution = endpoint.attributions.emplace_back();
endpointAttribution.html = html->value.GetString();
```

Cesium3DTilesSelection/src/TilesetLoadIonAssetEndpoint.cpp
```diff
@@ -183,44 +243,25 @@ Tileset::LoadIonAssetEndpoint::Private::mainThreadHandleResponse(
...
- * context.
- *
- * @param pContext The context
  * @param pIonResponse The response
- * @return Whether the update succeeded
+ * @return The access token if successful
  */
-bool updateContextWithNewToken(
-    TileContext* pContext,
+std::optional<std::string> GetNewAccessToken(
```
kring, line 261: Use camelCase for function names.

```suggestion
std::optional<std::string> getNewAccessToken(
```

kring (commented):
Looks good @joseph-kaile! Very minor comments below.

We should also extend this mechanism `IonRasterOverlay`, which has a nearly identical situation. But it can be a separate PR.

Cesium3DTilesSelection/src/TilesetLoadIonAssetEndpoint.cpp
joseph-kaile: Thanks for the tip.

kring:
Thanks @joseph-kaile!

PR #444, opened 2022-02-17
Exposed ion server as a parameter

xuelongmu:
To support projects that may be using another ion endpoint instead of the default cesium one, I have exposed the ion endpoint URL as a parameter. The default is `"https://api.cesium.com/"`, and it then constructs the url in `createEndpointResource` similar to https://github.com/CesiumGS/cesium/blob/34b0a1a245ddd4edfbf82c29e3c67baf019f7676/Source/Core/IonResource.js#L208


Cesium3DTilesSelection/src/TilesetLoadIonAssetEndpoint.cpp
```diff
@@ -62,21 +64,26 @@ struct Tileset::LoadIonAssetEndpoint::Private {
   }
 };
 
+std::string Tileset::LoadIonAssetEndpoint::Private::createEndpointResource(
+    Tileset& tileset) {
```
nithinp7, line 68: This argument can be const I think.

xuelongmu:
@nithinp7 Thanks for the feedback, I also found a bug in my boolean logic which I also fixed. Ready for another look

xuelongmu:
Bump

nithinp7:
@baothientran Oops sorry I didn't realize you were looking at this as well! Thanks @xuelongmu.

PR #447, opened 2022-02-24
Exclude skirt vertices from bounds

### LINKED ISSUE #405: Don't include skirts when computing bounding region
`GltfContent::computeBoundingRegion` and `GltfContent::createRasterOverlayTextureCoordinates` both use _every_ vertex to determine the bounds of the tile. When we've add skirts to hide cracks between tiles, including the skirt vertices in the computed min/max heights results in a larger-than-necessary bounding volume. We can improve our culling slightly by not doing this.

joseph-kaile:
Fixes #405


kring:
👍

PR #450, opened 2022-02-28
Add scope guard utility

### LINKED ISSUE #302: Better approach for avoiding allocations for distances during tile selection
In #283, I hastily added some code to avoid allocating a `vector<double>` to hold distances for each visited tile every frame. It works ok, but it's pretty ugly. We should see if there's a better way to avoid the allocation, or at least clean it up by using a proper RAII object rather than the dodgy unique_ptr+lambda trick.

baothientran:
fixes #302 

Provide ScopeGuard utility to automatically trigger a function upon exiting the scope

### DIFF
--- a/CesiumUtility/include/CesiumUtility/ScopeGuard.h
+++ b/CesiumUtility/include/CesiumUtility/ScopeGuard.h
@@ -39,7 +39,7 @@ public:
   ScopeGuard(ScopeGuard&& rhs) noexcept
       : _callExitFuncOnDestruct{rhs._callExitFuncOnDestruct},
         _exitFunc{std::move(rhs._exitFunc)} {
-    rhs.Release();
+    rhs.release();
   }
 
   ScopeGuard& operator=(const ScopeGuard& rhs) = delete;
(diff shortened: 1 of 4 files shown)


CesiumUtility/include/CesiumUtility/ScopeGuard.h
```diff
@@ -0,0 +1,84 @@
...
+    if (_callExitFuncOnDestruct) {
+      _exitFunc();
+    }
+  }
+
+  /**
+   * @brief Upon calling ScopeGuard::Release(), the guard will not execute the
+   * lambda function when exiting a scope
+   */
+  void Release() noexcept { _callExitFuncOnDestruct = false; }
```
kring, line 75:
```suggestion
  void release() noexcept { _callExitFuncOnDestruct = false; }
```

kring (commented):
Thanks @baothientran, this looks great. Just one nitpick below.

CesiumUtility/include/CesiumUtility/ScopeGuard.h
kring: No reason to wait another cycle for this. I made this tiny change myself and will merge this momentarily.

PR #452, opened 2022-02-28
Sort credits by frequency

### LINKED ISSUE #449: Sort credits based on the number of occurrences
A a credit that occurs in more tiles will be displayed before a tile that occurs in fewer tiles.

https://github.com/CesiumGS/cesium/issues/10128

joseph-kaile:
Fixes #449

Builds off of the "pass-gltf-copyright" branch because I am testing with credits in the gltf copyright.

### DIFF
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/CreditSystem.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/CreditSystem.h
@@ -66,7 +66,7 @@ public:
   /**
    * @brief Get the credits to show this frame.
    */
-  const std::vector<Credit>& getCreditsToShowThisFrame();
+  std::vector<Credit> getCreditsToShowThisFrame();
 
   /**
    * @brief Get the credits that were shown last frame but should no longer be
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/CreditSystem.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/CreditSystem.h
@@ -83,6 +83,7 @@ private:
   struct HtmlAndLastFrameNumber {
     std::string html;
     int32_t lastFrameNumber;
+    int count = 0;
   };
 
   std::vector<HtmlAndLastFrameNumber> _credits;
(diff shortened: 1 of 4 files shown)


Cesium3DTilesSelection/src/CreditSystem.cpp
```diff
@@ -54,6 +56,22 @@ void CreditSystem::addCreditToFrame(Credit credit) {
...
   _currentFrameNumber++;
 }
+
+const std::vector<Credit>& CreditSystem::getCreditsToShowThisFrame() noexcept {
+  // sort credits based on the number of occurrences
+  if (_creditsToShowThisFrame.size() < 2) {
+    return _creditsToShowThisFrame;
+  }
+  const auto& counts = _creditCounts;
+  std::sort(
```
kring, line 69: I think it's really important to use a stable sort here. Otherwise, credits with equal counts can jump around from frame to frame.

Unfortunately, reading the docs just now, I learned that std::stable_sort will allocate a scratch buffer (presumably on the heap) that is the same size as the container we're sorting. That's really not great for an operation like this that we do every frame.

So I think a good solution is to stick with std::sort and incorporate stability into the comparison function. Something like this (not tested):

```
[&counts](const Credit& a, const Credit& b) {
  int32_t aCounts = counts.at(a.id);
  int32_t bCounts = counts.at(b.id);
  if (aCounts == bCounts)
    return a.id < b.id;
  else
    return aCounts > bCounts;
}
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/CreditSystem.h
```diff
@@ -91,5 +90,6 @@ class CESIUM3DTILESSELECTION_API CreditSystem final {
   int32_t _currentFrameNumber = 0;
   std::vector<Credit> _creditsToShowThisFrame;
   std::vector<Credit> _creditsToNoLongerShowThisFrame;
+  std::unordered_map<size_t, int32_t> _creditCounts;
```
kring, line 93: Can you store the count in `HtmlAndLastFrameNumber`? I think it would be quite a bit more efficient that way. Credits are worth a bit of optimization, because they're used extensively every frame. Cozzi was just making fun of me today for my initial implementation of credits in CesiumJS showing up as a hotspot in the profiler. 😆

kring (commented):
Looks good @joseph-kaile, just a couple of suggestions below.

kring:
Thanks @joseph-kaile!

PR #456, opened 2022-03-07
Add Promise::resolve(const T&)

baothientran:
Fix the issue where CesiumAsync failed to compile when passing lvalue reference to Promise::resolve()

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### v0.14.0 - YYYY-MM-DD
+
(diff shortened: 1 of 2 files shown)


xuelongmu:
Thanks @baothientran ! I was able to compile my methods and they are working correctly at run-time.

baothientran:
I'm going to self-merge this as it's a straightforward change

baothientran:
I'm going to self-merge this as it's a straightforward change

PR #457, opened 2022-03-08
Show onscreen credits

joseph-kaile:
Part of https://github.com/CesiumGS/cesium-unreal/issues/765


kring:
I switched the base branch to make it easier to see what's new in this one. And it all looks good. I think the only thing missing compared to CesiumJS is the logic that turns the `collapsible==false` on a Cesium ion asset into `showOnScreen==true` on the credit.

kring:
👍

PR #459, opened 2022-03-15
Expose the swizzle string in feature texture property views

nithinp7:
Small PR to support: https://github.com/CesiumGS/cesium-unreal/pull/787

The channel offset representation can be useful for dynamically un-swizzling pixels from feature texture property, but the swizzle string itself can be useful to when generating shaders.


kring:
Thanks @nithinp7!

PR #461, opened 2022-03-17
Fix forbid holes bug

joseph-kaile:
This PR fixes the bug where the Cesium CWT would not load if the Forbid Holes option was enabled.


kring:
Thanks @joseph-kaile!

PR #462, opened 2022-03-18
Add concierge config to check first time contributors.

baothientran:
This is copied from https://github.com/CesiumGS/cesium-unreal/pull/791


kring:
Thanks @baothientran!

PR #463, opened 2022-03-18
Various improvements related to EXT_feature_metadata

nithinp7:
More small changes to support: https://github.com/CesiumGS/cesium-unreal/pull/698


Cesium3DTilesSelection/src/upsampleGltfForRasterOverlays.cpp
```diff
@@ -1161,4 +1171,136 @@ static bool upsamplePrimitiveForRasterOverlays(
...
+
+  for (BufferRangeToCopy& existingRange : bufferRanges) {
+    if (existingRange.sourceBuffer == parentBufferView.buffer) {
+      if (parentBufferView.byteOffset < existingRange.byteOffset) {
+        existingRange.byteOffset = parentBufferView.byteOffset;
+      }
+
+      if (parentBufferView.byteLength > existingRange.byteLength) {
+        existingRange.byteLength = parentBufferView.byteLength;
+      }
```
kring, line 1220: I think this logic is not quite right. Let's say the existingRange has a byteOffset of 10 and a byteLength of 10. Then the parentBufferView has a byteOffset of 0 and also a byteLength of 10. You'd want to copy bytes 0-19, but instead would only copy 0-9.

I think it might be easier and not really any less efficient to simply copy the bufferViews, and allocate a new buffer for each, rather than try to keep the buffers intact. In pathological cases, multiple bufferViews could refer to the same bytes in a buffer, but that would be pathological and we don't need to put too much effort into making it as efficient as it can be.

nithinp7: Ah yeah that part isn't correct. But that's a good point in general, maybe we can simplify it and just copy a different buffer for each metadata property buffer view (the vertex attributes will be reconstructed in the later step). The only case I was considering was interleaved property arrays. Each property array within a feature table will have the same count, so I figured it would be possible to interleave them. But now that appears to me to be a pathological case and maybe even explicitly disallowed, since there's no accessor.

kring:
Thanks @nithinp7!

PR #464, opened 2022-03-21
Finish onscreen credits

joseph-kaile:
Adds credits to tiles, and causes tiles to recieve credits from tilesets.
Also resized bing image (matches Cesium JS version)


Cesium3DTilesSelection/src/Tile.cpp
```diff
@@ -542,9 +557,19 @@ void Tile::processLoadedContent() {
...
-            *this->_pContent->model->asset.copyright,
-            this->getTileset()->getOptions().showCreditsOnScreen);
+        if (!perTileCredits.has_value()) {
+          perTileCredits = std::make_optional<std::vector<Credit>>();
+        }
+        const auto& copyright = *this->_pContent->model->asset.copyright;
+        size_t start = 0;
+        size_t end;
+        do {
+          end = copyright.find(',', start);
```
kring, line 567: This should change from `,` to `;`. See https://github.com/CesiumGS/cesium/pull/10147

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileContentLoadResult.h
```diff
@@ -77,7 +77,7 @@ struct TileContentLoadResult {
   /**
    * @brief Contains the copyright information of the glTF model.
    */
-  std::optional<Cesium3DTilesSelection::Credit> credit{};
+  std::optional<std::vector<Cesium3DTilesSelection::Credit>> credits{};
```
kring, line 80: Is the `std::optional` necessary? I think an empty vector would be just as good.

kring:
Thanks @joseph-kaile!

PR #467, opened 2022-03-24
Fix how Tile::computeByteSize handles external images

nithinp7:
Fixes a largely inconsequential bug that I ran across the other day.


kring:
Thanks @nithinp7!

PR #470, opened 2022-03-25
add WMS raster overlay

tomliangv2i:
a custom RasterOverlay that supports WMS.

-- I have signed the CLA.

### DIFF
--- a/Cesium3DTilesSelection/src/WebMapServiceRasterOverlay.cpp
+++ b/Cesium3DTilesSelection/src/WebMapServiceRasterOverlay.cpp
@@ -1,0 +1,410 @@
+#include "Cesium3DTilesSelection/WebMapServiceRasterOverlay.h"
+
+#include "Cesium3DTilesSelection/CreditSystem.h"
+#include "Cesium3DTilesSelection/QuadtreeRasterOverlayTileProvider.h"
+#include "Cesium3DTilesSelection/RasterOverlayLoadFailureDetails.h"
+#include "Cesium3DTilesSelection/RasterOverlayTile.h"
+#include "Cesium3DTilesSelection/TilesetExternals.h"
+#include "Cesium3DTilesSelection/spdlog-cesium.h"
+#include "Cesium3DTilesSelection/tinyxml-cesium.h"
+
+#include <CesiumAsync/IAssetAccessor.h>
+#include <CesiumAsync/IAssetResponse.h>
+#include <CesiumGeospatial/GlobeRectangle.h>
+#include <CesiumGeospatial/Projection.h>
+#include <CesiumGeospatial/WebMercatorProjection.h>
+#include <CesiumUtility/Uri.h>
+
+#include <cstddef>
+#include <sstream>
+
+using namespace CesiumAsync;
+using namespace CesiumGeometry;
+using namespace CesiumGeospatial;
+using namespace CesiumUtility;
+
+namespace Cesium3DTilesSelection {
+
+class WebMapServiceTileProvider final
+    : public QuadtreeRasterOverlayTileProvider {
+public:
+  WebMapServiceTileProvider(
+      RasterOverlay& owner,
+      const CesiumAsync::AsyncSystem& asyncSystem,
+      const std::shared_ptr<IAssetAccessor>& pAssetAccessor,
+      std::optional<Credit> credit,
+      const std::shared_ptr<IPrepareRendererResources>&
+          pPrepareRendererResources,
+      const std::shared_ptr<spdlog::logger>& pLogger,
+      const CesiumGeospatial::Projection& projection,
+      const CesiumGeometry::QuadtreeTilingScheme& tilingScheme,
+      const CesiumGeometry::Rectangle& coverageRectangle,
+      const std::string& url,
+      const std::vector<IAssetAccessor::THeader>& headers,
+			const std::string& version,
+      const std::string& layername,
+      const std::string& fileExtension,
+      uint32_t width,
+      uint32_t height,
+      uint32_t minimumLevel,
+      uint32_t maximumLevel)
+      : QuadtreeRasterOverlayTileProvider(
+            owner,
+            asyncSystem,
+            pAssetAccessor,
+            credit,
+            pPrepareRendererResources,
+            pLogger,
+            projection,
+            tilingScheme,
+            coverageRectangle,
+            minimumLevel,
+            maximumLevel,
+            width,
+            height),
+        _url(url),
+        _headers(headers),
+        _version(version),
+        _layername(layername),
+        _fileExtension(fileExtension) {}
+
+  virtual ~WebMapServiceTileProvider() {}
+
+protected:
+  virtual CesiumAsync::Future<LoadedRasterOverlayImage> loadQuadtreeTileImage(
+      const CesiumGeometry::QuadtreeTileID& tileID) const override {
+ 
+    LoadTileImageFromUrlOptions options;
+    options.rectangle = this->getTilingScheme().tileToRectangle(tileID);
+    options.moreDetailAvailable = tileID.level < this->getMaximumLevel();
+
+    //SPDLOG_LOGGER_INFO(getLogger(), "tileID: level>" + std::to_string(tileID.level) + "; x>" + std::to_string(tileID.x) + "; y>" + std::to_string(tileID.y));
+
+    const CesiumGeospatial::GlobeRectangle tileRectangle =
+        CesiumGeospatial::unprojectRectangleSimple(this->getProjection(), options.rectangle);
+
+    const std::string urlTemplate =
+        "{baseUrl}?request=GetMap&TRANSPARENT=TRUE&version={version}&service=wms&"
+        "format=image.png&styles="
+        "&width={width}&height={height}&bbox={minx},{miny},{maxx},{maxy}"
+        "&layers={layername}&crs=EPSG:4326";
+
+    const auto radiansToDegrees = [](double rad) { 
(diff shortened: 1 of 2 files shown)


kring:
Thanks for the pull request, @tomliangv2i! Based on a brief look, this looks very good. Have you implemented a WMS overlay component on the cesium-unreal side as well? If so, I'd love to see a pull request there too. It will make this one a bit easier to test.

tomliangv2i:
> Thanks for the pull request, @tomliangv2i! Based on a brief look, this looks very good. Have you implemented a WMS overlay component on the cesium-unreal side as well? If so, I'd love to see a pull request there too. It will make this one a bit easier to test.

Thanks Kevin,  just made a PR 
[tomliangv2i:features/WMS](https://github.com/CesiumGS/cesium-unreal/pull/799)

Cesium3DTilesSelection/src/WebMapServiceRasterOverlay.cpp
```diff
@@ -0,0 +1,410 @@
...
+    options.moreDetailAvailable = tileID.level < this->getMaximumLevel();
+
+    //SPDLOG_LOGGER_INFO(getLogger(), "tileID: level>" + std::to_string(tileID.level) + "; x>" + std::to_string(tileID.x) + "; y>" + std::to_string(tileID.y));
+
+    const CesiumGeospatial::GlobeRectangle tileRectangle =
+        CesiumGeospatial::unprojectRectangleSimple(this->getProjection(), options.rectangle);
+
+    const std::string urlTemplate =
+        "{baseUrl}?request=GetMap&TRANSPARENT=TRUE&version={version}&service=wms&"
+        "format=image.png&styles="
```
kring, line 88: `format` should be a MIME type, so `image/png` not `image.png`.

Cesium3DTilesSelection/src/WebMapServiceRasterOverlay.cpp
```diff
@@ -0,0 +1,410 @@
...
+  tinyxml2::XMLElement* pServiceName =
+      pService->FirstChildElement("Name");
+  if (!pServiceName) {
+    error = "Invalid web map service XML document (Service > Name is "
+            "missing) ";
+    return false;
+  }
+
+  const std::string serviceNameText = pServiceName->GetText();
+  if (serviceNameText != "WMS") {
```
kring, line 160: There is no requirement that the service be named `WMS`. Servers can return whatever they like here.

Cesium3DTilesSelection/src/WebMapServiceRasterOverlay.cpp
```diff
@@ -0,0 +1,410 @@
...
+      {"miny", radiansToDegrees(tileRectangle.getWest())},
+      {"layername", this->_layername},
+      {"width", std::to_string(this->getWidth())},
+      {"height", std::to_string(this->getHeight())}
+    };
+
+    std::string url = CesiumUtility::Uri::substituteTemplateParameters(
+        urlTemplate,
+        [
+          map=&urlTemplateMap
```
kring, line 111: It'd be a bit more idiomatic to write this as `&map = urlTemplateMap` and then `map` is a reference instead of a pointer.

Cesium3DTilesSelection/src/WebMapServiceRasterOverlay.cpp
```diff
@@ -0,0 +1,410 @@
...
+        ](const std::string& placeholder) {
+          return map->count(placeholder) == 0 ? "{" + placeholder + "}"
+                                              : map->at(placeholder);
```
kring, line 114: It's a little clearer and more performant (not that the performance improvement really matters here) to use iterators, e.g.:

```suggestion
          auto it = map.find(placeholder);
          return it == map.end() ? "{" + placeholder + "}" : it->second;
```

Cesium3DTilesSelection/src/WebMapServiceRasterOverlay.cpp
```diff
@@ -0,0 +1,410 @@
...
+  tinyxml2::XMLElement* pOptionalServiceMaxWidth =
+      pService->FirstChildElement("MaxWidth");
+  if (pOptionalServiceMaxWidth) {
+    const std::string maxWidthText = pOptionalServiceMaxWidth->GetText();
+    try {
+      const int maxWidth = std::stoi(maxWidthText);
+      const int optionalTileWidth = options.tileWidth.value_or(256);
+      if (optionalTileWidth > maxWidth) {
+        char buffer[512];
+        std::sprintf(
```
kring, line 174: Use `fmt::format` instead of `std::sprintf`. There's an example in `BingMapsRasterOverlay.cpp`.

Cesium3DTilesSelection/src/WebMapServiceRasterOverlay.cpp
```diff
@@ -0,0 +1,410 @@
...
+    return false;
+  }
+  tinyxml2::XMLElement* pLayer = pCapability->FirstChildElement("Layer");
+  if (!pLayer) { 
+    error = "Invalid web map service XML document";
+    return false;
+  } 
+
+  std::vector<std::string> validLayerNames;
+  for (tinyxml2::XMLElement* pQueryableLayer = pLayer->FirstChildElement("Layer");
```
kring, line 258: WMS allows layers to be arbitrarily nested inside other layers, and any of these can be named layers that can be requested individually. So we can't assume all of the layers of interest are children of the root Layer. They need to be discovered recursively.

But really, it's probably ok to just pass through whatever layers the user requested straight through to the server. Let the server validate them and reject them if necessary, rather than trying to validate them manually against the GetCapabilities ourselves.

Cesium3DTilesSelection/src/WebMapServiceRasterOverlay.cpp
```diff
@@ -0,0 +1,410 @@
...
+                coverageRectangle, 
+                url,
+                headers,
+                options.version.value_or("1.3.0"),
+                options.layers,
+                options.fileExtension.has_value() ? "." + options.fileExtension.value() : "",
+                options.tileWidth.value_or(256),
+                options.tileHeight.value_or(256),
+                options.minimumLevel.value_or(1),
+                options.maximumLevel.value_or(14)
```
kring, line 405: I think it would be better if these default values were set in `WebMapServiceRasterOverlayOptions`, rather than being optional there and quietly given values here. It would make the defaults more obvious.

Cesium3DTilesSelection/src/WebMapServiceRasterOverlay.cpp
```diff
@@ -0,0 +1,410 @@
...
+                tilingScheme,
+                coverageRectangle, 
+                url,
+                headers,
+                options.version.value_or("1.3.0"),
+                options.layers,
+                options.fileExtension.has_value() ? "." + options.fileExtension.value() : "",
+                options.tileWidth.value_or(256),
+                options.tileHeight.value_or(256),
+                options.minimumLevel.value_or(1),
```
kring, line 404: Shouldn't the minimum level be zero?

kring (commented):
Thanks @tomliangv2i! This is looking good, but I have a few suggestions below.

I used two different public WMS servers to test, one Geoserver and the other ArcGIS Server.

### Geoserver

URL: `https://nationalmap.gov.au/proxy/http://geoserver.nationalmap.nicta.com.au/geotopo_250k/ows`
Layers: `Hydrography:bores`
[CesiumJS Sandcastle demo](https://sandcastle.cesium.com/?src=Web%20Map%20Service%20(WMS).html)

### ArcGIS Server

URL: `https://nationalmap.gov.au/proxy/https://www.environment.gov.au/mapping/services/abares/CLUM_50m/MapServer/WMSServer`
Layers: `Catchment Scale Land Use 2020 - Primary classification`
[CesiumJS Sandcastle demo](https://sandcastle.cesium.com/#c=bVLRbuIwEPyVVZ5AAhtaUFtK0VXcy0mgqw71+hLptCSbYNWxI9uB4yr+vZuEVG2vT/aux7PjGSfW+AB7RQdycAeGDrAkr6pC/G56vThKmnppTUBlyMVR/zY2sZES7tMUEJ7WG1AF5uSOoPHICJM0rOfmqu55Jm+niA9tpvpQC0zTH23jwdm9SmsJBt4Le6LtGssNub1K6DP2pQYDVE7P2h1AHO1CKP1MSoNBWYO6wFLkdi+wkqWzf4+yAxwOB0Fmr5w1BZnQgRhfKpNL3870ErfoeFmuHtd/pqNCnvWQk2xGu4ujQSugscTPWMYSQ7KreWGToCZYoUnh0RNcjC5GMIQHx1awiYlG71WmkkbuG1GJDgsKDdlL97jg0Hg+YdYZFxUNupPMugJDPbcxWJYmf6M6NeupH5u3LDcBXQCbZaCtfebXAga4rzwP0ApFbM7pJazBofAU6g/S+p2SD6o1d9al9IuSgCbXJDJni++UOyLfa8ePxxMxvRmfxQwnU3F9eXWuxpNrcdMVw6m4uqy3fW6c+rcQDaK5D0dNi+6Z31RRWlbOifeEkIGKUmPgcLZV8kxBJN7Xb6yhc/n+6jxVe1Dp3RcfvE2AT7JK6436R3G0mEvG/3dVW0zZq58cOOdcw3bjxaptCiHmksuvbwZrNf+iT8yv).

kring:
Thanks @tomliangv2i! We really appreciate the pull request! I made a few tweaks to this on the way to main. Let me know if you have any concerns.

PR #471, opened 2022-03-26
Raster overlay filtering option

joseph-kaile:
Native side for the raster overlay renderer options PR (https://github.com/CesiumGS/cesium-unreal/pull/784)

### DIFF
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/RasterOverlay.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/RasterOverlay.h
@@ -4,6 +4,7 @@
 
 #include <CesiumAsync/IAssetAccessor.h>
 #include <CesiumGltf/Ktx2TranscodeTargets.h>
+#include <CesiumGltf/Sampler.h>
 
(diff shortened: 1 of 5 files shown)


Cesium3DTilesSelection/include/Cesium3DTilesSelection/RasterOverlay.h
```diff
@@ -86,6 +86,15 @@ struct CESIUM3DTILESSELECTION_API RasterOverlayOptions {
...
   bool showCreditsOnScreen = false;
+
+  /**
+   * @brief Arbitrary data that will be passed to {@link prepareRasterInLoadThread},
+   * for example, data to control the per-raster overlay client-specific texture
+   * properties.
+   *
+   * This pointer can and will be accessed by multiple threads simultaneously.
+   */
+  void* pRendererOptions = nullptr;
```
kring, line 97: Sorry I didn't think to mention this earlier, but how about using `std::any` instead of `void*`? I think that would make it easier to control the object's lifetime, as well as being generally safer.

kring:
Thanks @joseph-kaile!

PR #475, opened 2022-03-31
Use int64_t instead of uint32_t for Cesium ion asset IDs.

kring:
Because we should mostly avoid unsigned types, and because 32-bits for Cesium ion asset IDs might someday be insufficient.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,8 @@
 
 ##### Breaking Changes :mega:
 
-- Added a parameter to `prepareRasterInLoadThread`
+- Added a parameter to `prepareRasterInLoadThread`.
+- Changed the type of Cesium ion asset IDs from `uint32_t` to `int64_t`.
 
 ##### Fixes :wrench:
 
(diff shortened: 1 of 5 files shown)


kring:
I don't have strong feelings about it, but basically the reasons are:

1. As you said, it's "common wisdom" in C++ to avoid unsigned types. For a library, that matters a bit.
2. Our coding guidelines say to avoid them: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-nonnegative
3. As a library (again), "just turn on this warning to avoid misuse" isn't a great answer, particularly when we've made a non-standard choice.

As an illustration of (3), I opened this now because I temporarily enabled conversion warnings in Cesium for Unreal, and noticed that we were already using an int64_t for asset IDs in cesium-unreal, but were converting them to a uint32_t to send to cesium-native. Why don't we always flag these conversion warnings? Well, mostly because enabling these warnings causes a huge number of warnings in Unreal Engine code that we don't control. Which is (unfortunately) a common situation for a library to find itself in.

PR #477, opened 2022-03-31
Use intrusive linked list to remove visible tile from last frame

baothientran:
I think we are currently using `TileSelectionState::Result::Refine` to determine if the current rendered tile has any children that are rendered in the last frame. If there are, we traverse down the tile to find those children and grand children that were rendered in  the last frame to add them to the `ViewUpdateResult::tilesToNoLongerRenderInThisFrame`

This PR tries to experiment with intrusive linked list instead. Basically, after a frame is finished traversing, all the current rendered tiles will be added to this list. In the next frame, when we add a new tile to the render list, we remove this tile from the linked list that represents the list of visible tile in the last frame. After the current frame finish traversing, the linked list will contain only the list of tiles that need to be turned off. This way, we don't need to maintain the state machine of Refine

Open this as draft since I'm still unsure about the state machine for traversing

### DIFF
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tile.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tile.h
@@ -649,6 +649,7 @@ private:
   std::vector<RasterMappedTo3DTile> _rasterTiles;
 
   CesiumUtility::DoublyLinkedListPointers<Tile> _loadedTilesLinks;
+  CesiumUtility::DoublyLinkedListPointers<Tile> _visibleTilesLinks;
 
 public:
   /**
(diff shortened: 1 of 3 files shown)


kring:
I like that this is less code, and it's clearer in some ways. But I'm less thrilled about it adding 8 bytes to the `Tile`. I suspect it might not be a performance win over all even though it looks like less work is being done.

baothientran:
Yeah I don't think it's a performance improvement overall. I have some local code that tries to kick descendants out but sometimes forget to set them to refine state, which makes the parent and children to overlap. So hopefully this is a little easier since I don't forget to set the state of children correctly

cesium-concierge:
Thanks again for your contribution @baothientran! 

PR #485, opened 2022-04-13
Support for multiple layers of quantized-mesh terrain

### LINKED ISSUE #253: Add support for multiple layers of quantized mesh
CesiumJS has the support for multiple layers in the layer.json for Quantized mesh. It will allow client to use their own terrain data on top of our CWT.

### LINKED ISSUE #292: Support terrain tile availability in layer.json
This isn't needed for terrain coming from Cesium ion, because it now uses the "metadata" extension. However, old terrain tilers still put all the tile availability information in layer.json. cesium-native doesn't use this information, so it mistakenly thinks only the level zero tiles are available and refuses to refine.

### LINKED ISSUE #476: Immediate crash when attempting to load terrain from certain URL
Originally found here: https://community.cesium.com/t/cesium-terrain-for-unreal/17940/4

As noted in that thread, inputting either of the following into the URL slot of a Cesium3DTileset causes an instant crash with no logs.
`https://www.webatlas.no/wms-hoyde/tms/cesium_qmesh/14/16950/13876.terrain`
`https://www.webatlas.no/wms-hoyde/tms/cesium_qmesh/14/16950/13876/`

joseph-kaile:
This should fix #253 and also fix #292

In this picture, the top layer is in the foreground and the background layer (mountains) is using CWT.
image

Should fix #476 via commit e006e7a4a77e3c6772fe14c78df4


kring (commented):
Good progress, @joseph-kaile! A few things to address below.

Cesium3DTilesSelection/src/TilesetLoadTilesetDotJson.cpp
```diff
@@ -37,9 +37,15 @@ struct Tileset::LoadTilesetDotJson::Private {
       const std::shared_ptr<spdlog::logger>& pLogger,
       bool useWaterMask);
 
+  static void workerThreadLoadTileContext(
```
kring, line 40: This function is no longer purely synchronous, because it sometimes need to load parent layer.jsons. So it should return a future, as should `workerThreadHandleResponse`.

kring: As currently written, we'll start loading tiles before we know the chain of underlying layers, which can cause buggy, non-deterministic loading behavior depending on the timing of responses.

Cesium3DTilesSelection/src/TilesetLoadTilesetDotJson.cpp
```diff
@@ -430,9 +429,124 @@ void Tileset::LoadTilesetDotJson::Private::workerThreadLoadTerrainTile(
...
+  if (!parentUrl.empty()) {
+    std::string resolvedUrl = Uri::resolve(context.baseUrl, parentUrl);
+    // append forward slash if necessary
+    if (resolvedUrl[resolvedUrl.size() - 1] != '/') {
+      resolvedUrl += '/';
+    }
+    resolvedUrl += "layer.json";
+    auto pAsyncSystem = context.pTileset->getAsyncSystem();
+    auto pAssetAccessor = context.pTileset->getExternals().pAssetAccessor;
+    pAssetAccessor->get(pAsyncSystem, resolvedUrl, context.requestHeaders)
```
kring, line 473: This future should be returned here, too, so that the outer future doesn't resolve into the full chain of parent URL loading futures resolves.

kring: In the case where the parentUrl is empty, just return `asyncSystem.createResolvedFuture()`.

kring (commented):
The way the asynchronicity is handled here isn't quite right, and will lead to problems in certain cases. This stuff is a pain, let me know if anything is unclear and if it would help to have a little more background about how Futures and whatnot work and should be used.

Cesium3DTilesSelection/src/Tile.cpp
```diff
@@ -901,7 +904,8 @@ void Tile::update(
     // If this tile still has no children after it's done loading, but it does
     // have raster tiles that are not the most detailed available, create fake
     // children to hang more detailed rasters on by subdividing this tile.
-    if (moreRasterDetailAvailable && this->_children.empty()) {
+    if (moreRasterDetailAvailable && this->_children.empty() &&
+        this->_pContext->availabilityLoadsInProgress == 0) {
```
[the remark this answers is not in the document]
joseph-kaile: Thanks, this was one part which I was iffy about. I implemented the suggestion.

Cesium3DTilesSelection/src/TilesetLoadTilesetDotJson.cpp
```diff
@@ -430,24 +444,156 @@ void Tileset::LoadTilesetDotJson::Private::workerThreadLoadTerrainTile(
...
+             layerJson,
+             context,
+             asyncSystem,
+             pLogger,
+             useWaterMask)
+      .thenImmediately([&tile, &context]() {
+        tile.setContext(&context);
+        tile.setBoundingVolume(
+            context.implicitContext->implicitRootBoundingVolume);
+        tile.setGeometricError(999999999.0);
```
baothientran, line 573: this isn't thread-safe, loadImmediately() will certainly run on worker thread, but tile must be modified at main-thread

The lambda also captures the context by reference, that means there is a good chance that two more threads are accessing the same context at the same time, which is not thread safe as well

Cesium3DTilesSelection/src/TilesetLoadTilesetDotJson.cpp
```diff
@@ -261,16 +278,17 @@ LoadResult Tileset::LoadTilesetDotJson::Private::workerThreadHandleResponse(
         *pRootTile,
         tileset,
         *pContext,
+        asyncSystem,
```
baothientran, line 281: workerThreadLoadTerrainTile() now does spawn a new task, so it can't use the Tile, Tileset, and TileContext by reference as there are more than two threads access them both at the same time

baothientran (commented):
Thanks @joseph-kaile! Most looks good to me. The only thing that concerns me is TilesetLoadTilesetDotJson isn't thread-safe at the moment as the worker threads are using Tile and TileContext by reference which are thread-safe themselves

Cesium3DTilesSelection/src/TilesetLoadTilesetDotJson.cpp
joseph-kaile: > Thanks @joseph-kaile! Most looks good to me. The only thing that concerns me is TilesetLoadTilesetDotJson isn't thread-safe at the moment as the worker threads are using Tile and TileContext by reference which are thread-safe themselves

Thanks for the suggestions.

I do think that the future code is thread safe, I don't think tile would be modified by any other thread because the root tile won't be set until these futures complete.

I dont think any of the tasks that are spawned actually run concurrently:

workerThreadLoadTerrainTile runs workerThreadLoadTileContext, and then waits for it to finish before continuing.

workerThreadLoadTileContext potentially spawns off workerThreadLoadTileContext when it has finished.

And there's nothing else out there that would be running at the same time.

baothientran: I may miss something, but workerThreadLoadTerrainTile is called from another worker thread at line 277. The original thread doesn't wait for workerThreadLoadTerrainTile to finish, but resolved right away, which mean the root tile is already ended up in the main thread while workerThreadLoadTileContext is running

kring: @baothientran you're right, I may have broken this when I made my changes in that other branch. Let me take a look.

joseph-kaile: I see the problem now, I thought that I returned the future. Thanks.

PR #490, opened 2022-04-19
Fix sqlite crash

joseph-kaile:
In my case, when reading from `requestHeader`, the JSON string was empty, which caused the json document to have a parse error, and then the program would crash.

Now, it does not crash, it will query from the network instead, and attempt to store the value again. In my case, the value cannot be stored because the sqlite database became corrupt.

If SQL_STATUS is ever SQL_CORRUPT after a SQL_Step command, then it will attempt to delete the file, and then create the database again.

I was unable to call `destroyDatabase` during the `getEntry` because it is marked `const`.


kring:
Thanks @joseph-kaile!

PR #491, opened 2022-04-20
Update Tile from Content before calling prepareInMainThread

kring:
Previously, we called `IPrepareRendererResources::prepareInMainThread` _before_ we updated the `Tile` with the loaded content. Mostly this doesn't matter, but if `prepareInMainThread` used the tile's bounding volume, for example, and the loaded content included an updated bounding volume, this would result in `prepareInMainThread` using a stale bounding volume. Better to do the update first.


joseph-kaile (approved):
Makes sense

PR #492, opened 2022-04-20
Interfaces for hardware occlusion culling

nithinp7:
Renderer integrations with Cesium Native often may be doing sophisticated things to save draw calls on completely occluded geometry. Hardware occlusion queries are a common technique where a mesh's bounding volume is drawn into the previous frame's depth buffer, to check if the entire mesh was occluded (zero pixels drawn). Usually this information is used to save a much more expensive draw call on the mesh when it is found to be occluded. In our case, we would like to know if certain tile bounding volumes are entirely occluded so we can cull them and all their descendants, thereby saving not only the draw calls but also the load times and memory cost. This would especially be useful when examining a tileset from up close, where 100s of high-res tiles may be occluded by buildings, etc.

This PR adds interfaces that allow Cesium Native to manage a pool of renderer-side bounding volume objects that act as proxies for traversed Cesium Native tiles. The renderer integration should implement these proxies as invisible bounding volumes (OBBs, spheres, etc) that are to be tested for occlusion with hardware occlusion queries. These objects should not affect the rendering or anything else in any way, e.g. they should not show up on the screen, in the depth buffer, or as a physics objects. Further, when using this system the renderer may choose to explicitly disable hardware occlusion culling for the actually rendered tiles, since the feedback of the proxies in the traversal should not be rendering occluded tiles _anyways_.

This will be followed up soon with an implementation of this system in Cesium for Unreal...

Other considerations:
- For tilesets with many more visited tiles than bounding volume proxies, it will not be possible to assign a proxy to each visited tile. In that case, the current tile to proxy mapping will be inefficient since it will be in a depth-first order (as a result of the traversal order). It would be better to map the proxies breadth-first, so tiles further up the tree have an equal chance of culling all their descendants.


nithinp7:
@kring I think this is ready for a first-pass review. I still need to fix CI and work through some of the TODOs above before merging but taking this out of draft for review.

nithinp7:
Sorry for the somewhat open-ended rambling, but wanted to share a traversal-related issue I got stuck on and confirm my understanding. What are the cases where a tile gets to this point in this code in the traversal, but doesn't fulfill the `if(meetsSSE && !ancestorMeetsSSE)` criteria? 
https://github.com/CesiumGS/cesium-native/blob/52b181f94d3b330e3c8dfc123fd545d8f66d2123/Cesium3DTilesSelection/src/Tileset.cpp#L1156-L1163

If I could guess at a partial answer, there are cases where previously the parent (or ancestor in general) did not meet SSE, but after zooming out the ancestor now does. In this case, if we don't have that ancestor tile, we have to keep rendering the previous descendants (as an aside, sometimes we may have holes for the missing descendants right?). There's no use loading any more of the descendants, even though they are visited and meet the SSE, since we'd rather just wait for the missing ancestor and unrefine rather than load descendants to plug up holes. Is that correct so far?

Is there ever a case where we get to this point in the code where `ancestorMeetsSSE` is false, but `meetsSSE` is also false? I think this happens in the forbid holes case, where a non-leaf tile that doesn't meet SSE chooses not to refine since it is waiting for children to load. Since `meetsSSE` is false, I had thought wow this tile doesn't get added to the load queue in `_visitTile`? But the fact is that its parent must have loaded this tile before refining in the first place, so it is guaranteed to be loaded already! So in fact, `_visitTile` has different preconditions depending on if `forbidHoles` is enabled! Almost positive it is not currently worth refactoring `_visitTile` into functions with consistent preconditions / postconditions, but just wanted to describe something that tripped me up!

In my case a tile refused to refine because some of its children were occluded but didn't meet the SSE. When a tile is occluded we should just render it and not refine, so after the occlusion check of the child the `wantsToRefine` flag is set to false. This causes the child to enter this code path, decide _not_ to add itself to the loadQueue, causing the parent to perpetually wait to refine due to unloaded children. Phew! Glad the bug is caught now though! 

PS: @kring The bug was resolved, but if you could confirm my understanding when you get a chance that would be great!

kring:
>  There's no use loading any more of the descendants, even though they are visited and meet the SSE, since we'd rather just wait for the missing ancestor and unrefine rather than load descendants to plug up holes. Is that correct so far?

Yep, exactly!

> So in fact, _visitTile has different preconditions depending on if forbidHoles is enabled!

Hmm yeah I think I'd put it this way. `_visitTile` does not require that the tile it is given is already loaded, even in the forbid holes case. It's not a precondition. But when we _are_ in the forbid holes case, that just happens to be true.

Now, if that didn't "just happen to be true", then we'd have a bug, caused by the extra check for `meetsSse`. We should just remove that. `addTileToLoadQueue` already checks if the tile actually needs to be loaded, so we won't end up doing any extra loading (or even add anything extra to the load queue) in the forbid holes case where the tile is guaranteed to already be loaded.

I think removing that extra check fixes the problem you ran into with the occlusion test, too, right?

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -107,6 +111,11 @@ Tileset::Tileset(
   if (ionAssetID > 0) {
     CESIUM_TRACE_USE_TRACK_SET(this->_loadingSlots);
     this->notifyTileStartLoading(nullptr);
+    if (this->_externals.pTileOcclusionProxyPool) {
+      // TODO: find a reasonable number
+      this->_externals.pTileOcclusionProxyPool->initPool(
+          options.occlusionPoolSize);
```
kring, line 117: Perhaps the renderer should have control over the pool size?

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -612,68 +633,105 @@ static bool isVisibleInFog(double distance, double fogDensity) noexcept {
...
+    // this tile is occluded by fog so it is a culled tile
+    cullResult.culled = true;
+    if (this->_options.enableFogCulling) {
+      // fog culling is enabled so we shouldn't visit this tile
+      cullResult.shouldVisit = false;
+    }
+  }
+}
+
+CesiumUtility::ScopeGuard<std::function<void()>>
```
kring, line 729: I think returning a ScopeGuard here is pretty unusual. And it requires use of std::function, which is less efficient (not that it makes a big difference, probably). I think if you want to refactor this code so that the lifetime can't be managed by a scope anymore, then it shouldn't use ScopeGuard. Maybe instead use a small class that allocates a vector of distances from the pool, computes distances, and returns the vector to the pool when it goes out of scope.

kring (commented):
A few small comments, but overall looks great!

Cesium3DTilesSelection/src/Tileset.cpp
nithinp7: The pool size here comes from the `TilesetOptions` which is configurable on the renderer, would it be better if it was determined somewhere else? In Cesium for Unreal, this is configurable in the `Cesium3DTileset` details panel.

kring: It seems like it might be cleaner if the renderer were solely responsible for creating, sizing, and destroying the pool. Get rid of `initPool` completely.

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -63,12 +64,14 @@ Tileset::Tileset(
       _overlays(*this),
       _tileDataBytes(0),
       _supportsRasterOverlays(false),
-      _gltfUpAxis(CesiumGeometry::Axis::Y),
-      _distancesStack(),
-      _nextDistancesVector(0) {
+      _gltfUpAxis(CesiumGeometry::Axis::Y) {
```
kring, line 67: Missing initialization for `_distances` and `_childOcclusionProxies`.

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -106,9 +104,7 @@ Tileset::Tileset(
...
```
kring, line 107: Here too.

kring (commented):
Two very trivial comments here. The changes to distance handling look good to me. Just a couple of things from the last review that still need to be addressed (even if by saying "they wont be addressed") and then this is ready.

Cesium3DTilesSelection/src/Tileset.cpp
nithinp7: That makes sense, but that would also shift the responsibility of managing the pool (allocating from and freeing to the pool) to the renderer right? Currently the renderer only has to provide an implementation for creating individual proxies, but it doesn't have to manage the mapping to specific tiles and manage a free list.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileOcclusionRendererProxy.h
```diff
@@ -0,0 +1,114 @@
...
+  virtual void reset(const Tile* pTile) = 0;
+
+private:
+  bool _usedLastFrame = false;
+  TileOcclusionRendererProxy* _pNext = nullptr;
+};
+
+class CESIUM3DTILESSELECTION_API TileOcclusionRendererProxyPool {
+public:
+  virtual ~TileOcclusionRendererProxyPool(){};
```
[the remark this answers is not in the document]
nithinp7: `destroyPool` is called in the Tileset destructor, which is needed since `initPool` is called in the Tileset constructors. It probably makes sense to call it here too out of principle, just applied the change.

nithinp7:
@kring The only unresolved review comment is regarding initPool / destroyPool. Moving that responsibility to the renderer side would mean cesium native will no longer manage the tile to occlusion proxy map as well as the free proxy linked list right? I figured it made it easier for the renderer to _not_ have to worry about the pooling implementation, what do you think?

kring:
Thanks @nithinp7!

PR #494, opened 2022-04-28
Add support for 3DTILES_batch_table_hierarchy

kring:
This makes the cesium-native metadata system work with Cesium OSM Buildings, and other tilesets that use this extension. However, it doesn't support:

* Properties stored in binary. Only JSON properties are supported.
* The use of the `parentCounts` property.

Mostly because these would be a hassle to implement and I'm not currently aware of any tilesets that use them. Warnings are logged when these features are encountered, though, so at least the user isn't left wondering.

Properties are encoded in the narrowest type that can hold all of the property values. But in the common case where a feature is missing a property, its value will be deemed to be `null`, which will force the type of the entire property to be `STRING` in order to accomodate the value `"null"`.


kring:
Merging this for today's release.

PR #495, opened 2022-04-29
Make quantized-mesh availability tile loading part of the tile content load process

kring:
This is a PR into #485.

In the process of writing https://github.com/CesiumGS/cesium-native/blob/main/doc/implicit-tiling.md, I realized there might be a pretty straightforward way to make sure all of the necessary availability tiles are loaded. This PR is an attempt to implement it, and I think it worked out ok. @joseph-kaile please take a look, and if you're happy you can merge both this PR and #485. Let me know if you have any concerns.

The basic idea here is that whenever we're loading a quadtree tile, consider whether that tile is also an "availability tile" in any layers. If it is, download that layer's version of the tile and add its availability data to the TileContext. And make these extra loads prerequisites for moving the tile into the `ContentLoaded` state. Because we never create implicitly-tiled children before the parent hits `ContentLoaded`, this guarantees that we won't try to create children - upsampled or otherwise - before we know whether or not they're available.

This eliminates the need to separately track availability tile loads and simplifies the code quite a bit.

### DIFF
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/ImplicitTraversal.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/ImplicitTraversal.h
@@ -90,8 +90,6 @@ struct ImplicitTraversalInfo {
 
 namespace ImplicitTraversalUtilities {
 
-CesiumAsync::Future<void> loadAvailability(Tile& tile);
-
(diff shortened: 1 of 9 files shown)


baothientran:
I add about 4 layers with different resolutions and can confirm that it works as expected

In the image above, the wrinkled surfaces represent different layers on top the CWT
Capture

PR #496, opened 2022-05-03
Don't kick external tileset children.

kring:
@nithinp7 and @argallegos noticed recently that the New York City Buildings tileset in level 6 of the Cesium for Unreal Samples would sometimes get stuck. Tiles would be missing, no matter how long you wait. Moving the camera arounda  bit would usually fix it.

image

The problem turned out to be the tile selection's "kick" mechanism. It's usually a good idea to skip as many levels of detail as possible, and just directly load the tiles we need for the current view. But sometimes that's a very large number of tiles, and the user would have to sit there a long time waiting for them all. Our selection algorithm detects how many descendants of a given tile are waiting to load, and if it's a lot (a user-defined number), it kicks them all out of the load queue and instead loads and renders the current tile. As a result, the user sees detail sooner, even if it's lower quality. But it means the descendants won't start loading until after we finish loading and rendering the current tile. 

This works well enough, except if the "current tile" is not renderable at all. This happens when the tile represents an external tileset or has otherwise been deemed "unconditionally refine", because such tiles will never be renderable. In this scenario, we effectively end up waiting for a tile to become renderable, but that tile will _never_ become renderable, no matter how long we wait. Unless we move the camera, because that causes some of the tiles to be culled, which changes where the kick mechanism gets activated.

This PR fixes the problem by never kicking the descenants of an external tileset or unconditionally-refined tile.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,15 @@
 # Change Log
 
+### ? - ?
+
+##### Breaking Changes :mega:
+
+##### Additions :tada:
+
+##### Fixes :wrench:
+
+- Fixed a bug that could cause tiles in external tilesets to fail to load.
+
 ### v0.15.0 - 2022-05-02
 
 ##### Additions :tada:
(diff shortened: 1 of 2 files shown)


nithinp7:
@kring This works perfectly for NYC buildings, thanks for the investigation! I feel like I have to relearn how the kick mechanism works every time I look at it hahah, but this change looks solid!

PR #498, opened 2022-05-06
Third party library update

kring:
This PR updates thirdparty.json to match the new conventions. It also updates the versions of third party libraries used in a handful of cases.

### DIFF
--- a/ThirdParty.json
+++ b/ThirdParty.json
@@ -6,20 +6,6 @@
     "license": ["MIT"]
   },
   {
-    "name": "Catch2",
-    "url": "https://github.com/catchorg/Catch2",
-    "version": "2.13.9",
-    "license": ["Boost Software License 1.0"],
-    "notes": "Only used in the tests."
-  },
-  {
-    "name": "cmake-modules",
-    "url": "https://github.com/bilke/cmake-modules",
-    "version": "d6d1a778e41cb114e5cdf279b8a659fa0ce9a0d4",
-    "license": ["Boost Software License 1.0", "BSD-3-Clause"],
-    "notes": "Only used in the build system."
-  },
-  {
     "name": "cpp-httplib",
     "url": "https://github.com/yhirose/cpp-httplib",
     "version": "0.10.3",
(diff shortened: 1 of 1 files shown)


kring:
CC @shehzan10

ThirdParty.json
```diff
@@ -2,96 +2,119 @@
   {
     "name": "Async++",
     "url": "https://github.com/Amanieu/asyncplusplus",
-    "license": "MIT"
+    "version": "756893feb9e69c098225d5a454a668a2c139e4be",
```
shehzan10, line 5: @kring To confirm - Are we using commits throughout because we don't use release versions of the libraries?

ThirdParty.json
```diff
@@ -2,96 +2,119 @@
...
+    "license": ["MIT"]
   },
   {
     "name": "Catch2",
     "url": "https://github.com/catchorg/Catch2",
-    "license": "Boost Software License 1.0"
+    "version": "2.13.9",
+    "license": ["Boost Software License 1.0"],
+    "notes": "Only used in the tests."
```
shehzan10, line 13: We can skip these if we are not shipping them. The ThirdParty.json is meant to document dependencies that are included in releases and shipped. If testing code is for CI / Development only and never redistributed by Cesium Native, ie Dev Dependency, we don't have to list it here.

ThirdParty.json
```diff
@@ -2,96 +2,119 @@
...
+    "version": "4.0.0",
+    "license": ["MIT"]
   },
   {
     "name": "KTX-Software",
     "url": "https://github.com/KhronosGroup/KTX-Software",
-    "license": "Apache 2.0"
+    "version": "94764c270d43d90e4e4cedee6248753296f874b3",
+    "license": ["Apache 2.0"],
+    "notes": "Using Cesium's fork of this library, which has minor differences from the upstream: https://github.com/CesiumGS/KTX-Software/tree/cesium-native"
```
shehzan10, line 63: Should the URL point to Cesium's fork in this case? Thats what we do for [libcitygml](https://github.com/CesiumGS/libcitygml) for example. Of course with LGPL, it is mandatory to point to any custom modifications, so thats the slight difference.

shehzan10 (commented):
@kring Thanks, most of it looks good. Just a few questions / comments.

ThirdParty.json
kring: Yeah, I used commits when there is no specific version number we're using. In some cases that's because the library doesn't even release versions.

ThirdParty.json
kring: I thought it wouldn't hurt for completeness (or for the benefit of someone later comparing the `extern` directory to ThirdParty.json and wondering why they're different), but I'll remove them.

ThirdParty.json
kring: I wasn't sure, but I've now made that change.

kring:
This is ready as far as I know @shehzan10.

kring:
This is ready as far as I know @shehzan10.

PR #500, opened 2022-05-09
Fix terrain holes when using terrain cutout with no other layers

joseph-kaile:
This pull request should fix the problem described here: https://github.com/CesiumGS/cesium-unreal/issues/840.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Fixed a bug that could cause tiles in external tilesets to fail to load.
+- Fixed a bug where upsampled quadtree tiles could have siblings with mismatching projections.
 
 ### v0.15.0 - 2022-05-02
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,11 +1,16 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
+- Fixed a bug where upsampled quadtree tiles could have siblings with mismatching projections.
+
 ### v0.15.1 - 2022-05-05
 
 ##### Fixes :wrench:
 
 - Fixed a bug that could cause tiles in external tilesets to fail to load.
-- Fixed a bug where upsampled quadtree tiles could have siblings with mismatching projections.
 
 ### v0.15.0 - 2022-05-02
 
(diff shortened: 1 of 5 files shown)


Cesium3DTilesSelection/src/upsampleGltfForRasterOverlays.cpp
```diff
@@ -420,10 +424,13 @@ static bool upsamplePrimitiveForRasterOverlays(
 
   for (std::pair<const std::string, int>& attribute : primitive.attributes) {
     if (attribute.first.find("_CESIUMOVERLAY_") == 0) {
-      if (uvAccessorIndex == -1) {
-        uvAccessorIndex = attribute.second;
+      if (attribute.first.find(
+              "_CESIUMOVERLAY_" + std::to_string(textureCoordinateIndex)) ==
+          0) {
+        if (uvAccessorIndex == -1) {
```
kring, line 430: A few suggestions here:
- Move the construction of the attribute name outside the loop since it's loop-invariant and kinda expensive to build.
- Can you just do `attribute.first == textureCoordinateName` rather than using find? I can't see any reason why not. Should be both clearer and faster.
- Check the cheap thing (`uvAccessorIndex == -1`) before checking the expensive thing (`attribute.first == textureCoordinateName`). Won't make a huge difference, but it's a good habit.

Cesium3DTilesSelection/src/ImplicitTraversal.cpp
```diff
@@ -448,6 +449,46 @@ void createImplicitChildrenIfNeeded(
...
+            if (!pParentRegion) {
+              const BoundingRegionWithLooseFittingHeights* pParentLooseRegion =
+                  std::get_if<BoundingRegionWithLooseFittingHeights>(
+                      &tile.getBoundingVolume());
+              if (pParentLooseRegion) {
+                pParentRegion = &pParentLooseRegion->getBoundingRegion();
+              }
+            }
+
+            GltfContent::createRasterOverlayTextureCoordinates(
```
kring, line 483: This is a fairly costly operation being done in the main thread, but it's going to be hard to fix. This is basically the soonest we know we need these extra texture coordinates, and it's already too late to do an async operation. If you're not seeing any obvious hitches from it, we can probably live with it for now.

CHANGES.md
```diff
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Fixed a bug that could cause tiles in external tilesets to fail to load.
+- Fixed a bug where upsampled quadtree tiles could have siblings with mismatching projections.
```
kring, line 8: This is in the wrong section because v0.15.1 has already been released.

kring (commented):
Thanks @joseph-kaile!

kring:
Thanks @joseph-kaile!

PR #503, opened 2022-05-13
Upsample using first overlay with more detail

### LINKED ISSUE #385: Upsampling uses the first set of overlay texture coordinates, no matter their projection
cesium-native will create upsampled tiles whenever the geometry resolution is exceeded and _any_ of the overlays report that they have more detailed data. The tiles will initially be subdivided using the implicit tiling context's tiling scheme, if there is one. If not, the tiles are divided in half according to the Geographic projection.

However, it will then actually subdivide the _geometry_ at u=0.5 and v=0.5 in the _first_ overlay texture coordinates that is finds. This could be a Web Mercator projection or a Geographic projection, or others if and when we add them in the future. The difference is usually small; as long as leaf tiles that need subdivision are relatively deep in the tree, the difference between these two projections is tiny.

Even so, cesium-native should be more consistent about this. If only one raster overlay has more detail, it should subdivide in that raster overlay's projection. If multiple have more detail, it should ideally use the average of their centers to subdivide, but if that's infeasible then picking the first and being consistent about it would still be a big improvement.

joseph-kaile:
Pull request in order to fix #385. 

This creates bounding volumes for the four upsampled tiles using the first raster overlay tile that has more detail.

It will also upsample using that same projection.

To make sure that it uses the same projection, it makes sure that the previous raster overlay tiles in the list do not have more data.

Update: During upsampling, its posssible that all the primitives are clipped out. In this case, it returns a std::nullopt.

### DIFF
--- a/Cesium3DTilesSelection/src/Tile.cpp
+++ b/Cesium3DTilesSelection/src/Tile.cpp
@@ -665,12 +665,9 @@ getTileBoundingRegionForUpsampling(const Tile& parent) {
     // The tile will be subdivided by (0.5, 0.5) in the first overlay's
     // texture coordinates which overlay had more detail.
 
-    for (size_t i = 0; i < parent.getMappedRasterTiles().size(); ++i) {
-      RasterMappedTo3DTile* pMapped =
-          const_cast<RasterMappedTo3DTile*>(&parent.getMappedRasterTiles()[i]);
-      if (pMapped->update(*const_cast<Tile*>(&parent)) ==
-          RasterOverlayTile::MoreDetailAvailable::Yes) {
-        const Projection& projection = pMapped->getReadyTile()
+    for (const RasterMappedTo3DTile& mapped : parent.getMappedRasterTiles()) {
+      if (mapped.isMoreDetailAvailable()) {
+        const Projection& projection = mapped.getReadyTile()
                                            ->getOverlay()
                                            .getTileProvider()
                                            ->getProjection();
@@ -1044,9 +1041,8 @@ void Tile::upsampleParent(
       }
     }
   } else {
-    for (RasterMappedTo3DTile& mapped : pParent->getMappedRasterTiles()) {
-      if (mapped.update(*pParent) ==
-          RasterOverlayTile::MoreDetailAvailable::Yes) {
+    for (const RasterMappedTo3DTile& mapped : pParent->getMappedRasterTiles()) {
+      if (mapped.isMoreDetailAvailable()) {
         const Projection& projection = mapped.getReadyTile()
                                            ->getOverlay()
                                            .getTileProvider()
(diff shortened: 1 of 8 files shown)


Cesium3DTilesSelection/src/Tile.cpp
```diff
@@ -662,16 +662,26 @@ getTileBoundingRegionForUpsampling(const Tile& parent) {
...
-        glm::dvec3(centerProjected, 0.0));
-
-    return RegionAndCenter{details.boundingRegion, center};
+    // The tile will be subdivided by (0.5, 0.5) in the first overlay's
+    // texture coordinates which overlay had more detail.
+
+    for (size_t i = 0; i < parent.getMappedRasterTiles().size(); ++i) {
+      RasterMappedTo3DTile* pMapped =
+          const_cast<RasterMappedTo3DTile*>(&parent.getMappedRasterTiles()[i]);
+      if (pMapped->update(*const_cast<Tile*>(&parent)) ==
```
kring, line 671: A method called `getTileBoundingRegionForUpsampling` shouldn't have a side-effect where it updates the raster overlay mappings. The need for the const cast reinforces that. Is it needed? It looks to me like this method is always called after all the RasterOveralyTile's isMoreDetailAvailable field is updated, so we can avoid the side effect and some unnecessary work by using it.

Cesium3DTilesSelection/src/Tile.cpp
```diff
@@ -1026,6 +1043,21 @@ void Tile::upsampleParent(
         index = int32_t(it - parentProjections.begin());
       }
     }
+  } else {
+    for (RasterMappedTo3DTile& mapped : pParent->getMappedRasterTiles()) {
+      if (mapped.update(*pParent) ==
+          RasterOverlayTile::MoreDetailAvailable::Yes) {
```
kring, line 1049: I think this `update` is redundant as well.

Cesium3DTilesSelection/src/Tile.cpp
joseph-kaile: Thanks, I called Update out of convenience only to know whether the tile had more detail. I added a lightweight const-safe function that does just that.

Cesium3DTilesSelection/src/Tile.cpp
```diff
@@ -1054,8 +1082,9 @@ void Tile::upsampleParent(
             // We can't necessarily trust our original bounding volume, so
             // recompute it here. See:
             // https://github.com/CesiumGS/cesium-native/issues/385
-            pContent->updatedBoundingVolume =
-                GltfContent::computeBoundingRegion(*pContent->model, transform);
+            // pContent->updatedBoundingVolume =
```
kring, line 1092: Remove the commented-out code (and the now-obsolete comment).

kring (commented):
Looks good and works well! I added a comment for something that wasn't clear to me at first in review, so you may want to take a look and make sure I have it right.

Just two things left before merging this:

* Update CHANGES.md
* Fix the test compilation errors

joseph-kaile:
Thanks!

PR #504, opened 2022-05-23
Add option to invert polygon selections

nithinp7:
Adds an option to invert a `RasterizedPolygonOverlay` so that it rasterizes outside the polygon instead of inside. The `RasterizedPolygonsTileExcluder` also takes into account whether the polygon overlay is inverted. If so, tiles outside the polygons will be excluded instead of tiles within.

### DIFF
--- a/Cesium3DTilesSelection/src/TileUtilities.cpp
+++ b/Cesium3DTilesSelection/src/TileUtilities.cpp
@@ -199,8 +199,23 @@ bool outsidePolygons(
       }
     }
 
-    // Check if the polygon perimeter intersects the bounding globe rectangle
-    // edges.
+    // Check if an arbitrary point on the polygon is in the globe rectangle.
+    if (withinTriangle(
+            vertices[0],
+            rectangleCorners[0],
+            rectangleCorners[1],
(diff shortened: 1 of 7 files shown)


Cesium3DTilesSelection/src/TileUtilities.cpp
```diff
@@ -132,6 +141,107 @@ bool withinPolygons(
...
+      if (withinTriangle(
+              rectangleCorners[0],
+              vertices[indices[j - 2]],
+              vertices[indices[j - 1]],
+              vertices[indices[j]])) {
+        return false;
+      }
+    }
+
+    // Check if an arbitrary point on the polygon is in the globe rectangle.
```
kring, line 202: I think it would be better to check if the first polygon vertex is in the rectangle before checking if the first rectangle corner is in the polygon. Only because the former test will be much cheaper, and sometimes that cheaper test will save you from needing to do the more expensive one.

kring (commented):
Just one suggestion, looks great otherwise.

BTW I committed a small change myself to pass the initial value for a vector to `resize` rather than resizing and than memset'ing the initial value.

PR #509, opened 2022-06-01
Ignore layer.json "bounds" property.

kring:
Some third-party terrain tiling tools generate an incorrect "bounds" property in layer.json, and these tilesets still work in CesiumJS because it ignores this property. This PR makes cesium-native match the CesiumJS behavior.

I could argue that CesiumJS is wrong and should change, but because quantized-mesh / layer.json is basically a legacy format at this point (and had a rather informal specification to begin with), compatibility is more important than correctness.

See https://community.cesium.com/t/cesium-terrain-for-unreal/17940/18.


kring:
I'm going to merge this myself because it's a small change and would be good to include in today's release.

PR #510, opened 2022-06-06
quantized-mesh / layer.json loader for refactored loader system

kring:
I'm not sure this is as efficient as it can be, especially in terms of bouncing between the main thread and worker thread. And also the tile initializer probably isn't used optimally. But it's working, which seems like a good start for a PR into the main branch.

One part I struggled with was when to create child tiles. It's currently done at the end of the Future chain that loads the tile. This isn't great because:

1. We shouldn't need to wait for a tile to load in order to create its children, in most cases. But `main` has this problem as well at the moment.
2. It would likely be more efficient to this in the initializer rather than in a `thenInMainThread`.

I noticed that the existing implicit loaders create explicit Tile instances for every available tile at initializer time. I think this isn't great for implicit tiling, but it's especially bad for quantized-mesh / layer.json because CWT, for instance, chunks up availability data in groups of 10 levels. That means each level 10 tile we load would end up creating 1024 explicit Tile instances, and they'll never be unloaded.

I wonder if the loader interface can provide a separate method to create child tiles, making loader implementations both easier and more efficient. The method could return one of three ways: "ok, here are the tile's children", "nope, this tile has no children", or "no-can-do yet, ask again later".

### DIFF
--- a/Cesium3DTilesSelection/src/LayerJsonTerrainLoader.cpp
+++ b/Cesium3DTilesSelection/src/LayerJsonTerrainLoader.cpp
@@ -1,0 +1,439 @@
+#include "LayerJsonTerrainLoader.h"
+
+#include "calcQuadtreeMaxGeometricError.h"
+
+#include <CesiumAsync/IAssetResponse.h>
+#include <CesiumUtility/JsonHelpers.h>
+#include <CesiumUtility/Uri.h>
+
+#include <rapidjson/document.h>
+
+using namespace CesiumAsync;
+using namespace Cesium3DTilesSelection;
+using namespace CesiumGeometry;
+using namespace CesiumGeospatial;
+using namespace CesiumUtility;
+
+namespace {
+
+/**
+ * @brief Creates the query parameter string for the extensions in the given
+ * list.
+ *
+ * This will check for the presence of all known extensions in the given list,
+ * and create a string that can be appended as the value of the `extensions`
+ * query parameter to the request URL.
+ *
+ * @param extensions The layer JSON
+ * @return The extensions (possibly the empty string)
+ */
+std::string createExtensionsQueryParameter(
+    const std::vector<std::string>& knownExtensions,
+    const std::vector<std::string>& extensions) {
+
+  std::string extensionsToRequest;
+  for (const std::string& extension : knownExtensions) {
+    if (std::find(extensions.begin(), extensions.end(), extension) !=
+        extensions.end()) {
+      if (!extensionsToRequest.empty()) {
+        extensionsToRequest += "-";
+      }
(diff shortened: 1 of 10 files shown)


Cesium3DTilesSelection/src/LayerJsonTerrainLoader.cpp
```diff
@@ -0,0 +1,725 @@
...
+                  for (auto& range : loadResult.availableTileRectangles) {
+                    layer.availability.addAvailableTileRange(range);
+                  }
+                  return std::move(loadResult);
+                });
+  }
+
+  return std::move(futureQuantizedMesh)
+      .thenImmediately([](QuantizedMeshLoadResult&& loadResult) {
+        // TODO: why does TileLoadRequest need an IAssetRequest?
```
kring, line 507: I think I see why now: it's needed to determine that a load failure was the result of a 401, so we know to refresh the ion token (if this is an ion request). Getting the request into this lambda is super awkward, though.

baothientran:
> I wonder if the loader interface can provide a separate method to create child tiles, making loader implementations both easier and more efficient. The method could return one of three ways: "ok, here are the tile's children", "nope, this tile has no children", or "no-can-do yet, ask again later".

yeah that seems to be reasonable to me. I will try to add that method to the implicit loaders to see what happens

Cesium3DTilesSelection/src/LayerJsonTerrainLoader.cpp
```diff
@@ -0,0 +1,725 @@
...
+                });
+
+  return std::move(futureResult)
+      .thenInMainThread(
+          [this, &tile, pQuadtreeTileID](TileLoadResult&& result) {
+            // Now that all our availability is sorted out, create this tile's
+            // children.
+            // TODO: in most cases, we can create this tile's children before
+            // this tile is loaded, and it would be much more efficient to take
+            // advantage of that when we can.
```
baothientran, line 647: for most cases, do you mean the cases where the tile is in the middle of the subtree? Because if a tile is at the availability level or at the end of a subtree, it's still need to wait for the current layer to be updated with the new available rectangle before it can create the children right?

kring: Yep, exactly right. If availabilityLevels is 10, then there is availability data in tiles at level 0, 10, 20, and 30. At levels 1-9, we don't need to wait for any loading (beyond the level 0 tile) in order to create and traverse those tiles.

baothientran:
Thanks @kring! I updated the PR to use the new interface `updateTileContent` to create children tile without waiting for tile finished loading

PR #511, opened 2022-06-07
Allow tileset loader to update tile content

baothientran:
This PR addresses the feedback for [the refactor PR comment](https://github.com/CesiumGS/cesium-native/pull/508#discussion_r889857003)

The PR adds `TilesetContentLoader::updateTileContent(Tile &tile)` to allow loader to update the tile content piece by piece if needed. The function allows the implicit tileset to create the children of tile if needed, instead of creating all children in the subtree. The function returns a boolean to indicate if the tile should be continue updated in the next frame. For the implicit tileset, once it creates the children for the tile, it will return false to indicate that the `updateTileContent()` should no longer be called on this tile in the future.

Along with this PR, I move `std::function<Tile &> tileInitializer` to `TileExternalContent`. The original idea in the original PR is that the loader can pass some custom info to the initializer (e.g maybe updated bounding volume) during loading. However, the only current use for it is to create children for a tile. Using it to pass custom data from a loader to a tile maybe best suited for `std::any` and use the new `updateTileContent(Tile &tile)` to process that `std::any`, but that is a feature for the future. 

Using `TileExternalContent` and `std::function<Tile &>` to create a subtree can be even optimized further (not implemented in this PR). Since `TileExternalContent` is stored in the std::variant along with `CesiumGltf::Model` and `CesiumGltf::Model` has more than 512 bytes memory footprint, it would be nice if any functor that is passed to TileExternalContent is created in that memory. Currently `std::function<Tile &>` has way too small inlined buffer, so it will very likely allocate for any functor passed into it. We will need to create a custom one only for TileExternalContent to take advantage of this optimization

cc @kring

### DIFF
--- a/Cesium3DTilesSelection/src/ImplicitOctreeLoader.cpp
+++ b/Cesium3DTilesSelection/src/ImplicitOctreeLoader.cpp
@@ -319,28 +319,27 @@ CesiumAsync::Future<TileLoadResult> ImplicitOctreeLoader::loadTileContent(
     // subtree is not loaded, so load it now.
     std::string subtreeUrl =
         resolveUrl(_baseUrl, _subtreeUrlTemplate, subtreeID);
-    SubtreeAvailability::loadSubtree(
-        3,
-        asyncSystem,
-        pAssetAccessor,
-        pLogger,
-        subtreeUrl,
-        requestHeaders)
-        .thenInWorkerThread(
-            [this, subtreeID](std::optional<SubtreeAvailability>&&
-                                  subtreeAvailability) mutable {
-              if (subtreeAvailability) {
-                this->addSubtreeAvailability(
-                    subtreeID,
-                    std::move(*subtreeAvailability));
-              }
-            });
+    return SubtreeAvailability::loadSubtree(
+               3,
+               asyncSystem,
+               pAssetAccessor,
+               pLogger,
+               subtreeUrl,
+               requestHeaders)
+        .thenInMainThread([this, subtreeID](std::optional<SubtreeAvailability>&&
+                                                subtreeAvailability) mutable {
+          if (subtreeAvailability) {
+            this->addSubtreeAvailability(
+                subtreeID,
+                std::move(*subtreeAvailability));
+          }
 
-    // tell client to retry later
-    return asyncSystem.createResolvedFuture(TileLoadResult{
-        TileUnknownContent{},
-        TileLoadResultState::RetryLater,
-        nullptr});
+          // tell client to retry later
+          return TileLoadResult{
+              TileUnknownContent{},
+              TileLoadResultState::RetryLater,
+              nullptr};
+        });
   }
 
   // subtree is available, so check if tile has content or not. If it has, then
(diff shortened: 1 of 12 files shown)


Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -219,6 +219,10 @@ void TilesetContentManager::updateTileContent(Tile& tile) {
   default:
     break;
   }
+
+  if (content.shouldContentContinueUpdated()) {
+    content.setContentShouldContinueUpdated(_pLoader->updateTileContent(tile));
```
kring, line 224: I think it's also necessary to defer transitioning the Tile to the Done state until `updateTileContent` gives the OK. Or at least, we'll need some way to make sure we don't create upsampled children before the real children have been created.

kring:
Other than that one comment (which will only become an issue when raster overlays are available), this looks good to me. Feel free to merge it whenever is convenient for you.

baothientran:
@kring I just push some updates today:
- The quadtree and octree tileset will use the main thread continuation to add the subtree back to loader and tell the client to retry later for tile content. 
- I restore the tile initializer again since LayerJson tileset requires it to update the tile bounding volume. Also implicit tileset specs does have metadata to allow client to override the tile geometric error and bounding box as well which is not implemented at the moment. But I think the future implementation will need to use the tile initializer to pass those info to the tile
- On the tileset content manager side, task continuation returned by loader will be run immediately instead of in the worker thread. The immediate continuation is used to quickly determine if we need to spawn a worker thread to process render content. So in the case where loader uses main thread task, there is no need to spawn a new worker thread if the result of the previous task is not render content

Cesium3DTilesSelection/src/TilesetContentManager.cpp
baothientran: I can see it now. I think it maybe better to solve the transition in the raster overlay PR

Cesium3DTilesSelection/src/ImplicitOctreeLoader.cpp
```diff
@@ -380,67 +326,28 @@ CesiumAsync::Future<TileLoadResult> ImplicitOctreeLoader::loadTileContent(
...
+                subtreeID,
+                std::move(*subtreeAvailability));
+          }
+
+          // tell client to retry later
+          return TileLoadResult{
+              TileUnknownContent{},
+              TileLoadResultState::RetryLater,
+              nullptr,
+              {}};
```
baothientran, line 349: Subtree is now added back to the loader in the main thread as soon as possible. It doesn't need to wait for the tile to be loaded before adding back to the loader

Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -174,21 +173,48 @@ void TilesetContentManager::loadTileContent(
           _externals.pAssetAccessor,
           _externals.pLogger,
           _requestHeaders)
-      .thenInWorkerThread(
+      .thenImmediately(
```
baothientran, line 176: task returned by the loader will be run immediately, so that we can determine if we should spawn a new worker thread for render content. If the result is not a render content and not success, it can end up in the main thread much faster if the previous task is ran on main thread for some reasons

baothientran:
This does work quite well even for quantized mesh, so I'm confident to merge it now

PR #516, opened 2022-06-29
Add raster overlay back to the refactor code

baothientran:
This PR adds back raster overlay code to the refactored loader:

- Since Tile no longer stores the tileset, `IPrepareRendererResources` is passed to `MappedRasterToTile` directly to create render resources for raster
- In the main branch, upsampling code is used to handle both quantized mesh and raster overlay. In this branch, if the loader produced upsampled tile ID, it will be responsible for upsampling that tile. So in this case, layerjson will handle upsampling its own tile, and there is a RasterOverlayUpsampler which will handle the upsampling for raster overlay tiles
- TilesetContentManager will check if a tile has an upsampled tile ID or not. If it's an upsampled tile, it will check if the parent is loaded before loading this tile. This condition is for any generic loaders where it's required to upsample tiles (e.g voxel loader may requires upsampling), and not just for raster overlay upsampler, or quantized mesh. 
- `TileLoadResult` will have updated bounding volumes field as this information is needed to calculate Raster Overlay UVs

### DIFF
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/GltfUtilities.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/GltfUtilities.h
@@ -1,5 +1,8 @@
 #pragma once
 
+#include <Cesium3DTilesSelection/RasterOverlayDetails.h>
+#include <CesiumGeospatial/BoundingRegion.h>
+#include <CesiumGeospatial/GlobeRectangle.h>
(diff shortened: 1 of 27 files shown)


kring (commented):
Looks good, I'm going to merge this. Minor comments below can be addressed in the target branch.

baothientran:
Thanks @kring! I addressed all the PR comments in the new `pr-review-integration` branch

PR #520, opened 2022-06-30
Add 3D Tiles 1.1 support to serializer/deserializer

lilleyse:
Added 3D Tiles 1.1 support to the 3D Tiles serializer/deserializer and dropped support for 3D Tiles Next extensions that have been promoted to core. It would have been too complicated to support both simultaneously and without much benefit.


kring:
Thanks @lilleyse!

PR #523, opened 2022-07-05
Remove explicit initPool step.

kring:
@nithinp7 this is a PR into #492 showing what I had in mind for removing `initPool`. The base-class, cesium-native code still manages the pool, but the items in the pool are only created when they're needed (rather than all up front) and the derived class controls the total pool size.


nithinp7:
This looks cleaner to me and seem to work well, thanks @kring!

PR #525, opened 2022-07-14
Add API call for current tileset load progress

miatang13:
- This PR contains adding a new API `computeLoadProgress` to the `Tileset` class.
- We added a counter keeping track of the current number of loaded tiles in the tileset to avoid counting the doubly linked list of loaded tiles every frame. 
- The load progress computation currently takes into account the three loading queues (`low`, `medium`, `high`), current `loadsInProgress`.


nithinp7:
Thanks for this PR @miatang13! Some of this code will have to be moved into the `TileContentManager` when integrating with the tile loading refactor but we can merge this first!

Issue #481, opened 2022-04-08
Tileset and Tile Content Loader Refactor Overview


baothientran:
### Why we need to refactor our tile and tileset content loader?
  - Currently the code that loads tile and tileset json does quite many things which may not be related to each other. For example:  
    - `LoadTilesetDotJson` struct is used to load tileset.json and layer.json formats, which are two different formats for 3D Tiles and Quantized Mesh respectively
    - `LoadSubtree` struct is used to load a subtree of implicit quadtree and octree format. However, at most, only one of them is used for the entire life cycle of the tileset
    - ~`Tile::requestContent()` does upsampling for quantized mesh, but the code seems to never be used for normal tileset, implicit quadtree, and octree~ (Current implementation provides upsampling for all tilesets)
    - `TileContext` stores some unrelated info for an external tileset. It stores the info of  a normal tileset.json, a layer.json, a quadtree, and an octree. But at most, only one of them is used
    - `ImplicitTraversalUtilities::createImplicitChildrenIfNeeded()` creates the children for a tile of quantized mesh, implicit quadtree, and octree. `Tile::processLoadedContent()` also create a tile's children in case it's a normal external tileset.json only. They have the same responsibility, but appears in different code path
    
  -  The state machine for keeping track of the tile loading process is [leaked out to the Tileset class](https://github.com/CesiumGS/cesium-native/blob/75ca82a48f3b566770b8d1e550d5e5643d8438ab/Cesium3DTilesSelection/src/Tileset.cpp#L1374). This should be encapsulated within the tile itself, and never exposed to the client. Otherwise, we will never know who change the state of whom, and if not careful, will leak resources

### What is the goal of the refactor?
  - We should ensure that the tileset loader of one type will be able to load the tile content of that type only and nothing else. That means:
    -  tileset.json, implicit quadtree, and octree will deal with loading only b3dm, cmpt, gltf, glb tile and no quantized mesh. Layer.json will deal with quantized mesh tile only.
    - tileset.json will load json as an external tileset. Implicit quadtree deals with quadtree subtree. And implicit octree deals with octree subtree. Layer.json doesn't have explicit external tileset and deal with rectangle availability metadata as an external subtree
    - ~Only layer.json will deal with upsampling~ (Current implementation provides upsampling for all tilesets)
   - Tile State machine should be put into one place and not leaked out
   - We should handle other similar types of tileset in the future as well like voxel, etc without modifying existing codes

### Proposed architecture change
  - Diagram: 
[...]

PR #531, opened 2022-07-26
Remove dependency between tileset and Raster Overlay Collection

baothientran:
This PR is part of the refactor series #481. Raster Overlay collection now only uses `TilesetExternals` and `Tile::LoadedLinkList` to attach raster overlay to tile. This makes TilesetContentManager easier to test without involving Tileset since the manager depends on the raster overlay collection which requires tileset to be constructed in the previous implementation.

This is also the last dependency between the traversal (residing in tileset itself) and loading systems. By removing it, client can totally build their own traversal outside of Cesium Native and reuse the current loading system as they see fit

### DIFF
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/RasterOverlayCollection.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/RasterOverlayCollection.h
@@ -1,8 +1,10 @@
 #pragma once
 
-#include "Library.h"
-#include "RasterOverlay.h"
-#include "RasterOverlayTileProvider.h"
+#include <Cesium3DTilesSelection/Library.h>
+#include <Cesium3DTilesSelection/RasterOverlay.h>
+#include <Cesium3DTilesSelection/RasterOverlayTileProvider.h>
+#include <Cesium3DTilesSelection/Tile.h>
+#include <Cesium3DTilesSelection/TilesetExternals.h>
 
 #include <gsl/span>
 
@@ -10,8 +12,6 @@
 #include <vector>
 
 namespace Cesium3DTilesSelection {
-
-class Tileset;
 
 /**
  * @brief A collection of {@link RasterOverlay} instances that are associated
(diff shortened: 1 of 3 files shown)


Cesium3DTilesSelection/include/Cesium3DTilesSelection/RasterOverlayCollection.h
```diff
@@ -1,8 +1,10 @@
 #pragma once
 
-#include "Library.h"
-#include "RasterOverlay.h"
-#include "RasterOverlayTileProvider.h"
+#include <Cesium3DTilesSelection/Library.h>
```
kring, line 3: These were using quotes before because they were relative paths to files in the same directory as this header file. The idea being that Cesium3DTilesSelection should include other parts of itself from the same directory, rather than resolving `Cesium3DTilesSelection` anew. I don't have strong feelings on this - it's pretty unlikely to make a difference to anyone - but we should try to be consistent about it.

Cesium3DTilesSelection/src/RasterOverlayCollection.cpp
```diff
@@ -28,14 +41,14 @@ void RasterOverlayCollection::add(std::unique_ptr<RasterOverlay>&& pOverlay) {
   this->_overlays.push_back(std::move(pOverlay));
 
   pOverlayRaw->loadTileProvider(
-      this->_pTileset->getAsyncSystem(),
-      this->_pTileset->getExternals().pAssetAccessor,
-      this->_pTileset->getExternals().pCreditSystem,
-      this->_pTileset->getExternals().pPrepareRendererResources,
-      this->_pTileset->getExternals().pLogger);
+      this->_externals.asyncSystem,
```
kring, line 44: Would it be better to introduce a `RasterOverlayExternals` struct to hold these? Or perhaps just rename `TilesetExternals` to something less Tileset-specific?

Cesium3DTilesSelection/src/RasterOverlayCollection.cpp
```diff
@@ -8,9 +8,22 @@ using namespace CesiumGeometry;
 using namespace CesiumGeospatial;
 
 namespace Cesium3DTilesSelection {
+namespace {
+template <class Function>
```
kring, line 12: `Tileset.h` is still included above, but I don't think it's needed anymore.

kring (commented):
More minor comments that can be addressed i the main branch.

baothientran:
Thanks @kring! I addressed the comment in `pr-review-integration` branch. I haven't renamed the TilesetExternals to something else though, but that can be in the future if it's needed

PR #533, opened 2022-07-27
Change tileset load percentage to float

miatang13:
- A follow up update for https://github.com/CesiumGS/cesium-unreal/issues/774 
- Update the API for getting the current loaded percentage from returning an unsigned int to a float.

### DIFF
--- a/Cesium3DTilesSelection/src/Tileset.cpp
+++ b/Cesium3DTilesSelection/src/Tileset.cpp
@@ -367,7 +367,7 @@ void Tileset::notifyTileUnloading(Tile* pTile) noexcept {
   }
 }
 
-uint32_t Tileset::computeLoadProgress() noexcept {
+float_t Tileset::computeLoadProgress() noexcept {
   uint32_t queueSizeSum = (uint32_t)(this->_loadQueueLow.size() +
                                 this->_loadQueueMedium.size() +
                                 this->_loadQueueHigh.size());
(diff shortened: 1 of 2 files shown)


Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -367,7 +367,7 @@ void Tileset::notifyTileUnloading(Tile* pTile) noexcept {
   }
 }
 
-uint32_t Tileset::computeLoadProgress() noexcept {
+float_t Tileset::computeLoadProgress() noexcept {
```
nithinp7, line 370:
```suggestion
float Tileset::computeLoadProgress() noexcept {
```

Not sure I'm completely familiar with the difference between `float` and `float_t`, but I think it's more consistent with the rest of the codebase to stick with float. From my quick googling of the difference just now, it appears platforms are allowed to implement `float_t` as double-precision, which forces the compiler to have to treat all intermediary results involving `float_t` as doubles, just in case. I think `float` on the other hand is guaranteed to have a size of 4 bytes. Doesn't particularly matter in this case, but it's good to be consistent regardless.

This is what I was reading on the difference between the two: https://stackoverflow.com/questions/5390011/whats-the-point-of-float-t-and-when-should-it-be-used

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -376,7 +376,7 @@ uint32_t Tileset::computeLoadProgress() noexcept {
   uint32_t totalNum = this->_loadedTilesCount + inProgressSum;
   float_t percentage = static_cast<float>(this->_loadedTilesCount) /
```
nithinp7, line 377:
```suggestion
  float percentage = static_cast<float>(this->_loadedTilesCount) /
```

Might as well change it here too.

nithinp7 (changes requested):
Looks good apart from the below comment!

miatang13:
Updated! @nithinp7

PR #534, opened 2022-07-28
Let loader interface to accept const reference of tile

baothientran:
This is part of the refactor series #481. The loader now only accepts the const reference of tile in `TilesetContentLoader::loadTileContent()`. `TilesetContentLoader::updateTileContent()` is changed to `TilesetContentLoader::createTileChildren()` and accepts the const ref for tile as well

### DIFF
--- a/Cesium3DTilesSelection/src/LayerJsonTerrainLoader.cpp
+++ b/Cesium3DTilesSelection/src/LayerJsonTerrainLoader.cpp
@@ -24,10 +24,12 @@ TileLoadResult convertToTileLoadResult(QuantizedMeshLoadResult&& loadResult) {
       TileRenderContent{std::move(loadResult.model)},
       loadResult.updatedBoundingVolume,
       std::nullopt,
+      std::nullopt,
+      nullptr,
+      {},
       loadResult.errors.hasErrors() ? TileLoadResultState::Failed
                                     : TileLoadResultState::Success,
-      nullptr,
-      {}};
+  };
 }
 
 size_t maxSubtreeInLayer(uint32_t maxZooms, int32_t availabilityLevels) {
@@ -89,6 +91,38 @@ void addRectangleAvailabilityToLayer(
   }
 
   layer.loadedSubtrees[subtreeLevelIdx].insert(subtreeMortonIdx);
+}
+
+void generateRasterOverlayUVs(
+    const BoundingVolume& tileBoundingVolume,
+    const glm::dmat4& tileTransform,
+    const Projection& projection,
+    TileLoadResult& result) {
+  if (result.state != TileLoadResultState::Success) {
+    return;
+  }
+
+  const BoundingRegion* pParentRegion = nullptr;
+  if (result.updatedBoundingVolume) {
+    pParentRegion =
+        std::get_if<BoundingRegion>(&result.updatedBoundingVolume.value());
+  } else {
+    pParentRegion = std::get_if<BoundingRegion>(&tileBoundingVolume);
+  }
+
+  TileRenderContent& renderContent =
+      std::get<TileRenderContent>(result.contentKind);
+  if (renderContent.model) {
+    result.rasterOverlayDetails =
+        GltfUtilities::createRasterOverlayTextureCoordinates(
+            *renderContent.model,
+            tileTransform,
+            0,
+            pParentRegion ? std::make_optional<GlobeRectangle>(
+                                pParentRegion->getRectangle())
+                          : std::nullopt,
+            {projection});
+  }
 }
 
 /**
@@ -614,7 +648,7 @@ Future<int> loadTileAvailability(
 } // namespace
 
 Future<TileLoadResult> LayerJsonTerrainLoader::loadTileContent(
-    Tile& tile,
+    const Tile& tile,
     const TilesetContentOptions& contentOptions,
     const AsyncSystem& asyncSystem,
     const std::shared_ptr<IAssetAccessor>& pAssetAccessor,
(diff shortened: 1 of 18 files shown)


Cesium3DTilesSelection/src/RasterOverlayUpsampler.cpp
```diff
@@ -85,26 +82,29 @@ CesiumAsync::Future<TileLoadResult> RasterOverlayUpsampler::loadTileContent(
...
+            {},
+            TileLoadResultState::Success};
+      });
 }
 
-bool RasterOverlayUpsampler::updateTileContent([[maybe_unused]] Tile& tile) {
-  return false;
+TileChildrenResult
+RasterOverlayUpsampler::createTileChildren([[maybe_unused]] const Tile& tile) {
+  return {{}, TileLoadResultState::Success};
```
baothientran, line 108: the code for subdividing tile for upsampling in `TilesetContentManager` can be put in here. The only concern I have is `createQuadtreeSubdividedChildren()` is called when `RasterMappedTo3DTiles::update()` indicates if more raster is available. So when moving `createQuadtreeSubdividedChildren()` to `TilesetContentLoader::createTileChildren()`, it will be called before `RasterMappedTo3DTiles::update()`. There is `RasterMappedTo3DTiles::isMoreDetailAvailable()`, can I use that to determine if we should create more children instead of using the update method?

kring: Yes, I believe that should work just fine. Just by aware that if `isMoreDetailAvailable` returns false, that doesn't necessarily mean "no". It can also mean "I don't know yet."

Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -295,13 +294,43 @@ std::optional<RasterOverlayDetails> calcRasterOverlayDetailsInWorkerThread(
...
       getBoundingRegionFromBoundingVolume(contentBoundingVolume);
 
+  // remove any projections that are already used to generated UV
+  int32_t firstRasterOverlayTexCoord = 0;
+  if (result.rasterOverlayDetails) {
+    const auto& existingProjections =
+        result.rasterOverlayDetails->rasterOverlayProjections;
+    firstRasterOverlayTexCoord =
+        static_cast<int32_t>(existingProjections.size());
+    for (size_t i = 0; i < projections.size(); ++i) {
```
kring, line 304: I think this can be written more cleanly with `std::remove_if`. Something like this (totally untested):

```
projections.erase(
  std::remove_if(
    projections.begin(),
    projections.end(),
    [&existingProjections](const Projection& projection) {
      return std::find(
        existingProjections.begin(),
        existingProjections.end(),
        projection) != existingProjections.end();
    }),
  projections.end());
```

Maybe no one but me finds that cleaner? 😆

kring (commented):
This looks good to me. One minor suggestion below.

More generally, I'd prefer to avoid using `auto` so much. The reason is that, in my view, types are a form of documentation. Suppressing them from the code makes it harder to read. Things like this are fine:

```
auto pFoo = std::make_unique<Foo>();
```

Because it's no less clear and avoids naming Foo on both sides of the equals. But something like this:

```
const auto& pLogger = loadInput.pLogger;
```

I think is less clear because the actual type of pLogger is excellent documentation about what that thing is and how it should be used. Naming the type makes the code easier to read. Obviously any IDE will show the type on hover anyway, but it's still not quite the same. And GitHub's review interface doesn't show types on hover.

But you don't need to go back and change all the autos, just mentioning it for future reference.

Cesium3DTilesSelection/src/LayerJsonTerrainLoader.cpp
```diff
@@ -727,69 +766,108 @@ Future<TileLoadResult> LayerJsonTerrainLoader::loadTileContent(
...
   }
 
-  return false;
+  return {{}, TileLoadResultState::Failed};
 }
 
-void LayerJsonTerrainLoader::createTileChildren(Tile& tile) {
-
-  if (tile.getChildren().empty()) {
+bool LayerJsonTerrainLoader::doesTileHasUpsampledChild(const Tile& tile) const {
```
kring, line 865: `doesTileHaveUpsampledChild`

or alternatively: `tileHasUpsampledChild` is fine too.

kring (commented):
Merging this, too. One more small thing:

baothientran:
Thanks @kring! I addressed the PR comments in pr-review-integration branch

Cesium3DTilesSelection/src/TilesetContentManager.cpp
javagl:
```
/**
 * Returns whether the given container contains the given element.
 * 
 * @param c The container
 * @param t The element
 * @return Whether the container contains the element
 * @tparam <C> The `Container` type
 * @tparam <T> The element type
 */
template<typename C, typename T> 
bool contains(const C& c, const T& t) {
  return std::find(c.begin(), c.end(), t) != c.end(); 
}

/**
 * Removes from the given vector all elements that are contained
 * in the given container.
 * 
 * @param v The vector
 * @param toRemove The container containing the elements to remove
 * @tparam <C> The `Container` type
 * @tparam <T> The element type
 */
template<typename C, typename T> 
void removeAll(std::vector<T>& v, const C& toRemove) {
  auto& newEnd = std::remove_if(v.begin(), v.end(), 
    [&](const T& e) { return contains(toRemove, e); });
  v.erase(newEnd, v.end());
}
```
Call as
`removeAll(projections, existingProjections);`

Akin to [`removeAll`](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#removeAll-java.util.Collection-) / [`contains`](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#contains-java.lang.Object-). 

BTW: This has quadratic complexity. But I assume that there will be very few projections...

baothientran: Thanks @javagl! good idea, I will add this to the utility library

PR #535, opened 2022-07-28
Only occlusion test tiles when they are within the frustum

nithinp7:
Frustum culled tiles usually never reach _visitTile. When they do, they usually never need to refine. This means they don't need to check for occlusion. However, after #492 we use children BVs when possible to determine the visibility of a tile (both for frustum and occlusion culling). This lets us get a better hitrate on culling tiles not needed for the current view.

A tile is not considered frustum culled if only some of the children BVs are out of the frustum. This creates a situation where children BVs that are out of the frustum are still being occlusion tested. This creates two equally uncomfortable options for the renderer client:
- The client might issue a few unnecessary occlusion queries if the implementation does no independent frustum culling.
- If the client does its own frustum culling (e.g., Unreal Engine), the resulting occlusion results might come back invalid (since they may have never been issued). It will be hard then, to decide what to do in that case: do we wait because there will eventually be results, or do we just treat it as not occluded? If we treat a tile's child as unoccluded, but it turns out to just be out of the frustum, we may miss a chance to cull the tile.

Luckily, it's easy enough on our end to avoid occlusion queries for BVs out of the frustum: we can just cache the results of the frustum checks in _frustumCull and reuse them in _checkOcclusion.

### DIFF
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
@@ -601,6 +601,11 @@ private:
   // selection.
   std::vector<double> _distances;
 
+  // Caches frustum check results of children for the currently traversed tile.
+  // Allows for occlusion culling to reuse frustum results - occlusion queries
+  // only have to be issued for tiles that are within the frustum.
+  std::vector<bool> _childrenFrustumChecks;
+
   // Holds the occlusion proxies of the children of a tile. Store them in this
   // scratch variable so that it can allocate only when growing bigger.
   std::vector<const TileOcclusionRendererProxy*> _childOcclusionProxies;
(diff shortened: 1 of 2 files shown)


kring:
The only potential problem I can think of is if cesium-native's frustum check says "it's visible", but then the renderer's says "nope it's not actually visible" because it checks with a tighter-fitting bounding volume. That's pretty unlikely for Unreal because it uses AABBs, but it wouldn't be shocking if a renderer computed a bounding volume from the geometry or somesuch.

Can this really happen?

> the resulting occlusion results might come back invalid (since they may have never been issued)

I mean, if an occlusion check is not issued because it was frustum culled, surely we should be able to detect that and treat it as occluded for our purposes, right? That seems like the correct policy for the occlusion itnerface: occlusion results must come back, even if the renderer decides that the occlusion check is pointless because the entire thing was frustum culled.

If it is possible to detect this in Unreal, then this change might still be useful to avoid issuing pointless occlusion queries. But it's just a small optimization rather than a matter of correctness.

kring:
In Unreal, occlusion results are per view, right? So what happens if the occlusion query gets culled in some views but not in others? Will the occlusion results be incorrect in the views in which the query was culled?

cesium-concierge:
Thanks again for your contribution @nithinp7! 

nithinp7:
This is no longer needed after changes to Cesium for Unreal that took renderer-level frustum culling into account.

PR #536, opened 2022-08-03
Update SqliteCache last access time correctly

kring:
Because of an error in a SQL statement, the last access time in the request cache was not updated on cache hit. This could lead to the wrong requests being evicted from the cache.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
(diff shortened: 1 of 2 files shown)


kring:
@baothientran can you please review this? I want to do a v1.16.2 release and include this PR and https://github.com/CesiumGS/cesium-unreal/pull/936.

baothientran:
Look good to me

PR #539, opened 2022-08-12
Add unit test to the refactored loading branch

baothientran:
This PR is part of the refactor series #481. This adds unit tests for all the following:
- fix the QuantizedMeshContent tests
- SubtreeAvailability loading code
- TilesetJsonLoader interface
- ImplicitQuadtreeLoader
- ImplicitOctreeLoader
- LayerJsonTerrainLoader
- TilesetContentManager

API for the loader has changed to make the unit test easier:
- `TileContentLoaderResult` is now templated to point to the original type of the loader, instead of pointing to the base class `TilesetContentLoader`. This is needed because the unit test needs to access to the real loader type to examine the loader's field when creating loader

Other changes to address the PR comment in #508:
- Address [the issue where cesium-native doesn't log when error callback option is missing](https://github.com/CesiumGS/cesium-native/pull/508#discussion_r888518846).
- Address [the issue where `TileContent` is allocated when creating tile](https://github.com/CesiumGS/cesium-native/pull/508#discussion_r889841160). `TileContent` is now just a variant of `TileUnknownContent`, `TileEmptyContent`, `TileExternalContent`, and `std::unique_ptr<TileRenderContent>`. So its size is now only at least 9 bytes and will be constructed on the stack. Constructing `TileExternalContent` or `TileEmptyContent` won't no longer allocate. Along with this change, I move `TileLoadState` and `TilesetContentLoader *pLoader` fields from previous `TileContent` to the `Tile` class. They can reside within the new `TileContent`, but either case, the Tile size is still larger than previous implementation
- Address [the issue where `TileRenderContent` has optional gltf model](https://github.com/CesiumGS/cesium-native/pull/508#discussion_r889843328). This is now removed and only stores `CesiumGltf::Model`
- [Use std::visit to subdivide bounding box of octree and quadtree](https://github.com/CesiumGS/cesium-native/pull/508#discussion_r889855125)
- [Use std::get to retrieve octree and quadtree ID instead of std::get_if](https://github.com/CesiumGS/cesium-native/pull/508#discussion_r889856069)
- [Expose TilesetContentLoader to include directory and in the Tileset constructor to let client pass custom loader to the tileset](https://github.com/CesiumGS/cesium-native/pull/508#discussion_r889872291)
- [Add `TilesetContentLoader::getTileUpAxis(Tile)` to account for tileset up axis](https://github.com/CesiumGS/cesium-native/pull/508#discussion_r889875783). This will handle the case where different external tilesets will have different up axis. The up axis is also added to gltf extra similar to the current implementation in main.
- [Separate log error and warning into a file, so that different loader can use them](https://github.com/CesiumGS/cesium-native/pull/508#discussion_r889880768)


Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tile.h
```diff
@@ -387,9 +464,20 @@ class CESIUM3DTILESSELECTION_API Tile final {
...
+      TileLoadState loadState,
+      TilesetContentLoader* pLoader,
+      TileContentArgs&&... args);
+
+  void setParent(Tile* pParent) noexcept;
+
+  void setState(TileLoadState state) noexcept;
 
-  void setParent(Tile* pParent) noexcept { this->_pParent = pParent; }
+  bool shouldContentContinueUpdated() const noexcept;
```
kring, line 477: Suggest `shouldContentContinueUpdating`.

kring:
Looks great @baothientran. A massive - and much needed - effort on the testing here! I didn't go through every test in extreme detail, but looked at a couple closely and skimmed the rest. The general approach looks good, and I'm very pleased with how little time the tests take to run as well. If there's anything in particular you're unsure of and want me to look at closely, let me know. In any case, I'm going to merge this, let you merge in main, and then do a final review of the whole thing together before we :shipit:!

PR #541, opened 2022-08-17
Let content manager owns root tile and raster overlay collection

baothientran:
This PR is part of the refactor series #481. In the current refactor-branch, the content manager doesn't own the root tile and the raster overlay collection. The tileset itself owns those fields. However, it is quite awkward to manage the lifetime of those components:
- Because the manager only captures the pointer of `RasterOverlayCollection`, client of the manager (in this case, it's the `Tileset`) has to make sure the collection doesn't die; otherwise, we have a dangling pointer. But the `Tileset` itself doesn't use the collection in its implementation, so it doesn't make sense to let the tileset own it. Instead, the manager should own this collection, since it is the main client of the collection.
- The reason for moving the root tile to the manager is that the manager will create render resources for tile, and those resources are raw pointer, so [it requires client to explicitly unload the tile when not used](https://github.com/CesiumGS/cesium-native/blob/ef87387b6d85ddb57e737ddcfac230db2a71c1c7/Cesium3DTilesSelection/src/Tileset.cpp#L162). In addition, when loading a tile, the manager needs that tile to be alive at least when the loading returns to the main thread. So it probably makes sense to let the manager manages the lifetime of the tile as well, instead of letting client of the manager do it

### DIFF
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/RasterOverlayCollection.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/RasterOverlayCollection.h
@@ -39,6 +39,36 @@ public:
       Tile::LoadedLinkedList& loadedTiles,
       const TilesetExternals& externals) noexcept;
 
+  /**
+   * @brief Deleted Copy constructor.
+   *
+   * @param rhs The other instance.
+   */
+  RasterOverlayCollection(const RasterOverlayCollection& rhs) = delete;
+
+  /**
+   * @brief Move constructor.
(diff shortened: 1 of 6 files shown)


kring:
Looks good! One thing that surprised me a little is that TilesetContentManager now knows about Cesium ion and about tileset.json, and constructs an appropriate loader for each. I thought that responsibility might stay in Tileset, and the content manager would just deal with a loader and a root tile (wherever they come from). This doesn't necessarily need to be changed, I just wanted to mention it.

PR #542, opened 2022-08-18
WebP support

### LINKED ISSUE #75: Add support for WebP images
Currently we're using [stb](https://github.com/nothings/stb) (via tinygltf) for image decoding, but it doesn't support every format. Most notably missing is WebP. We should add support for this format for images inside glTFs and as raster overlays.

joseph-kaile:
Fixes #75


nithinp7:
Thanks @joseph-kaile, this looks great! I tested it with the Sponza model tiled on Cesium Ion with the WebP option enabled. It didn't work and I realized Ion outputs webp images wrapped in the `EXT_texture_webp`. So I generated that extension here and made some small adjustments on the Unreal side to use the texture source found in the extension, if it exists. Webp images should work both ways now, within the extension or just within the regular glTF texture, just like KTX.

nithinp7:
The only other thing I noticed was that libwebp runs its own tests every time I configure the CMake project, is there a way to disable that?

joseph-kaile:
Thanks @nithinp7. I didn't see the Webp option in Ion before but now I do. Unchecking WEBP_ENABLE_SIMD disables the tests. I can set that.

nithinp7:
@joseph-kaile Looks like that last change broke CI for linux + clang and Mac. Do you know how to fix that? We could also just revert to the last commit and I can merge this, up to you.

nithinp7:
If we find a better way to get rid of those extra tests we can add that in a separate PR. Merging the latest commit that passed CI, thanks again @joseph-kaile!

PR #550, opened 2022-08-25
Support layer.json when passing url

baothientran:
Previously when a url is passed to tileset, it will be able to load the layer.json format as well. This PR adds that support for the refactor


kring:
Thanks @baothientran!

PR #556, opened 2022-08-31
Update raster tile right after content has glTF render resources

### LINKED ISSUE #555: Tiles sometimes rendering before raster tile is ready
When flying to a new area, I can occasionally see the flash of a white tile on the screen for a fraction of a second. It appears to be a tile being rendered without the raster imagery. I don't recall seeing this last month, so it might be a regression from #508.

baothientran:
Fixes #555 

The reason is the tile is transitioned to a new state after per frame instead of transitioning to Done within a single frame. So the glTF render resources maybe created when transitioning fron ContentLoaded -> Done, but the raster overlay hasn't attached to it yet before the tile is rendered


kring:
Thanks @baothientran!

PR #557, opened 2022-09-02
Flip the range for fade-out

nithinp7:
Previously, fading in tiles went from 0.0 --> 1.0 and fading out tiles went from 1.0 --> 0.0. When these values are used to threshold a dither texture in a renderer, this small discrepancy might cause floating point disagreement between fading in tiles and fading out tiles. So instead, we now want fading in and fading out tiles to have the exact same dither threshold (fade percentage) and simply invert the resulting sample if it is fading out. Since the dither texture is sampled in screen-space, this allows the fading out tiles and fading in tiles to blend together perfectly without appearing transparent.

@kring came up with the idea and it works quite nicely!

### DIFF
--- a/Cesium3DTilesSelection/src/Tileset.cpp
+++ b/Cesium3DTilesSelection/src/Tileset.cpp
@@ -188,24 +188,25 @@ void Tileset::_updateLodTransitions(
           (*tileIt)->getLastSelectionState().getResult(
               frameState.currentFrameNumber);
       if (selectionResult == TileSelectionState::Result::Rendered) {
-        // This tile will already be on the render list. Just fade it in from
-        // the current fade percentage, no need to set it to 0.0f.
+        // This tile will already be on the render list.
+        pRenderContent->setLodTransitionFadePercentage(0.0f);
         tileIt = result.tilesFadingOut.erase(tileIt);
         continue;
       }
 
       float currentPercentage =
           pRenderContent->getLodTransitionFadePercentage();
-      if (currentPercentage <= 0.0f) {
+      if (currentPercentage >= 1.0f) {
         // Remove this tile from the fading out list if it is already at 0%.
         // The client will already have had a chance to stop rendering the tile
         // last frame.
+        pRenderContent->setLodTransitionFadePercentage(0.0f);
         tileIt = result.tilesFadingOut.erase(tileIt);
         continue;
       }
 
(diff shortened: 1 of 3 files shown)


Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -188,24 +189,25 @@ void Tileset::_updateLodTransitions(
...
 
       float currentPercentage =
           pRenderContent->getLodTransitionFadePercentage();
-      if (currentPercentage <= 0.0f) {
-        // Remove this tile from the fading out list if it is already at 0%.
+      if (currentPercentage >= 1.0f) {
+        // Remove this tile from the fading out list if it is already done.
         // The client will already have had a chance to stop rendering the tile
         // last frame.
+        pRenderContent->setLodTransitionFadePercentage(0.0f);
```
kring, line 204: So when the fade is complete, the fade percentage is reset to 0.0...

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -1402,7 +1404,7 @@ void Tileset::_unloadCachedTiles() noexcept {
     const TileRenderContent* pRenderContent =
         pTile->getContent().getRenderContent();
     if (_options.enableLodTransitionPeriod && pRenderContent &&
-        pRenderContent->getLodTransitionFadePercentage() > 0.0f) {
+        pRenderContent->getLodTransitionFadePercentage() < 1.0f) {
```
kring, line 1407: ... but tiles with a fade less than 1.0 will not be unloaded.

So... tiles will pretty much never be unloaded, right?

kring (commented):
This looks great except I think there's a memory leak. Memory usage goes up steadily even with the cache size at 1 byte, and it drops immediately when LOD transitions are disabled. See below for the likely cause.

nithinp7:
@kring Thanks for spotting that, I just simplified the logic that keeps fading out tiles from being unloaded. Now it just checks if a tile is in the fading-out list (actually an unordered_set) before deciding to unload it. The logic seems cleaner, the memory leak is fixed, and the extra `std::unordered_set::find` call looks fast enough in the profiler. What do you think?

kring:
That sounds good to me, thanks @nithinp7!

PR #558, opened 2022-09-02
Add `rendererOptions` to `TilesetOptions`

### LINKED ISSUE #480: Adopt `rendererOptions` pattern from `RasterOverlayOptions` for `TilesetOptions`, too
In #471, we added a new `rendererOptions` field to `RasterOverlayOptions`. This allows a renderer (e.g. Unreal Engine / Cesium for Unreal) to pass data to an overlay that is then passed back to the renderer via `IPrepareRendererResources::prepareRasterInLoadThread`. This is important because many objects in both cesium-native and the renderer itself are unsafe to access from this background thread.

We should add a similar field to `TilesetOptions` and pass it to `IPrepareRendererResources::prepareInLoadThread`, for the same reason. See discussion here:
https://github.com/CesiumGS/cesium-unreal/pull/698#discussion_r844726486

j9liu:
Closes #480. This PR adds a `rendererOptions` field to `TilesetOptions`, which gets passed to `IPrepareRendererResources::prepareInLoadThread` similar to how `RasterOverlayOptions` has its `rendererOptions` passed to `IPrepareRendererResources::prepareRasterInLoadThread`.

I couldn't find an equivalent unit test for `RasterOverlayOptions / RasteryOverlayTileProvider` -- would it be good to write one for this PR?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,15 @@
 # Change Log
 
+### ? - ?
+
+##### Breaking Changes :mega:
+
+- Added a new parameter, `rendererOptions`, to `IPrepareRendererResources::prepareInLoadThread`.
+
+##### Additions :tada:
+
+- Added a `rendererOptions` property to `TilesetOptions` to pass arbitrary data to `prepareInLoadThread`.
+
 ### v0.19.0 - 2022-09-01
 
 ##### Breaking Changes :mega:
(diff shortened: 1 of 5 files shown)


kring:
This looks good to me. Please just update CHANGES.md. We should also be able to use this to remove the `_encodedMetadataDescription` field from `ACesium3DTileset` in cesium-unreal, as mentioned in https://github.com/CesiumGS/cesium-unreal/pull/698#discussion_r844726486 but that can be a separate PR.

As for the test, I'm not too fussed as there's probably not too much chance of test-breakage here. It's just weaving an instance through. But if it looks straightforward to add a test then it certainly can't hurt!

j9liu:
@kring updated! I looked to see if I could add a test, but nothing in native actually uses the values of `rendererOptions`, so it's hard to test that they were passed through.

kring:
Thanks @j9liu!

PR #559, opened 2022-09-04
CesiumGltfWriter - remove special case for `byteOffset`

lilleyse:
In CesiumGltfWriter we had a special case where `byteOffset` would always get written out even if the value was 0. This causes problems when you have accessors that don't point to buffer views since it's invalid to have `byteOffset: 0` if no `bufferView` is defined. I noticed this when writing out glTFs with `KHR_draco_mesh_compression`.


kring:
Looks good to me. Probably worth mentioning in CHANGES.md, though, right?

lilleyse:
@kring updated!

PR #560, opened 2022-09-14
Refresh Token If Expired

### LINKED ISSUE #551: Layer.json terrain breaks after Cesium ion token expiration
After CesiumGS/cesium-native#508, `CesiumIonTilesetLoader` handles an HTTP 401 from a tile content loading request by attempting to refresh the Cesium ion token. So far, so good. The problem is that it doesn't also do this for the layer.json request. This is usually not a problem because the layer.json is requested first thing, before the token could possibly expire. However, the Cesium ion asset endpoint response (which is where the token comes from) is cached across Tilesets. So the following results in a broken Tileset:

1. Create a Tileset for, say, Cesium World Terrain.
2. Wait for the token to expire (1 hour?)
3. Delete the Tileset without loading any new tiles.
4. Create a new Cesium World Terrain tileset.

The new tileset in step 4 will reuse the asset endpoint response with the expired token to attempt to download layer.json. It will fail with a 401, and give up without refreshing the token. I don't think there's any way to recover from this other than by restarting the application.

I don't know offhand if we have the same problem when downloading the root tileset.json, or if it's specific to layer.json. It's also possible that we had this bug before CesiumGS/cesium-native#508, too.

joseph-kaile:
Fixes #551

### DIFF
--- a/Cesium3DTilesSelection/src/CesiumIonTilesetLoader.cpp
+++ b/Cesium3DTilesSelection/src/CesiumIonTilesetLoader.cpp
@@ -541,15 +541,11 @@ CesiumIonTilesetLoader::refreshTokenIfNeeded(
     bool showCreditsOnScreen,
     TilesetContentLoaderResult<CesiumIonTilesetLoader>&& result) {
   if (result.errors.hasErrors()) {
-    std::string errorString = fmt::format(
-        "Received status code {} for tile content {}",
-        401,
-        ionAssetEndpointUrl);
+    std::string errorString = "Received status code 401";
     for (std::string error : result.errors.errors) {
-      if (errorString == error) {
+      if (error.find(errorString) != std::string::npos){
         CesiumIonTilesetLoader* pIonLoader =
             static_cast<CesiumIonTilesetLoader*>(result.pLoader.get());
-
         pIonLoader->refreshTokenInMainThread(
             externals.pLogger,
             externals.pAssetAccessor,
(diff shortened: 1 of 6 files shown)


Cesium3DTilesSelection/src/CesiumIonTilesetLoader.cpp
```diff
@@ -492,4 +530,35 @@ CesiumIonTilesetLoader::createLoader(
...
+    int64_t ionAssetID,
+    const std::string& ionAccessToken,
+    const std::string& ionAssetEndpointUrl,
+    const AuthorizationHeaderChangeListener& headerChangeListener,
+    bool showCreditsOnScreen,
+    TilesetContentLoaderResult<CesiumIonTilesetLoader>&& result) {
+  if (result.errors.hasErrors()) {
+    std::string errorString = "Received status code 401";
+    for (std::string error : result.errors.errors) {
+      if (error.find(errorString) != std::string::npos) {
```
kring, line 546: Searching for some error text in the list of errors seems fragile. For example, if we change the error message slightly, it'll break silently. Can we modify `mainThreadLoadLayerJsonFromAssetEndpoint` and `mainThreadLoadTilesetJsonFromAssetEndpoint` to include the HTTP status code instead, so we can avoid this?

Cesium3DTilesSelection/src/CesiumIonTilesetLoader.cpp
```diff
@@ -492,4 +530,35 @@ CesiumIonTilesetLoader::createLoader(
...
+        endpointCache.erase(createEndpointResource(
```
kring, line 547: This function is invoked by a `thenImmediately`, so it might end up running any thread. Accessing `endpointCache` from anywhere but the main thread is not allowed, as it can cause undefined behavior. I think the easy fix in this case is to invoke it with `thenInMainThread` instead.

kring:
Thanks @joseph-kaile!

Since this isn't the first time this has broken, I think it's well worth writing a test for it. It can be a separate PR, though.

PR #562, opened 2022-09-22
Add LocalHorizontalCoordinateSystem

kring:
This is kind of like the cesium-native side of cesium-unreal's CesiumGeoreference.

You give it a point on the globe and tell it which direction (west, south, east, north, up, down) each coordinate axis points, and it creates a coordinate system centered on the point and with the axes you specified. Depending on the directions you choose for the axes, the coordinate system may be right- or left-handed. You can also specify a `scale` parameter if the coordinate system is not in meters.

For example, it's easy to construct a left-handed, East-South-Up coordinate system in centimeters, like the one we use in Unreal Engine:

```
LocalHorizontalCoordinateSystem coordinateSystem(
    Cartographic::fromDegrees(-105.25737, 39.736401, 2250.0),
    LocalDirection::East,
    LocalDirection::South,
    LocalDirection::Up,
    1.0 / 100.0);
```

It then provides methods for transforming between the constructed coordinate system and ECEF.

The three axes directions must be orthogonal, i.e., it makes no sense for X to point West and Y to point East. This is only enforced with assertions, as it's almost certainly a developer error.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@
 ##### Additions :tada:
 
 - Added a `rendererOptions` property to `TilesetOptions` to pass arbitrary data to `prepareInLoadThread`.
+- Added `LocalHorizontalCoordinateSystem`, which is used to create convenient right- or left-handeded coordinate systems with an origin at a point on the globe.
 
 ##### Fixes :wrench:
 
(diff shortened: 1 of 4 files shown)


nithinp7:
This looks solid thanks @kring!

PR #563, opened 2022-09-22
Make `IPrepareRendererResources::prepareInLoadThread` asynchronous, return a Future

kring:
This is a PR into  #554.

Previously, `prepareInLoadThread` was invoked in the load thread (as is hopefully obvious from the name), but it had no option to do any truly asynchronous work. Once the function returned, the load-thread part of the process was considered complete.

This PR changes the paradigm so that `prepareInLoadThread` now returns a Future, and the load-thread part of the process isn't complete until the Future resolves. This allows additional network requests, work in other threads, etc. without requiring the load thread to block while those other operations are in progress. For example, it could be used to dispatch part of the work to a render thread, thread pool, or game thread.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -56,3 +56,6 @@
 [submodule "extern/libmorton"]
 	path = extern/libmorton
 	url = https://github.com/Forceflow/libmorton.git
+[submodule "extern/expected-lite"]
+	path = extern/expected-lite
+	url = https://github.com/martinmoene/expected-lite.git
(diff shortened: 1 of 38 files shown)


nithinp7:
Thanks @kring, this looks good to me, I'll merge this and #554 once we get the comment in the other PR resolved!

nithinp7:
@kring Since some of the fixes for issues discussed in #554 are committed here, feel free to self-merge this branch into that one. I'm happy with the future based loading changes, so up to you if you want to just merge it into the other PR to be reviewed altogether when it's done - whatever is more convenient for you!

nithinp7 (approved):

PR #565, opened 2022-09-29
Add resource creation throttling + mip map generation

nithinp7:
Primary Changes:
- Throttle main-thread part of tile loading - particularly throttles transition between `ContentLoaded` and `Done`. Note that this changes some semantics that we previously relied on - previously a `ContentLoaded` tile that was traversed was _always_ pushed to `Done` immediately, this is not the case anymore. We used to rely on that fact in sometimes subtle ways...
- Throttle `unloadCachedTiles`.
- (Both types of throttling can be controlled from `TilesetOptions`.)
- Moved mip-map generation code into Cesium Native, it is now a utility function in `CesiumGltfReader.cpp` that can be called by clients.
- Client implementations of the asynchronous "resource preparer" can now modify the tile and glTF (e.g., they can generate mipmaps if wanted directly in `ImageCesium`).
- Added ability for clients to override our tracing macros (e.g., allows Cesium for Unreal to trace through Cesium Native with Unreal Insights).


Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -1118,6 +1152,18 @@ void TilesetContentManager::updateContentLoadedState(
...
+  } else if (content.isRenderContent()) {
+    TileRenderContent* pRenderContent = content.getRenderContent();
+    if (pRenderContent) {
+      // add copyright
+      pRenderContent->setCredits(GltfUtilities::parseGltfCopyright(
+          *_externals.pCreditSystem,
+          pRenderContent->getModel(),
+          tilesetOptions.showCreditsOnScreen));
+
+      tile.setState(TileLoadState::CreatingResources);
```
kring, line 1165: I think this new state isn't needed at all. Just leave the tile in the `ContentCreated` state until we're ready to create resources, and then do this `setCredits` bit plus the `createRenderResources` before transitioning to `Done`. No reason AFAICT that setCredits needs to happen earlier.

kring:
Adding a comment here doesn't submit the review, does it? (sorry for the noise, just testing something)

Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -466,6 +468,13 @@ TileLoadResultAndRenderResources postProcessGltfInWorkerThread(
     model.generateMissingNormalsSmooth();
   }
 
+  // generate mipmaps
+  if (tileLoadInfo.contentOptions.generateMipMaps) {
+    for (CesiumGltf::Image& image : model.images) {
+      CesiumGltfReader::GltfReader::generateMipMaps(image.cesium);
```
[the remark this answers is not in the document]
nithinp7: That's a good point. I changed the mip-map generation to be an optional client-invoked helper function during loading. This also mean that clients can modify the image during loading now (in both glTFs and raster tiles).

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -1387,12 +1391,14 @@ void Tileset::_processLoadQueue() {
 }
 
 void Tileset::_unloadCachedTiles() noexcept {
+  // TODO: this is a hack, time-budget this similar to tickResourceCreation
```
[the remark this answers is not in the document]
nithinp7: I think a time-based throttling might (hypothetically) still be useful if unload time per-tile is more than main-thread load-time. But in the case of Cesium for Unreal, the destruction bottleneck was elsewhere and it's been fixed. So removed this hack throttling.

Cesium3DTilesSelection/src/TilesetContentManager.cpp
nithinp7: This was a good idea and cleaned up the logic a good bit!

nithinp7:
@kring This is updated and ready for a review as well now!

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -339,7 +340,11 @@ Tileset::updateView(const std::vector<ViewState>& frustums, float deltaTime) {
     pOcclusionPool->pruneOcclusionProxyMappings();
   }
 
-  this->_unloadCachedTiles();
+  this->_pTilesetContentManager->tickMainThreadLoading(
+      this->_options.mainThreadLoadingTimeLimit,
+      this->_options);
+
+  this->_unloadCachedTiles(this->_options.tileCacheUnloadTimeLimit);
```
kring, line 347: Can we call `_unloadCachedTiles` before `tickMainThreadLoading`? It should reduce peak memory usage a little to do it in that order.

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -1416,6 +1432,16 @@ void Tileset::_unloadCachedTiles() noexcept {
...
 
     pTile = pNext;
+
+    // A time budget of 0.0 indicates we shouldn't throttle cache unloads.
+    std::chrono::time_point<std::chrono::system_clock> time =
+        std::chrono::system_clock::now();
+    if (timeBudget > 0.0 &&
+        std::chrono::duration_cast<std::chrono::duration<double, std::milli>>(
+            time - start)
+                .count() >= timeBudget) {
```
kring, line 1442: This can be simplified by computing the _end_ time once outside the loop, and then just checking the current time against it inside the loop. The unlimited budget case can be handled by setting the end time to infinity or max value or a million years from now or whatever is convenient.

Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -1072,6 +1077,37 @@ bool TilesetContentManager::tileNeedsLoading(const Tile& tile) const noexcept {
          anyRasterOverlaysNeedLoading(tile);
 }
 
+void TilesetContentManager::tickMainThreadLoading(
+    double timeBudget,
+    const TilesetOptions& tilesetOptions) {
+  CESIUM_TRACE("TilesetContentManager::tickMainThreadLoading");
+
+  // A budget of 0.0 indicates that all ready tiles should finish loading.
+  if (timeBudget == 0.0) {
```
kring, line 1086: Same here, no need to handle 0.0 as a separate loop.

Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -1072,6 +1077,37 @@ bool TilesetContentManager::tileNeedsLoading(const Tile& tile) const noexcept {
...
+    for (MainThreadLoadTask& task : this->_finishLoadingQueue) {
+      finishLoading(*task.pTile, tilesetOptions);
+    }
+    return;
```
kring, line 1090: The queue isn't cleared in this case.

kring: It looks like tiles only get pushed into `_finishLoadingQueue` when the time budget is not 0.0. But in that case, this loop is unnecessary, right?

Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -1146,24 +1195,39 @@ void TilesetContentManager::updateContentLoadedState(
...
+    const TilesetOptions& tilesetOptions) {
+  // Run the main thread part of loading.
+  TileContent& content = tile.getContent();
+  TileRenderContent* pRenderContent = content.getRenderContent();
+
+  assert(pRenderContent != nullptr);
+
+  // add copyright
+  pRenderContent->setCredits(GltfUtilities::parseGltfCopyright(
+      *_externals.pCreditSystem,
```
kring, line 1214: `this->` was removed.

Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -1146,24 +1195,39 @@ void TilesetContentManager::updateContentLoadedState(
...
+      pRenderContent->getModel(),
+      tilesetOptions.showCreditsOnScreen));
 
+  void* pWorkerRenderResources = pRenderContent->getRenderResources();
+  void* pMainThreadRenderResources =
+      _externals.pPrepareRendererResources->prepareInMainThread(
```
kring, line 1220: `this->` was removed here too.

kring (commented):
Looks great @nithinp7, just a few minor changes and this should be good to merge.

Cesium3DTilesSelection/src/TilesetContentManager.cpp
nithinp7: Sorry, I was starting to get used to some other conventions from non-Cesium code. I was starting to think the underscore and "this->" together were redundant since I _thought_ an underscore implies a private member. But in fact, the accessibility is unrelated to member-ship, since private static variables are possible!

Cesium3DTilesSelection/src/Tileset.cpp
nithinp7: Good idea, it looks cleaner now! Also switched out those long, verbose std::chrono types with "auto", if that's okay.

Cesium3DTilesSelection/src/TilesetContentManager.cpp
nithinp7: Yup good point, removed it and added a clarifying comment!

Cesium3DTilesSelection/src/Tileset.cpp
nithinp7: Good idea! I was initially worried there would be some thrashing if we do that, but I think it's unfounded - since only visited tiles end up on in the finishLoadingQueue and only non-visited tiles are evicted in unloadCachedTiles.

kring:
Thanks @nithinp7!

PR #572, opened 2022-11-03
Create TMS and WMS tile providers in the main thread.

kring:
`TileMapServiceRasterOverlay` and `WebMapServiceRasterOverlay` (unlike all the other raster overlay types) were both creating their tile provider in a `thenInWorkerThread` continuation, capturing an intrusive pointer to the `RasterOverlay` in order to do this. This can trigger an assertion failure in `ReferenceCountedNonThreadSafe` when it detects an attempt to decrement the reference count from the wrong thread, and could also (rarely) cause the reference count itself to get corrupted.

The fix in this PR is to do this small amount of work in the main thread instead. If response parsing for these raster overlay types ever turns out to be a main thread bottleneck, we can refactor it into a `thenInWorkerThread` continuation that does the parsing followed by a `thenInMainThread` that creates the tile provider. But I didn't bother doing that in this PR because it's unlikely to ever be a concern.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
+- Fixed a bug that could cause an assertion failure - and on rare occasions a more serious problem - when creating a tile provider for a `TileMapServiceRasterOverlay` or a `WebMapServiceRasterOverlay`.
+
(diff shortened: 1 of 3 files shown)


j9liu:
This PR seems straightforward enough to merge. I tried it out on Cesium for Unreal, and it no longer crashes when I add either overlay component. Thanks @kring!

PR #578, opened 2022-12-01
Use snprintf instead of sprintf.

kring:
It's safer and Xcode 14 reports a warning on use of sprintf which turns into an error, breaking the build.


joseph-kaile:
Looks good

PR #579, opened 2022-12-06
Detach raster overlay tiles before freeing the geometry tiles they're attached to

kring:
This is the only remaining change to cesium-native for Unity. Background in CesiumGS/cesium-unity#123.


joseph-kaile:
Thanks @kring, makes total sense!

PR #580, opened 2022-12-07
Add optional field for custom ion endpoint in IonRasterOverlay

xuelongmu:
For an internal application, we need to be able to set a custom ion endpoint to retrieve imagery with an access token. This PR sets an optional field to do so, with the default being cesium ion, similar to how it's done in [Tileset.cpp](https://github.com/CesiumGS/cesium-native/blob/main/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h#L75)


nithinp7:
Looks good to me, thanks @xuelongmu!

PR #595, opened 2023-02-22
Use libturbo-jpeg instead of stb for faster jpg decompression

joseph-kaile:


kring:
@joseph-kaile can you please get some numbers for how much this reduces JPEG decoding time with some of our typical tilesets? You should be able to use the cesium-native tracing framework to get some stats and a chart of the time distributions. See here:
https://github.com/CesiumGS/cesium-native/pull/267

Overall what you've done here looks good, but I noticed that Unreal includes a copy of libjpeg-turbo, which means that there's the potential for conflicts (causing compiler or runtime errors) if we use our own, different version of it. It's in `C:\Program Files\Epic Games\UE_4.27\Engine\Source\ThirdParty\libjpeg-turbo` or similar.

If it's a big win, we'll surely figure it out. But right now I don't have a good sense of how big a win it is.

joseph-kaile:
Got 17,056 jpeg images from the San Diego 3d tiles, ranging from 254 kb to 1 kb, the median size is around 60 kb.

Turbo jpeg takes 31 seconds to load these images and stb_image.h takes 78 seconds.

Then profiling with Visual Studio, panning around San Diego for about two minutes:
stb
turbo

With libjpeg-turbo, the worker threads are running ReadImage 16% of the time, and with stb-image around 41% of the time.

So, with San Diego, the jpeg decoding is around 2.5 times faster.

kring:
Wow, that's a very nice improvement! Just to confirm, this was a RelWithDebInfo build, right?

joseph-kaile:
Yes, thats with RelWithDebInfo. I think the visual improvement is that there is less trickling in.

CesiumGltfReader/src/GltfReader.cpp
```diff
@@ -639,39 +640,71 @@ ImageReaderResult GltfReader::readImage(
...
-      stbi_image_free(pImage);
+      if (tjDecompress2(
+              tjInstance,
+              reinterpret_cast<const unsigned char*>(data.data()),
+              static_cast<unsigned long>(data.size()),
+              reinterpret_cast<unsigned char*>(image.pixelData.data()),
+              image.width,
+              0,
+              image.height,
+              TJPF_RGBX,
```
kring, line 668: The docs says `TJPF_RGBX` leaves the alpha component undefined. That seems like a pretty big problem, unless I'm missing something.
https://rawcdn.githack.com/libjpeg-turbo/libjpeg-turbo/1.5.x/doc/html/group___turbo_j_p_e_g.html#gac916144e26c3817ac514e64ae5d12e2a

ThirdParty.json
```diff
@@ -65,6 +65,12 @@
     "version": "0.2.10",
     "license": ["MIT"]
   },
+  {
+    "name": "libjpeg-turbo",
+    "url": "https://github.com/libjpeg-turbo/libjpeg-turbo",
+    "version": "2.1.91",
+    "license": ["BSD-3-Clause"]
```
kring, line 72: I think it's more complicated than this:
https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/LICENSE.md

If I'm reading this correctly, different parts of the code are covered by different licenses. So we should list all three: `IJG`, `BSD-3-Clause`, and `zlib`. Two of the three we're using already, and are definitely fine for us (BSD and zlib). The IJG license is an uncommon one, but seems to be BSD-ish and fine:
https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/README.ijg

kring (commented):
Looks good, just a few comments below...

I also _very briefly_ looked at using Conan for our dependencies, now that Conan 2.0 is out. The good news is that 2.0 looks greatly improved; I think trying to use it for cesium-native will be much less frustrating than it was the last time I tried to use Conan. The bad news is that not all the third party packages on Conan Center support 2.0 yet, including some that we use. Hopefully that will improve quickly.

But I _think_ we could use it just to install libjpeg-turbo in the short term, and it wouldn't be too hard to set that up. So if you have trouble getting this cmake approach working reliably, let me know, and we can give Conan a go.

CesiumGltfReader/src/GltfReader.cpp
```diff
@@ -639,39 +640,71 @@ ImageReaderResult GltfReader::readImage(
...
-        image.channels);
-    if (pImage) {
-      CESIUM_TRACE(
-          "copy image " + std::to_string(image.width) + "x" +
-          std::to_string(image.height) + "x" + std::to_string(image.channels) +
-          "x" + std::to_string(image.bytesPerChannel));
-      // std::uint8_t is not implicitly convertible to std::byte, so we must use
-      // reinterpret_cast to (safely) force the conversion.
+
+    tjhandle tjInstance = tjInitDecompress();
```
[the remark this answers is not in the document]
joseph-kaile: Its not clear but the corresponding tjDestroy is after the large if/else statement which will always execute. I don't recall seeing this in the profiler.

joseph-kaile:
I think this is ready for review. I tested this in:

Unity: https://github.com/CesiumGS/cesium-unity/pull/239
UE4: https://github.com/CesiumGS/cesium-unreal/pull/1058
UE5: https://github.com/CesiumGS/cesium-unreal/pull/1057

joseph-kaile:
FYI: this relys on https://github.com/joseph-kaile/libjpeg-turbo, however I think its not necessary to modify libjpeg-turbo after all (https://github.com/CesiumGS/cesium-native/tree/libjpeg-turbo-og). Just need to verify it works (https://github.com/CesiumGS/cesium-unreal/tree/libjpeg-turbo-ue5-og).

kring:
Ok, cool. If we still need the fork, I'll move it into the CesiumGS org before I merge this.

kring:
Looks great @joseph-kaile! Looks like there's still a problem with Unity on iOS, but I'll merge this once that is resolved.

kring:
Thanks @joseph-kaile! 🎉

PR #597, opened 2023-02-23
Avoid holes when frustum culling is disabled, even with unconditionally-refined tiles

kring:
Way back in #488, we added some logic to detect empty tiles with a geometric error greater than their parent's geometric error, and treat these specially.

If such a tile's parent's geometric error is too high, then, by definition, this tile's geometric error will be too (since it's greater), and therefore we should immediately refine past that tile.

Except! the tile's children may not be loaded yet. Prior to #488, we would make the mistake of rendering the placeholder while waiting for the children to load. This created totally unnecessary holes in the tileset, even with `forbidHoles` enabled. So the special logic added in #488 detected this case and marked the tile "unconditionally refined." And special logic in the traversal guaranteed we wouldn't select such a tile for rendering.

But during the tile content loader refactoring - I think! - this system gained a subtle bug where we accidentally treated these unconditionally-refined tiles as renderable. It didn't cause any problems in the `forbidHoles` mode, but when disabling frustum culling we would sometimes get these temporary holes in the tileset.

So this PR fixes that bug, making it so any unconditionally-refined tile - not just an external tileset - is considered unrenderable. But then this caused a problem in `forbidHoles` mode because the check `if (!child.isRenderable() && !child.isExternalContent())`  went from false to true for these unconditionally-refined tiles, skipping some necessary logic. So this PR fixes that as well.

This is fairly difficult to test well, but I did add some basic tests to confirm it is working correctly both with and without `forbidHoles`.


joseph-kaile:
In queueChildrenRequiredForForbidHoles, is the check for external content missing? I don't see it checked in isRenderable or isUnconditionallyRefined.

kring:
External tilesets are always marked as unconditionally refined:
https://github.com/CesiumGS/cesium-native/blob/6be63bf6bae3cac4afc7057a9628cb1bd44d0087/Cesium3DTilesSelection/src/TilesetContentManager.cpp#L1205

So I don't think we need to check for external tilesets explicitly.

PR #600, opened 2023-03-02
Fix missing b3dm property values

### LINKED ISSUE #598: Assertion failure / crash when a batch table property has too few values
When a property in a JSON batch table has too few values, we treat it as a string:

```cpp
  if (propertyValue.size() == 0 || propertyValue.size() < featureTable.count) {
    // No property to infer the type from, so assume string.
    updateExtensionWithJsonStringProperty(
        gltf,
        classProperty,
        featureTable,
        featureTableProperty,
        propertyValue);
    return;
  }
```

But the first line of that function is:

```cpp
assert(propertyValue.size() >= featureTable.count);
```

Which clearly contradicts the condition that caused us to invoke this function in the first place, so the assertion will _always_ fail in this case. And the code that follows assumes the assertion holds, so we'll crash in release builds.

Instead, missing values should be treated as nulls, which should in turn be treated as empty strings.

Such data is not compliant with the 3D Tiles spec, but it can happen, and we shouldn't crash.

As reported here:
https://community.cesium.com/t/3dtiles-from-ion-crashing-unity/22606

joseph-kaile:
Fixes #598

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### v0.23.0 - 2023-04-01
+
+##### Fixes :wrench:
+
+- Fixed crash when a batch table property has too few values.
+
 ### v0.22.0 - 2023-03-01
 
 ##### Breaking Changes :mega:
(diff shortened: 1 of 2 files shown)


kring:
Thanks @joseph-kaile! I confirmed that this fixes the user tileset that originally prompted #598.

PR #601, opened 2023-03-02
expose the bufferStart and stride of the accesorview's buffer

ZackOfAllTrades:
* giving access to these properties would allow more flexibility to the developers
personally, I'm building a bounded volume hierarchy to do height sampling off the unity main thread, I would like to reinterperet_cast the underlying buffer and not take the overhead of creating the specific bvh::Vector3 form the glm::vec3 returned by the AccessorView.h operator[], as well as I don't want to take the overhead of two if checks, 2 additions and a multiplication for every vertex in the model

```
const T& operator[](int64_t i) const {
    if (i < 0 || i >= this->_size) {
      throw std::range_error("index out of range");
    }

    return *reinterpret_cast<const T*>(
        this->_pData + i * this->_stride + this->_offset);
  }
```


CesiumGltf/include/CesiumGltf/AccessorView.h
```diff
@@ -191,6 +191,21 @@ template <class T> class AccessorView final {
    */
   int64_t size() const noexcept { return this->_size; }
 
+  /**
+   * @brief Returns the stride (number of bytes per component) of this accessor.
```
kring, line 195: I don't think "number of bytes per component" is quite accurate. It's actually the number of bytes from one value to the next, which can be greater than the number of bytes of each element if other elements are interleaved. So how about:

```suggestion
   * @brief Returns the stride of this accessor, which is the number of bytes from the start of one element to the start of the next.
```

CesiumGltf/include/CesiumGltf/AccessorView.h
```diff
@@ -191,6 +191,21 @@ template <class T> class AccessorView final {
...
+   *
+   * @returns The stride.
+   */
+  int64_t stride() const noexcept { return this->_stride; }
+
+  /**
+   * @brief Returns the start of this AccesorViews data into the buffer.
```
kring, line 202:
```suggestion
   * @brief Returns a pointer to the first byte of this accessor view's data. The elements are stored contiguously, so the next one starts {@link stride} bytes later.
```

CesiumGltf/include/CesiumGltf/AccessorView.h
```diff
@@ -191,6 +191,21 @@ template <class T> class AccessorView final {
...
+   *
+   * @returns The start of this view.
+   */
+  const std::byte* bufferStart() const noexcept {
```
kring, line 206: Let's call this `data` to be consistent with typical STL terminology.

```suggestion
  const std::byte* data() const noexcept {
```

kring (commented):
Thanks for the PR @ZackOfAllTrades! Some minor suggestions below. Also:

* Please add a note about this change in CHANGES.md. It should go in Additions, so copy one of those from a previous version section.
* Please run clang-format on the modified source file. The easiest way to do this to run install node.js and then run `npm install && npm run format` in the repo's root directory.

ZackOfAllTrades:
Thank you for the feedback! Will you please check the CHANGES.md, im still not quite clear on how to document the change. I'll continue to work through the build errors. Thank you for your patience!!

kring:
Thanks @ZackOfAllTrades!

PR #604, opened 2023-03-10
Fix priority inversion in main thread tile loading

### LINKED ISSUE #602: Priority inversion in main thread tile loading
The main thread portion of tile loading is throttled so that it won't tank the frame rate. But this loading is managed with a single queue, rather than the low, medium, and high priority queues used by worker-thread loading. Unfortunately, this leads to a priority inversion in some cases, where low-priority work (such as loading ancestor tiles we're not planning to render anytime soon) can block high priority work (such as rendering the currently-visible level of detail, or loading an external tileset so that we can start loading all its children). This makes load times significantly longer than they need to be.

A quick-hack fix of this bug made Melbourne load 2+ times faster in Cesium for Unity! See the `priority-inversion` branches in cesium-native and cesium-unity.

I noticed this while benchmarking CesiumGS/cesium-unity#239. A reload of the Melbourne tileset in the default view with a warm cache would take 4-6 seconds. And during that time, there were 1-2 periods of almost a second each where it seemed to be doing... nothing! No new tiles loaded by a worker thread. Maybe it would render one additional tile after the pause, but not some big chunk of new data that would justify the pause. Turns out that during this apparent pause it was busily doing main-thread load work for a bunch of low priority tiles we didn't actually need right then, while the selection algorithm waited patiently for one or two high priority tiles that were near the end of the queue.

kring:
Fixes #602 
This is a PR into #599 so merge that first.

Measuring load times for the Melbourne Photogrammetry tileset, in the default view in 02_CesiumMelbourne in Cesium for Unity Samples. Looking at the time from when the tileset is created until the last tile that will be selected _is_ selected. In other words, when the `Rendered` count stops changing. I pressed "Refresh Tileset" 5 times, waiting for it to finish refining in between each, then did it 10 more times and recorded the time for each. So the cache was warmed up in all cases.

Before this PR:

| *Refresh Number* | *Time (ms)* |
|---------------------|----------|
| 1 | 5602 |
| 2 | 6166 |
| 3 | 7007 |
| 4 | 6938 |
| 5 | 6217 |
| 6 | 8885 |
| 7 | 5161 |
| 8 | 6649 |
| 9 | 6769 |
| 10 | 6577 |
| Min | 5161 |
| Max | 8885 |
| Median | 6613 |
| Median reloads per hour | 544 |

After this PR:

| *Refresh Number* | *Time (ms)* |
|---------------------|----------|
| 1 | 1869 |
| 2 | 1742 |
| 3 | 2999 |
| 4 | 3055 |
| 5 | 2139 |
| 6 | 2091 |
| 7 | 2207 |
| 8 | 3224 |
| 9 | 2302 |
| 10 | 1911 |
| Min | 1742 |
| Max | 3224 |
| Median | 2173  |
| Median reloads per hour | 1657 |

So based on the median times, this branch is over 200% faster!

Note that the times are kind of all over the place in both branches. I don't know why there is so much variability, and I'm a bit concerned about it. Still, the trend is obvious!


kring:
Here is a similar timing for Cesium World Terrain in 01_CesiumWorld.

Before this PR:

| *Refresh Number* | *Time (ms)* |
|---------------------|--------------|
| 1 | 985 |
| 2 | 966 |
| 3 | 952 |
| 4 | 1090 |
| 5 | 1002 |
| 6 | 1103 |
| 7 | 920 |
| 8 | 1035 |
| 9 | 936 |
| 10 | 953 |
| Min | 920 |
| Max | 1103 |
| Median | 975.5 |
| Median reloads per hour |  3690 |

After this PR:

| *Refresh Number* | *Time (ms)* |
|---------------------|--------------|
| 1 | 849 |
| 2 | 998 |
| 3 | 810 |
| 4 | 882 |
| 5 | 869 |
| 6 | 919 |
| 7 | 801 |
| 8 | 855 |
| 9 | 822 |
| 10 | 948 |
| Min | 801 |
| Max | 998 |
| Median | 862 |
| Median reloads per hour | 4176 |

A more modest improvement here, about 13% faster.

kring:
The trouble with inserting them in sorted order in the first place is that it makes the "kick" mechanism much harder (or at least slower) to implement. When all the tiles for a subtree are at the end of the load queue, it's easy and cheap to remove them all.

javagl:
I'm waaay out of the code (and wouldn't even claim to have been "in" the code at any point in time), and only vaguely remember the "kick" mechanism. But even if the tile load tasks have to be maintained in a linear list in order to make the `kick` easy and efficient, one could consider options to _avoid_ the `sort` call. 

Specifically: When there's a strong invariant that the queue is always sorted, then a 'sorted insert' of new elements could be implemented trivially, and done in O(n) time. (Of course, that "sorted insert" would not be implemented on a `vector`, but a `linked list` - an indexed insert for a vector has O(n) on its own, whereas it's O(1) for a linked list).

Whether or not that brings any benefit is hard to say, and depends on all the obvious factors: How many elements are in the queue, how many are added, how many are removed (individually or 'kicked' in bulk). But it might be worth considering, at least.

kring:
To make the kick easy, the list has to be maintained in _traversal_ order (i.e. depth-first, pre-order). But for loading, they need to be in _priority_ order (which is very different from traversal order). I get what you're saying about inserting in sorted order, and that being more efficient (possible cache-thrashing of linked list data structures aside). I just haven't quite seen a way to do it efficiently yet given the need for the two different orderings for different purposes.

javagl:
OK, beyond that, I could only brainstorm about this on an even higher level, at least as long as I don't refresh my memories. (That _would_ start with looking at the "kick" thing, and how that can operate on a list in traversal order, when the actual list is _sorted_ occasionally, thus "destroying" any other order that it might have had). This is not going to happen soon, but maybe I can at least allocate a chunk of time to update https://github.com/javagl/cesium-cpp to work with the latest state, and do some "smaller" benchmarks at some point...

joseph-kaile:
I created a test to time Tileset Refresh and here are the results:

(Melbourne -- No Priority Inversion, Forbid holes, preload ancestors, preload siblings)

4.976 

4.824

4.756

4.787


4.817

-------------------------------------------------------

(Melbourne -- Forbid holes, preload ancestors, preload siblings)

5.034

5.093

4.939

5.093

4.982

-------------------------------------------------------

(Melbourne -- No Priority Inversion, No Forbid holes, No preload ancestors, No preload siblings)


2.979

2.995

3.089


-------------------------------------------------------

(Melbourne -- No Forbid holes, No preload ancestors, No preload siblings)

3.015

2.970

3.194

-------------------------------------------------------

This is the scene that it loads:

HighresScreenshot00000


And the script to time the Tileset refresh:

```
// Fill out your copyright notice in the Description page of Project Settings.

#pragma once

#include "CoreMinimal.h"
#include "GameFramework/Actor.h"
#include "Cesium3DTileset.h"
#include "MyActor.generated.h"

UCLASS()
class MYPROJECT_API AMyActor : public AActor
{
	GENERATED_BODY()
	
public:	
	// Sets default values for this actor's properties
	AMyActor();

	UPROPERTY(EditAnywhere)
		ACesium3DTileset* Tileset;

protected:
	// Called when the game starts or when spawned
	virtual void BeginPlay() override;

	UFUNCTION()
	void ProcessLoad();

	int NumRefreshes = 0;

	float AccumulatedTime = 0;

	float NextLoadCountDown = 10.0f;

	bool FirstLoad = true;

	const int32 TotalNumRefreshes = 10;

	FTimerHandle NextLoadHandle;

	float CurrentTimeElapsed = 0;

	UFUNCTION()
		void OnCountDown();

};


// Fill out your copyright notice in the Description page of Project Settings.


#include "MyActor.h"

// Sets default values
AMyActor::AMyActor()
{
 	// Set this actor to call Tick() every frame.  You can turn this off to improve performance if you don't need it.
	PrimaryActorTick.bCanEverTick = true;

}

// Called when the game starts or when spawned
void AMyActor::BeginPlay()
{
	Super::BeginPlay();
	Tileset->MaximumCachedBytes = 0;
	FScriptDelegate delegate;
	delegate.BindUFunction(this, FName("ProcessLoad"));
	Tileset->OnTilesetLoaded.Add(delegate);

	OnCountDown();
}
void AMyActor::ProcessLoad()
{
	CurrentTimeElapsed = GetWorldTimerManager().GetTimerElapsed(NextLoadHandle);

	UE_LOG(LogTemp, Warning, TEXT("Time elapsed: %f"), CurrentTimeElapsed);
}

void AMyActor::OnCountDown()
{

	if (FirstLoad) {
		FirstLoad = false;
		GetWorld()->GetTimerManager().ClearTimer(NextLoadHandle);
		GetWorld()->GetTimerManager().SetTimer(NextLoadHandle, this, &AMyActor::OnCountDown, 10.0f, true);
		Tileset->RefreshTileset();
	}
	else {
		AccumulatedTime += CurrentTimeElapsed;
		CurrentTimeElapsed = 0;

		if (NumRefreshes++ == TotalNumRefreshes) {

			float AverageTime = AccumulatedTime / float(TotalNumRefreshes);
			UE_LOG(LogTemp, Warning, TEXT("Average Time: %f"), AverageTime);
			Tileset->Destroy();
		}
		else {
			GetWorld()->GetTimerManager().ClearTimer(NextLoadHandle);
			GetWorld()->GetTimerManager().SetTimer(NextLoadHandle, this, &AMyActor::OnCountDown, NextLoadCountDown, true);
			Tileset->RefreshTileset();
		}
	}
}
```

PR #548, opened 2022-08-23
Final merge of tile content refactoring into main

kring:
This has all already been reviewed, this PR is just the final merge.

PR #606, opened 2023-03-13
Fix compile error when tracing is enabled, disable trace tracks for raster overlays

kring:
cesium-native didn't compile with tracing enabled (see #267 for details of how to set that up), so this PR fixes that.

It also disables the use of tracks for raster overlay loading. The idea behind a "track" is to visualize each of our loading slots as a virtual thread, so we can see the breakdown of the entire loading process even as it progresses across multiple threads. But we lost the ability to do this for tileset content loading in #548. And for raster overlays, we still had it in theory, but it didn't work right after #325 because multiple overlay tiles could end up loaded in parallel in a single track when multiple tiles from the raster overlay service overlapped a single terrain/geometry tile.

Rather than fixing all that, this PR just removes the use of tracks entirely. So at least we can see what work the threads are doing and how long it takes, which is still very useful.

### DIFF
--- a/Cesium3DTilesSelection/src/RasterOverlayCollection.cpp
+++ b/Cesium3DTilesSelection/src/RasterOverlayCollection.cpp
@@ -47,7 +47,7 @@ RasterOverlayCollection::~RasterOverlayCollection() noexcept {
 
 void RasterOverlayCollection::add(
     const CesiumUtility::IntrusivePointer<RasterOverlay>& pOverlay) {
-  CESIUM_TRACE_USE_TRACK_SET(this->_loadingSlots);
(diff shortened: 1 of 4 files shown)


joseph-kaile:
In Cesium Unreal 5.1, I am getting a crash during shutdown. Then when I open it again, during startup, it crashes in alloc.c.

kring:
@joseph-kaile as we discussed offline, this is probably caused by specifying `CESIUM_TRACE` for the cesium-native CMake build, but not in cesium-unreal's CesiumRuntime.build.cs. If that doesn't work, let me know!

joseph-kaile:
Yes, that was it. I opened the trace file and it outputted a helpful file.

PR #607, opened 2023-03-14
Improve support for KTX2 and WebP textures.

kring:
As originally reported here:
https://community.cesium.com/t/compressed-texture-support-webp-ktx-in-unity/22681

Cesium for Unity didn't support KTX2 or WebP compressed textures. I initially found this surprising, because Cesium for Unreal does, and cesium-native does the decoding, so why wouldn't Cesium for Unity? But it turns out the client (e.g. Cesium for Unity) still had to be aware of the extension and check it for the `source` property in order to identify which (already decoded) image to use. This was a bit silly, because by the time the client sees it, cesium-native has already decoded the image to raw pixels (or a supported compressed texture format if requested) and a KTX2 or WebP image isn't meaningfully different from a JPEG or PNG.

So this PR makes cesium-native propagate the extension's `source` to the main Texture's `source` after decoding, so that clients don't need to be aware of all the supported texture formats. This brings KTX2 and WebP support to Unity and allows us to add new formats in the future without any changes to clients.


j9liu:
@kring looks good, I'll merge once CI passes. thanks!

PR #611, opened 2023-03-23
Fix a bug causing holes in some tilesets, like San Francisco.

kring:
This fixes a bug that causes holes to sometimes appear when flying around the San Francisco tileset, even with "Forbid Holes" enabled.

Unfortunately, I don't know exactly why it happens. The immediate cause is that the selection algorithm is choosing to "render" an external tileset Tile, which is not actually renderable because it has no content. It does this because it can't render the tileset's children, because they're not loaded yet. This situation is supposed to be prevented by the "Forbid Holes" mode's requirement that all sibling tiles - even culled ones - be loaded before any of them can be rendered. This should be further enforced by the "kick" mechanism, which should "un-select" a subtree that doesn't meet this requirement. But in this case, the kicking doesn't (and can't) happen because some of the tiles in the subtree were already rendered in the last frame. This is the part that has me stumped - I haven't been able to figure out why it happens! I don't know how an external tileset Tile would ever get marked rendered in the first place.

But what I do know is that these tiles that were "rendered" in the last frame weren't actually renderable yet, and therefore they weren't "really" rendered after all. So this PR essentially says "yeah, previously-rendered tiles need to block kicking (otherwise, detail would disappear from one frame to the next), but that's actually true only for tiles that _actually_ rendered something." So when traversing, we now only set the `anyWereRenderedLastFrame` flag when the tile was marked as rendered last frame, _and_ it actually had something to render.

In short, this feels like a bit of a band-aid over another problem, but it's not an unreasonable change, and it shouldn't have any downsides.

### DIFF
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
@@ -442,7 +442,7 @@ private:
       TileLoadPriorityGroup priorityGroup,
       double priority);
 
-  TraversalDetails createTraversalDetailsForSingleTile(
+  static TraversalDetails createTraversalDetailsForSingleTile(
       const FrameState& frameState,
       const Tile& tile,
       const TileSelectionState& lastFrameSelectionState);
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
@@ -442,6 +442,11 @@ private:
       TileLoadPriorityGroup priorityGroup,
       double priority);
 
+  TraversalDetails createTraversalDetailsForSingleTile(
+      const FrameState& frameState,
+      const Tile& tile,
+      const TileSelectionState& lastFrameSelectionState);
+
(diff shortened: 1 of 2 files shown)


joseph-kaile:
@kring , I think I'm still seeing holes (UE 5.1). I'm going to check again tomorrow morning to make sure.

kring:
Yeah me too @joseph-kaile! They're not totally gone. But rarer, I think.

joseph-kaile:
The new function `createTraversalDetailsForSingleTile` is used in `_visitTileIfNeeded`, `_renderLeaf`,  `_renderInnerTile`. Should it also be used in `_kickDescendantsAndRenderTile`?

joseph-kaile:
I have one question about the Mark Tile as unrendered functions. I don't see where the tile is actually being marked. Is it supposed to modify the tile selection state? Also another question is the kick condition. It looks like it will kick if any of it's children are unrenderable, which makes sense. But also it will only kick if all of its children were not rendered last frame. That means it won't kick if at least one tile in the sub tree was rendered last frame, and so unready tiles could still be rendered. Getting rid of the second condition gets rid of the holes, however maybe the second condition is true because of a bug in Mark Tile as unrendered function?

PR #613, opened 2023-03-23
Add ITileExcluder::startNewFrame

kring:
This gives a tile excluder an opportunity to do once-per-frame updates, such as computing a new transformation if a tileset moved.


j9liu (approved):
@kring just noticed a typo, otherwise this looks good!

PR #621, opened 2023-04-13
Check for implicit root tile if tileset is 1.1 format

### LINKED ISSUE #527: Implicit tiling is no longer an extension in 3D Tiles 1.1
It is harmless and potentially useful to continue supporting the `3DTILES_implicit_tiling` extension. But in 3D Tiles 1.1 `implicitTiling` is an optional property in the `Tile` schema itself, so we should support tilesets that use that new property.

joseph-kaile:
Fixes #527


Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -489,34 +489,43 @@ std::optional<Tile> parseTileJsonRecursively(
...
-      tile.setRefine(tileRefine);
-
-      parseImplicitTileset(
-          implicitExtensionIt->value,
-          contentUri,
-          tile,
-          currentLoader);
-
-      return tile;
+  const rapidjson::Value* implicitExtensionJson = nullptr;
```
j9liu, line 492: Nitpick but can this be renamed:

```suggestion
  const rapidjson::Value* implicitTilingJson = nullptr;
```

that way, we reinforce that the `implicitTiling` object is no longer an extension.

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -489,34 +489,43 @@ std::optional<Tile> parseTileJsonRecursively(
...
+  const auto implicitTilingIt = tileJson.FindMember("implicitTiling");
+  if (implicitTilingIt != tileJson.MemberEnd() &&
+      implicitTilingIt->value.IsObject()) {
+    implicitExtensionJson = &implicitTilingIt->value;
+  }
+  if (!implicitExtensionJson) {
```
j9liu, line 498: You can consolidate to an else statement:
```suggestion
  }
  else {
```

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -489,34 +489,43 @@ std::optional<Tile> parseTileJsonRecursively(
...
+    const auto extensionIt = tileJson.FindMember("extensions");
+    if (extensionIt != tileJson.MemberEnd()) {
+      // this is an external tile pointing to an implicit tileset
```
j9liu, line 501: Can you move this comment to the first if statement (the one checking for `implicitTiling`)? Then here, the comment should indicate that this is the legacy 3D Tiles Next implicit tiling extension.

j9liu (commented):
@joseph-kaile this looks good from a glance, but could use some unit tests to confirm that the changes work.

I see that in `Cesium3DTilesSelection/test/data` there's an ImplicitTileset that uses the extension. Can you do the following:
- [x] Create a copy of `tileset.json` that uses `implicitTiling` instead of the legacy extension. Name this according to the convention that we use in CesiumJS's specs: the legacy version is called `tileset_1.0.json`, and the 1.1 version is `tileset_1.1.json`. 
- [ ] Make sure all of the existing unit tests use `tileset_1.1.json` instead.
- [ ] Add a backwards compatibility test for `tileset_1.0.json`. The legacy test can be really simple, but we just want to make sure it parses correctly.

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -488,35 +488,41 @@ std::optional<Tile> parseTileJsonRecursively(
     }
   }
 
-  // determine if tile points to implicit tiling extension
-  const auto extensionIt = tileJson.FindMember("extensions");
-  bool hasImplicitContent = false;
-  if (extensionIt != tileJson.MemberEnd()) {
+  // determine if tile is an implicit tile
```
j9liu, line 491:
```suggestion
  // determine if tile has implicit tiling
```

Cesium3DTilesSelection/test/data/ImplicitTileset/tileset_1.1.json
```diff
@@ -0,0 +1,34 @@
+{
+  "asset": {
+    "version": "1.0"
```
j9liu, line 3:
```suggestion
    "version": "1.1"
```

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -488,35 +488,41 @@ std::optional<Tile> parseTileJsonRecursively(
...
-      parseImplicitTileset(
-          implicitExtensionIt->value,
-          contentUri,
-          tile,
-          currentLoader);
-
-      return tile;
+    implicitTilingJson = &implicitTilingIt->value;
+  }
+  if (!implicitTilingJson) {
```
j9liu, line 499: Changing this to an `else` works for me -- can you double-check if it works for you?

j9liu (commented):
I see two other implicit tilesets in here that need to be changed, so I'll push a commit with those changes + the ones I note in my feedback here.

PR #622, opened 2023-04-17
Handling when query string is included in WMS url

tomlee-dev:
Hello.

I found that some wms service needs key in url.

for example.

https://api.wms.example.com/req/wms?KEY={MY-PRIVATE-KEY}&DOMAIN={MY-OWN-DOMAIN}

This changes is for if wms url already has query string(?) in url string, then replace with & instead of ? in following parameter string.

please review this PR.

Thanks!


kring:
Thanks @tomlee-dev!

tomlee-dev:
@kring you're welcome! my pleasure!

PR #626, opened 2023-04-22
Enable position-independent code for libjpeg-turbo (fix #625)

### LINKED ISSUE #625: cesium-native can't be linked into shared library on Linux
The reason is that the recently added libjpeg-turbo is built without `-fPIC`

TheMostDiligent:


TheMostDiligent:
> Would it make sense to set the PIC property in the external project based on the current one, as we do with other variables?

Totally, I updated the branch.

kring:
Thanks @TheMostDiligent!

PR #627, opened 2023-04-26
Fix ninja build for libjpeg-turbo

pjanetzek:
Hello,
this change fixes the build when using ninja. otherwise it doesn't find the target for
`cesium-native/extern/libjpeg-turbo/RelWithDebInfo/libturbojpeg.a` automatically.  
The solution was suggested here https://cmake.org/pipermail/cmake/2015-November/062055.html


kring:
Thanks for the PR @pjanetzek!
@joseph-kaile can you please review and merge it.

joseph-kaile:
@pjanetzek Thank you. I confirmed that it fixes the Ninja error.

PR #640, opened 2023-05-15
Add classes to convert between C++ and `EXT_structural_metadata` property types

j9liu:
This PR adds `EXT_structural_metadata` counterparts to the `PropertyType.h` and `PropertyTypeTraits.h` files under anew files and a "StructuralMetadata" namespace. The idea is that once everything is done, the old files will be deleted and the new files will be renamed.

@kring It's a pretty monotonous PR but I'd love to know if there's any way to make this more... nice? Templated? If at all possible. Especially with the `vecN` and `matN` permutations. I also was incorrect, there _are_ integer `matN`s and my Visual Studio Intellisense wasn't 100% working 🙃 

There's a Visual Studio warning in `TestStructuralMetadataPropertyTypeTraits` that the function takes up a lot of memory on the stack -- should I address this?


kring:
Looks good!

PR #642, opened 2023-05-16
Use partial template specialization to reduce code duplication

kring:
This is a PR into #640.


j9liu:
Thank you for the help @kring !

PR #643, opened 2023-05-17
Add `StructuralMetadataPropertyView`

j9liu:
This PR creates a `StructuralMetadata` counterpart to `MetadataPropertyView`. This handles `vecN` and `matN` properties in addition to the existing ones. There are slight rewrites of the documentation and internal code for clarity.


kring:
Looks good! Just one minor suggestion that can be implemented separately. Given all the other assertions in `MetadataPropertyView::get`, it seems sensible to also assert that `index < size()`.

Issue #248, opened 2021-05-14
Publish cesium-native to conan and vcpkg


lilleyse:
Once cesium-native reaches some level of stability we should consider publishing it to conan and vcpkg like how we publish CesiumJS to npm.

lilleyse:
First step would be using conan for third party libraries. When not possible we could either find alternative libraries or publish third party libraries on those libraries' behalf.

Here's the initial conan PR: https://github.com/CesiumGS/cesium-native/pull/261

Library|In Conan|Notes
--|--|--
asyncplusplus|:heavy_check_mark:
Catch2|:heavy_check_mark:
cmake-modules|:x:
cpp-httplib|:heavy_check_mark:
CSPRNG|:x:
draco|:heavy_check_mark:
earcut|:heavy_check_mark:
glm|:heavy_check_mark:
GSL|:heavy_check_mark:
modp_b64|:x:|Alternative: https://conan.io/center/base64
PicoSHA2|:x:
rapidjson|:heavy_check_mark:|Use `cci.20211112` or newer
s2geometry|:x:|Look at CesiumJS which has no external dependency
spdlog|:heavy_check_mark:
sqlite3|:heavy_check_mark:
stb|:heavy_check_mark:
tinyxml2|:heavy_check_mark:
uriparser|:heavy_check_mark:

kring:
After having it on the back-burner for awhile, we're starting to look into this again, because an important customer wants to consume cesium-native via Conan. After spending lots of time reading the Conan documentation and looking at various examples, I have more questions than answers. It's crazy complicated. But here are the goals I'm thinking of for cesium-native Conan integration:

1. Each cesium-native library (CesiumGltf, CesiumGeometry, Cesium3DTiles, etc.) is a separate Conan package with its own dependencies on other cesium-native packages and on third-party packages.
2. Conan is used to install dependencies, but is not strictly required. You could install the dependencies another way, put them in your CMake module path, and just run the cesium-native build.
3. Easy development workflow, just hit build after making changes. Ideally using the built-in CMake support for Visual Studio and Visual Studio Code.
4. Minimal repetition, e.g. dependencies stated in only one place.
5. Flexibility to build cesium-native with weird settings, such as "Debug builds use the Release MSVCRT" as required by Unreal Engine. Third party libraries must use these weird settings, too. Which will require building them from source.
[...]

PR #644, opened 2023-05-17
Migrating dependencies to vcpkg

jherico:
This PR is a work in progress to address #248 by moving the dependencies to a packaging system, in this case [VCPKG](https://vcpkg.io/).  The intent is to make the project fully buildable using vcpkg dependencies and also to make it easily publishable to vcpkg as well.

The current work migrates 
* glm
* uriparser
* earcut
* Catch2
* GSL
* tinyxml2
* async++
* spdlog
* rapidjson
* stb
* httplib
* PicoSHA2
* KTX
* libwebp
* libmorton
* expected-lite
* libjpeg-turbo

Remaining libraries are
* s2geometry
* CSPRNG

The project builds on Windows x64, but has not been tested on other platforms.  The test suite runs with some failures and those might be related to API changes between the version of Catch2 originally in use and the one currently in vcpkg.  I'm investigating further.


kring:
Thanks for the PR @jherico, really excited to see this! To be able to merge it, there are some details we'll need to sort out:

1. How will this impact our downstream projects, like [Cesium for Unreal](https://github.com/CesiumGS/cesium-unreal), [Cesium for Unity](https://github.com/CesiumGS/cesium-unity), and [Cesium for Omniverse](https://github.com/CesiumGS/cesium-omniverse)? 
2. How will it impact third-party applications using cesium-native, such as [vsgCs](https://github.com/timoore/vsgCs)?
3. We at Cesium have at least one internal project that uses cesium-native, and also uses Conan. Is that likely to present difficulties?
4. What does vcpkg support look like on non-Windows platforms? Are we likely to have any particular difficulties?

Thanks again for exploring this, @jherico!

jherico:
@kring 

> How will this impact our downstream projects

If you're talking about licensing, that shouldn't be an issue because I'm using the same upstream libraries.  If you're talking about build processes, it should be fairly painless. My goal is to have the new VCPKG based CMake build be a drop-in replacement so that a `cmake --build` would produce basically the same thing that the current branch does.  The biggest difference is that while the current main branch creates install targets for lots of the dependency sub-projects, in the VCPKG world those shared libraries will be in the VCPKG `bin` folder (or `debug/bin` folder for debug builds) and the installer mechanism must account for that when gathering binary dependencies.  I'll look into required syntax for doing so.  Alternatively, it would be possible to have cesium-native forcibly require static building and linking for all upstream dependencies.  This has the advantage of preventing conflicts with downstream users that might want to use different versions of the same dependencies.

> How will it impact third-party applications using cesium-native

As mentioned above, downstream projects should be able to continue building from source as they were.  They'd also have the option of using a VCPKG managed version of the dependency.

> We at Cesium have at least one internal project that uses cesium-native, and also uses Conan. Is that likely to present difficulties.

I don't believe there's any problem posed by this.  Even if you build cesium-native from source and it is using VCPKG for dependency management, the resulting output should be the same, so from the perspective of a downstream project using Conan there wouldn't be any difference.

> What does vcpkg support look like on non-Windows platforms? Are we likely to have any particular difficulties?

VCPKG has [first-party support](https://github.com/microsoft/vcpkg/tree/master/triplets) for x64 Linux, OSX and Windows as well as Windows and UWP for some 32-bit platforms.  It has additional community support for [lots more platforms](https://github.com/microsoft/vcpkg/tree/master/triplets/community), including Android, iOS, etc.

You can browse the available packages [here](https://vcpkg.io/en/packages.html) and see that they're all annotated with exactly which of the first-party platforms are supported by each.

I actually got into using vcpkg because I was working on an open source project with lots of dependencies, including OpenSSL and historically building that on Windows has been something of a hellscape.  VCPKG smoothed out a lot of the platform specific snarls and made dependency management a lot easier for me.  

My only complaint was managing vcpkg manually, which is why I created a CMake file to allow one to just declare a bunch of VCPKG dependencies directly at the beginning of your root CMakeLists.txt and have them automatically install (it downloads and boostraps vcpkg if necessary).  This approach was also useful because I put the VCPKG folder outside of the build directory, meaning that multiple different checkouts could share the same set of external dependencies without having to constantly rebuild them, even if I told the project to "clean".  Eventually if you accept this approach, I could also work on updating the Travis configuration so that it caches the VCPKG build folders between runs, meaning you'd get drastically reduced build times.

kring:
Thanks for the answers!

> The biggest difference is that while the current main branch creates install targets for lots of the dependency sub-projects, in the VCPKG world those shared libraries will be in the VCPKG bin folder (or debug/bin folder for debug builds) and the installer mechanism must account for that when gathering binary dependencies.

Just as a heads up, most of our downstream products consume cesium-native as a pile of static libraries. It's not currently set up well for dynamic linking at all, actually. In particular, it's missing the necessary dllexports on Windows.

> VCPKG has [first-party support](https://github.com/microsoft/vcpkg/tree/master/triplets) for x64 Linux, OSX and Windows as well as Windows and UWP for some 32-bit platforms. It has additional community support for [lots more platforms](https://github.com/microsoft/vcpkg/tree/master/triplets/community), including Android, iOS, etc.

We have some unusual platforms in the game world, such as Android on AMD/Intel x64 for the Magic Leap 2. I don't think it will be a problem, though, as long as vcpkg can (cross-)compile libraries from source when binaries are not available.



Bigger picture, though... I should mention that Cesium has _tentatively_ been heading down a Conan road. I'm not sure to what extent we've considered vcpkg as an alternative. But if you have some familiarity with Conan as well, perhaps you could share some thoughts on why vcpkg may be a better choice?

jherico:
> as long as vcpkg can (cross-)compile libraries from source when binaries are not available.

vcpkg _only_ builds from source. As far as I know it doesn't support binary caching at all.

>  perhaps you could share some thoughts on why vcpkg may be a better choice?

I'll give it some thought and reply at length later.

jherico:
@kring I guess my main motivation for preferring VCPKG over Conan is that it feels more self-contained.  I remember experiencing this before when I was looking at package managers before adopting VCPKG.  Specifically, reading the Conan [introductory](https://docs.conan.io/2/introduction.html) material makes it feel like there are a lot of moving parts... a client, a server, a central repository and an artifact server.  Additionally, their recommended installation for Conan is to use `pip`, which means you have to deal with the whole Python ecosystem (what if Conan has python dependencies you don't want on your global installation.. do you have to use a virtualenv just to run Conan?).  I'm also not clear on the path you need to take if you want to customize the dependencies you're getting through Conan.  If you want some obscure tool or some weird version of common dependency and you know it's not going to get integrated into the main Conan repository, I don't know what the path is (I'm assuming there IS one).

By comparison, with VCPKG, the only thing you need is the ability to clone _a_ VCPKG repository and to have basic build tools installed on your system... xcode, VS, gcc, etc.  That's it.  VCPKG will bootstrap itself by downloading a binary if it can, but by building from source if it can't.  There's no binary hosting server because VCPKG always works by building from source (strictly speaking you _could_ have a VCPKG port that was created by fetching binaries, but I don't know of anything in the central repository that works like that).

I particularly like how easy it is to customize your dependencies.  If you need something that isn't covered by the VCPKG main repo, you just fork their repository and make changes to your fork.  If you're adding new ports that are likely to get accepted upstream, then you can concurrently work on submitting those ports, but you're not blocked on continuing your own project while waiting for them to be accepted.  

That all said, I haven't really spent much time interacting with Conan, so it's entirely possible I'm missing some key information about ease of use, but again, just from the introduction, it still looks "more complicated to use" than VCPKG, particularly when using my CMake VCPKG wrapping functionality.

kring:
Thanks @jherico, really appreciate the writeup!

kring:
Thanks for your work on this @jherico. Just curious: did you run into problems that made it difficult to continue? Hopefully I didn't inadvertently discourage you from working on it!

jherico:
@kring I mostly didn't have time to work on it with my day job, and didn't want to saddle you guys with a dead PR if I couldn't be responsive to changes.  I've since worked a little bit on my branch and almost have the CI passing without issues.  Once I've done that I may open a new PR for this.

kring:
Thanks for all your work on this @jherico!

jherico:
Thank you @kring .  I'm currently trying to get a PR added to s2geometry to fix the windows build problems so it can be sourced from VCPKG as well.

PR #645, opened 2023-05-18
Automatically unzip data if gzipped data is detected

### LINKED ISSUE #634: Automatically detect gzipped payloads and ungzip them
Many 3D Tiles and quantized-mesh generation tools gzip each file (e.g., tileset.json, tile.b3dm) by default. These gzipped files cause problems for cesium-native applications:

1. If the files are coming from a web server, that web server must be configured to include `Content-Encoding: gzip` in the response header, or else cesium-native will not be able to understand it.
2. If the files are loaded locally, there's no way to add a response header. Instead, the user has to manually gunzip all the files.

This comes up surprisingly often on the community forum. Someone tries to load some tileset, often from the local filesystem, and it doesn't work. The error message is generic, because all cesium-native knows is that the content makes no sense to it. So the user posts to the forum, we ask them to share their tileset with us, they do, we download it, hopefully notice rather quickly that the files are gzipped, tell the user this, and then they have to do one of the two things above to work around it.

So all in all, this is a hassle for users and also ends up taking up a lot of our time doing support for it. But we could easily avoid the whole thing just by looking for the magic bytes at the start of the file that indicate gzip, and automatically ungzipping the payload. Or, at the very least, we could tell the user their files are gzipped so they don't have to go to the community forum to find out.

joseph-kaile:
Fixes #634

Unreal: https://github.com/CesiumGS/cesium-unreal/pull/1097
Unity: https://github.com/CesiumGS/cesium-unity/pull/323

### DIFF
--- a/CesiumAsync/src/GunzipAssetAccessor.cpp
+++ b/CesiumAsync/src/GunzipAssetAccessor.cpp
@@ -5,11 +5,14 @@
 #include "CesiumUtility/Gunzip.h"
 
 namespace CesiumAsync {
+
+namespace {
+
 class GunzippedAssetResponse : public IAssetResponse {
 public:
   GunzippedAssetResponse(const IAssetResponse* pOther) noexcept
       : _pAssetResponse{pOther} {
-    _dataValid = CesiumUtility::gunzip(
+    this->_dataValid = CesiumUtility::gunzip(
         this->_pAssetResponse->data(),
         this->_gunzippedData);
   }
@@ -27,8 +30,8 @@ public:
   }
 
   virtual gsl::span<const std::byte> data() const noexcept override {
-    return _dataValid ? gsl::span(_gunzippedData.data(), _gunzippedData.size())
-                      : _pAssetResponse->data();
+    return this->_dataValid ? this->_gunzippedData
+                            : this->_pAssetResponse->data();
   }
 
 private:
@@ -39,8 +42,9 @@ private:
 
 class GunzippedAssetRequest : public IAssetRequest {
 public:
-  GunzippedAssetRequest(std::shared_ptr<IAssetRequest>& pOther)
-      : _pAssetRequest(pOther), _AssetResponse(pOther->response()){};
+  GunzippedAssetRequest(std::shared_ptr<IAssetRequest>&& pOther)
+      : _pAssetRequest(std::move(pOther)),
+        _AssetResponse(_pAssetRequest->response()){};
   virtual const std::string& method() const noexcept override {
     return this->_pAssetRequest->method();
   }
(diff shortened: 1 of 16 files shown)


kring:
This looks great @joseph-kaile! Just one suggestion: let's make the gunzip capability into a separate asset accessor decorator rather than building it into the CacheAssetAccessor decorator. There's no reason those two orthogonal behaviors - caching and gunzipping - need to inhabit the same class. With a separate `GunzipAssetAccessor`, it's easy to compose an accessor with all the behaviors we need, like `GunzipAssetAccessor(CachingAssetAccessor(UnrealAssetAccessor))`.

CesiumAsync/src/GunzipAssetAccessor.cpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+Future<std::shared_ptr<IAssetRequest>> GunzipAssetAccessor::request(
+    const AsyncSystem& asyncSystem,
+    const std::string& verb,
+    const std::string& url,
+    const std::vector<THeader>& headers,
+    const gsl::span<const std::byte>& contentPayload) {
+  return this->_pAssetAccessor
+      ->request(asyncSystem, verb, url, headers, contentPayload);
```
kring, line 96: This one should gunzip the response as well. Unless there's some particular difficulty there that I haven't thought of? It should be almost identical to `get`, I think.

CesiumAsync/src/GunzipAssetAccessor.cpp
```diff
@@ -0,0 +1,100 @@
...
+    const std::string& url,
+    const std::vector<THeader>& headers) {
+  return this->_pAssetAccessor->get(asyncSystem, url, headers)
+      .thenImmediately([](std::shared_ptr<IAssetRequest>&& pCompletedRequest) {
+        const IAssetResponse* pResponse = pCompletedRequest->response();
+        if (!pResponse) {
+          return std::move(pCompletedRequest);
+        }
+        if (CesiumUtility::isGzip(pResponse->data())) {
+          pCompletedRequest = std::make_shared<GunzippedAssetRequest>(
```
kring, line 82: gunzip is pretty fast, so maybe it doesn't matter... but the `thenImmediately` here means that the actual gunzip operation may be run in any old thread. I think it would make sense to delegate the gunzipping operation to an actual worker. But you don't want to switch to a worker for the `isGzip`, so you can't just change the `thenImmediately` to a `thenInWorkerThread`. Instead, leave it as a `thenImmediately` and change this section to something like this (untested):

```cpp
if (CesiumUtility::isGzip(pResponse->data())) {
  return asyncSystem.runInWorkerThread([pCompletedRequest]() mutable {
    return std::make_shared<GunzippedAssetRequest>(
GunzippedAssetRequest(std::move(pCompletedRequest)));
  });
}
```

CesiumAsync/src/GunzipAssetAccessor.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+private:
+  const IAssetResponse* _pAssetResponse;
+  std::vector<std::byte> _gunzippedData;
+  bool _dataValid;
+};
+
+class GunzippedAssetRequest : public IAssetRequest {
+public:
+  GunzippedAssetRequest(std::shared_ptr<IAssetRequest>& pOther)
```
kring, line 42: This should be `&&`, and the place that calls it will need a std::move.

kring (commented):
This is looking great! Just a couple more small things.

kring:
Thanks @joseph-kaile!

PR #646, opened 2023-05-18
Always log warning/errors

joseph-kaile:


kring:
Thanks @joseph-kaile! A couple of changes I'd like to see here:

* It should log the warnings even if there are no errors. Currently all the logging is wrapped in `if (result.errors)`, so it'll only happen if there's at least one error.
* The logging should happen before the `loadErrorCallback` is invoked. I think that will help minimize confusion, especially if the callback fixes something. We could end up reporting an error that has already been fixed.
* There's very similar code to this in `RasterOverlayCollection.cpp`. It should also be changed to log warnings and errors before invoking the loadErrorCallback.

kring:
Thanks @joseph-kaile!

Issue #650, opened 2023-05-23
Allow a way to specify an external sqlite3 library


keyboardspecialist:
Currently cesium-native uses its own internal sqlite3 lib. We would like a way to specify our own version from sources such as Conan, so all of our dependencies are aligned across each project.

keyboardspecialist:
To add more context:

PRIVATE_CESIUM_SQLITE was added to the cmake build to namespace against Unreal's include sqlite. However, this doesn't protect the windows asset-pipeline build. Even though it's linked as private the build pulls it in as a public dependency on citygmlswayze causing linker conflicts.

Pointing CesiumAsync to the Conan sqlite is a viable solution though we need to ensure that the cesium-sqlite.h header still works as we expect.

Windows asset-pipeline builds should do something like: set PRIVATE_CESIUM_SQLITE to OFF, link against the included conan lib, and add the internal sqlite3 dir to the CesiumAsync header dirs for cesium-sqlite.h.

kring:
> However, this doesn't protect internal projects from having linker conflicts.

Are you sure? It should. When that is defined, every symbol in SQLite gets renamed.

You'll end up with two copies of SQLite, which may not be ideal, but it should _work_.

azrogers:
It seems like this issue was resolved, so I will close the issue.

PR #651, opened 2023-05-23
Add support for using external sqlite3 libraries, issue #650

keyboardspecialist:
sqlite3 is checked to see if its aliased, if so, don't attempt the install. Otherwise, we install the packaged version in extern/sqlite3

In cases where we are aliased, we need to include the extern/sqlite3 dir in our include dir list.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -177,7 +177,14 @@ install(DIRECTORY ${CESIUM_NATIVE_SPDLOG_INCLUDE_DIR}/spdlog DESTINATION "${CMAK
 
 install(TARGETS ${CESIUM_NATIVE_DRACO_LIBRARY})
 
-install(TARGETS sqlite3)
+#if we're aliased then someone else has a sqlite3 lib for us to use, otherwise install our own
+get_target_property(_aliased sqlite3 ALIASED_TARGET)
+if(NOT _aliased)
+    message("sqlite3 not aliased, using ours in extern/sqlite3")
+    install(TARGETS sqlite3)
+else()
+    message("sqlite3 aliased as ${_aliased}")
+endif()
 
 install(TARGETS modp_b64)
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -177,7 +177,14 @@ install(DIRECTORY ${CESIUM_NATIVE_SPDLOG_INCLUDE_DIR}/spdlog DESTINATION "${CMAK
 
 install(TARGETS ${CESIUM_NATIVE_DRACO_LIBRARY})
 
-install(TARGETS sqlite3)
+#if we're aliased then someone else has a sqlite3 lib for us to use, otherwise install our own
+get_target_property(_aliased sqlite3 ALIASED_TARGET)
+if(NOT _aliased)
+    message("sqlite3 not aliased, using ours in extern/sqlite3")
```
kring, line 183: I don't think we need a message in this normal case.

CesiumAsync/CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ target_include_directories(
         ${CMAKE_CURRENT_LIST_DIR}/include
     PRIVATE
         ${CMAKE_CURRENT_LIST_DIR}/src/
+        ${CMAKE_CURRENT_LIST_DIR}/../extern/sqlite3/
```
kring, line 38: I'm not clear on why this is needed. Shouldn't the dependency on the (aliased) sqlite3 via `target_link_library` automatically add sqlite3's include directory? Just like it does in the non-aliased case?

CMakeLists.txt
kring: I'm curious, too: why is it a problem to install the targets from an aliased sqlite3? Why does sqlite3 needs this special code, when other libraries used by both cesium-native and another project (such as Draco?) don't need this?

CesiumAsync/CMakeLists.txt
keyboardspecialist: It was a side effect of skipping the install command. Once it was using the aliased version it was no longer looking into the extern/sqlite directory. Only the conan version.

keyboardspecialist:
We can hold off on this until a cleaner solution can be found. I like the idea of only building with one sqlite, but we can get by with the private flag for now.

I will close this for now.

PR #652, opened 2023-05-24
Add a method to compute a transformation between LocalHorizontalCoordinateSystem instances

kring:
This is useful in switching Cesium for Unreal to use this new cesium-native class.

Also implemented a small optimization to the construction of `LocalHorizontalCoordinateSystem` (fewer matrix multiplications). And marked the various methods `noexcept`.


j9liu:
Looks good to me @kring , thanks!

PR #654, opened 2023-05-24
Upgrade `EXT_feature_metadata` to `EXT_structural_metadata`

### LINKED ISSUE #407: Support EXT_mesh_features
@kring I haven't fully dived into the spec differences between this and EXT_feature_metadata yet, but just to take inventory of the things that need to be done:
* Add the EXT_mesh_features extension
* Refactor upgradeBatchTableToFeatureMetadata so that it actually upgrades to mesh_features instead.
* On the Unreal side, refactor FCesiumMetadataPrimitive (maybe just the constructor?) Will the in-memory representation of primitive metadata be the same, do the actual structures and metadata access api need any change? 
* Add FCesiumMetadataTexture (or some such name) and an api to access this on the cpu.
* Styling? Is this just for metadata textures or for per-vertex metadata as well? 
  * For metadata textures: can we send the texture itself straight to Unreal with a NEAREST filter, then in the fragment shader sample the feature id for that texel, and then do a color lookup in the fragment shader based on the id? Can the fragment shader part be deferred to the user, who presumably knows more about the tileset and the desired id-based coloring?
  * For per-vertex metadata styling, we'll need to hack a way to send feature ids as vertex data, probably sneak it through as a UV coordinate.

Does this list sound about right? Are there any huge pieces I'm missing?

j9liu:
Closes #407. This PR is a **DRAFT** that I'm opening to track the remaining tasks on the staging branch.

When complete, this PR will handle metadata based on `EXT_structural_metadata`, the 3D Tiles 1.1 metadata extension. ~`EXT_feature_metadata` should still be parsed, but it will be internally converted to `EXT_structural_metadata`~ Batch tables in `.b3dm` and `.pnts` tiles will be converted to property tables in `EXT_structural_metadata`. Feature IDs will be indicated by `EXT_mesh_features`.

`EXT_feature_metadata` will no longer be supported.

**Progress**
- Store batch IDs in `EXT_mesh_features`
- Add `EXT_structural_metadata` counterparts to underlying metadata retrieval implementation
- [x] #655 
- [x] #658 
- [x] #668 
- [x] #669 
- [x] #667 
- [x] #713 
- [x] #716
- [x] #719 

Additionally, Cesium for Unreal and Cesium for Unity's metadata systems should be updated to account for differences in the implementation.

Observations
- `schemaUri` doesn't seem to be handled for `EXT_feature_metadata`, and `EXT_structural_metadata` requires either it or `schema` to be present. Support should be added in a future PR
- Property views on `ENUM` types are not supported


kring:
Amazing work @j9liu!

PR #655, opened 2023-05-25
Convert batch tables to `EXT_structural_metadata`

j9liu:
This PR rewrites how batch tables in `b3dm` and `pnts` formats are parsed. They are now stored as a property table in a glTF with the `EXT_mesh_features` and `EXT_structural_metadata` extensions. The tests have also been rewritten to accommodate this difference.

### DIFF
--- a/Cesium3DTilesSelection/src/BatchTableToGltfStructuralMetadata.cpp
+++ b/Cesium3DTilesSelection/src/BatchTableToGltfStructuralMetadata.cpp
@@ -163,15 +163,15 @@ struct CompatibleTypes {
    * Merges a MaskedArrayType into this CompatibleTypes.
    */
   void operator&=(const MaskedArrayType& maskedArrayType) {
-    if (!isArray()) {
-      makeIncompatible();
+    if (isArray()) {
+      MaskedArrayType& arrayType = std::get<MaskedArrayType>(maskedType);
+      arrayType &= maskedArrayType;
       return;
     }
 
-    MaskedArrayType* pMaskedArrayType =
-        std::get_if<MaskedArrayType>(&maskedType);
-    if (pMaskedArrayType) {
-      *pMaskedArrayType &= maskedArrayType;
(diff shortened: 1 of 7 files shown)


Cesium3DTilesSelection/src/BatchTableToGltfStructuralMetadata.cpp
```diff
@@ -50,21 +236,71 @@ struct BinaryProperty {
...
+                 FLOAT32,
+             sizeof(float)}},
+        {"DOUBLE",
+         GltfPropertyTableComponentType{
+             ExtensionExtStructuralMetadataClassProperty::ComponentType::
+                 FLOAT64,
+             sizeof(double)}},
 };
 
 int64_t roundUp(int64_t num, int64_t multiple) noexcept {
```
j9liu, line 306: Technically `EXT_structural_metadata` property tables don't need to be aligned to 8-bytes. Quote:

> Implementation note: Authoring tools may choose to align all buffer views to 8-byte boundaries for consistency, but client implementations should only depend on 8-byte alignment for buffer views containing 64-bit component types.

 Should I keep this rounding code? Or okay to remove?

kring: There's still a 4-byte alignment requirement though, right? That might be required of all bufferViews in general.

kring: It's a bit more nuanced than that, actually:
https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#data-alignment

j9liu: Ah right, I forgot about that. We can just leave the rounding behavior at 8-bytes for simplicity.

kring:
Thanks @j9liu!

PR #656, opened 2023-05-30
Add views for property textures and their properties

j9liu:
This adds `EXT_structural_metadata` counterparts of `FeatureTextureView` and `FeatureTexturePropertyView`. It also renames the `PropertyTableView` and `PropertyTablePropertyView` classes, and their error enums, to be consistent with the property texture naming scheme.

I kept most of the implementation of `FeatureTextureView` and `FeatureTexturePropertyView` because I'm not familiar with how it interfaces with Unreal, and I don't want to make potentially incompatible changes. But I'm not happy with the inconsistencies between property table / property texture implementations. For instance,
- `PropertyTableView` doesn't validate all of its properties on construction, but `PropertyTextureView` does.
- `PropertyTextureView` is marked invalid if one of its properties is invalid, but `PropertyTableView` does not do the same.
- `PropertyTextureProperty` is in charge of validating its own elements, but `PropertyTableView` handles all of that and simply passes the status to `PropertyTablePropertyView`.
 
Perhaps these inconsistencies are necessary / not that bad, but I'd like to revisit them once I switch Cesium for Unreal to this implementation, so that they are more standardized.

### DIFF
--- a/Cesium3DTilesSelection/src/B3dmToGltfConverter.cpp
+++ b/Cesium3DTilesSelection/src/B3dmToGltfConverter.cpp
@@ -176,7 +176,7 @@ rapidjson::Document parseFeatureTableJsonData(
   return document;
 }
 
-void convertB3dmMetadataToGltfFeatureMetadata(
+void convertB3dmMetadataToGltfStructuralMetadata(
     const gsl::span<const std::byte>& b3dmBinary,
     const B3dmHeader& header,
     uint32_t headerLength,
@@ -246,7 +246,7 @@ GltfConverterResult B3dmToGltfConverter::convert(
     return result;
   }
 
-  convertB3dmMetadataToGltfFeatureMetadata(
+  convertB3dmMetadataToGltfStructuralMetadata(
       b3dmBinary,
       header,
       headerLength,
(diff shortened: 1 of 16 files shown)


j9liu:
Hm, not sure why the formatting check is failing... when I use "Format Document" in Visual Studio in the mentioned files, it doesn't change anything?

kring:
Hmm is clang-format enabled in VS on your new computer?
image

kring:
Also, formatting code in cesium-native while some other solution (like cesium-unreal) is open, probably won't work. It won't necessarily find the right .clang-format file. You're better off opening the cesium-native repo (or the cesium-unreal extern directory) with VS Code and doing the formatting there.

j9liu:
My VS settings look exactly like yours. But I got the formatting to work by cutting and pasting the entire file in place. Very strange :shrug:

Anyways, hopefully CI accepts the formatting and makes this reviewable :')

j9liu:
Converting to draft until #662 is merged.

j9liu:
@kring this is ready for review, when you get the chance :pray:

kring:
This all looks reasonable to me, thanks @j9liu!

PR #668, opened 2023-06-16
Rename generated `EXT_feature_metadata` and `EXT_mesh_features` files

j9liu:
This PR does the following:

1. Prefixes the existing `EXT_feature_metadata` files with `ExtensionExtFeatureMetadata`.
2. Removes the `ExtensionExtMeshFeatures` prefix from the feature ID / feature ID texture classes.

PR #669, opened 2023-06-16
Rename structural metadata files

j9liu:
Following #668, this PR removes the `ExtensionExtStructuralMetadata` prefix from the generated classes files. This should make the code read more neatly in the new metadata implementations.


joseph-kaile:
Looks good. The renaming amounts to around 1,000 less characters, and thus less time to write code.

PR #675, opened 2023-06-27
include generated headers in doxygen documentation

### LINKED ISSUE #672: documentation missing multiple headers
Going through the documentation that doxygen builds, at least a few classes in the `CesiumGltf` namespace are missing, but the headers exist and have doxygen-style comments. Notably `Model` and `NamedObject` (which do not appear to be generated by `generate-classes`) are in the docs, but `Mesh` and `MeshPrimitive` are not (which are generated). Is it possible that headers generated by `generate-classes` are not making their way into some doxygen related config?

mattelser:
fixes #672. @kring pointed out that these missing headers just need to be added to the documentation `CMakeLists.txt`. Looking for other generated headers I found the following dirs:
```sh
Cesium3DTiles/generated/include
Cesium3DTilesReader/generated/src
Cesium3DTilesWriter/generated/src
CesiumGltf/generated/include
CesiumGltfReader/generated/src
CesiumGltfWriter/generated/src
```
Looking through the `src` dirs I didn't see anything with doxygen-style comments so I only added the two with `include` subdirs.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -4,6 +4,7 @@ if (DOXYGEN_FOUND)
     set(
         LIB_DIRS
         ../Cesium3DTiles/include
+        ../Cesium3DTiles/generated/include
         ../Cesium3DTilesReader/include
         ../Cesium3DTilesWriter/include
         ../Cesium3DTilesSelection/include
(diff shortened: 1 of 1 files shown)


kring:
Thanks @mattelser!

PR #678, opened 2023-06-29
Rework `PropertyTexturePropertyView`

j9liu:
`PropertyTexturePropertyView` used to return a fixed 4-component array no matter what the property's actual type was. You would manually check which components were valid, and only `uint8_t` component types were allowed.

This has all been reworked by making `PropertyTexturePropertyView` a templated property view, similar to `PropertyTablePropertyView`. The values it returns (assuming the view is valid) are exactly the type as specified by the schema. There are definitely unsupported types, but this makes it easier to explicitly indicate which ones are unsupported.

`PropertyTextureView` no longer stores a map of these `PropertyTexturePropertyView`s. Now it is similar to `PropertyTableView`, where you can call `getPropertyView` and `forEachProperty` to get individual or all property views respectively.

This also includes changes for `FeatureIdTextureView`, to match some changes made in `PropertyTexturePropertyView`.


j9liu:
This is ready for a review. I struggled with figuring out how to best handle unsupported properties -- `PropertyTexturePropertyView` only supports a small subset of the types that `PropertyTablePropertyView` supports. My initial thought was that someone could construct a `PropertyTexturePropertyView<bool>` and it would compile, but return an invalid view. But this requires implementing fallback return values for unsupported types, which feels wrong?

So I cleaned up `PropertyTextureView` to only handle the actually supported types, similar to how `PropertyTableView` is a bunch of `if constexpr`s expressions with no fallback. This means calling `getPropertyView<T>` with an unsupported `T` will not make code compile, because of the unresolved template. But I think this more strongly reinforces which types are supported by the view.

CesiumGltf/src/PropertyTextureView.cpp
```diff
@@ -78,4 +44,71 @@ PropertyTextureView::getClassProperty(const std::string& propertyName) const {
 
   return &propertyIter->second;
 }
+
+PropertyTexturePropertyViewStatus PropertyTextureView::getTextureSafe(
+    const int32_t textureIndex,
+    int32_t& samplerIndex,
+    int32_t& imageIndex) const noexcept {
```
kring, line 51: This _might_ be a good use for `nonstd::expected`. So the signature of the function would be:

```cpp
nonstd::expected<std::reference_wrapper<Texture>, PropertyTexturePropertyViewStatus> getTextureSafe(const int32_t textureIndex) const noexcept;
```

The idea is that the function returns either a `Texture` reference (on success) or `PropertyTexturePropertyViewStatus` (on error) indicating the reason for failure.

That reference_wrapper is a bit unfortunate, though. It's needed because expected-lite apparently doesn't allow the success type to be a reference. It could be a `Texture*` instead, but it will never be null. I dunno, maybe this is too fancy and a bad idea, but I thought I'd mention it because using the Either monad for error handling is a cool technique if we don't use it here!

kring (commented):
Looks great! A couple comments, but nothing that needs to hold up merging it.

PR #682, opened 2023-07-07
Fix handling of KTX2 images without mipmaps.

kring:
Fixed a bug where `GltfReader::readImage` would always populate `mipPositions` when reading KTX2 images, even when the KTX2 file indicated that it had no mip levels that and they should be created, if necessary, from the base image. As a result, `generateMipMaps` wouldn't generate any mipmaps for the image.

Fixes CesiumGS/cesium-unity#274


j9liu:
@kring I just noticed two typos, so I'll push a commit for those. But otherwise, looks good to me!  🐶 

(posted this on the Unity thread by accident)

PR #683, opened 2023-07-09
Don't let zlib dirty its repo on configure

kring:
On configure, zlib renames `zconf.h` to `zconf.h.included` for no good reason, causing git to report its repo as dirty. The dirty repo then propagates up through cesium-native into whatever projects are using it via submodule, which is really annoying. There are a [bunch of issues](https://github.com/madler/zlib/issues?q=is%3Aissue+is%3Aopen+zconf.h.included) written for this, along with multiple PRs to fix it in different ways, but seemingly no interest from the maintainers.

So the hacky approach I used here is to simply rename it back after the zlib `add_subdirectory`. Since the rename was pointless to begin with, this doesn't do any harm and avoids a dirty repo.

### DIFF
--- a/CesiumUtility/CMakeLists.txt
+++ b/CesiumUtility/CMakeLists.txt
@@ -35,8 +35,8 @@ target_include_directories(
         ${CESIUM_NATIVE_RAPIDJSON_INCLUDE_DIR}
     PRIVATE
         ${CMAKE_CURRENT_LIST_DIR}/src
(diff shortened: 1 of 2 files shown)


kring:
Well, my original approach didn't work, because compilers apparently check the current directory for quoted includes before checking anywhere else. So now I've implemented another approach: copy the zlib source into the build dir and let it modify that copy.

PR #686, opened 2023-07-17
Treat fully-loaded, empty, unconditionally-refined tiles with no children as renderable

kring:
The Google Maps Photorealistic 3D Tiles contain some _leaf_ tiles with a) no content, and b) a geometric error larger than their parent. This is unusual and not very useful, but it's not invalid. But it exposes a bug in cesium-native's selection algorithm.

Ignoring the "no children" bit, such tiles are intentionally treated quite specially. Rendering them would cause a hole in the tileset, so we very carefully avoid doing so, even if none of the tile's children are loaded yet. To that end, these tiles are marked as unrenderable. Not just "there's nothing to render" (which is also true), but "we're not alowed to render these." Not being allowed to render such a tile _also_ prevents us from rendering the tile's siblings under many circumstances, because otherwise by simply refining the tile's parent we would be creating a hole. This is intended to be a temporary situation. Once the tile's children finish loading, we can render _them_, and therefore we can refine the unconditionally-refined tile's parent.

But of course this all falls apart if an unconditionally-refined tile has no children. In that case, the unconditionally-refined tile prevents rendering of its subtree (as intended), but there are no children to eventually load and allow the subtree to be rendered.

Note that we have to be careful to distinguish between "this unconditionally-refined tile has no children _yet_", which is common when the tile is an external tileset, and "this unconditionally-refined tile will _never_ have children."

So this PR resolves this by treating tiles that are unconditionally-refined, completely done loading, and have no children as renderable. They'll render as "nothing" because they have no content, but that's fine. It allows the rest of the subtree to render.

I started looking into this because of user reports of missing tiles or entire tilesets when recording videos with the Google tileset. I was able to reproduce it, and this was the cause. In this PR, I also tweaked the `updateViewOffline` method to be more certain that all loading has completed. I'm not sure if this part is strictly necessary, but it should be safer and not do any harm.


j9liu:
Looks good, thanks @kring !

PR #688, opened 2023-07-20
Remove credits argument

javagl:
The `showCreditsOnScreen` argument was passed all the way down through the call chains in `LayerJsonTerrainLoader`, and apparently, it was not used.

Opening this as a DRAFT, because I don't know whether...
- the fact that it was passed on was unintentional
- the fact that it was not used was unintentional
- (Maybe I overlooked some place where it was used, but if this is the case, then there was no test that took this flag into account - all tests still seem to be passing...)

Maybe someone wants to have a look...

### DIFF
--- a/Cesium3DTilesSelection/src/CesiumIonTilesetLoader.cpp
+++ b/Cesium3DTilesSelection/src/CesiumIonTilesetLoader.cpp
@@ -167,8 +167,7 @@ mainThreadLoadLayerJsonFromAssetEndpoint(
              externals,
              contentOptions,
              url,
-             requestHeaders,
-             showCreditsOnScreen)
+             requestHeaders)
       .thenImmediately([credits = std::move(credits),
                         requestHeaders,
                         ionAssetID,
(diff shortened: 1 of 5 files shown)


javagl:
Well, I'll not look closer into build errors that say
`The command "choco install doxygen.install --version=1.9.2" failed and exited with 148 during .` (sic!)
unless someone confirms that this is worth pursuing....

kring:
Thanks @javagl. It's clearly not used, so it clearly won't do any harm to merge this, and I'm happy to do so. As for how we got here, I looked briefly and I'm still not sure. I think it was originally added here:
https://github.com/CesiumGS/cesium-native/commit/62fb4984988c86bf0e43011970a73b7f94ac20d3

And at some point we gained the ability to force all credits to show on the screen. So I would guess that's where it became unused, but I wasn't able to pin down the smoking commit.

In any case, mark this ready for review and I'll merge it!

javagl:
The likelihood for something like this could be decreased when people felt that itchy urge to add a...
```
/**
 * ...
 * @showCreditsOnScreen This flag causes... oh...  wait a minute... it doesn't cause anything...
 */
```
everywhere (and even more so when looong lambdas with looong capture lists were actually _methods_). But that may involve judgement calls here and there.

kring:
Thanks @javagl!

PR #689, opened 2023-07-20
Add support for Meshopt compression

### LINKED ISSUE #685: Add support for EXT_meshopt_compression
Gltf with EXT_meshopt_compression cannot be loaded in UE, as mesh data compressed through EXT_meshopt_compression by gltfpack. However CesiumJS has a meshopt_decoder module to decode the compressed data in buffer, so same Gltf can be loaded in CesiumJS. 

Is there any plan to add this kind decoder?

joseph-kaile:
Fixes  #685


CesiumGltfReader/src/decodeQuantized.cpp
```diff
@@ -0,0 +1,119 @@
...
+} // namespace
+
+void CesiumGltfReader::decodeQuantized(Model& model) {
+  for (Mesh& mesh : model.meshes) {
+    for (MeshPrimitive& primitive : mesh.primitives) {
+      for (pair<const string, int32_t>& attribute : primitive.attributes) {
+        Accessor* pAccessor =
+            Model::getSafe(&model.accessors, attribute.second);
+        if (pAccessor) {
+          decodeAccessor(model, *pAccessor);
```
kring, line 113: I don't think it's appropriate to decode _every_ integer accessor. Only the ones with the specific semantics and accessor types listed in the spec.

CesiumGltfReader/src/decodeQuantized.cpp
```diff
@@ -0,0 +1,119 @@
...
+void unquantizeFloat(float* fPtr, xVec<T, N>& quantizedView) {
+  for (int i = 0; i < quantizedView.size(); i++) {
+    const auto& q = quantizedView[i];
+    for (int j = 0; j < q.length(); j++) {
+      *fPtr++ = intToFloat<T>(q[j]);
+    }
+  }
+}
+
+template <typename T> float intToFloat(T t) { return intToFloat(t); }
```
kring, line 28: These conversion are all assuming that the attribute is normalized, which is not necessarily true. Non-normalized attributes should simply be cast to float, i.e., integer 2 becomes 2.0f.

kring: BTW, if you ever called `intToFloat` with an unsupported type, it would cause a stack overflow due to endless recursion. You can turn this into a compiler error instread by using template specialization. Something like this (untested):

```
template <typename T> float intToFloat(T c);
template <> float intToFloat<int8_t>(int8_t c) { return max(c / 127.0f, -1.0f); }
// ...
```

CesiumGltfReader/src/decodeQuantized.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+namespace CesiumGltf {
+struct Model;
+}
+
+namespace CesiumGltfReader {
+
+void decodeQuantized(CesiumGltf::Model& model);
```
kring, line 9: It wouldn't hurt to be a little more descriptive here. `decodeQuantizedMeshExtension`, maybe.

CesiumGltfReader/src/transformTexture.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+namespace CesiumGltf {
+struct Model;
+}
+
+namespace CesiumGltfReader {
+
+void transformTexture(CesiumGltf::Model& model);
```
kring, line 9: Same here, `decodeTextureTransformExtension`.

CesiumGltfReader/src/transformTexture.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+void CesiumGltfReader::transformTexture(Model& model) {
+  for (Mesh& mesh : model.meshes) {
+    for (MeshPrimitive& primitive : mesh.primitives) {
+      Material* pMaterial =
+          Model::getSafe(&model.materials, primitive.material);
+      if (pMaterial) {
+        if (pMaterial->pbrMetallicRoughness) {
+          std::optional<TextureInfo> textureInfo =
+              pMaterial->pbrMetallicRoughness->baseColorTexture;
```
kring, line 87: There are other `TextureInfo` instances beyond `baseColorTexture`. Those should be handled as well.

CesiumGltfReader/src/transformTexture.cpp
```diff
@@ -0,0 +1,96 @@
...
+      *uvs++ = glm::vec2((matrix * glm::vec3(accessorView[i], 1)));
+    }
+  }
+}
+
+void processTextureInfo(
+    Model& model,
+    MeshPrimitive& primitive,
+    TextureInfo& textureInfo) {
+  ExtensionKhrTextureTransform* textureTransform =
```
kring, line 48:
```suggestion
  ExtensionKhrTextureTransform* pTextureTransform =
```

CesiumGltfReader/src/transformTexture.cpp
```diff
@@ -0,0 +1,96 @@
+#include "transformTexture.h"
+
+#include "CesiumGltf/ExtensionKhrTextureTransform.h"
+#include "CesiumGltfReader/GltfReader.h"
+
+#include <CesiumGltf/AccessorView.h>
+using namespace glm;
```
kring, line 7: I think it's better to fully-qualify this one everywhere, too.

CesiumGltfReader/src/transformTexture.cpp
```diff
@@ -0,0 +1,96 @@
...
+        0,
+        0,
+        1);
+    mat3 scale = mat3(Scale.x, 0, 0, 0, Scale.y, 0, 0, 0, 1);
+    mat3 matrix = translation * rotation * scale;
+
+    vec2* uvs = reinterpret_cast<glm::vec2*>(buffer.cesium.data.data());
+
+    for (int i = 0; i < accessorView.size(); i++) {
+      *uvs++ = glm::vec2((matrix * glm::vec3(accessorView[i], 1)));
```
kring, line 39: What if these texture coordinates are used from multiple places with different transforms?

CesiumGltfReader/CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ target_include_directories(
     SYSTEM PUBLIC
         # Necessary for `draco/draco_features.h`
         ${CESIUM_NATIVE_DRACO_INCLUDE_DIR}
+        ${CESIUM_NATIVE_MESHOPTIMIZER_INCLUDE_DIR}
```
kring, line 49: Based on a quick look at meshoptimizer's CMakeLists.txt, I don't think this should be necessary. It adds it itself with `target_include_directories`.

CesiumGltfReader/include/CesiumGltfReader/GltfReader.h
```diff
@@ -101,6 +101,10 @@ struct CESIUMGLTFREADER_API GltfReaderOptions {
    */
   bool decodeDraco = true;
 
+  bool decodeQuantized = true;
```
kring, line 104: Need doc. And a more descriptive name.

kring (commented):
Good progress @joseph-kaile!

CesiumGltfReader/src/dequantizeMeshData.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+namespace CesiumGltf {
+struct Model;
+}
+
+namespace CesiumGltfReader {
+
+void dequantizeMeshData(CesiumGltf::Model& model);
```
kring, line 9: Add doc.

CesiumGltfReader/src/transformTextureCoords.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+namespace CesiumGltf {
+struct Model;
+}
+
+namespace CesiumGltfReader {
+
+void transformTexture(CesiumGltf::Model& model);
```
kring, line 9: Add doc. Also name it `transformTextureCoords` to match the file. Or even a more descriptive name (for both the function and file) wouldn't hurt!

CesiumGltfReader/src/transformTextureCoords.cpp
```diff
@@ -0,0 +1,119 @@
...
+  for (Mesh& mesh : model.meshes) {
+    for (MeshPrimitive& primitive : mesh.primitives) {
+      Material* pMaterial =
+          Model::getSafe(&model.materials, primitive.material);
+      if (pMaterial) {
+        if (pMaterial->pbrMetallicRoughness) {
+          processTextureInfo(
+              model,
+              primitive,
+              pMaterial->pbrMetallicRoughness->baseColorTexture);
```
kring, line 113: Looks like this is still only handling `baseColorTexture`. There are several other textures that it applies to.

CesiumGltfReader/src/transformTextureCoords.cpp
```diff
@@ -0,0 +1,119 @@
...
+    return;
+  }
+  const BufferView* pBufferView =
+      Model::getSafe(&model.bufferViews, pAccessor->bufferView);
+  if (!pBufferView) {
+    return;
+  }
+  Accessor& accessor = model.accessors.emplace_back(*pAccessor);
+  Buffer& buffer = model.buffers.emplace_back();
+  buffer.cesium.data.resize(static_cast<size_t>(pBufferView->byteLength));
```
kring, line 91: Move these two lines inside the `if` below.

CesiumGltfReader/test/data/DucksMeshopt/README.md
```diff
@@ -0,0 +1,7 @@
+# GLTF Model for Testing
+
+This repository contains the Duck GLTF model that is used for testing various features and scenarios. The model is taken from the [glTF-Sample-Assets](https://github.com/KhronosGroup/glTF-Sample-Assets) repository, which provides a set of sample models for testing and demonstrating the glTF format.
+
+## License Information
+
+The Duck model is in the public domain and can be used for any purpose. The source is [here](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/Duck).
```
kring, line 7: The license on the linked page definitely doesn't say it's in the public domain! 😬

kring (commented):
Getting close now! Just a few more small things to fix it up.

joseph-kaile:
@kring, Ready for review.

kring:
This is looking great @joseph-kaile! I just committed a few really tiny tweaks that weren't really worth commenting on.

I think it would be really useful to flesh out the tests, beyond the basic "can it load without reporting any warnings?" test that you already have. Maybe just verify that the meshes have the expected attributes and that their values fall in the expected range? I'm tempted to merge this even without adding those tests, but it's better to wait for them. I think we can still get this into the August release if you're able to do a bit of work on the tests today.

joseph-kaile:
The test has been updated. The unreal pull request is here: https://github.com/CesiumGS/cesium-unreal/pull/1183

kring:
Thanks @joseph-kaile, the new tests look good and the feature looks great in general. Merging!

PR #690, opened 2023-07-20
Support stale-while-revalidate

### LINKED ISSUE #659: Add basic support for `stale-while-revalidate` cache control header
The new Google 3D Tiles include a cache header like this:

```
Cache-Control: private, max-age=0, stale-while-revalidate=86400, no-transform
```

cesium-native currently treats this as totally uncacheable, due to the `max-age=0`. But actually the header says "you can use the cached response for 24 hours, but you should also, in the background, revalidate it."

Implementing this exactly as the spec says is somewhat complicated, due to the need to return a result _and then also do the request_. And the value is dubious because that revalidation request competes with other requests for content, making the theoretical lower latency unlikely to really pan out in practice.

But there is one way in which this header is clearly better than "no caching at all." It allows the response to be stored in a cache and then, when we later request it again, we can include the `If-None-Match` header with the cached Etag. If they match, the server can respond with a 304 code and an empty response. This still requires a round-trip to the server (which is the bulk of the time in some environments), but doesn't require the actual data to be downloaded again. On slower connections, at least, this can be a pretty good win. And we can add support for it pretty easily, without any of the complexity of the deferred revalidation that `stale-while-revalidate` suggests.

In fact, we should be able to do this Etag-based validation in a wide variety of cache header contexts, not just `stale-while-revalidate`.

csciguy8:
Fixes and missing checks to caching logic to improve performance by enabling on disk caching for certain tilesets (Google 3D Tiles).

1. Fix cases where response headers with both `max-age=0` and an intentionally out of date `Expires` header would cause responses to never get cached on disk. Ex. Google 3D Tiles root response.
2. Fix cases where response headers with `max-age=0` and `stale-while-revalidate` would not get cached to disk, even though they are eligible with revalidation. Ex. Google 3D Tiles response.
3. Fix handling of `no-cache` directive to let response go to cache `shouldCacheRequest` and always cause a revalidate in `shouldRevalidateCache`. Previous logic did not let it go to cache at all.
4. Modify `ResponseCacheControl` class to store `stale-while-revalidate` directive. Also let it contain new methods to individually check for existence / value of certain params.
5. Add 5 new unit tests

Closes #659


CesiumAsync/src/CachingAssetAccessor.cpp
```diff
@@ -311,17 +311,21 @@ bool shouldCacheRequest(
...
 
     maxAge = cacheControl->maxAge();
+    staleWhileRevalidate = cacheControl->staleWhileRevalidate();
   }
 
-  // check response header contains expires if maxAge is not specified
-  const HttpHeaders& responseHeaders = pResponse->headers();
-  if (maxAge == 0) {
+  // Prefer cache control directives over older Expires header
+  if (maxAge != 0 || staleWhileRevalidate != 0) {
```
kring, line 325: I don't think we need to check for `staleWhileRevalidate`. Even with a `maxAge` of zero, it's still valid to _store_ the object as long as both of these are true:

1. It does _not_ have the `no-store` directive. It's fine if it has `no-cache`, because that just means you have to revalidate it each time.
2. It has an ETag or Date. In the former case we revalidate with `If-None-Match`, and in the latter we revalidate with `If-Modified-Since`.

If you haven't already seen it, this page is super helpful:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching

Basically, as long as we're not forbidden from storing it (`no-store`), and as long as we'll have _some_ way of revalidating it in the future, we should go ahead and store it in the cache because doing so is allowed and _may_ help us get a 304 from the server in the future.

CesiumAsync/src/CachingAssetAccessor.cpp
```diff
@@ -343,8 +347,14 @@ std::string calculateCacheKey(const IAssetRequest& request) {
 std::time_t calculateExpiryTime(
     const IAssetRequest& request,
     const std::optional<ResponseCacheControl>& cacheControl) {
+
+  // Prefer cache control directives over older Expires header
   if (cacheControl) {
-    if (cacheControl->maxAge() != 0) {
+    // A specified maxAge indicates exact expiration.
+    // A specified staleWhileRevalidate still indicates expiration with maxAge,
+    // even if it is 0.
```
kring, line 355: `stale-while-revalidate` doesn't actually indicate anything about expiration. It just declares a (kinda complicated) policy for what you can do with it _after_ it's expired. We can ignore that special policy for now, which really means ignoring `stale-while-revalidate` and instead revalidating immediately and always.

kring (commented):
A couple comments, which will hopefully lead more to deleting code than adding it!

CesiumAsync/src/CachingAssetAccessor.cpp
csciguy8: All good points. This first pass did just enough to get Google 3D tiles going to disk cache, but it looks prudent to go a bit further.

I've committed more changes, and good news, looks like any stale-while-revalidate logic isn't strictly needed, although I kept storage for it in `ResponseCacheControl`.

While the original problem of tiles not going to disk cache was caused by `max-age=0`, it was actually our implementation of `shouldCacheRequest` that was the problem. It would always return false for this response since the was no `Expires` header. This is point 2) in my PR description.

In addition, the root request had a slightly different issue which I noted in point 1). `max-age=0` and an old `Expires` value of `Mon, 01 Jan 1990 00:00:00 GMT`. This also caused `shouldCacheRequest` to always return false. Fixing this lets us cache our google sessionId across application runs, which I am very happy about.

Question for you... for rules that we follow in code that are backed by documentation, should we cite it in comments? For example, there is Mozilla documentation that states when to ignore the `Expires` header, which I am following.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires

CesiumAsync/src/CachingAssetAccessor.cpp
csciguy8: Agreed. I've changed this.

csciguy8:
@kring All ready for re-review, when you're ready.

csciguy8:
Here are some numbers that anticipate the real-world improvements that users could see...
@shehzan10 

**Plugin:** Cesium for Unreal
**Repro:** Open map in editor, let it sit until network access stops
**Test Map:** cesium-unreal-samples -> \Content\CesiumSamples\Maps\12\12_CesiumGoogleMapsTiles.umap

_First run:_ Disk cache empty, tile content loading from network only
  Network bytes downloaded: `14,320,916`

_Second run:_ Disk cache populated, content is revalidated, and previous content reused
  Network bytes downloaded: `308,364`

In this example, our network usage reduced by **~97.8%** when our visible tiles were already in our disk cache. In addition, tiles should appear to load faster, as there is less to download, but times will vary based on your processor and network connection.

This is a substantial win for network bandwidth in the best cases. That being said, network lag is still present in all cases, as we do have to ask the server if the content has changed.

CesiumAsync/src/CachingAssetAccessor.cpp
kring: > for rules that we follow in code that are backed by documentation, should we cite it in comments?

kring:
> This is a substantial win for network bandwidth in the best cases.

14MB -> 300KB is indeed a great improvement! That's enough of a difference that, even if you have a decent network that can download multiple megabytes per second, it may still save a few seconds. Out of curiosity, were you able to measure any reduction in total load time?

CesiumAsync/src/CachingAssetAccessor.cpp
```diff
@@ -309,23 +311,24 @@ bool shouldCacheRequest(
...
-  // check response header contains expires if maxAge is not specified
-  const HttpHeaders& responseHeaders = pResponse->headers();
-  if (maxAge == 0) {
-    HttpHeaders::const_iterator expiresHeader = responseHeaders.find("Expires");
-    if (expiresHeader == responseHeaders.end()) {
+  // Check older Expires header
+  if (!ignoreExpiresHeader) {
+    const HttpHeaders& headers = pResponse->headers();
+    HttpHeaders::const_iterator expiresHeader = headers.find("Expires");
+    if (expiresHeader == headers.end())
```
kring, line 330: I think something is missing here. What if we have these headers:

```
Cache-Control: no-cache
ETag: whatever
```

And no `Expires`.

This should be cached, and it should be revalidated with `If-None-Match`. But currently this logic will return false and it will not be stored.

Or what if we have this:

```
Cache-Control: no-cache
Last-Modified: [some date]
```

Same thing: it should be stored and revalidated by `If-Modified-Since`, but currently it wouldn't be cached at all.

Unless I'm missing something (very possible), I think the logic should be:

1. If it's `no-store`, don't store (because we're not allowed to), else...
2. If it has a `max-age` or an `Expires` that puts its expiration in the future, then store it (because it might be reusable without revalidation in the time window), else...
3. If it has an `ETag` or a `Last-Modified`, then store it (because it might be reusable with revalidation even though it's already expired)

Note that a response we previously cached for rule 2 might be still be reusable with revalidation after its expiration if it also has an `ETag` or a `Last-Modified`. But we don't need to check when deciding whether to store it.

CesiumAsync/src/CachingAssetAccessor.cpp
```diff
@@ -343,20 +346,24 @@ std::string calculateCacheKey(const IAssetRequest& request) {
...
-    return convertHttpDateToTime(expiresHeader->second);
-  }
+  // If there is a Cache-Control header with the max-age or s-maxage directive
+  // in the response, the Expires header is ignored
+  bool preferCacheControl =
+      cacheControl &&
+      (cacheControl->maxAgeExists() || cacheControl->sharedMaxAgeExists());
+
+  if (preferCacheControl) {
+    return std::time(nullptr) + cacheControl->maxAgeValue();
```
kring, line 357: Technically we should subtract the value of the `Age` header from the max-age, before adding it to the current time. There's a note about this on the [Mozilla cache-control page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control):

> Note that max-age is not the elapsed time since the response was received; it is the elapsed time since the response was generated on the origin server. So if the other cache(s) — on the network route taken by the response — store the response for 100 seconds (indicated using the Age response header field), the browser cache would deduct 100 seconds from its [freshness lifetime](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#freshness_lifetime).

Probably not urgent, but would you mind writing an issue for it?

CesiumAsync/src/ResponseCacheControl.cpp
```diff
@@ -14,17 +14,25 @@ ResponseCacheControl::ResponseCacheControl(
     bool accessControlPublic,
     bool accessControlPrivate,
     bool proxyRevalidate,
-    int maxAge,
-    int sharedMaxAge)
+    bool maxAgeExists,
+    int maxAgeValue,
```
kring, line 18: Consider using `std::optional<int>` instead of the bool/int pairing.

kring (commented):
Looking good @csciguy8! Sorry to nitpick the caching details, but I think it's worthwhile to spend a little more time to iron out some last loose ends.

csciguy8:
> > This is a substantial win for network bandwidth in the best cases.
> 
> 14MB -> 300KB is indeed a great improvement! That's enough of a difference that, even if you have a decent network that can download multiple megabytes per second, it may still save a few seconds. Out of curiosity, were you able to measure any reduction in total load time?

I didn't, but I can. Timing will vary a lot with your particular CPU and network provider, even from run to run, but I can measure something so we can have at least one data point.

CesiumAsync/src/ResponseCacheControl.cpp
csciguy8: Done

CesiumAsync/src/CachingAssetAccessor.cpp
```diff
@@ -268,13 +269,14 @@ bool shouldRevalidateCache(const CacheItem& cacheItem) {
...
-  }
+  if (cacheControl && cacheControl->noCache())
+    return true;
+
+  bool cacheIsStale = isCacheStale(cacheItem);
+  if (cacheControl && cacheIsStale && cacheControl->mustRevalidate())
+    return true;
 
-  return isCacheStale(cacheItem);
+  return cacheIsStale;
```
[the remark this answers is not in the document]
csciguy8: Agreed. I didn't see much point in this check either. Committed changes to simplify with some explanation in the comments.

CesiumAsync/src/CachingAssetAccessor.cpp
csciguy8: Committed two more unit tests to cover these no-cache examples, as well as the changed logic to cache them correctly.

CesiumAsync/src/CachingAssetAccessor.cpp
csciguy8: Done. #694

CesiumAsync/src/CachingAssetAccessor.cpp
kring: I think it covers my specific examples now, but the logic is still not quite right. I still stand by the algorithm I described above, although I should have added a step 4, "else don't store". Here's a new specific example that the current algorithm would get wrong (but this is definitely not meant to be an exhaustive inventory of cases):

```
ETag: whatever
```

No cache control, no expires. But it should still be cached because it can be revalidated by ETag later.

kring:
One more small thing I forgot to mention before: please update CHANGES.md to mention this change.

CesiumAsync/src/CachingAssetAccessor.cpp
csciguy8: >  although I should have added a step 4, "else don't store"

Ah yes, that is a non-trivial change at the end of `shouldCacheRequest`, but doable. I've added your 2 new examples as unit tests, and changed the logic to match your algorithm closer.

Thanks for tolerating this back and forth. I tend to prefer test driven development with logic like this, thus the new unit tests. But having intuitive and well thought through algorithms is important too. Hopefully we get the best of both worlds.

csciguy8:
> Thanks @csciguy8! I tweaked the `shouldCacheRequest` function just a liiiitle more, and wrote some tests to demonstrate the difference. Please check my commit to make sure I didn't miss anything, but I'm going to go ahead and merge this.

Looks sensible to me. Like the addition of even more tests. Funny how this started as supporting the `stale-while-revalidate` directive, but has very little to do with it in the end. Lots of great changes for our caching engine.

PR #692, opened 2023-07-24
Remove logs when registering magic headers of file extensions

### LINKED ISSUE #679: Hide log messages about "registering magic header" and "registering file extension"
Originally reported here:
* https://github.com/CesiumGS/cesium-unity/issues/340

These messages are not particularly useful in most cases, so we should either move them to a "verbose" logging level or get rid of them entirely.

csciguy8:
These logs appear to be informational in nature and are visible in the console when using Cesium for Unity. I've also noticed these in the unit tests output for Cesium for Unreal.

Ultimately, I decided to remove these completely. In the context our official releases, I didn't see much use. Our supported magic headers and file extensions are not configurable by a user. So really it's just verifying the call to `Cesium3DTilesSelection::registerAllTileContentTypes()`, which is informative, but isn't particularly complicated. For example, in the Cesium for Unreal plugin, this call only happens once, and always during plugin startup.

Fixes #679

### DIFF
--- a/Cesium3DTilesSelection/src/GltfConverters.cpp
+++ b/Cesium3DTilesSelection/src/GltfConverters.cpp
@@ -12,14 +12,12 @@ std::unordered_map<std::string, GltfConverters::ConverterFunction>
 void GltfConverters::registerMagic(
     const std::string& magic,
     ConverterFunction converter) {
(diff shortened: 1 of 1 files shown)


j9liu:
@csciguy8 looks fine to me, thanks!

Issue #676, opened 2023-06-28
Support for `KHR_materials_variants` in `cesium-native` and implementations


javagl:
**tl;dr** Summary:

- The `KHR_material_variants` extension allows different material configurations for glTF assets that can be selected _at runtime_
- The same could be applied to a _tileset_, to allow selecting a material variant for all its glTF tile contents
- The information about the variants would be stored in the tileset JSON (e.g. as metadata)
- Support for _processing_ this material variant information and displaying the selected variant will have to be added to the runtime engines, e.g. `cesium-unreal` and `cesium-unity`

---

### Goal

[`KHR_material_variants`](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_variants/README.md) is a ratified glTF extension that allows defining different material variants for meshes:

- The extension defines a set of "variant names" (identified by their name) at the top level of the glTF asset
- For each mesh primitive, the extension defines the _material_ that should be used for this primitive when a certain variant is selected

It should be possible to apply the concept of "material variants" to the more coarse-grained level of _tilesets_. The tileset should contain information about the variants that are expected to be present in the glTF assets that it refers to. Runtime engines should offer a mechanism to switch between these variants (i.e. the materials). 

### `cesium-native` glTF level

There already is _very_ basic support for `KHR_materials_variants` in `CesiumGltf`, as of https://github.com/CesiumGS/cesium-native/pull/630 : This PR added the classes that are auto-generated from the schema, allowing to obtain the top-level information about the available variants, and the mapping between materials and variants for each primitive.

### `cesium-native` tileset level

There could be information about the avaialable material variants _for a tileset_, that summarizes the variants that are available in the glTF assets that are used as the tile content of the tileset.

In the most simple case, there could be an array of (unique) strings that match the names of the material variants that are supposed to be contained in the glTF assets. The user should then be able to select one of these material variants, and this variant should be activated for all glTF assets. When a glTF does not contain the respective variant name, then it should be rendered with its default material. 

Some technical details still have to be sorted out. But the rough idea is to store the information about the variants as [Tileset- or Group metadata](https://github.com/CesiumGS/3d-tiles/tree/main/specification#metadata). So the information could be stored as
```
{
  "schema": {
    "classes": {
      "materialVariants": {
        "properties": {
          "material_variants": {
            "type": "STRING",
            "array": true,
            "description": "Names of material variants to be expected in the glTF assets"
          }
        }
      }
    }
  }
}
[...]

PR #693, opened 2023-07-24
Material variants support

javagl:
Addressing https://github.com/CesiumGS/cesium-native/issues/676

Maybe it is possible to split the _conceptual_ questions from some lower-level questions. I'll add a comment with some _conceptual_ questions in the issue. 

The lower-level questions: 

I tried to find the right place for extracting the (metadata) information from the tileset JSON, and pass it on to the `Tileset` class. The right place for _parsing_ it may be a bit arbitrary. Since this is currently solely operating on the `rapidjson` document for the tileset, the magic 
`MaterialVariants parseMaterialVariants(const rapidjson::Document& tileset)`
function can basically be inserted anywhere (currently in an anonymous namespace of the `TilesetContentManager`). **But**... this has to be changed for the case that the metadata `schema` is **not** contained in the tileset JSON, but referred to via a `schemaUri`. This is related the the question about the "proper way" for passing this information from the place where it is loaded (or where it has to be loaded in the future) to the `Tileset` object. Having short descriptions about what a `TilesetContentManager` actually **is** (beyond `/* A class that manages tileset content */`), or the role that a `TilesetContentLoaderResult` plays (beyond that it is `/* The result of loading tileset content */`) could be really, really, really helpful.

### DIFF
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/MaterialVariants.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/MaterialVariants.h
@@ -1,0 +1,18 @@
+#pragma once
+
+#include "Library.h"
+
+#include <string>
+#include <vector>
+
+namespace Cesium3DTilesSelection {
+
+/**
+ */
+struct CESIUM3DTILESSELECTION_API MaterialVariants {
+public:
+  std::vector<std::string> tilesetMaterialVariantNames;
+  std::vector<std::vector<std::string>> groupsMaterialVariantNames;
+};
+
+} // namespace Cesium3DTilesSelection
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "Library.h"
+#include "MaterialVariants.h";
 #include "RasterOverlayCollection.h"
 #include "Tile.h"
 #include "TilesetContentLoader.h"
@@ -153,6 +154,8 @@ public:
 
   /** @copydoc Tileset::getOverlays() */
   const RasterOverlayCollection& getOverlays() const noexcept;
+
+  const MaterialVariants& getMaterialVariants() const noexcept;
 
   /**
    * @brief Updates this view but waits for all tiles that meet sse to finish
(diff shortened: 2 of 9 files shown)


javagl:
Even though this is only an early draft, it could be good if someone (@kring !?) could have a _quick_ look whether the overall approach is not entirely wrong:

- There is a plain struct 
  ```
  struct MaterialVariants {
    std::vector<std::string> tilesetMaterialVariantNames;
    std::vector<std::vector<std::string>> groupsMaterialVariantNames;
  };
  ```
  Some details TBD here. The main point for now is: It is supposed to be a simple structure that offers a bunch of strings and string arrays that are extracted from metadata 

- An instance of this is offered by the `Cesium3DTilesSelection::Tileset` class

- The data is read, directly from the `tileset.json`, by the `TilesetContentManager` class - [here (in the part that is 'hidden' by GitHub](https://github.com/CesiumGS/cesium-native/pull/693/files#diff-e9f2f7fd16a07a1c4c8e35035731874c9d86b178a833191e6a38aa5d3a17fc41R919)

- It is stored in the `TilesetContentLoaderResult`, which appears to be the only way to 'transport' this information back to the caller (which has to happen on the main thread, via `propagateTilesetContentLoaderResult`)

---

One open question is:

The `schema` is currently assumed to be in the `tileset.json`. But the `tileset.json` might also contain a `schemaUri`, in which case the schema has to be resolved from an external file. I assume that this will involve some juggling with `TilesetJsonLoader` and the surrounding infrastructure (`createLoader`, `addChildLoader` etc...), and I'm not sure whether this conflicts with the current approach of using the `TilesetContentManager` to "transport" the data that is loaded back to the `Tileset` instance.


---

Longer-term points that are related to this PR:

- The `TilesetJsonLoader` could eventually be replaced by the `Cesium3DTilesLoader`. But for now, it is pretty specific and deeply entangled in the loader infrastructure, and can probably not easily be replaced
- This PR only adds easy, top-level access to the `MaterialVariants` data. This data is extracted from the metadata. Thus, it should be considered only as a "convenience functionality" that is offered _on top of whatever metadata handling classes will exist in the future_. Longer term, there might be a `Cesium3DTilesMetadata` library that contains the [auto-generated classes that currently exist only for glTF](https://github.com/CesiumGS/cesium-native/blob/89a317d9f9bb2249e7082c68a2c72fa95a1ea512/CesiumGltfReader/generated/src/ExtensionExtStructuralMetadataClassJsonHandler.h), but which will also have to exist for 3D Tiles at some point. But regardless of what happens there, "under the hood", the `MaterialVariants` could still be exposed in the form that is drafted here, with no changes for the clients that consume it.

cesium-concierge:
Thanks again for your contribution @javagl! 

PR #696, opened 2023-07-28
Add LocalHorizontalCoordinateSystem constructors taking ECEF<->Local matrices.

kring:
These matrices are the actual data representation of the class. It's helpful to have these for some refactoring I'm doing in Cesium for Unreal.

### DIFF
--- a/CesiumGeospatial/include/CesiumGeospatial/LocalHorizontalCoordinateSystem.h
+++ b/CesiumGeospatial/include/CesiumGeospatial/LocalHorizontalCoordinateSystem.h
@@ -73,6 +73,30 @@ public:
       const Ellipsoid& ellipsoid = Ellipsoid::WGS84);
 
   /**
+   * @brief Create a new coordinate system with a specified transformation to
+   * the Earth-Centered, Earth-Fixed frame. This is an advanced constructor and
+   * should be avoided in most cases.
+   *
+   * @param localToEcef The transformation matrix from this coordinate system to
+   * the ECEF frame.
+   */
+  explicit LocalHorizontalCoordinateSystem(const glm::dmat4& localToEcef);
+
+  /**
+   * @brief Create a new coordinate system with the specified transformations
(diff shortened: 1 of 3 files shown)


CesiumGeospatial/include/CesiumGeospatial/LocalHorizontalCoordinateSystem.h
```diff
@@ -72,6 +72,30 @@ class CESIUMGEOSPATIAL_API LocalHorizontalCoordinateSystem {
...
+   *
+   * @param localToEcef The transformation matrix from this coordinate system to
+   * the ECEF frame. This must be the inverse of `ecefToLocal`.
+   * @param ecefToLocal The transformation matrix from the ECEF frame to this
+   * coordinate system. This must be the inverse of `localToEcef`.
+   */
+  LocalHorizontalCoordinateSystem(
+      const glm::dmat4& localToEcef,
+      const glm::dmat4& ecefToLocal);
+
```
csciguy8, line 98: I appreciate the direct access to the members from the ctor.
- Is it possible to mention a use case in the comments that this would be typically used for?
- Also, this states that the two matrices (`ecefToLocal localToEcef`) must be inverse of each other, but there doesn't seem to be any code in the class that enforces this. Maybe _must_ really means _typically is_

kring: > Is it possible to mention a use case in the comments that this would be typically used for?

Ok, I added a note.

> Also, this states that the two matrices (ecefToLocal localToEcef) must be inverse of each other, but there doesn't seem to be any code in the class that enforces this. Maybe must really means typically is?

No, they must be. It's a precondition. Enforcing it would be costly and unnecessary. It doesn't make any sense to have two transformations that are meant to be the opposite of each other yet one is not the inverse of the other. I tweaked the doc to make this clearer. I considered adding an assertion, but that quickly gets into floating point equality land and it's just not worth the trouble.

kring:
This is ready for another review.

PR #700, opened 2023-08-02
Prevent empty error message from getting propagated to loadErrorCallback

lilleyse:
There was a small regression in https://github.com/CesiumGS/cesium-native/pull/646 where an invalid error would get propagated to the tileset's `loadErrorCallback`. Fixed by wrapping in a `if (result.errors)` check.

Feel to free to tweak for style as needed (e.g. maybe the if statements should be combined into one?).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### v0.27.0 - 2023-09-01
+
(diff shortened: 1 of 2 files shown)


kring:
Thanks @lilleyse!

PR #705, opened 2023-08-16
Update npm dependencies and regenerate 3DTiles/glTF classes.

kring:
I just updated npm dependencies by removing package-lock.json from the root and from tools/generate-classes and running npm install again. And then I regenerated both the 3D Tiles classes and the glTF classes to make sure the generator was still working. Regenerating picked up some doc changes, and there are also some very minor formatting changes, possibly due to a clang-format upgrade.

There are no functional changes here, but the doc improvements are useful and this should also resolve the dependabot PRs recently opened. My motivation in doing this at all was mostly just to make sure I can still run the generator successfully before I start making changes to it, but the result in this PR is worth merging.

### DIFF
--- a/Cesium3DTiles/generated/include/Cesium3DTiles/BoundingVolume.h
+++ b/Cesium3DTiles/generated/include/Cesium3DTiles/BoundingVolume.h
@@ -31,15 +31,20 @@ struct CESIUM3DTILES_API BoundingVolume final
   /**
    * @brief An array of six numbers that define a bounding geographic region in
    * EPSG:4979 coordinates with the order [west, south, east, north, minimum
-   * height, maximum height]. Longitudes and latitudes are in radians, and
-   * heights are in meters above (or below) the WGS84 ellipsoid.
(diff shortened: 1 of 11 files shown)


csciguy8 (commented):
Very nice, thanks for this. The changes look good, but did have some related questions,
- Is there documentation how to generate these classes? If not, can we get a blurb about it?
- Is generation a step we can do on CI so we never have to manually run it?

kring:
> Is there documentation how to generate these classes? If not, can we get a blurb about it?

I added some doc about this to the main README.md.

> Is generation a step we can do on CI so we never have to manually run it?

Well... maybe! As I mentioned offline, there are some downsides to this:

1. Developers would need Node.js installed in order to build cesium-native. Currently it's not required.
2. We'd need to lock down the particular version of the specs we're generating code from, so that changes to the specs can't break our build.
3. It takes a little time to run the generator.

Even with these caveats, it could be a worthwhile tradeoff. Maybe. Definitely not as part of this PR, though.

kring:
Ready for another look.

PR #706, opened 2023-08-16
Capture unknown properties when reading JSON for glTF, 3D Tiles, etc.

kring:
This is a PR into #705 so merge that first.

When reading JSON from a glTF, tileset.json, etc., it's sometimes useful to be able to capture properties that are in the JSON but that aren't officially part of the spec. For example:

1. Deprecated properties like `gltfUpAxis`.
2. Experimental properties, without bothering to update the spec.
3. When writing a validator, capture all unknown properties so that you can inform the user that they're unknown.

With this PR, such properties are put into the `unknownProperties` field that exists on every `ExtensibleObject`. There's an option on each of the readers to instead ignore them completely, as we did in previous versions.

This PR looks really big but it's almost all changes to generated code. If you ignore anything in a `generated` subdirectory the review should be much easier.


Cesium3DTilesReader/test/TestTilesetReader.cpp
```diff
@@ -310,3 +310,85 @@ TEST_CASE("Reads custom extension") {
...
+
+TEST_CASE("Reads tileset JSON with unknown properties and ignores them when "
+          "requested") {
+  using namespace std::string_literals;
+
+  std::filesystem::path tilesetFile = Cesium3DTilesReader_TEST_DATA_DIR;
+  tilesetFile /= "tileset-with-unsupported-properties.json";
+  std::vector<std::byte> data = readFile(tilesetFile);
+  Cesium3DTilesReader::TilesetReader reader;
+  reader.getOptions().setCaptureUnknownProperties(false);
```
csciguy8, line 385: For my curiosity, is there a compelling reason to have the option to turn on / off the capture? My guess is it saves performance / memory having it off.

CesiumGltfReader/include/CesiumGltfReader/GltfReader.h
```diff
@@ -210,7 +208,7 @@ class CESIUMGLTFREADER_API GltfReader {
   generateMipMaps(CesiumGltf::ImageCesium& image);
 
 private:
-  CesiumJsonReader::ExtensionReaderContext _context;
+  CesiumJsonReader::JsonReaderOptions _context;
```
csciguy8, line 211: I noticed that `TilesetReader` contains these same changes. What are your thoughts on having them derive from a shared base class? (or really any way to avoid code duplication)

CesiumGltfReader/test/TestGltfReader.cpp
```diff
@@ -332,7 +332,7 @@ TEST_CASE("Can deserialize KHR_draco_mesh_compression") {
 
   // Repeat test but this time the extension should be deserialized as a
   // JsonValue.
-  reader.getExtensions().setExtensionState(
+  reader.getOptions().setExtensionState(
```
csciguy8, line 335: In this test, in theory the reader options could be set the same as the TilesetReader test does, `reader.getOptions().setCaptureUnknownProperties(true or false);`
From the looks of this PR, I assume that the GLTF reader is not supporting unknown properties?
If so, will the default state (true) of _captureUnknownProperties cause any problems?

csciguy8 (commented):
Looks good. Just a few clarifying questions, but looks like nice added functionality.

Ran this in Cesium for Unreal, verified the new test is running on CI, and tested generating the gltf and 3d tiles classes locally (output matched this PR).

CesiumGltfReader/include/CesiumGltfReader/GltfReader.h
kring: I've moved to generating the reader classes in #711. GltfReader is still a bit special, though.

Cesium3DTilesReader/test/TestTilesetReader.cpp
kring: Yeah that's about it, though the effect should be small (barring weird edge cases like a glTF will a huge number of unknown properties for some reason). I probably could have skipped this.

CesiumGltfReader/test/TestGltfReader.cpp
kring: I'm not quite sure what you're asking here. `setCaptureUnknownProperties` should work on `GltfReader` the same as it does on `TilesetReader`. Unless I've missed something?

kring:
I don't think any code changes are needed based on the review here, so this should still be ready to merge. But let me know if I've missed something.

CesiumGltfReader/include/CesiumGltfReader/GltfReader.h
csciguy8: I see it, TilesetReader.h looks to be deleted. Nevermind this comment then.

CesiumGltfReader/test/TestGltfReader.cpp
csciguy8: Setting the capture flag on the options should work, yes. But for glTF reading, I didn't see a way to fetch the unknown properties object. 

This capture might show what I'm saying best...
image


Also, I didn't see a glTF reader test for unknown properties. For both these reasons, I assumed it wasn't supported.

kring: I still don't understand the problem. The glTF case is no different from the 3D Tiles one. Every generated class (including the glTF `Model` class) has an `unknownProperties` property. I added [two new tests](https://github.com/CesiumGS/cesium-native/pull/706/commits/2babed53b106290bc0906b1bd037e4d606412951) to demonstrate this, but it's definitely not worth adding such a test for _every_ generated class.

csciguy8: You got it with the tests. It wasn't obvious to me how unknown properties were fetched with gltf results, but they are now. And there are tests for it too, nice.

PR #710, opened 2023-08-23
Allow reading statically-typed classes from `radidjson::Value`, avoid diamond-shaped inheritance

kring:
This is a PR into #706, so merge that first.

Two changes to the JSON reading system here:

## Don't inherit `IExtensionJsonHandler` from `IJsonHandler`.

glTF / 3D Tiles extension `*JsonHandler` classes inherit from both `ExtensibleObjectJsonHandler` and `IExtensionJsonHandler`. Previously, both of these types inherited `IJsonHandler`, which made for diamond-shaped inheritance and ambiguity when upcasting to `IJsonHandler`. Now, `IExtensionJsonHandler` instead has a pure virtual method, `getHandler`, to get the `IJsonHandler` reference. All implementations will simply return `*this`. Another way to do this would be to dynamic_cast where necessary, but that would require RTTI to be enabled. The approach implemented here will work anywhere.

Most of the changes in this PR are to generated code in response to this change. Previously we needed to define a bunch of methods in every extension JsonHandle to resolve the IJsonHandler ambiguity. Now we only need to define `getHandler`.

## Allow reading statically-typed classes from `rapidjson::Value`

Previously, we could only read statically-typed classes from a byte buffer containing JSON. Now, we can also read them from an already-parsed `rapidjson::Value`. This is useful (in a future PR) in order to allow us to parse a tileset.json into a rapidjson::Document (as we're already doing) but then parse its metadata fields into statically-typed classes.

## Add `ValueType` type alias to `ArrayJsonHandler` classes

Most JsonHandlers already had this, so this is just making the array handler consistent with the rest. Another small changes that makes sense in isolation but will become necessary in a future PR.


CesiumJsonReader/src/JsonReaderOptions.cpp
```diff
@@ -25,6 +25,8 @@ class AnyExtensionJsonHandler : public JsonObjectJsonHandler,
         &std::any_cast<CesiumUtility::JsonValue&>(value));
   }
 
+  virtual IJsonHandler& getHandler() override { return *this; }
+
   virtual IJsonHandler* readNull() override {
```
j9liu, line 30: Just want to make sure, these other `override` methods are supposed to be here? Is there a reason why `getHandler()` can't return the `JsonObjectJsonHandler` ?

j9liu (commented):
@kring looks good to me! Just had one clarification question, and one typo I caught, which I'll just fix

CesiumJsonReader/src/JsonReaderOptions.cpp
kring: > Just want to make sure, these other override methods are supposed to be here?

You're right, they're not needed anymore. Removed.

> Is there a reason why getHandler() can't return the JsonObjectJsonHandler ?

None of the statically-typed extension JsonHandlers derived from JsonObjectJsonHandler.

But it might be sensible for `AnyExtensionJsonHandler` to output an ExtensibleObject with lots of `unknownProperties`, rather than a `JsonValue`, now that we have that unknown properties thing. And then IExtensionJsonHandler::getHandler could return an `ExtensibleObjectJsonHandler`. Probably not really worth the effort to change now, though, and definitely not part of this PR.

j9liu:
Merging now. Thanks @kring !

PR #713, opened 2023-08-25
Create `PropertyView` base class with normalization handling

j9liu:
The branch name is misleading, this expanded into a larger rework than I had originally anticipated :') This PR does the following:

- Adds `PropertyView<T, bool>`, a templated class that acts as a base for `PropertyTablePropertyView` and `PropertyTexturePropertyView`.
  - Includes support for parsing `offset`, `scale`, `min`, `max`, `noData`, and `default` from the `EXT_structural_metadata` spec.
  - Allows `offset`, `scale`, `min`, `max` to be overridden by values from the `PropertyTableProperty` and `PropertyTextureProperty` themselves
  - Allows normalized properties to parse values as `double`s, instead of whatever integer type `T` their data is
- Added `PropertyTransformations.h`, which includes functions for applying normalization, offset, and scale to property values
- Changed the retrieval of values in `PropertyTablePropertyView` and `PropertyTexturePropertyView`
  - The previous `get` implementation is now `getRaw`. This returns the value as-is from the data, without transformations
  - The new `get` implementation returns a `std::optional`. This returns the value with transformations. If a `noData` value is defined in the property, this returns std::nullopt for values that equal `noData`.
  - For normalized properties, `get` returns the `double` form of the property type. (e.g. `dvec2` for `ivec2`)


CesiumGltf/include/CesiumGltf/PropertyArrayView.h
```diff
@@ -69,7 +81,7 @@ template <> class PropertyArrayView<bool> {
   /**
    * @brief Constructs an empty array view.
    */
-  PropertyArrayView() : _values{}, _bitOffset{0}, _size{0} {}
+  PropertyArrayView() : _values{} {}
```
j9liu, line 84: I changed this around and didn't properly revert it, so ignore the missing initialization / random `= 0`'s in the variable declaration

CesiumGltf/include/CesiumGltf/PropertyArrayView.h
```diff
@@ -125,7 +154,7 @@ template <> class PropertyArrayView<std::string_view> {
       : _values{values},
         _stringOffsets{stringOffsets},
         _stringOffsetType{stringOffsetType},
-        _size{size} {}
+        _size(size) {}
```
j9liu, line 157: leftover from a previous change, revert this

CesiumGltf/include/CesiumGltf/PropertyTablePropertyViewTypes.h
```diff
@@ -0,0 +1,294 @@
...
+#define View PropertyTablePropertyView
+
+/**
+ * @brief An alias for all of the potential types of
+ * {@link PropertyTablePropertyView} possible without normalization.
+ * Can be useful for applications that want to implement abstract
+ * representations of PropertyTablePropertyView, without the mess of templated
+ * types.
+ */
+typedef std::variant<
```
j9liu, line 24: I added this anticipating that we'll need this `typedef` for Unreal / Unity, but it's not used anywhere in Native right now

j9liu (commented):
noticed a few things, leaving comments for me to clean up tomorrow

CesiumGltf/include/CesiumGltf/PropertyView.h
```diff
@@ -0,0 +1,2375 @@
...
+   */
+  static const PropertyViewStatusType ErrorInvalidNoDataValue = 10;
+
+  /**
+   * @brief The property provided an invalid default value.
+   */
+  static const PropertyViewStatusType ErrorInvalidDefaultValue = 11;
+};
+
+namespace {
```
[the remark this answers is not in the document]
j9liu: Ah, when I did that I was trying to convey that these helper functions are intended solely for this `.h` file, but I didn't realize it was bad form. I'll remove the namespace.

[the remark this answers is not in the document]
j9liu: I actually keep running into linking errors when I remove the anonymous namespace (for some function). This is the sort of thing that happens:
image

It seems like the templated functions are fine, but any non-templated ones result in multiple definitions in the file. I'm not sure why this is happening... in any case, I'm keeping the anonymous namespace for the problematic functions, unless there is a better method of preventing this error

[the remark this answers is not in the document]
j9liu: on @csciguy8's suggestion, I'll append `inline` to the functions that are causing these linker errors. I think they're small enough to warrant the `inline` anyways

CesiumGltf/include/CesiumGltf/PropertyTableView.h
```diff
@@ -107,21 +108,21 @@ class PropertyTableView {
    * @return A {@link PropertyTablePropertyView} of the property. If no valid property is
    * found, the property view will be invalid.
    */
-  template <typename T>
-  PropertyTablePropertyView<T>
+  template <typename T, bool Normalized = false>
```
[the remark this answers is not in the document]
j9liu: Yep, it does return an invalid `PropertyTablePropertyView`! But I'll update the documentation so that's more apparent. I also added a more proper `ErrorNormalizationMismatch` error code, separate from the `ErrorInvalidNormalization` code. (The latter should be used for floating point components erroneously marked as normalized ).

j9liu:
Thanks for the review @kring ! 😄 No worries about only skimming the unit tests, I know that there are a ton I added.

>  Did you have some sort of process for producing the tests for all the property type variations? Even if it was just copy/paste/search/replace?

Yes, so for every numeric property type I have two tests: normalized, and non-normalized. I'm going through the tests and cleaning them up, so I'll summarize the tests from each file:

- Add normalized versions of every numeric property test.
  - For example, `"Test scalar PropertyTableProperty"` and `"Test scalar PropertyTableProperty (normalized)"`.
  - Includes tests for fixed and variable length numeric arrays, too.
- Added tests for `offset`, `scale`, `min`, `max`, both for normalized and non-normalized properties
- Added tests for a property instance's `offset`, `scale`, `min`, `max` overriding the values of its `ClassProperty` (when specified)
- Added tests for `noData`, both with and without `default` defined
- Added tests for callbacks on invalidly normalized properties (e.g. `float` property erroneously marked `normalized`). This returns an invalid property view. It's not possible to construct a `PropertyView<T, true>` if `T` has no corresponding `TypeToNormalizedType<T>` defined.

I used helper functions where I could, but ultimately there were ~2x the number of unit tests needed in each relevant file. Plus the new unit tests for `PropertyView.h`. That's probably why the diff is so huge :')

In any case, I self-reviewed the unit tests today and caught some mistakes, so don't feel the need to pore over them. Hopefully CI will pass with my changes and then this will be ready for another look 🤞

CesiumGltf/include/CesiumGltf/PropertyView.h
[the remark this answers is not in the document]
kring: Yeah you can either declare them inline, or you can move them to a .cpp file. Generally the latter is better because doing that helps keep compile times lower (not that it'll make a big difference in this case). Declaring them inline mostly only makes sense if you want them inlined for performance reasons.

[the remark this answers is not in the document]
j9liu: Maybe something's terribly wrong with my Visual Studio 2022 install, but moving them to the .cpp file keeps resolving in "unresolved external symbol" errors :/

Not sure what's wrong, this works just fine for `PropertyType.h`, which contains functions whose bodies are implemented in a `.cpp` file. Maybe it's got to do with the templating?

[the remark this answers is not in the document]
kring: Ah yep, sorry, I didn't realize those are templates. Templatize functions can't be defined in a .cpp file like that.

But also, including the complete definition in a header should _not_ result in multiply-defined symbols for template functions. I think the reason it is in your case is because you've declared some of the functions `static`. Remove that and it will work fine (I think).

CesiumGltf/include/CesiumGltf/PropertyArrayView.h
```diff
@@ -58,6 +56,20 @@ template <typename ElementType> class PropertyArrayView {
         _values);
   }
 
+  bool operator==(const PropertyArrayView<ElementType>& other) const noexcept {
```
kring, line 59: Was this added primarily for the tests? If so, it might be better to define a free-standing `operator==` in the tests, rather than here. Or maybe even a comparison function rather than an overloaded operator? It's a little unusual to overload an operator to to element-wise comparison. `std::vector` doesn't have this operator, for instance. But if you think it's generally useful as part of the API, then you should also define `operator!=` too. It can just call `operator==` and negate it.

CesiumGltf/include/CesiumGltf/PropertyTexturePropertyView.h
```diff
@@ -284,173 +419,257 @@ template <typename ElementType> class PropertyTexturePropertyView {
...
-  double applySamplerWrapS(const double u, const int32_t wrapS) const noexcept {
-    if (wrapS == Sampler::WrapS::REPEAT) {
-      double integral = 0;
-      double fraction = std::modf(u, &integral);
-      return fraction < 0 ? 1.0 - fraction : fraction;
-    }
+    double wrappedU = applySamplerWrapS(u, this->_pSampler->wrapS);
+    double wrappedV = applySamplerWrapT(v, this->_pSampler->wrapT);
+
+    // TODO: account for sampler's filter (can be nearest or linear)
```
kring, line 601: Does this need to be addressed before we can merge? I'm guessing we should at least inform the user that it's not supported, right?

CesiumGltf/include/CesiumGltf/PropertyTexturePropertyView.h
```diff
@@ -284,173 +419,257 @@ template <typename ElementType> class PropertyTexturePropertyView {
...
+    int64_t pixelIndex = this->_pImage->bytesPerChannel *
+                         this->_pImage->channels *
+                         (y * this->_pImage->width + x);
+
+    // TODO: Currently stb only outputs uint8 pixel types. If that
+    // changes this should account for additional pixel byte sizes.
+    const uint8_t* pValue = reinterpret_cast<const uint8_t*>(
+        this->_pImage->pixelData.data() + pixelIndex);
+
+    std::vector<uint8_t> channelValues(this->_channels.size());
```
kring, line 632: This allocation looks like it can be quite costly. The number of channels is always going to be bounded, right? Like <= 4? In that case, a `std::array<uint8_t, 4>` will be a _lot_ more efficient (no heap allocation). Can wrap it up in a span of the right size to pass to `assembleValueFromChannels` to indicate how many elements are actually used.

kring (commented):
Looking great!

CesiumGltf/include/CesiumGltf/PropertyArrayView.h
j9liu: I did this to enable the code here:
https://github.com/CesiumGS/cesium-native/blob/a56f7eddec540003d5ee5b92ec48a63065f55138/CesiumGltf/include/CesiumGltf/PropertyTablePropertyView.h#L272-L274

This way, it can compare a `PropertyArrayView` value to the optional `PropertyArrayView` `noData` value.

But good point, I'll define `operator!=` too.

CesiumGltf/include/CesiumGltf/PropertyTexturePropertyView.h
j9liu: We got away with not having it in the previous implementation so I thought it might be ok this round. And I wasn't sure what the official linear (bilinear?) sampling algorithm was, so I ignored it at the time.

I'll see if I can implement it quickly today, to eliminate surprises for the user.

j9liu: Ah, I did have some questions about this... The separation of a min / mag filter makes it possible for one to have a `NEAREST` filter and one to have `LINEAR`. So which one would we use?

I feel like accounting for this setting was awkward because this mainly exists for rendering the texture at different distances. At what point does the user expect the data to be interpolated? Perhaps we could add a parameter to the `get` functions that specifies whether to use linear filtering. It would just have to be user-specified, and I'm not sure how important this functionality is to a user.

I think for now, I'll forgo adding support, and just document that it retrieves data from the nearest pixel.

CesiumGltf/src/PropertyTexturePropertyView.cpp
```diff
@@ -58,4 +58,45 @@ double applySamplerWrapT(const double v, const int32_t wrapT) {
...
+  int64_t pixelIndex =
+      image.bytesPerChannel * image.channels * (y * image.width + x);
+
+  // TODO: Currently stb only outputs uint8 pixel types. If that
+  // changes this should account for additional pixel byte sizes.
+  const uint8_t* pValue =
+      reinterpret_cast<const uint8_t*>(image.pixelData.data() + pixelIndex);
+
+  std::array<uint8_t, 4> channelValues{0, 0, 0, 0};
+  for (size_t i = 0; i < channels.size(); i++) {
```
kring, line 95: I'm not sure if the caller will always guarantee that channels.size() <= 4, but a small tweak here will make this safer with very little cost. I'll go ahead and add that to save another cycle, and then merge this PR!

kring:
Thanks @j9liu!

PR #718, opened 2023-08-30
Update readme with more CMake / VS build notes

csciguy8:
Add more detail about generating projects from CMake for use with Visual Studio


kring:
Thanks @csciguy8!

PR #716, opened 2023-08-29
Add `PropertyAttributeView` and `PropertyAttributePropertyView`.

j9liu:
This adds views for `PropertyAttribute` and `PropertyAttributeProperty`, similar to the views for property tables and textures. Unlike the others, which can construct views on the `Model` extension alone, the`PropertyAttributeView` requires a `MeshPrimitive` to construct a `PropertyAttributePropertyView` on. This is because the data exists as an attribute on the primitive-level, not as data on the root model.

Added some unit tests as well. Feel free to ignore the changes in the non-`PropertyAttribute` specs. Most of it was reorganizing the tests, cleaning up code, and correcting some uncaught errors from the previous PR.

PR #719, opened 2023-08-30
Miscellaneous `EXT_structural_metadata` changes

j9liu:
This PR is based on #716, so merge that first.

This PR includes a batch of changes related to `EXT_structural_metadata` support:

1. Log warning about `EXT_feature_metadata` when glTFs are loaded.
2. Add ability to access `name`, `description`, `semantic` properties from `PropertyView`
3. Add support for sentinel values when upgrading batch table. If possible, `null` can be converted into a sentinel value that is put as the `noData` value of the class property.
4. Add some missing checks to `TestPntsToGltfConverter`. The lines were originally written for `EXT_feature_metadata` and were not properly replaced.
5. Add additional `typedef`s for `PropertyTexturePropertyView` and `PropertyAttributePropertyView` variants. Even though they could probably fit into one variant `typedef` each, I wanted them to have symmetry with `PropertyTablePropertyView`, which has too many types to fit in one variant.


kring:
Looks great, happy to merge this one too, after the others.

j9liu:
While testing this out in Cesium for Unreal, I triggered an assert with New York 3D Buildings. This was due to an attempt to handle an instance of the "Year" property that contained all null values. I just pushed a commit with some code to catch this and ignore the property if this is the case.

PR #721, opened 2023-09-01
Finish loading parent tile when needed for upsampling.

kring:
This fixes the Unreal Movie Render Queue freezing that several users have noticed in Cesium for Unreal v1.29.0:
https://community.cesium.com/t/movie-render-queue-freeze-at-lower-values-for-maximum-screen-space-error/26331

Here's what's happening:

1. A tile needs to be upsampled in order to hang more detailed raster overlays on it.
2. That tile can't be upsampled until its parent enters the Done state, because we need the parent geometry in order to do the upsampling.
3. `TilesetContentManager::loadTileContent` detects this case, and explicitly calls `loadTileContent` on the parent tile. Because there's no guarantee the parent tile would ever load otherwise.
4. `loadTileContent` allows the parent tile to complete its request and worker thread work, but by itself it will only get the tile as far as the `ContentLoaded` state. The transition from `ContentLoaded` to `Done` happens in the course of visiting the tile during the main traversal.
5. Except (4) isn't quite true. It used to work that way, but after #565 this transition is throttled and prioritized, in order to avoid tanking the frame rate by loading too many tiles at once. We'll only do this work for tiles that we actually want to render.
6. Because the parent tile in this scenario isn't actually selected for rendering, it never transitions from `ContentLoaded` to `Done`.
7. So the parent tile's failure to load blocks loading of the upsampled tile, which blocks the movie rendering from advancing to the next frame because loading isn't completed.

This is quite an old problem. However it just became apparent to users in Cesium for Unreal v1.29.0 because we tightened up the logic for waiting for tiles to load when rendering movies. Previously it wouldn't have waited, now it does (forever).

### DIFF
--- a/Cesium3DTilesSelection/src/TilesetContentManager.cpp
+++ b/Cesium3DTilesSelection/src/TilesetContentManager.cpp
@@ -878,6 +878,12 @@ void TilesetContentManager::loadTileContent(
     if (pParentTile) {
       if (pParentTile->getState() != TileLoadState::Done) {
         loadTileContent(*pParentTile, tilesetOptions);
+
+        // Finalize the parent if necessary, otherwise it may never reach the
+        // Done state.
+        if (pParentTile->getState() == TileLoadState::ContentLoaded) {
+          finishLoading(*pParentTile, tilesetOptions);
+        }
         return;
       }
     } else {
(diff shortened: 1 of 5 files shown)


Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -878,6 +878,12 @@ void TilesetContentManager::loadTileContent(
     if (pParentTile) {
       if (pParentTile->getState() != TileLoadState::Done) {
         loadTileContent(*pParentTile, tilesetOptions);
+
+        // Finalize the parent if necessary, otherwise it may never reach the
+        // Done state.
+        if (pParentTile->getState() == TileLoadState::ContentLoaded) {
+          finishLoading(*pParentTile, tilesetOptions);
```
csciguy8, line 885: Look sane.

Verified
* finishLoading can be called from this function (looks to be able to be deferred or called immediately)
* finishLoading is still getting called from the same thread that other callers are using it (main thread)
* Adding this call won't cause any infinite recursion

PR #723, opened 2023-09-05
Fix libjpeg-turbo build on Android

TheMostDiligent:
Android build mostly works out-of-box except for the libjpeg-turbo. The problem is that the library does not respect the `CMAKE_ANDROID_ARCH_ABI` variable. Instead, it expects the custom [`ANDROID_ABI`](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/BUILDING.md#building-libjpeg-turbo-for-android) variable.


kring:
Thanks for the PR @TheMostDiligent! Can I bug you to sign the Contributor License Agreement so that we can review and merge this? I don't think we have one from you yet.

https://github.com/CesiumGS/cesium/blob/master/CONTRIBUTING.md#contributor-license-agreement-cla

TheMostDiligent:
@kring Done

kring:
Thanks @TheMostDiligent!

PR #726, opened 2023-09-13
Support omitted `EXT_structural_metadata` properties

j9liu:
This PR adds support for properties that are omitted from a property table / texture / attribute instance. This is allowed by the spec so long as the `ClassProperty` is not `required`. A special case happens if an omitted `ClassProperty` defines a default value; ideally, this value should be used by for metadata picking / styling in our runtime engines. However, in the current implementation it creates `ErrorNonexistentProperty` instances for these omitted properties. Technically, this is not wrong, but it may be confusing when trying to convey what data is actually available to users. 

My initial approach was to restrict the `forEachProperty` functions in `PropertyTable`, `PropertyTexture` and `PropertyAttribute`. They would only handle properties that were actually defined in the instance, and ignore anything else in the metadata `Class`. It would be up to the runtime engine to cross-reference this with the `Class` definition using the `getClass` functions. This also implied that `Class` and `ClassProperty` should be made accessible to users in the runtime engines, since they'd need the `Class` definition to know if properties were omitted. It would also reduce the need for `forEachProperty`, since we'd prefer to iterate over all of the `ClassProperty`s instead.

I thought about this some more today, and realized the approach would put more work on the runtime implementations than necessary. Fundamentally, `PropertyTable`, `PropertyTexture`, and `PropertyAttribute` are just ways of implementing a metadata `Class` with a certain data storage. The views we create on their properties already consolidate the `ClassProperty` information with the `PropertyView`s. So I think it is fitting to fold in the omitted property definitions into the `PropertyTableView`, etc. classes. Because even if they omit properties that are included in the `Class`, there should still be some way to quickly indicate that the property has data.

That's my thinking, anyways. Feel free to push back on it if the design is questionable. If we want, we still can expose `Class` and `ClassProperty` to users in the runtime engines, but at least this doesn't mandate it. You can still get the full picture of a metadata entity even if it has omitted properties. Ultimately, there is a distinction between a property that *should* be found but wasn't, and a property that is okay to omit.

### DIFF
--- a/CesiumGltf/include/CesiumGltf/PropertyAttributePropertyView.h
+++ b/CesiumGltf/include/CesiumGltf/PropertyAttributePropertyView.h
@@ -127,7 +127,7 @@ public:
   /**
    * @brief Constructs an invalid instance for an erroneous property.
    *
-   * @param status The value of {@link PropertyAttributePropertyViewStatus} indicating the error with the property.
+   * @param status The code from {@link PropertyAttributePropertyViewStatus} indicating the error with the property.
    */
   PropertyAttributePropertyView(PropertyViewStatusType status) noexcept
       : PropertyView<ElementType, false>(status), _accessor{}, _size{0} {
--- a/CesiumGltf/include/CesiumGltf/PropertyAttributePropertyView.h
+++ b/CesiumGltf/include/CesiumGltf/PropertyAttributePropertyView.h
@@ -25,66 +25,66 @@ public:
    * @brief This property view was initialized from an invalid
    * {@link PropertyAttribute}.
    */
-  static const int ErrorInvalidPropertyAttribute = 13;
+  static const int ErrorInvalidPropertyAttribute = 14;
 
   /**
    * @brief This property view is associated with a {@link ClassProperty} of an
    * unsupported type.
    */
-  static const int ErrorUnsupportedProperty = 14;
+  static const int ErrorUnsupportedProperty = 15;
 
   /**
    * @brief This property view was initialized with a primitive that does not
    * contain the specified attribute.
    */
-  static const int ErrorMissingAttribute = 15;
+  static const int ErrorMissingAttribute = 16;
 
   /**
    * @brief This property view's attribute does not have a valid accessor index.
    */
-  static const int ErrorInvalidAccessor = 16;
+  static const int ErrorInvalidAccessor = 17;
 
   /**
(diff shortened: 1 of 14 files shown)


lilleyse:
> If we want, we still can expose `Class` and `ClassProperty` to users in the runtime engines, but at least this doesn't mandate it.

Out of curiosity (and a bit of a tangent) - are [metadata statistics](https://github.com/CesiumGS/3d-tiles/tree/main/specification#metadata-statistics) exposed to users? They could be useful for creating color ramps and other effects.

j9liu:
> Out of curiosity (and a bit of a tangent) - are [metadata statistics](https://github.com/CesiumGS/3d-tiles/tree/main/specification#metadata-statistics) exposed to users? They could be useful for creating color ramps and other effects.

@lilleyse and I talked about this offline, and the quick answer is no, we don't support accessing the `statistics` property in tilesets. But we could add support if it become a higher priority.

j9liu:
Pushed some unrelated changes to remove the `channels` parameter from `PropertyTexturePropertyView`. This is available through `PropertyTextureProperty`, so no need to include it anymore.

j9liu:
Another unrelated change, just added `PropertyTexturePropertyView::getSampler` so the sampler is accessible in Unreal Engine.

CesiumGltf/include/CesiumGltf/PropertyAttributePropertyView.h
```diff
@@ -122,34 +122,68 @@ class PropertyAttributePropertyView<ElementType, false>
...
+    // Don't override the status / size if something is wrong with the class
+    // property's definition.
+    if (this->_status != PropertyAttributePropertyViewStatus::Valid) {
+      return;
+    }
+
+    assert(
+        classProperty.defaultProperty &&
+        "Cannot construct a valid property view for an empty property with no "
+        "default value.");
```
kring, line 162: Given that this is part of the class's public API, would it be a bit safer to set the status to `ErrorNonexistentProperty` when there's no default? As written, it will assert in debug, which is fair enough because it's probably a developer error. But in a release build, it will pretend like there is a default, which doesn't seem great.

kring (commented):
Just one small comment, otherwise this makes sense to me! I'll go ahead and merge it and you can make the suggested change in a separate PR if you agree.

kring:
I'm merging this even though the Travis macOS runner seems to be unable to start right now.

PR #729, opened 2023-09-19
Fix `EmptyPropertyWithDefault` constructors

j9liu:
Followup to #726. This addresses @kring 's suggestion for improving the `EmptyPropertyWithDefault` constructor. Now when the constructor is called and `classProperty.defaultProperty` is not defined, the property view status is set to `ErrorNonexistentProperty`.


kring:
Thanks @j9liu!

PR #730, opened 2023-09-20
Switch from Travis to GitHub Actions

kring:


.github/workflows/build.yml
```diff
@@ -0,0 +1,165 @@
...
+    name: "Quick Checks"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      - name: Check source formatting
+        run: |
+          npm install
+          npm run format -- --dry-run -Werror
+  Documentation:
```
csciguy8, line 14: I appreciate this single documentation step as opposed to the older version which had it with each build.

.github/workflows/build.yml
```diff
@@ -0,0 +1,165 @@
...
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: Generate Documentation
+        run: |
+          cmake -B build -S .
+          cmake --build build --target cesium-native-docs
+      - name: Publish Documentation Artifact
+        if: ${{ success() }}
+        uses: actions/upload-artifact@v3
```
csciguy8, line 30: I appreciate this single documentation step as opposed to the older version which had it with each build. Also like how easy it is to grab the artifact

.github/workflows/build.yml
```diff
@@ -0,0 +1,165 @@
...
+          submodules: recursive
+      - name: Compile Debug Configuration
+        run: |
+          cmake -B build -S .
+          cmake --build build --config Debug --parallel 4
+      - name: Test Debug Configuration
+        run: |
+          cd build
+          ctest -V
+      - name: Compile RelWithDebInfo Configuration
```
csciguy8, line 53: Looks good here, but question for my sanity, did Travis ever build release? Because I didn't see it in the logs.

.github/workflows/build.yml
```diff
@@ -0,0 +1,165 @@
...
+        run: |
+          cmake --build build --config RelWithDebInfo --parallel 4
+      - name: Test RelWithDebInfo Configuration
+        run: |
+          cd build
+          ctest -V
+  VS2022:
+    name: "Windows + VS2022"
```
csciguy8, line 61: Saw the addition of VS2022 and drop of VS2017. Sounds good, but is there anything we need to note in changes?

.github/workflows/build.yml
```diff
@@ -0,0 +1,165 @@
...
+        run: |
+          sudo apt-get install nasm
+      - name: Check out repository code
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: Compile Debug Configuration
+        run: |
+          cmake -B build-debug -S . -DCMAKE_BUILD_TYPE:STRING=Debug
+          cmake --build build-debug --parallel 4
```
csciguy8, line 100: I noticed that the travis version is  `cmake --build . --config Debug`. I assume this is equivalent?

csciguy8 (approved):
Look like equivalent changes to travis, with some improvements. I'm excited to use it.

Just some assorted questions due to my own ignorance. Looks good to go though.

csciguy8:
One more observation. All the jobs are slower than their equivalent travis jobs. All but the macOS build are ~2x slower. Even with this, the times still seem reasonable.

.github/workflows/build.yml
kring: No, Travis previously only built Debug.

csciguy8: Fantastic improvement then, especially since tests are now run in release too.

.github/workflows/build.yml
kring: Yep. CMake didn't have `-S` before v3.13, but our build system requires 3.15+ anyway so we can use the new syntax.

.github/workflows/build.yml
kring: Good idea, I updated the changelog and the README.

kring:
> All but the macOS build are ~2x slower.

I'm guessing this is because we're building/testing the release configuration, too, now. If the extra build time is a problem, we could separate the debug and release configurations into separate jobs. Or even remove the release build entirely, to go back to the way things were on Travis.

kring:
This is ready for another look.

csciguy8:
> > All but the macOS build are ~2x slower.
> 
> I'm guessing this is because we're building/testing the release configuration, too, now. If the extra build time is a problem, we could separate the debug and release configurations into separate jobs. Or even remove the release build entirely, to go back to the way things were on Travis.

Makes sense. More work, more time needed. 

I like the idea of catching release build problems, release test problems. Sounds worth it to me.

csciguy8:
@kring The `Replace asserts with ifs` commit looks a little out of place. Was that meant for this PR?

Sounds like an ok thing to do, I just don't know the context of why it was needed.

csciguy8:
Weird, looks like the latest Linux + Clang build is failing.

I don't see any commits that could have broken it since the last good build, but I do see that the runner image version has changed (now 20230924.1.0)

image

kring:
> looks like the latest Linux + Clang build is failing.

Ha yeah apparently GitHub just removed Clang 12 support from their Linux runner:
https://github.com/actions/runner-images/issues/8263

Should be able to switch to `ubuntu-20.04`. I'll try it.

PR #731, opened 2023-09-22
Remove hardcoded size logic

ZackOfAllTrades:
Right now the size of each tile is hardcoded to the size of the gltf model. Specifically in Unity but probably Unreal too there is no real reason to keep the gltf model in memory after the IPrepareRendererResources runs and you have created your game engine-specific resources. Our devices have very little working memory so it is imperative we free the gltf model immediately after using it to create the Unity meshes and textures.
I wanted to present the proposed changes as a pr but this is actually a bigger change than I thought and would love to discuss different solutions.
Right now I'm working with the idea that TileRenderContent now has a byteSize that can be set by the end user inside the IPrepareRenderResources. This can still be set to the gltf model like before or something else, this allows more flexibility as I can set it to a calculated value I get from my Unity resources as well as the size of  RenderResources, in my case a collision mesh for each tile. 
This works well except the notifyTileDoneLoading() which increments the TilesetContentManager _tilesDataUsed after the prepareInLoadThread but before the prepareInMainThread which is a problem as the final true size of the tile may not be known until after prepareInMainThread.

I attached an example of how I'm using setByteSize at the end of my prepareInMainThread
example

### DIFF
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileContent.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileContent.h
@@ -163,6 +163,20 @@ public:
   void setRenderResources(void* pRenderResources) noexcept;
 
   /**
+   * @brief Set the byte size of the render content
+   *
+   * @param byteSize The size of the render content
+   */
+  void setByteSize(int64_t byteSize) noexcept;
+
+  /**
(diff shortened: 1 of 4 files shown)


cesium-concierge:
Thanks again for your contribution @ZackOfAllTrades! 

ZackOfAllTrades:
Closing PR and creating an issue instead

PR #733, opened 2023-09-27
Fix Tileset::ComputeLoadProgress incorrectly reporting done, under some conditions

csciguy8:
Addresses cases where `::ComputeLoadProgress` would report 100% done prematurely, before all work in the view had completed. 

(I noticed this in cesium-unreal performance tests, with Google 3D Tiles)

With certain tilesets and our "kicking' logic (`::_kickDescendantsAndRenderTile`), we can run into frames where all work has been kicked out of the main / worker threads, but no new work has been added yet. This could lead to a 100% value being returned, in error.

The fix is to consider frames that kicked tiles as not complete yet. This gives us at least one more view traversal to find remaining work.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,7 +10,8 @@
 ##### Fixes :wrench:
 
 - Fixed the handling of omitted metadata properties in `PropertyTableView`, `PropertyTextureView`, and `PropertyAttributeView` instances. Previously, if a property was not `required` and omitted, it would be initialized as invalid with the `ErrorNonexistentProperty` status. Now, it will be treated as valid as long as the property defines a valid `defaultProperty`. A special instance of `PropertyTablePropertyView`, `PropertyTexturePropertyView`, or `PropertyAttributePropertyView` will be constructed to allow the property's default value to be retrieved, either via `defaultValue` or `get`. `getRaw` may not be called on this special instance.
-
+- Fixed corner cases where `Tileset::ComputeLoadProgress` can report done (100%) in error, before all work is actually complete in the current view
+  
 ### v0.28.0 - 2023-09-08
 
 ##### Breaking Changes :mega:
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/ViewUpdateResult.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/ViewUpdateResult.h
@@ -55,6 +55,7 @@ public:
   uint32_t tilesCulled = 0;
   uint32_t tilesOccluded = 0;
   uint32_t tilesWaitingForOcclusionResults = 0;
+  uint32_t tilesKicked = 0;
   uint32_t maxDepthVisited = 0;
   //! @endcond
 
--- a/Cesium3DTilesSelection/src/Tileset.cpp
+++ b/Cesium3DTilesSelection/src/Tileset.cpp
@@ -314,6 +314,7 @@ Tileset::updateView(const std::vector<ViewState>& frustums, float deltaTime) {
   result.tilesCulled = 0;
   result.tilesOccluded = 0;
   result.tilesWaitingForOcclusionResults = 0;
+  result.tilesKicked = 0;
   result.maxDepthVisited = 0;
 
   if (!_options.enableLodTransitionPeriod) {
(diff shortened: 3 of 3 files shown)


kring:
This change looks safe, and it should fix the problem you were seeing. I think a better fix might be possible, though.

First a little background. In case the kick mechanism wasn't already complicated enough, there are actually two "levels" to it:

### Kick Level 1

The purpose of Kick Level 1 is to avoid unnecessary holes in the mesh. It is triggered whenever _any_ of the selected descendants of a tile are not ready to render, _and none_ of them were rendered last frame. If we didn't kick in this scenario, but instead rendered those descendants (some of which aren't renderable yet), it's hopefully pretty clear that we'd have visible holes in the mesh.

At Kick Level 1, we only kick tiles out of the render queue; they are left in the load queue. So the progress tracking is already accounting for them, and I don't think this can be the source of the progress bug you observed.

### Kick Level 2

The logic for the second kick level starts with the comment "If we're waiting for heaps of descendants..." in the `_kickDescendantsAndRenderTile` function. The purpose of this form of kicking is to give the user feedback during the load progress by sometimes loading and rendering intermediate tiles in the bounding volume hierarchy. Without this kicking, a user starting with a zoomed-out view would stare at a blank screen until all tiles at the correct LOD were loaded, and then they would all appear at once. So the goal of Kick Level 2 is essentially "mostly load just the tiles we need, in order to minimize the total tiles loaded, but if that ends up being a ton of tiles then load a smaller number of lower detail tiles first so we can get something on the screen sooner."

Now, the way Kick Level 2 is implemented, it kicks out all the descendants from the load queues, too. Which means that the progress tracking isn't aware that they're needed at all.

So I think the progress bug occurs in this situation. Imagine we have a tile hierarchy like this:

```
  A
  |
  B
  |
 C-Z
```

A is the root, B is a child tile of A, and a bunch of tiles (C through Z) are children of B. We're zoomed in close enough that all of these tiles should be visible.

#### Frame 1

Tile A is loaded and ready to render. None of the other tiles are. Because C through Z are a bunch of tiles, they all get kicked (level 2!) in favor of B. So C-Z are not represented in the render or load queues.

But then B also gets kicked (level 1 this time!) because it's not ready to render. B stays in the load queue though. Only A is in the render queue.

#### Frame 2

Tile B has finished loading and is now ready to render. C-Z are still not loaded (we haven't even started loading them yet). We still visit tiles C-Z, and initially add them to the load queue. But because they're not ready to render yet, we end up in `_kickDescendantsAndRenderTile` for tile B. It kicks C-Z out of the render queue (level 1, no problem). But then wet get to the level 2 logic.

Because tile B wasn't rendered _last frame_, and C-Z is still heaps of descendants, we go through with the full level 2 kick. C-Z are removed from the load queue _again_. But this time there's really no good reason for it, because even though B wasn't rendered last frame, it's ready to go this frame. There's no good reason to wait until next frame to start loading C-Z.

Furthermore, it causes the load progress bug. Tiles A and B are already loaded, while C-Z were kicked out of the load queue. So no tiles are in need of loading, and so we're 100% loaded... right?

#### Frame 3

Tile B was rendered last frame, so Tiles C-Z now only get a level 1 kick. They're added to the load queue. We went from 100% progress down to less than 100%.

---

With that background, a think a slightly better solution than the one implemented here is to only count the level 2 kick against the progress. Currently it counts any kick.

Initially I thought it might be useful to change the level 2 kick logic. Rather than level 2 kicking when the current tile wasn't rendered last frame, we could instead only do it if the current tile can't be rendered this frame. That should avoid the progress bug. And at first thought, it sounds like a small performance improvement, too. It would mean, in the example above, that we start loading tiles C-Z a frame earlier.

However, I think that might actually be a step backwards. In the scenario described above, there's certainly no reason to hold off loading C-Z for another frame. However, what if Tile B has a sibling (B2) that also needs to be loaded? And maybe we can't even render tile B at all until tile B2 is loaded. In other words, both B and B2 will be level 1 kicked and A will be rendered. In that scenario, filling up the load queue with C-Z will actually increase the time before useful content appears on the screen, because tiles we can't use yet (C-Z) will compete for load resources with a tile we desperately need (B2). Waiting until tile B is _really rendered_ before starting the load C-Z avoids this situation, at the cost of just one frame in the case where it's unnecessary.

kring:
To be really clear about what I'm proposing:

- Don't increment `tilesKicked` for tiles removed from the `renderList`.
- Instead, increment it for tiles removed from the `_mainThreadLoadQueue` and `_workerThreadLoadQueue` below that.

PR #734, opened 2023-09-28
Tweak compute load progress

csciguy8:
Addresses cases where `::ComputeLoadProgress` would report 100% done prematurely, before all work in the view had completed. 

(I noticed this in cesium-unreal performance tests, with Google 3D Tiles)

With certain tilesets and our "kicking' logic (`::_kickDescendantsAndRenderTile`), we can run into frames where all work has been kicked out of the main / worker threads, but no new work has been added yet. This could lead to a 100% value being returned, in error.

The fix is to consider frames that kicked tiles as not complete yet. This gives us at least one more view traversal to find remaining work.


kring:
Thanks @csciguy8!

PR #736, opened 2023-10-03
Updated ion authorization url to the current url

weegeekps:
I noticed that the authorization URL used in cesium-native is an old one that's doing a redirect last week. I checked with @mramato and confirmed that we should be using `https://ion.cesium.com/oauth`. This is just a PR to make this update.

P.S. I also added `.idea` to the `.gitignore` to prevent git from trying to scoop up the Clion prefs on my machine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ CMakeSettings.json
 *.DS_Store
 test.db
 build-wsl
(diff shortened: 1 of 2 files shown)


kring:
Thanks @weegeekps!

PR #742, opened 2023-10-20
Install uriparser includes to the right location.

kring:
In cesium-unreal, the uriparser header files were being installed to `Source/ThirdParty/include` instead of `Source/ThirdParty/include/uriparser`. This PR fixes that.


csciguy8 (approved):
Tested functionality in cesium-unreal. Looks good.

PR #744, opened 2023-10-23
Extract Value From URI

joseph-kaile:
Add utility function to extract a value from a URI given a key.

### DIFF
--- a/CesiumIonClient/src/Connection.cpp
+++ b/CesiumIonClient/src/Connection.cpp
@@ -15,6 +15,7 @@
 #include <rapidjson/error/en.h>
 #include <rapidjson/stringbuffer.h>
 #include <rapidjson/writer.h>
+#include <uriparser/Uri.h>
 
 #include <thread>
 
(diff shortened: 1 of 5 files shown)


CesiumIonClient/src/Connection.cpp
```diff
@@ -373,6 +374,43 @@ Asset jsonToAsset(const rapidjson::Value& item) {
...
+          if (parseJsonObject(pResponse, d) && d.IsObject()) {
+            const auto itr = d.FindMember("apiHostname");
+            if (itr != d.MemberEnd() && itr->value.IsString()) {
+              return std::make_optional<std::string>(itr->value.GetString());
+            }
+          }
+        }
+
+        UriUriA newUri;
+        if (uriParseSingleUriA(&newUri, ionUrl.c_str(), nullptr) !=
```
kring, line 402: This result needs to be freed in order to avoid a memory leak.

CesiumIonClient/src/Connection.cpp
```diff
@@ -373,6 +374,43 @@ Asset jsonToAsset(const rapidjson::Value& item) {
...
+  return pAssetAccessor->get(asyncSystem, configUrl)
+      .thenImmediately([ionUrl](std::shared_ptr<IAssetRequest>&& pRequest) {
+        const IAssetResponse* pResponse = pRequest->response();
+        if (pResponse && pResponse->statusCode() >= 200 &&
+            pResponse->statusCode() < 300) {
+          rapidjson::Document d;
+          if (parseJsonObject(pResponse, d) && d.IsObject()) {
+            const auto itr = d.FindMember("apiHostname");
+            if (itr != d.MemberEnd() && itr->value.IsString()) {
+              return std::make_optional<std::string>(itr->value.GetString());
```
kring, line 396: I think this will be just a hostname, e.g., `api.cesium.com`, rather than a full URL, like `https://api.cesium.com`.

CesiumUtility/src/Uri.cpp
```diff
@@ -100,6 +100,37 @@ std::string Uri::addQuery(
   // uriFreeUriMembersA(&baseUri);
 }
 
+std::string Uri::getQueryValue(const std::string& url, const std::string& key) {
```
kring, line 103: It would be worthwhile to write some quick tests for this.

kring (commented):
Looks good, just a few suggestions.

kring:
Thanks @joseph-kaile!

PR #747, opened 2023-11-02
First round of refactoring for enabling use of cesium-native for loading/manipulating/saving 3D Tiles

kring:
Previously, the various classes and functions to load and convert 3D Tiles tile content were considered implementation details of the `Cesium3DTilesSelection` library and couldn't be used from outside it. This PR moves most of that functionality into a new `Cesium3DTilesContent` library and makes it public. This new library doesn't depend on either `Cesium3DTilesSelection` or `Cesium3DTiles`; the idea is it can be used with either of them.

Later PRs will add doc, clean up the APIs, and add some new ones. This first PR is noisy (lots of changed files) but actually does very little.


Cesium3DTilesSelection/test/TestTilesetJsonLoader.cpp
```diff
@@ -17,6 +17,8 @@
 
 using namespace CesiumAsync;
 using namespace Cesium3DTilesSelection;
+using namespace CesiumNativeTests;
+using namespace CesiumNativeTests;
```
j9liu, line 21: Duplicate line
```suggestion
```

j9liu (commented):
Looks good to me @kring ! Just caught a few extraneous lines but I'll remove them on your behalf

PR #751, opened 2023-11-07
Fix logical errors in byteStride calculations

Yeah-Zi:
Fixed loading failure issue when byteStride  description is 0 in buffer view

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@
 
 - Added `getClass` to `PropertyTableView`, `PropertyTextureView`, and `PropertyAttributeView`. This can be used to retrieve the metadata `Class` associated with the view.
 - Added `PropertyViewStatus::EmptyPropertyWithDefault` to indicate when a property contains no data, but has a valid default value.
+- A glTF `bufferView` with a `byteStride` of zero is now treated as if the `byteStride` is not defined at all. Such a glTF technically violates the spec (the minimum value is 4), but the new behavior is sensible enough and consistent with CesiumJS.
 
 ##### Fixes :wrench:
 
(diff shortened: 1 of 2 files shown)


kring:
@Yeah-Zi can you explain more why this is needed? If a `bufferView` has `byteStride` explicitly specified to be zero, that's a broken glTF. The value is _required_ to be greater than or equal to 4:
https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#_bufferview_bytestride

Yeah-Zi:
Thank you for providing the standard format specification for gltf.
Why I need this is because the customer provides data with a byteStride of 0, it can be loaded normally in CesiumJS, while Cesium For Unreal cannot

Yeah-Zi:
@kring Perhaps some log information can be added here, as loading failure does not result in any errors. It took me some time to debug this error

Yeah-Zi:
i found how cesiumjs due with this situation:
https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Scene/GltfPipeline/getAccessorByteStride.js

kring:
> Perhaps some log information can be added here, as loading failure does not result in any errors.

I'm not opposed to adding some more logging but, in general, cesium-native is a loader, not a validator. A proper glTF validator (like this one: https://github.khronos.org/glTF-Validator/) should easily detect invalid glTF files like this.

However, in this case, since your change matches what CesiumJS is already doing, and it has very little cost, I'm happy to merge it. Can you please sign the Contributor License Agreement first:
https://github.com/CesiumGS/cesium/blob/master/CONTRIBUTING.md#contributor-license-agreement-cla

Thank you for the PR!

Yeah-Zi:
@kring I have signed the CLA

kring:
Thanks @Yeah-Zi!

PR #750, opened 2023-11-06
3D Tiles bounding volume improvements

kring:
This is a PR into #747 so merge that first.

Fixes #647 

- Added `TileBoundingVolumes` class to `Cesium3DTilesContent`, making it easier to create the rich bounding volume types in `CesiumGeometry` and `CesiumGeospatial` from the simple `std::vector` representations in `Cesium3DTiles`
- Added `transform` method to `CesiumGeometry::BoundingSphere`.
- Added `toSphere`, `fromSphere`, and `fromAxisAligned` methods to `CesiumGeometry::OrientedBoundingBox`.
- Added `TileTransform` class to `Cesium3DTilesContent`, making it easier to create a `glm::dmat4` from the `transform` property of a `Cesium3DTiles::Tile`.

PR #752, opened 2023-11-08
Expose implicit tiling related functionality

kring:
This is a PR into #750 so merge that first.

Previously, a lot of logic for working with implicit tilesets was buried in `ImplicitQuadtreeLoader` and `ImplicitOctreeLoader`, which are closely tied to the selection algorithm. With this PR, a lot of the functionality is now available in the `Cesium3DTilesContent` library and can be used independent of selection. This is useful when programmatically loading and manipulating a 3D Tiles implicit tileset.


Cesium3DTilesContent/include/Cesium3DTilesContent/ImplicitTilingUtilities.h
```diff
@@ -0,0 +1,346 @@
...
+   * @param tileID The tile ID of the parent tile for which to get children.
+   * @return The children.
+   */
+  static OctreeChildren
+  getChildren(const CesiumGeometry::OctreeTileID& tileID) noexcept {
+    return OctreeChildren{tileID};
+  }
+
+  /**
+   * @brief Computes the bounding volume for an implicit tile with the given ID.
```
j9liu, line 291: I think these comments could use a few extra words so it's easier to distinguish the functions at a glance. I'll add some changes in a commit.

j9liu (commented):
@kring looks good to me! I'm not familiar with subtree math so I couldn't verify the mathematical correctness of everything. It looks like all the same tests pass, though, so that's probably good enough :')

I noticed two things but I'll just push a commit for them, then merge once CI passes.

PR #755, opened 2023-11-13
Fix `PropertyAttributeView::forEachProperty`

lilleyse:
Fixes `PropertyAttributeView::forEachProperty`. The parameter order just needed to be swapped.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,10 @@
 
 - Added `Cesium3DTilesContent` library and namespace. It has classes for loading, converting, and manipulating 3D Tiles tile content.
 
+##### Fixes :wrench:
+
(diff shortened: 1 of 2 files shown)


csciguy8:
Looks like a good fix @lilleyse.

Would be nice if our unit tests caught these compile errors first though. @j9liu any ideas?

j9liu:
@csciguy8 Yeah I didn't write tests for `forEachProperty` for any of the property views. It seemed a little awkward to test because each property is of a different type, so if we were going to thoroughly test multiple properties then you'd have to write a super specific / big callback.

I don't think that it should fall on @lilleyse to fix that so I'm happy to merge once CI passes. Thanks for the catch!

PR #756, opened 2023-11-13
Add conversion logic to cesium-native

j9liu:
`CesiumMetadataConversions.h` in Cesium for Unreal contains many conversions between metadata value types. Similar logic will be needed for the metadata systems in other engines, like Unity. Therefore, I've copied many of the functions over to Native from Cesium for Unreal.

Notes:
- Instead of using the `defaultValue` parameter, I have decided to make the functions return a `std::optional`.  Using a default value obfuscates the meaning of the returned value; if the conversion was not successful, there's no reliable way of telling. The default value behavior can always be implemented in the runtime engines using `value_or` on the returned optional.
- `vecN` -> string and `matN` -> string methods are included, and it uses `glm::to_string`. However, every game engine has a different way of printing vectors / matrices to the console, so ultimately this method may not be used. It's included here for completeness.
- Along the same vein, string -> `vecN` is not included because of the different engine formats. It can be implemented by the Cesium for Unreal plugin since Unreal already has functions to construct vectors from strings. Whether it's worth doing for Unity is TBD.


kring:
Thanks @j9liu!

PR #754, opened 2023-11-13
Improve support for reading implicit tiling subtrees, and manipulating them after they're loaded

kring:
This is a PR into #752 so merge that first.

This PR makes "querying" of implicit tile availability usable from outside the Cesium3DTilesSelection library and the selection-specific "loader" classes. It also adds the ability to modify availability information that is already loaded.

PR #757, opened 2023-11-14
Create CesiumRasterOverlays library.

kring:
This PR builds on #754 so merge that first.

Like #747, this PR doesn't do much other than move things around. A followup PR will make some smaller changes to raster overlays to allow them to be used outside of Cesium3DTilesSelection. Specifically, it will be possible to bake raster overlays into a tile glTF.


CHANGES.md
```diff
@@ -4,12 +4,13 @@
...
 - Moved `GltfUtilities` from `Cesium3DTilesSelection` to `Cesium3DTilesContent`.
 - Moved `createRasterOverlayTextureCoordinates` method from `GltfUtilities` to a new `RasterOverlayUtilities` class in the `Cesium3DTilesSelection` library.
 - `GltfUtilities::parseGltfCopyright` now returns the credits as a vector of string_views. Previously it took a `CreditSystem` and created credits directly.
 - The `SubtreeAvailability` constructor and `loadSubtree` static method now take an `ImplicitTileSubdivisionScheme` enumeration parameter instead of a `powerOf2` parameter. They also now require a `levelsInSubtree` parameter, which is needed when switching from constant to bitstream availability. Lastly, the constructor now takes a `Subtree` parameter instead of a `std::vector<std::vector<std::byte>>` representing the buffers.
 - `SubtreeConstantAvailability`, `SubtreeBufferViewAvailability`, and `AvailabilityView` are now members of `SubtreeAvailability`.
+- Moved `RasterOverlay`, `RasterOverlayTileProvider`, `RasterOverlayTile`, `QuadtreeRasterOverlayTileProvider`, and all of the `RasterOverlay`-derived types to a new `CesiumRasterOverlays` library and namespace.
```
j9liu, line 13:
```suggestion
- Moved `RasterOverlay`, `RasterOverlayTileProvider`, `RasterOverlayTile`, `QuadtreeRasterOverlayTileProvider`, `RasterOverlayLoadFailure`, and all of the `RasterOverlay`-derived types to a new `CesiumRasterOverlays` library and namespace.
```

CesiumRasterOverlays/include/CesiumRasterOverlays/IPrepareRasterOverlayRendererResources.h
```diff
@@ -0,0 +1,74 @@
...
+   */
+  virtual void* prepareRasterInMainThread(
+      RasterOverlayTile& rasterTile,
+      void* pLoadThreadResult) = 0;
+
+  /**
+   * @brief Frees previously-prepared renderer resources for a raster tile.
+   *
+   * This method is always called from the thread that called
+   * {@link Tileset::updateView} or deleted the tileset.
```
j9liu, line 58: Now that the functions have been separated from the tileset implementation, is this still true?

j9liu (commented):
@kring I had a gripe about the way that `IPrepareRendererResources` inherits from `IPrepareRasterOverlayRendererResources`, but that's the only major thing. Let me know what you think!

Cesium3DTilesSelection/include/Cesium3DTilesSelection/IPrepareRendererResources.h
```diff
@@ -46,7 +50,8 @@ struct TileLoadResultAndRenderResources {
  * {@link TilesetExternals} structure that can be obtained
  * via {@link Tileset::getExternals}.
  */
-class CESIUM3DTILESSELECTION_API IPrepareRendererResources {
+class CESIUM3DTILESSELECTION_API IPrepareRendererResources
+    : public CesiumRasterOverlays::IPrepareRasterOverlayRendererResources {
```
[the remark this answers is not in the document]
kring: Certainly inheritance should model IS-A. In this case, we're saying "a thing that can create renderer resources for a tileset _IS A_ thing that can create renderer resources for a raster overlay." Which is true enough. And composition doesn't really work for an interface like this, because an interface can't "contain" anything. `IPrepareRendererResources` could have a method that gets `IPrepareRasterOverlayRendererResources` instead of inheriting from it, which would at least allow implementors to compose the implementation of one inside the other. But then that either breaks backward compatibility (by removing the `prepareRasterInLoadThread` method and friends from `IPrepareRendererResources`), or it ends up with `IPrepareRendererResources` being a `IPrepareRasterOverlayRendererResources` anyway (by leaving `prepareRasterInLoadThread` and friends, which also exist on this interface). In the latter case, we end up modeling IS-A anyway, so why make our lives harder by fighting it?

kring:
I've addressed all the comments, ready for another look.

csciguy8 (approved):
Updated documentation looks good. Merging.

PR #759, opened 2023-11-14
Fix property related docs

lilleyse:
The main thing here is renaming `propertyName` to `propertyId` for `getClassProperty`, `getPropertyView`, and related functions. This is to disambiguate the property id (the dictionary key) from the property name (its `name` field). When calling these functions you want to pass in the property id.

I also fixed some other doc issues that I noticed along the way.


j9liu:
@lilleyse Looks great, thanks for all the fixes!

PR #763, opened 2023-11-17
Add thenPassThrough method to Future and SharedFuture

kring:
When using cesium-native's AsyncSystem, it often happens that you have some values computed inside one continuation that you want to pass to the next continuation in the chain. You can do this by bundling them up in a struct and returning that struct from the current continuation. Or even use a `std::tuple`, if defining a struct seems like too much ceremony. 

Unfortunately, that breaks down when the continuation itself returns another future. For example:

```cpp
whatever.thenInMainThread([](Whatever&& whatever) {
  SomeType value = computeSomeValue(whatever);
  return startAnAsyncOperation(value);
}).thenInMainThread([](AsyncOperationResult&& result) {
  // ... do something with the result of the async operation
});
```

What if the "do something with the result of the async operation" also requires access to the `SomeType value`? It gets awkward fast! Here's how it can be done:

```cpp
whatever.thenInMainThread([](Whatever&& whatever) {
  SomeType value = computeSomeValue(whatever);
  return startAnAsyncOperation(value)
    .thenImmediately([value = std::move(value)](AsyncOperationResult&& result) mutable {
      return std::make_tuple(std::move(value), std::move(result));
    });
}).thenInMainThread([](std::tuple<SomeType, AsyncOperationResult>>&& tuple) {
  auto& [value, result] = tuple;
  // now we have access to both the SomeType value and the async op result
});
```

It works, but it's pretty ugly, and there are a lot of details to get right. Forgetting the mutable or any of the std::moves will make it a lot less efficient.

This PR adds a new method, `thenPassThrough`, on `Future` and `SharedFuture` that makes this a little easier to deal with:

```cpp
whatever.thenInMainThread([](Whatever&& whatever) {
  SomeType value = computeSomeValue(whatever);
  return startAnAsyncOperation(value)
    .thenPassThrough(std::move(value));
}).thenInMainThread([](std::tuple<SomeType, AsyncOperationResult>>&& tuple) {
  auto& [value, result] = tuple;
  // now we have access to both the SomeType value and the async op result
});
```

You can pass as many parameters as you like to `thenPassThrough`, and they all get wrapped up in the tuple. There are a lot less details to get right.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@
 - Added `TileTransform` class to `Cesium3DTilesContent`, making it easier to create a `glm::dmat4` from the `transform` property of a `Cesium3DTiles::Tile`.
 - Added `ImplicitTilingUtilities` class to `Cesium3DTilesContent`.
 - Added overloads of `isTileAvailable` and `isContentAvailable` on the `SubtreeAvailability` class that take the subtree root tile ID and the tile ID of interest, instead of a relative level and Morton index.
+- Added `Future<T>::thenPassThrough`, used to easily pass additional values through to the next continuation.
 
 ##### Fixes :wrench:
 
--- a/CesiumAsync/include/CesiumAsync/Future.h
+++ b/CesiumAsync/include/CesiumAsync/Future.h
@@ -202,6 +202,30 @@ public:
   }
 
   /**
+   * @brief Passes through one or more additional values to the next
+   * continuation.
+   *
+   * The next continuation will receive a tuple with each of the provided
+   * values, followed by the result of the current Future.
+   *
+   * @tparam TPassThrough The types to pass through to the next continuation.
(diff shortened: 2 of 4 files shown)


csciguy8 (approved):
Looks like a good option if you need it.

Do you have any specific use cases in mind?

In the example you gave, my only thought was that if `value` was really needed in the lambda that handles the asycn op result, shouldn't we just add `value` to `AsyncOperationResult`?

kring:
Here's a case where it's helpful (and the reason I added it):
https://github.com/CesiumGS/cesium-native/pull/766/files#diff-30d11ceb9f485751cb1aefabb5ce7185a7500cb59e15d2e10692acb8a587728dR159

csciguy8:
Nice, I can see how the code is much cleaner by using this.

PR #764, opened 2023-11-17
Add `AccessorUtility.h`

j9liu:
This PR adds `AccessorUtility.h`, which is just the `GltfAccessors.h` file from Cesium for Unreal. It provides `typedef`s for the possible types of feature ID, index, and texcoord accessors, as well as visitors to retrieve their data. This is useful for retrieving feature IDs and texture coordinates without requiring them to be baked into mesh data, enabling picking.

This is a PR into #756 , so merge that first.


kring:
This file is still using Unreal conventions for function names (starting with a capital letter) instead of the cesium-native convention (starting with a lowercase letter). Other than that, this looks good to me!

kring:
Thanks @j9liu!

PR #767, opened 2023-11-20
Fix VS 2022 17.8 warnings/errors by defining _SILENCE_STDEXT_ARR_ITERS_DEPRECATION_WARNING

kring:
Longer term, we should upgrade fmt. Our copy is included in spdlog.
https://github.com/fmtlib/fmt/issues/3540


kring:
I'm merging this because every branch is broken without it after upgrading to VS 2022 17.8.

PR #771, opened 2023-11-28
Fix VS 2022 17.8 warnings/errors in v0.27.x by defining _SILENCE_STDEXT_ARR_ITERS_DEPRECATION_WARNING

kring:
This is the same as #767 except it's a PR into the v0.27.x branch, which we're still using in Cesium for Unity until we can update to the new metadata system.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### v0.27.4 - ?
+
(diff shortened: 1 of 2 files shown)


csciguy8 (approved):
Verified these are the same changes. Looks good. Merging....

PR #773, opened 2023-12-04
Add support for new defaults service to CesiumIonClient.

kring:
This is a PR into #772 so merge that first.

Adds support for Cesium ion's new `defaults` service which will be used to populate the Quick Add panels. The new service is not available in the public Cesium ion just yet, though.

### DIFF
--- a/CesiumIonClient/include/CesiumIonClient/Defaults.h
+++ b/CesiumIonClient/include/CesiumIonClient/Defaults.h
@@ -1,0 +1,95 @@
+#pragma once
+
+namespace CesiumIonClient {
+
+/**
+ * @brief The default assets.
+ */
+struct DefaultAssets {
+  /**
+   * @brief The asset id of the default imagery asset.
+   */
+  int64_t imagery = -1;
+
+  /**
+   * @brief The asset id of the default terrain asset.
+   */
+  int64_t terrain = -1;
+
+  /**
+   * @brief The asset id of the default buildings asset.
+   */
+  int64_t buildings = -1;
+};
+
+/**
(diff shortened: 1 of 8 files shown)


CesiumIonClient/include/CesiumIonClient/Defaults.h
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * @brief The raster overlays available for this asset
+   */
+  std::vector<QuickAddRasterOverlay> rasterOverlays{};
+};
+
+struct Defaults {
+  DefaultAssets defaultAssets{};
+  std::vector<QuickAddAsset> quickAddAssets{};
+};
```
j9liu, line 93: Is `defaultAssets` or the `DefaultAssets` struct actually used for anything? Can we remove it and rename `Defaults -> DefaultAssets`? It would make this more intuitive to read in the Unreal Editor code

j9liu: I know this rename would be wide-reaching, so if you agree with the decision I am happy to do the name replacements on the Unreal / Unity side

kring: We don't currently plan to use the `DefaultAssets` for anything in Unreal or Unity, at least, but it's part of ion's `/v1/defaults` service, so we should include it. Also, as mentioned offline, I think we need to stick to the names that the service documentation uses, so we shouldn't rename `Defaults` to something else. It's namespaced (`CesiumIonClient::Defaults`) so hopefully not _too_ bad.

Issue #348, opened 2021-09-18
Functions are declared noexcept, even though they may throw


javagl:
This issue is a special instance of https://github.com/CesiumGS/cesium-native/issues/30 , and refers to warnings

> Warning C26447 The function is declared 'noexcept' but calls function ... which may throw exceptions (f.6).

I think that these are somewhat "critical", because throwing from a `noexcept` will trigger an `std::terminate()`, and this is... usually not desired. 

There are currently 118 functions that have this problem, which I'll just dump here, to give an overview:

```
Warning	C26447	The function is declared 'noexcept' but calls function 'normalize<3,double,0>()' which may throw exceptions (f.6).	CesiumGeospatial	C:\Dev\cesium-native\CesiumGeospatial\src\Ellipsoid.cpp	14	
Warning	C26447	The function is declared 'noexcept' but calls function 'mix<double,double>()' which may throw exceptions (f.6).	CesiumGeospatial	C:\Dev\cesium-native\CesiumUtility\include\CesiumUtility\Math.h	329	
Warning	C26447	The function is declared 'noexcept' but calls function 'operator "" s()' which may throw exceptions (f.6).	Cesium3DTilesSelection	C:\Dev\cesium-native\Cesium3DTilesSelection\src\Tile.cpp	35	
Warning	C26447	The function is declared 'noexcept' but calls function 'all_of<std::_Vector_const_iterator<std::_Vector_val<std::_Simple_types<Cesium3DTilesSelection::RasterMappedTo3DTile> > >,<lambda_338a4d015d4c98edbecfa7a0bc4540b3> >()' which may throw exceptions (f.6).	Cesium3DTilesSelection	C:\Dev\cesium-native\Cesium3DTilesSelection\src\Tile.cpp	110	
Warning	C26447	The function is declared 'noexcept' but calls function 'make_unique<Cesium3DTilesSelection::RasterOverlayTile,Cesium3DTilesSelection::RasterOverlay & __ptr64,0>()' which may throw exceptions (f.6).	Cesium3DTilesSelection	C:\Dev\cesium-native\Cesium3DTilesSelection\src\RasterOverlayTileProvider.cpp	34	
[...]

PR #777, opened 2023-12-13
Add triangle/ray and gltf/ray intersect function with tests

joseph-kaile:
Added two new utility functions, `rayTriangle` and `intersectRayGltfModel`. They have unit tests to verify their correctness.

This PR can really be thought of as two distinct efforts:
* The implementation of ray triangle intersections in a mathematically correct and performant way
* glTF traversals that are robust and support all reasonable use cases. A good bit of coding was devoted this. Also notable, the glTF test data was generated using python scripts.


javagl:
- The part at https://github.com/CesiumGS/cesium-native/pull/777/files#diff-6fc212582e9b885efd06f87042cb7db30c537e565c2369d606ad8d968c83178bR271 seems to cover `TRIANGLE` and `TRIANGLE_STRIP`, but the latter via the `else` case. It does not cover `TRIANGLE_FAN`. The `LINE...` cases should probably be ignored.
- The `indices` are assumed to be present at https://github.com/CesiumGS/cesium-native/pull/777/files#diff-6fc212582e9b885efd06f87042cb7db30c537e565c2369d606ad8d968c83178bR264 - the geometry may be non-indexed
- The point about [`EXT_mesh_gpu_instancing`](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Vendor/EXT_mesh_gpu_instancing/README.md) will come up sooner or later, but may not be in scope yet...

CesiumGltfContent/include/CesiumGltfContent/GltfUtilities.h
```diff
@@ -121,5 +125,23 @@ struct CESIUMGLTFCONTENT_API GltfUtilities {
...
+   * This function only handles primitives with TRIANGLES or TRIANGLE_FAN mode.
+   * Other modes are ignored.
+   *
+   * @param ray The ray.
+   * @param gltf The glTF model.
+   * @param cullBackFaces An optional boolean flag to indicate whether to cull
+   * backfaces or not. Defaults to true.
+   * @param return The intersection point along the ray, if any.
+   */
+  static std::optional<glm::dvec3> intersectRayGltfModel(
```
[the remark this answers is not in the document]
javagl: *Iff* (if and only iff) the barycentric coordinates fall out of the intersection test, that could also be useful. 
(If not, they can be computed post-hoc where necessary, but ... if my dusty memories are not misleading me, I think that this is something that comes "(nearly) for free" for some implementations of ray/triangle intersection tests)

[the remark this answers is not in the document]
csciguy8: The use case that I thought was useful was if you have a hit, and needed traverse back into the model to do more.

For example, I hit something, and have a hit point, but want to know the entire triangle I hit because I want to show it to the user. Similarly, you hit something, and want to show the whole mesh that was hit.

I've made changes to this function to return more info (`HitResult` struct) and modified the tests to check that we can dive into the model with this info after the fact.

CesiumGltfContent/src/GltfUtilities.cpp
```diff
@@ -971,6 +970,11 @@ void intersectRayScenePrimitive(
   if (!pPositionAccessor)
     return;
 
+  // Ignore non-float positions, these are unsupported
+  // Although valid from the glTF spec, they aren't generally useful here
+  if (pPositionAccessor->componentType != Accessor::ComponentType::FLOAT)
```
javagl, line 975: According to the table at https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#meshes-overview, the `POSITION` accessor must use VEC3/float. But 1. it can still make sense to have this check here (and maybe even log this as an error instead of "doing nothing"), and 2. maybe there are extensions that relax this constraint (like the (pending!) extension at https://github.com/KhronosGroup/glTF/pull/2397 )

csciguy8: Nice catch! I missed that part of the spec. I'll update the comments with your notes.

It could be nice to have some warnings logged somewhere when traversing a gltf with bogus values. There are other checks like this that "do nothing" and try our best when encountering bad data.

lilleyse: ... unless the glTF is using [KHR_mesh_quantization](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_mesh_quantization/README.md)

csciguy8: Thanks, I'll add a note about that extension, and see how hard it would be to extend this PR to cover it.

javagl: @lilleyse That was one reason why I mentioned extensions (although I might have been more explicit)
(I would have to refresh my memory about the `AccessorView` - I don't know whether this already decodes quantized data - i.e. where _exactly_ the quantization extension is handled - but being aware of that point might be enough...)

csciguy8: It wasn't too bad to support quantized meshes (non-float positions). Change have been committed.

csciguy8:
Opening this up for review (non-draft).

All previous comments should have been addressed and any known major functionality is done.

I may add some more commits from self-review later on, but it's ready for some eyes on it.

CesiumGeometry/include/CesiumGeometry/IntersectionTests.h
```diff
@@ -94,6 +97,129 @@ class CESIUMGEOMETRY_API IntersectionTests final {
...
+   * @param cullBackFaces Ignore triangles that face away from ray.
+   * @return True if a hit occurred, false otherwise.
+   */
+  static bool rayTriangleParametric(
+      const Ray& ray,
+      const glm::dvec3& p0,
+      const glm::dvec3& p1,
+      const glm::dvec3& p2,
+      double& t,
+      bool cullBackFaces = false);
```
kring, line 141: `rayTriangle` returns an optional, while `rayTriangleParametric` returns a boolean plus an out parameter. Is there a good reason for this difference? The optional seems like the cleaner interface to me, unless we know it to have a significant performance impact (I know this function can be quite performance critical).

CesiumGeometry/include/CesiumGeometry/Ray.h
```diff
@@ -2,6 +2,7 @@
 
 #include "Library.h"
 
+#include <glm/mat4x4.hpp>
```
kring, line 5: Better to include `<glm/fwd.hpp>`. It might be possible to remove the `glm/vec3.hpp` below as well.

CesiumGeometry/src/IntersectionTests.cpp
```diff
@@ -100,6 +106,209 @@ std::optional<glm::dvec2> IntersectionTests::rayEllipsoid(
...
+}
+
+bool IntersectionTests::rayOBBParametric(
+    const Ray& ray,
+    const OrientedBoundingBox& obb,
+    double& t) {
+
+  const glm::dmat3x3& inverseHalfAxis = obb.getInverseHalfAxes();
+  glm::dmat4x4 transformation(
+      glm::normalize(glm::dvec4(inverseHalfAxis[0], 0.0)),
```
kring, line 232: Bit of a nitpick, but swapping the order of the normalize and dvec4 construction would save a multiply and an add (for each column) by normalizing a dvec3 instead of a dvec4.

CesiumGeometry/src/Ray.cpp
```diff
@@ -19,6 +20,16 @@ Ray::Ray(const glm::dvec3& origin, const glm::dvec3& direction)
   //>>includeEnd('debug');
 }
 
+glm::dvec3 Ray::pointFromDistance(double distance) const noexcept {
+  return _origin + distance * _direction;
```
kring, line 24: Nitpick!
```suggestion
  return this->_origin + distance * this->_direction;
```

CesiumGeometry/src/Ray.cpp
```diff
@@ -19,6 +20,16 @@ Ray::Ray(const glm::dvec3& origin, const glm::dvec3& direction)
...
+}
+
+Ray Ray::transform(const glm::dmat4x4& transformation) const noexcept {
+  return Ray(
+      glm::dvec3(transformation * glm::dvec4(_origin, 1.0)),
+      glm::normalize(glm::dvec3(transformation * glm::dvec4(_direction, 0.0))));
```
kring, line 30:
```suggestion
      glm::dvec3(transformation * glm::dvec4(this->_origin, 1.0)),
      glm::normalize(glm::dvec3(transformation * glm::dvec4(this->_direction, 0.0))));
```

CesiumGltf/include/CesiumGltf/Model.h
```diff
@@ -117,7 +117,9 @@ struct CESIUMGLTF_API Model : public ModelSpec {
       Model& gltf,
       Node& node,
       Mesh& mesh,
+      const int meshId,
       MeshPrimitive& primitive,
+      const int primitiveId,
```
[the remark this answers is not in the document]
kring: Also please avoid `int` and use the appropriate specific size type. In this case it should either by `size_t` (if we think of these as indices into a vector) or `int32_t` (if we think of these as glTF object IDs).

CesiumGltfContent/include/CesiumGltfContent/GltfUtilities.h
```diff
@@ -191,5 +195,44 @@ struct CESIUMGLTFCONTENT_API GltfUtilities {
...
+  struct IntersectResult {
+    struct Hit {
+      glm::dvec3 primitivePoint = {};
+      glm::dmat4x4 primitiveToWorld = {};
+
+      glm::dvec3 worldPoint = {};
+      double rayToWorldPointDistanceSq = -1;
+
+      int meshId = -1;
+      int primitiveId = -1;
```
kring, line 211:
```suggestion
      int32_t meshId = -1;
      int32_t primitiveId = -1;
```

CesiumGltfContent/include/CesiumGltfContent/GltfUtilities.h
```diff
@@ -191,5 +195,44 @@ struct CESIUMGLTFCONTENT_API GltfUtilities {
    * @param bufferIndex The index of the buffer to compact.
    */
   static void compactBuffer(CesiumGltf::Model& gltf, int32_t bufferIndex);
+
+  /**
+   * @brief Hit result data for intersectRayGltfModelParametric
+   */
+  struct IntersectResult {
+    struct Hit {
+      glm::dvec3 primitivePoint = {};
```
kring, line 204: These fields could use some doc comments.

CesiumGltfContent/include/CesiumGltfContent/GltfUtilities.h
```diff
@@ -191,5 +195,44 @@ struct CESIUMGLTFCONTENT_API GltfUtilities {
...
+  /**
+   * @brief Intersects a ray with a glTF model and returns the first
+   * intersection point.
+   *
+   * Supports all mesh primitive modes.
+   * Points and lines are assumed to have no area, and are ignored
+   *
+   * @param ray A ray in world space.
+   * @param gltf The glTF model to intersect.
+   * @param cullBackFaces Ignore triangles that face away from ray.
```
kring, line 228: Perhaps the default should be to do what the primitive's material says to do?
https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#double-sided

In that case you'd need an enumeration instead of a bool if you also want to let the user override it.

CesiumGltfContent/include/CesiumGltfContent/GltfUtilities.h
```diff
@@ -191,5 +195,44 @@ struct CESIUMGLTFCONTENT_API GltfUtilities {
...
+   * @param gltfTransform Optional matrix to apply to entire gltf model.
+   * @param return IntersectResult describing outcome
+   */
+  static IntersectResult intersectRayGltfModel(
```
kring, line 232: Should we provide a function that returns all of the intersections, not just the first one? A common use-case for that is ignoring hits against translucent geometry.

CesiumGltfContent/src/GltfUtilities.cpp
```diff
@@ -901,4 +1228,265 @@ void GltfUtilities::compactBuffer(
...
+        // Bail on invalid view
+        if (positionView.status() != AccessorViewStatus::Valid) {
+          warnings.push_back(
+              "Skipping mesh with an invalid position component type");
+          return;
+        }
+
+        bool hasIndexedTriangles = primitive.indices != -1;
+        if (hasIndexedTriangles) {
+          assert(primitive.indices >= 0);
```
kring, line 1318: An assertion is not appropriate here because this is user-provided data. You must check - even in release builds - that indices is valid before using it to index into model.accessors.

There's a getSafe method on Model to make this easy.

CesiumGltfContent/src/GltfUtilities.cpp
```diff
@@ -901,4 +1228,265 @@ void GltfUtilities::compactBuffer(
...
+                // Bail on invalid view
+                if (indexView.status() != AccessorViewStatus::Valid) {
+                  warnings.push_back(
+                      "Could not create accessor view for mesh indices");
+                  return;
+                }
+
+                findClosestIndexedRayHit<
+                    decltype(positionView),
+                    decltype(indexView)>(
```
kring, line 1349: The compiler should be able to deduce these template arguments automatically, no need for this awkward explicit instantiation with decltypes.

CesiumGltfContent/src/GltfUtilities.cpp
```diff
@@ -901,4 +1228,265 @@ void GltfUtilities::compactBuffer(
...
+              min[1],
+              min[2],
+              max[0],
+              max[1],
+              max[2]),
+          t)) {
+    return std::optional<glm::dvec3>();
+  }
+
+  double tClosest = -1;
```
kring, line 1291:
```suggestion
  double tClosest = -1.0;
```

CesiumGltfContent/src/GltfUtilities.cpp
```diff
@@ -901,4 +1228,265 @@ void GltfUtilities::compactBuffer(
...
+          findClosestRayHit<decltype(positionView)>(
+              transformedRay,
+              positionView,
+              primitive,
+              cullBackFaces,
+              tClosest,
+              warnings);
+        }
+      });
+  assert(tClosest >= -1);
```
kring, line 1369:
```suggestion
  assert(tClosest >= -1.0);
```

Seems nitpicky, and maybe it is, but without the `.0` you're basically relying on an int->double conversion.

CesiumGltfContent/src/GltfUtilities.cpp
```diff
@@ -901,4 +1228,265 @@ void GltfUtilities::compactBuffer(
...
+
+  if (tClosest == -1)
```
kring, line 1371:
```suggestion
  if (tClosest == -1.0)
```

CesiumGltfContent/src/GltfUtilities.cpp
```diff
@@ -901,4 +1228,265 @@ void GltfUtilities::compactBuffer(
...
+            homogeneousWorldPoint.y,
+            homogeneousWorldPoint.z);
+
+        glm::dvec3 rayToWorldPoint = worldPoint - ray.getOrigin();
+        double rayToWorldPointDistanceSq =
+            glm::dot(rayToWorldPoint, rayToWorldPoint);
+
+        // Use as closest if it's first, or closer
+        bool setClosest =
+            closestHit.rayToWorldPointDistanceSq == -1 ||
```
kring, line 1474:
```suggestion
            closestHit.rayToWorldPointDistanceSq == -1.0 ||
```

CesiumGltfContent/src/GltfUtilities.cpp
```diff
@@ -901,4 +1228,265 @@ void GltfUtilities::compactBuffer(
...
+          closestHit.primitivePoint = std::move(*primitiveHitPoint);
+          closestHit.primitiveToWorld = std::move(primitiveToWorld);
+          closestHit.worldPoint = std::move(worldPoint);
+          closestHit.rayToWorldPointDistanceSq = rayToWorldPointDistanceSq;
+          closestHit.meshId = meshId;
+          closestHit.primitiveId = primitiveId;
+        }
+      });
+
+  if (closestHit.rayToWorldPointDistanceSq != -1)
```
kring, line 1486:
```suggestion
  if (closestHit.rayToWorldPointDistanceSq != -1.0)
```

CesiumGltfContent/src/GltfUtilities.cpp
```diff
@@ -901,4 +1228,265 @@ void GltfUtilities::compactBuffer(
...
+GltfUtilities::IntersectResult GltfUtilities::intersectRayGltfModel(
+    const CesiumGeometry::Ray& ray,
+    const CesiumGltf::Model& gltf,
+    bool cullBackFaces,
+    const glm::dmat4x4& gltfTransform) {
+  glm::dmat4x4 rootTransform = applyRtcCenter(gltf, gltfTransform);
+  rootTransform = applyGltfUpAxisTransform(gltf, rootTransform);
+
+  IntersectResult result;
+  IntersectResult::Hit closestHit;
```
kring, line 1389: Minor thing, but it seems like it would be cleaner to capture `&result` in the lambda below, and then use the hit embedded in it. That avoids the need to compare `rayToWorldPointDistanceSq` to -1.0 to see if there's already a hit (instead you can use the optional), and it avoids the need to move the hit into the actual result at the end.

CesiumGltfContent/test/TestIntersectRayGltf.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+  std::streamsize size = file.tellg();
+  file.seekg(0, std::ios::beg);
+
+  std::vector<std::byte> buffer(static_cast<size_t>(size));
+  file.read(reinterpret_cast<char*>(buffer.data()), size);
+
+  return buffer;
+}
+} // namespace
```
kring, line 33: Use the implementation of this already in `readFile.h` / `readFile.cpp` instead of copying it here.

kring (changes requested):
A few comments below, but overall looks great! I appreciate the extensive testing.

CesiumGltf/include/CesiumGltf/Model.h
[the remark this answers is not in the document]
lilleyse: For consistency shouldn't `node` also become `nodeId`?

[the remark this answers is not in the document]
lilleyse: If we're switching to indices everywhere, `ForEachNodeInSceneCallback` should also be updated.

joseph-kaile:
> Opening this up for review (non-draft).
> 
> 
> 
> All previous comments should have been addressed and any known major functionality is done.
> 
> 
> 
> I may add some more commits from self-review later on, but it's ready for some eyes on it.

Hello team,

I (and ChatGPT) noticed an edge case not handled. In the ray-OBB code, halfAxes could be singular. See the comment in the OBB constructor. I would recommend rewriting it to not use halfAxis.

Thanks

kring:
Hi @joseph-kaile, good to hear from you, hope you're doing well! Thanks for the note about the edge case. Are you referring to `IntersectionTests::rayOBBParametric` specifically?

joseph-kaile:
> Hi @joseph-kaile, good to hear from you, hope you're doing well! Thanks for the note about the edge case. Are you referring to `IntersectionTests::rayOBBParametric` specifically?

I’m doing well, thanks. Yes, I’m referring to that function.

I found this out recently after intersecting the large root tiles of CWT.

kring:
> I found this out recently after intersecting the large root tiles of CWT.

Ok, cool. It's not obvious to me why the CWT root tile OBB half axes would be singular. But we should probably handle that possibility in the intersect function in any case.

javagl:
The question about the handling of (near) singular matrices is essentially leaked through from [an old, unaddressed `TODO`](https://github.com/CesiumGS/cesium-native/blob/5513845b2cc69f9f88fdeac56c5c7b3a1dc69445/CesiumGeometry/include/CesiumGeometry/OrientedBoundingBox.h#L39). My understanding is that this is undefined behavior and a crash waiting to happen. 

But... one could argue whether this should be addressed in _this_ PR or ... more holistically. Yes, the `glm::inverse` function [brutally divides by (potentially) zero](https://github.com/g-truc/glm/blob/33b4a621a697a305bc3a7610d290677b96beb181/glm/gtc/matrix_inverse.inl#L114). The fact that the`OrientedBoundingBox` constructor is declared `noexcept` therefore raises a bunch of questions, somewhat related to https://github.com/CesiumGS/cesium-native/issues/348 and the [comment in the glm docs](https://github.com/g-truc/glm/blob/master/manual.md#-710-why-some-glm-functions-can-crash-because-of-division-by-zero) that I already linked to. Still, the code in this PR just _relied_ on existing functionality. Nobody can expect someone who uses the `OrientedBoundingBox` to look into the constructor code and see whether there might be an open `TODO` (of which nobody knows how to resolve it).

(But of course, all that is not on me to decide...)

CHANGES.md
```diff
@@ -172,6 +172,11 @@
 
 - Add `defaults` method to `CesiumIonClient::Connection`.
 
+##### Additions :tada:
+
+- Added `rayTriangle` intersection function that returns the intersection point between a ray and a triangle.
+- Added `intersectRayGltfModel` intersection function that returns the first intersection point between a ray and a glTF model.
+
```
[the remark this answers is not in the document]
csciguy8: Fixed

CesiumGltf/include/CesiumGltf/Model.h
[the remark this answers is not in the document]
csciguy8: > Adding explicit index parameters is fine, but it's definitely a breaking change, so it needs to be documented as such.
> Also please avoid int and use the appropriate specific size type

Done and done

CesiumGeometry/include/CesiumGeometry/IntersectionTests.h
csciguy8: `rayTriangle` is definitely cleaner.

`rayTriangleParametric` is a little different, in that it's testing an infinite ray. It will return true (hit), even if the hit is behind the ray origin. The caller can interpret this as finite or not, which is what `rayTriangle` is doing

CesiumGeometry/include/CesiumGeometry/Ray.h
csciguy8: > Better to include <glm/fwd.hpp>

Why?

CesiumGeometry/include/CesiumGeometry/IntersectionTests.h
```diff
@@ -94,6 +97,129 @@ class CESIUMGEOMETRY_API IntersectionTests final {
...
       glm::dvec3& barycentricCoordinates) noexcept;
+
+  /**
+   * @brief Tests if a ray hits a triangle and returns the hit point.
+   *
+   * @param ray The ray.
+   * @param p0 The first vertex of the triangle.
+   * @param p1 The second vertex of the triangle.
+   * @param p2 The third vertex of the triangle.
+   * @param cullBackFaces Ignore triangles that face away from ray.
```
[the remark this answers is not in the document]
csciguy8: Documentation updated to answer this question

CesiumGltfContent/test/TestIntersectRayGltf.cpp
csciguy8: Done. Also fixed up some other duplicate occurrences

CesiumGeometry/include/CesiumGeometry/Ray.h
javagl: Header files should usually always use forward declarations when available (mainly due to shorter compilation times and less risk of namespace issues, circular dependencies and such)

csciguy8: Makes sense. I've changed this, as well as a handful of other header file occurrences that could do the same thing.

CesiumGltfContent/src/GltfUtilities.cpp
csciguy8: Good catch. This was leftover from some other in-progress code that eventually got removed.

decltypes are now removed

CesiumGeometry/src/IntersectionTests.cpp
csciguy8: Nah, not a nitpick. Valid optimization. Committed

CesiumGltfContent/include/CesiumGltfContent/GltfUtilities.h
csciguy8: Indeed. Documentation committed

CesiumGltfContent/src/GltfUtilities.cpp
csciguy8: I like it. Changes committed

CesiumGltfContent/src/GltfUtilities.cpp
csciguy8: Agreed. Looks like this validation slipped through the cracks.

Changes committed.

CesiumGltfContent/include/CesiumGltfContent/GltfUtilities.h
csciguy8: That's a great idea, but I think I'd like to write a new issue for it.

This PR is already fairly big. But also, I wouldn't want to short change the "translucent geometry" use case. It could use some unit tests and testing against Ion assets. It may even overlap with that earlier comment about supporting double sided materials.

csciguy8:
> Still, the code in this PR just relied on existing functionality. Nobody can expect someone who uses the OrientedBoundingBox to look into the constructor code and see whether there might be an open TODO (of which nobody knows how to resolve it).

I'm thinking we should make this a separate issue, and not handle it for this PR.

I do have opinions of course (like letting the OBB constructor code throw), but this feels like something to be discussed outside of ray intersection tests.

csciguy8:
@kring, ready for another look. All comments have been resolved or commented on.

CesiumGeometry/include/CesiumGeometry/IntersectionTests.h
kring: Yeah I understand the difference between the two functions as far as an infinite versus finite ray, and returning the intersect distance (which may be negative) instead of the intersection position. But I'm not sure that justifies the different API here. Couldn't `rayTriangleParametric` be equivalently written as:

```cpp
static std::optional<double> rayTriangleParametric(
      const Ray& ray,
      const glm::dvec3& p0,
      const glm::dvec3& p1,
      const glm::dvec3& p2,
      bool cullBackFaces = false);
```

There's no need to impose the hassle of the out parameter on callers, is there? Unless it makes a measurable performance difference in th tight loops where this function is called. It's not obvious to me that would be the case, but I wouldn't be shocked either.

CesiumGltfContent/include/CesiumGltfContent/GltfUtilities.h
kring: Ok, that's fine. I'm slightly concerned about the backward compatibility implications of adding it later, but I guess it could be a separate function anyway, like `intersectRayGltfModelMultiple`?

CesiumGltf/include/CesiumGltf/Model.h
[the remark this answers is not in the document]
kring: I agree with Sean here. If we're going to add ID parameters, we should add all of them, to all the functions. The "we'll do it later when we need it" approach is weak because a) it may be our users who need it, not us, and b) if we decide to add a node ID later, that will be _another_ breaking changes for us and our users. If we're going to do a breaking change, we should try to do all of the related breaking changes that we can foresee at the same time, in order to minimize churn for our users (and ourselves).

I'm also ok with just removing all the IDs, and users can use pointer subtraction if they need them. But if we're going to add _any_, we should add _all_.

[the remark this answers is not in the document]
lilleyse: I would be fine with removing the IDs as well, since it feels a bit redundant.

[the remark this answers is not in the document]
csciguy8: Removing the ID was definitely the way to go. 

Tried this, and in doing so, realized how much code it reverted, when all we needed is indexes in one function. Doing pointer arithmetic is fine there.

CesiumGeometry/include/CesiumGeometry/IntersectionTests.h
csciguy8: Done. Change committed

csciguy8:
@kring Review changes committed. Ready again for another look.

CesiumGeometry/include/CesiumGeometry/IntersectionTests.h
```diff
@@ -99,6 +102,130 @@ class CESIUMGEOMETRY_API IntersectionTests final {
...
+   * ray origin, negative if it is behind it, or zero if the two points
+   * coincide.
+   *
+   * @param ray The ray.
+   * @param aabb The axis aligned bounding box.
+   * @param t Returned parametric value, if a hit is detected.
+   * @return True if a hit occurred, false otherwise.
+   */
+  static bool
+  rayAABBParametric(const Ray& ray, const AxisAlignedBox& aabb, double& t);
```
kring, line 174: This should also return std::optional, same as `rayTriangleParametric` now does. Same with the other parametric intersection tests below.

csciguy8: Fixed up the whole file. Should be t-less now...

csciguy8:
@kring Ready for one more pass!

kring:
Looks good, thanks! I'll merge as soon as CI passes.

PR #778, opened 2023-12-14
Handle `std::string` in `MetadataConversions`

j9liu:
In Cesium for Unity, `System::String::ToStlString()` returns a `std::string`, which is used in a `variant` for enabling conversions in `CesiumMetadataValueImpl`. Although you could construct a `std::string_view` from the `std::string` before converting, it's more convenient and won't require special handling in `std::visit`. Plus, the `string -> int` conversions require that the string is stored in `std::string` anyway.

This also adds checks to prevent negative numbers from being returned as valid unsigned integers.


CesiumGltf/test/TestMetadataConversions.cpp
```diff
@@ -49,7 +49,16 @@ TEST_CASE("Test MetadataConversions for boolean") {
...
 
-  SECTION("returns std::nullopt for incompatible strings") {
+  SECTION("converts from string") {
+    REQUIRE(MetadataConversions<bool, std::string>::convert("true") == true);
+    REQUIRE(MetadataConversions<bool, std::string>::convert("yes") == true);
+    REQUIRE(MetadataConversions<bool, std::string>::convert("1") == true);
+    REQUIRE(MetadataConversions<bool, std::string>::convert("false") == false);
+    REQUIRE(MetadataConversions<bool, std::string>::convert("no") == false);
+    REQUIRE(MetadataConversions<bool, std::string>::convert("0") == false);
+  }
```
csciguy8, line 59: Maybe add a test for special characters? (! & # ^ space ... etc)

CesiumGltf/include/CesiumGltf/MetadataConversions.h
```diff
@@ -441,20 +560,13 @@ template <> struct MetadataConversions<double, std::string_view> {
...
-    std::string temp(from);
-
-    char* pLastUsed;
-    double parsedValue = std::strtod(temp.c_str(), &pLastUsed);
-    if (pLastUsed == temp.c_str() + temp.size() && !std::isinf(parsedValue)) {
-      // Successfully parsed the entire string as a double.
-      return parsedValue;
-    }
-    return std::nullopt;
+    return MetadataConversions<double, std::string>::convert(std::string(from));
```
csciguy8, line 569: The comment block preceding this line needs to be moved to line 535.

CesiumGltf/include/CesiumGltf/MetadataConversions.h
```diff
@@ -363,15 +460,8 @@ template <> struct MetadataConversions<float, std::string_view> {
...
-    std::string temp(from);
-
-    char* pLastUsed;
-    float parsedValue = std::strtof(temp.c_str(), &pLastUsed);
-    if (pLastUsed == temp.c_str() + temp.size() && !std::isinf(parsedValue)) {
-      // Successfully parsed the entire string as a float.
-      return parsedValue;
-    }
-    return std::nullopt;
+    return MetadataConversions<float, std::string>::convert(
```
csciguy8, line 463: The comment block preceding this line needs to be moved to line 430.

CesiumGltf/include/CesiumGltf/MetadataConversions.h
```diff
@@ -210,64 +235,107 @@ struct MetadataConversions<
...
+      // Return early. Otherwise, empty strings will be parsed as 0, which is
+      // misleading.
+      return std::nullopt;
+    }
+
+    // Amazingly, C++ has no* string parsing functions that work with strings
+    // that might not be null-terminated. So we have to copy to a std::string
+    // (which _is_ guaranteed to be null terminated) before parsing.
+    // * except std::from_chars, but compiler/library support for the
+    //   floating-point version of that method is spotty at best.
```
csciguy8, line 334: This comment block can be deleted

CesiumGltf/include/CesiumGltf/MetadataConversions.h
```diff
@@ -210,64 +235,107 @@ struct MetadataConversions<
...
+
+    errno = 0;
 
     char* pLastUsed;
-    uint64_t parsedValue = std::strtoull(temp.c_str(), &pLastUsed, 10);
-    if (pLastUsed == temp.c_str() + temp.size()) {
-      // Successfully parsed the entire string as an integer of this type.
-      return CesiumUtility::losslessNarrow<TTo, uint64_t>(parsedValue);
+    uint64_t parsedValue = std::strtoull(from.c_str(), &pLastUsed, 10);
+    if (errno != EINVAL && errno != ERANGE) {
```
csciguy8, line 274: From browsing the documentation on `strtoull`, I do see `ERANGE` as one of the error values. I don't see `EINVAL` mentioned.

Also, I see that if no conversion can be performed, `0` or `ULONG_MAX or ULLONG_MAX` could be returned too.

CesiumGltf/include/CesiumGltf/MetadataConversions.h
```diff
@@ -210,64 +235,107 @@ struct MetadataConversions<
...
-      double truncated = glm::trunc(parsedDouble);
-
-      uint64_t asInteger = static_cast<uint64_t>(truncated);
-      double roundTrip = static_cast<double>(asInteger);
-      if (roundTrip == truncated) {
-        return CesiumUtility::losslessNarrow<TTo, uint64_t>(asInteger);
+    errno = 0;
+
+    double parsedDouble = std::strtod(from.c_str(), &pLastUsed);
+    if (errno != EINVAL && errno != ERANGE) {
```
csciguy8, line 287: From browsing the documentation on `strtod`, I do see ERANGE as one of the error values. I don't see EINVAL mentioned.

Also, I see that if no conversion can be performed, `0.0` or `HUGE_VAL` could be returned too.

CesiumGltf/include/CesiumGltf/MetadataConversions.h
```diff
@@ -152,56 +174,59 @@ struct MetadataConversions<
...
-    if (pLastUsed == temp.c_str() + temp.size()) {
-      // Successfully parsed the entire string as a double.
-      // Convert it to an integer if we can.
-      double truncated = glm::trunc(parsedDouble);
-
-      int64_t asInteger = static_cast<int64_t>(truncated);
-      double roundTrip = static_cast<double>(asInteger);
-      if (roundTrip == truncated) {
-        return CesiumUtility::losslessNarrow<TTo, int64_t>(asInteger);
+    double parsedDouble = std::strtod(from.c_str(), &pLastUsed);
```
csciguy8, line 217: From browsing the documentation on `strtod`, I do see ERANGE as one of the error values. I don't see EINVAL mentioned.

Also, I see that if no conversion can be performed, `0.0` or `HUGE_VAL` could be returned too.

CesiumGltf/include/CesiumGltf/MetadataConversions.h
```diff
@@ -152,56 +174,59 @@ struct MetadataConversions<
...
-    //   floating-point version of that method is spotty at best.
-    std::string temp(from);
+    errno = 0;
 
     char* pLastUsed;
-    int64_t parsedValue = std::strtoll(temp.c_str(), &pLastUsed, 10);
-    if (pLastUsed == temp.c_str() + temp.size()) {
-      // Successfully parsed the entire string as an integer of this type.
-      return CesiumUtility::losslessNarrow<TTo, int64_t>(parsedValue);
+    int64_t parsedValue = std::strtoll(from.c_str(), &pLastUsed, 10);
```
csciguy8, line 204: From browsing the documentation on `strtoll`, I do see `ERANGE` as one of the error values. I don't see `EINVAL` mentioned.

Also, I see that if no conversion can be performed, `0` or ` LONG_MAX, LONG_MIN, LLONG_MAX or LLONG_MIN` could be returned too.

CesiumGltf/test/TestMetadataConversions.cpp
```diff
@@ -192,6 +252,18 @@ TEST_CASE("Test MetadataConversions for float") {
         std::string_view()));
   }
 
+  SECTION("returns std::nullopt for invalid strings") {
```
csciguy8, line 255: Complexity comment. 

This section and the one preceding are almost the same, other than one converts from `string` and the other from `string_view`.

Can we restructure to reduce code, but crucially reduce the duplication of input test data?

Ex.
```
 REQUIRE(!MetadataConversions<float, std::string_view>::convert(std::string_view("10.00f hello")));
  ...
 REQUIRE(!MetadataConversions<float, std::string>::convert("10.00f hello"));
```
becomes
```
TestStringConversion<float>("10.00f hello", std::nullopt);

void TestStringConversion<toVal> (const std::string& input, std::optional<toVal> result) {
   REQUIRE(MetadataConversions<toVal, std::string>::convert(input) == result);
   REQUIRE(MetadataConversions<toVal, std::string_view>::convert(std::string_view(input)) == result);
}
```

This thought could extend to all tests. For example, it looks like all numerical conversions fail with mixed types of input, or input with invalid characters. It would be nice to have a defined set of "invalid conversion input" that we could reuse for all the related tests.

The ultimate goal: easier to maintain code with more test coverage

csciguy8 (changes requested):
Nice changes @j9liu.

A few comments. The only things that might be worth changing could be the return values from some of the `std` conversion functions.

Other than than, just some comments about comments, and a suggestions to help reduce test complexity.

CesiumGltf/include/CesiumGltf/MetadataConversions.h
j9liu: This comment block initially annotated the conversion from `std::string_view` -> `float`. The point was that, due to a strange bug, you could not convert from the `std::string_view` directly and had to copy the data to a `std::string`. This is why I kept it in the `std::string_view` function, because the `std::string_view` must still be converted to a `std::string` before being converted to a `float`.

CesiumGltf/include/CesiumGltf/MetadataConversions.h
```diff
@@ -210,64 +235,107 @@ struct MetadataConversions<
...
 struct MetadataConversions<
     TTo,
-    std::string_view,
+    std::string,
     std::enable_if_t<
         CesiumGltf::IsMetadataInteger<TTo>::value && !std::is_signed_v<TTo>>> {
   /**
-   * @brief Converts the contents of a std::string_view to an signed integer.
-   * This assumes that the entire std::string_view represents the number, not
+   * @brief Converts the contents of a std::string to an signed integer.
```
j9liu, line 247:
```suggestion
   * @brief Converts the contents of a std::string to an unsigned integer.
```

CesiumGltf/include/CesiumGltf/MetadataConversions.h
```diff
@@ -7,7 +7,9 @@
 #include <glm/common.hpp>
 #include <glm/gtx/string_cast.hpp>
 
+#include <cerrno>
 #include <cstdint>
+#include <iostream>
```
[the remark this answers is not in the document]
j9liu: `<iostream>` was included for debugging, so I will remove that. But `<cerrno>` should probably be kept just to be explicit about what the file requires.

CesiumGltf/include/CesiumGltf/MetadataConversions.h
j9liu: This comment block initially annotated the conversion from `std::string_view` -> `double`. The point was that, due to a strange bug, you could not convert from the `std::string_view` directly and had to copy the data to a `std::string`. This is why I kept it in the `std::string_view` function, because the `std::string_view` must still be converted to a `std::string` before being converted to a float.

CesiumGltf/include/CesiumGltf/MetadataConversions.h
j9liu: > I don't see EINVAL mentioned.

Some websites mention that these functions return `EINVAL` if no conversion could be performed (i.e., string did not contain any number). [This one for example](https://man7.org/linux/man-pages/man3/strtol.3.html). I wasn't sure if the functions still had this kind of behavior, so I erred on the side of caution. 

On second glance, most of the sites that mention `EINVAL` look to be written for C and not C++, and it says that `EINVAL` is not used in C99. I don't really know what that means relative to C++, but it seems like the pointer check (`pLastUsed`) is probably sufficient for checking if the entire string was numerical. So I'll remove the references to `EINVAL`.

> Also, I see that if no conversion can be performed, 0 or ULONG_MAX or ULLONG_MAX could be returned too.

I avoid checking for those values because they are valid values that could be read from a valid string. At the end of the day, either `pLastUsed` or `errno` will tell us if something went wrong. I believe that will be sufficient.

CesiumGltf/test/TestMetadataConversions.cpp
```diff
@@ -59,6 +68,15 @@ TEST_CASE("Test MetadataConversions for boolean") {
     REQUIRE(!MetadataConversions<bool, std::string_view>::convert(stringView));
```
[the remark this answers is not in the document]
j9liu: As much as I am a stickler for code clarity, I use the `!` syntax to check for `std::optional` checks across the entire file, and I think the effort to replace all of them is out-of-scope for this PR :')

j9liu:
I responded to most of the comments, and also realized I didn't add conversion tests for `std::string` conversions, so I appended those to the file. Assuming CI passes this should be ready.

CesiumGltf/include/CesiumGltf/MetadataConversions.h
csciguy8: (not used in C99) <-- Yeah, that's a little ambiguous. Does it mean it's older and stopped being used at C99 and up? Or just C99? I don't know.

Either way, this is an interesting point related to what C++ reference documentation we use. I'd lean Microsoft's [doc](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/strtoull-strtoull-l-wcstoull-wcstoull-l?view=msvc-170#return-value) (updated 2/16/2023), since we develop daily on it. On the other hand, it also implements functions outside of the standard, and since we build on Linux too, and generally don't use those in our platform independent code.

And that doc somewhat ambiguously states that 0 can be returned if no conversation can be performed. In this case would ERANGE get set? Would it still move the pointer in this case? (again, don't know).

I'll leave these thoughts with you.

CesiumGltf/include/CesiumGltf/MetadataConversions.h
csciguy8: I'm still a little concerned that a conversion failure could happen here, but our code would interpret success.

From the [linux doc](https://linux.die.net/man/3/strtod),

```
Since 0 can legitimately be returned on both success and failure, the calling program should set errno to 0 before the call, and then determine if an error occurred by checking whether errno has a nonzero value after the call.
```

This hints that we should use something like...
```
if (errno == 0)
   // Check if the entire string was parsed.
   ...
```

CesiumGltf/test/TestMetadataConversions.cpp
```diff
@@ -59,6 +68,15 @@ TEST_CASE("Test MetadataConversions for boolean") {
...
   }
 
+  SECTION("returns std::nullopt for incompatible strings") {
+    // invalid number
+    REQUIRE(!MetadataConversions<bool, std::string_view>::convert("11"));
```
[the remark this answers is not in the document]
csciguy8: Like the test refactoring. Very clean. Also resolves this comment.

csciguy8:
I like all the changes.

The only thing that I still would love to see would be broader error checking for `strtod` and `strtoull`. The docs aren't super clear on what goes on, but I think that changing

```
if (errno != ERANGE)
```

to 
```
if (errno == 0)
```

should cover any cases where overflow, underflow, or unnamed conversion errors occur.

Sorry, not trying to be pedantic here. But there's probably some metadata conversions that we don't know about yet, and could be very time costly to trace it back to here.

j9liu:
@csciguy8 Pushed a commit that addresses your last comment. Let me know if there's anything else this needs in order to get it merged.

CesiumGltf/include/CesiumGltf/MetadataConversions.h
```diff
@@ -340,6 +395,35 @@ template <> struct MetadataConversions<float, double> {
...
+  static std::optional<float> convert(const std::string& from) {
+    if (from.size() == 0) {
+      // Return early. Otherwise, empty strings will be parsed as 0, which is
+      // misleading.
+      return std::nullopt;
+    }
+
+    char* pLastUsed;
+    float parsedValue = std::strtof(from.c_str(), &pLastUsed);
+    if (pLastUsed == from.c_str() + from.size() && !std::isinf(parsedValue)) {
```
csciguy8, line 419:
```suggestion
    if (errno == 0 && pLastUsed == from.c_str() + from.size() && !std::isinf(parsedValue)) {
```
I think we missed talking about this line from the last review. Would benefit from the same errno logic.

CesiumGltf/include/CesiumGltf/MetadataConversions.h
```diff
@@ -423,6 +500,35 @@ template <> struct MetadataConversions<double, float> {
...
+  static std::optional<double> convert(const std::string& from) {
+    if (from.size() == 0) {
+      // Return early. Otherwise, empty strings will be parsed as 0, which is
+      // misleading.
+      return std::nullopt;
+    }
+
+    char* pLastUsed;
+    double parsedValue = std::strtod(from.c_str(), &pLastUsed);
+    if (pLastUsed == from.c_str() + from.size() && !std::isinf(parsedValue)) {
```
csciguy8, line 524:
```suggestion
    if (errno == 0 && pLastUsed == from.c_str() + from.size() && !std::isinf(parsedValue)) {
```

I think we missed talking about this line from the last review. Although looks like this isn't new code from this PR, would benefit from the same errno logic.

csciguy8:
Two more tweaks @j9liu ?

Found a couple more instances where we can use the same `errno == 0 ` logic. Sorry I missed these from before, but looks like the same logic can be applied.

Fresh new year's eyes I suppose?

j9liu:
@csciguy8 done!

csciguy8:
Thanks @j9liu ! Merging...

Issue #768, opened 2023-11-22
Support for `KHR_texture_transform` is broken


javagl:
The following is a simple test glTF asset that uses the `KHR_texture_transform` extension:

[textureTransformTest.zip](https://github.com/CesiumGS/cesium-native/files/13440816/textureTransformTest.zip)

Trying to load this with `cesium-native` will create an invalid model: It will not create a valid `AccessorView` for the texture coordinates.

The first part of the issue is that the code block in [`applyKHRTextureTransform::processTextureInfo`](https://github.com/CesiumGS/cesium-native/blob/378fab71e989fa727acc03b6c6e1d3c2a91d2897/CesiumGltfReader/src/applyKHRTextureTransform.cpp#L94) does not work when the data is _interleaved_. My understanding of this code block, based on <strike>the `// Inlined comments`</strike> quickly skimming over it is...

- It replaces the accessor with a new one (that is created by copying the properties of the old one)
- It creates a new buffer view that is supposed to _only_ contain data for that new accessor (also copying the properties of the old one)
- It creates a new buffer that is supposed to _only_ contain the data for the new accessor

Now, these 'copies' that are made there fall apart when the data is interleaved: They receive wrong byte offsets and byte strides.  Inserting
```
    // XXX
    accessor.byteOffset = 0;
    bufferView.byteOffset = 0;
    bufferView.byteStride = std::nullopt;
    // XXX 
```
at the end of that block ""fixes"" it insofar that the proper data can be accessed, but of course, that's not a viable solution: The old buffer will still contain the (now unused) accessor data (interleaved into the data that is still used). 

Even if this is fixed, I _think_ that the implementation of [`applyKHRTextureTransform::applyKHRTextureTransform`](https://github.com/CesiumGS/cesium-native/blob/378fab71e989fa727acc03b6c6e1d3c2a91d2897/CesiumGltfReader/src/applyKHRTextureTransform.cpp#L113) itself is broken: It "bakes" the texture transform that is defined in the **texture** into the texture coordinates of the **mesh primitive**. This will probably have undesired effects in several cases:

- There may be a texture coordinates accessor that is used in two mesh primitives, where one mesh primitive refers to a texture that **does** have the `KHR_texture_transform` extension, and the other one has a texture that does **not** have this extension. Modifying the texture coordinate accessor values will affect both primitives
[...]

PR #781, opened 2023-12-21
Apply texture transform for main

joseph-kaile:


javagl:
This seems to be related to https://github.com/CesiumGS/cesium-native/issues/768 , but does not fix it, right? 
(I just stumbled over this, and may have to read the changes here more thoroughly...)

kring:
Right, it doesn't fix it by itself. But it should fix it (I haven't tested or reviewd it yet) when combined with these PRs:
* https://github.com/CesiumGS/cesium-unreal/pull/1315
* https://github.com/CesiumGS/cesium-unity/pull/385

kring:
Thanks @joseph-kaile!

PR #783, opened 2024-01-02
Add height-query function to Tileset

### LINKED ISSUE #636: Add a function to compute the terrain height at a point
Similar to CesiumJS's `sampleTerrainMostDetailed`.

joseph-kaile:
Adds new functionality to `Tileset` to enable terrain height queries.

Builds on previous cesium-native work for ray / triangle and ray / gltf intersections.

Approach
-----------------------------------------
To find the height at a latitude / longitude, we must first load the terrain data for this location, construct a ray at that coordinate pointing towards the center of the earth, then intersect all the content in the tile, using the closest hit as our discovered height.

Height queries can happen for tiles not selected from a camera, so we introduce a parallel tile loading strategy that builds on the same system. Just as we use a camera frustum for `_visitTileIfNeeded`, we use our new height query requests and call `visitHeightRequests`. This discovers which tiles are needed to load and adds them to the same priority queues that the camera-based tiles go to.

Once all relevant tiles are loaded, we ray intersect against all of their contents.

Fixes #636

TO DO
-
- [x] Solve priority questions, how does it co-exist with camera tile loads?
- [x] Reuse `addTileToLoadQueue` for tiles added from query traversal
- [x] Performance - How do we know how much time is loading vs. intersecting? Intersect code should be minimal compared to url content requests. How can we verify this?
- [x] Performance - Are we wasting time creating physics meshes on the game engine side? (Ex. Unreal). 
- [x] Fix incremental `findCandidateTiles` for additive-refined tilesets.
- [x] Support for implicit tilesets
- [x] ~Support for instanced models~ Wrote an issue: #948 
- [x] ~Return all heights, not just the highest~ Not urgent at this time.
- [x] Write unit tests
- [ ] Peer review
- [ ] Merge


csciguy8:
@kring This is a good candidate for early review.

The bulk of the work is done enough to see the direction this is going, and it works pretty well when used with the related automated tests from https://github.com/CesiumGS/cesium-unreal/pull/1421

Transitioning to making some reasonable unit tests for this functionality...

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -294,6 +310,127 @@ Tileset::updateViewOffline(const std::vector<ViewState>& frustums) {
...
+  // Add a load request for tiles that are needed, and haven't started
+  TileLoadPriorityGroup priorityGroup = TileLoadPriorityGroup::Urgent;
+  double priority = 0.0;
+  for (Tile* pTile : tilesNeedingLoading) {
+    TileLoadState loadState = pTile->getState();
+
+    CESIUM_ASSERT(loadState != TileLoadState::Done);
+
+    // Push to appropriate queue based on state
+    if (loadState == TileLoadState::Unloaded) {
```
kring, line 401: Can we just call `addTileToLoadQueue` here? It seems like we've repeated some of that logic here. I guess the main problem is that `addTileToLoadQueue` assumes (asserts) that the tile is not already in the load queue, whereas here we don't know whether that is true. But that check is moderately expensive, so maybe there's a way we can eliminate it anyway?

Cesium3DTilesSelection/src/TerrainQuery.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+  // If tile failed to load, this means we can't complete the intersection
+  if (pTile->getState() == TileLoadState::Failed) {
+    warnings.push_back("Tile load failed during query. Ignoring.");
+    return;
+  }
+
+  if (pTile->getChildren().empty()) {
+    // This is a leaf node, add it to the list
+    candidateTiles.push_back(pTile);
```
kring, line 95: What if it's only temporarily a leaf node? For example, it's an external tileset that hasn't been loaded yet. Some mechanism will need to ensure a) the external tileset gets loaded, and b) we don't consider the height available until we load it and then potentially load some of the tiles it contains. Is there already such a mechanism in place? Will it also work with implicit tilesets that need to load subtrees? With quantized-mesh where tiles are created on demand based on availability, and availability is embedded in tiles at certain levels?

kring (commented):
Looking good so far! I did a fairly high level review and have some comments/questions.

Cesium3DTilesSelection/src/TerrainQuery.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+    // If additive refinement, add parent to the list with children
+    if (pTile->getRefine() == TileRefine::Add)
+      candidateTiles.push_back(pTile);
+
+    // Traverse children
+    for (Tile& child : pTile->getChildren()) {
+      auto& contentBoundingVolume = child.getContentBoundingVolume();
+
+      // If content bounding volume exists and no intersection, we can skip it
```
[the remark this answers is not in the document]
csciguy8: Thanks for the clarification! I'll make the related changes...

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -294,6 +310,127 @@ Tileset::updateViewOffline(const std::vector<ViewState>& frustums) {
...
+    HeightRequest& request,
+    std::set<Tile*>& tilesNeedingLoading) {
+  Tile* pRoot = _pTilesetContentManager->getRootTile();
+
+  bool tileStillNeedsLoading = false;
+  std::vector<std::string> warnings;
+  for (TerrainQuery& query : request.queries) {
+    query.candidateTiles.clear();
+
+    query.findCandidateTiles(pRoot, warnings);
```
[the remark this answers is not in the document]
csciguy8: It's not optimal, that's for sure.

We could just for the traversal once, and then, if something changes, like a tile we're waiting on loads its children, do it again, or continue where we left off.

I'll see if I can implement something a little smarter

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -294,6 +310,127 @@ Tileset::updateViewOffline(const std::vector<ViewState>& frustums) {
...
+      ++it;
+    } else {
+      auto deleteIt = it;
+      ++it;
+      _heightRequests.erase(deleteIt);
+    }
+  }
+
+  // Add a load request for tiles that are needed, and haven't started
+  TileLoadPriorityGroup priorityGroup = TileLoadPriorityGroup::Urgent;
```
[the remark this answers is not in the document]
csciguy8: Fair point. I can some use cases where you want this to be urgent, like loading a bunch of terrain clamped assets before the user experience starts. But also, there might be cases where we want somewhat "normal" priority like you mentioned. For example, a car is driving along the road, clamped, and has to play nice with the tiles that camera needs to see.

How about letting the caller determine the priority? (then letting it get handled by the existing tile load priority system that is in place)

Cesium3DTilesSelection/src/Tileset.cpp
csciguy8: That would be ideal. Might need to change `addTileToLoadQueue` slightly to make sure it all works. But it's reasonable to consider `addTileToLoadQueue` the official way to do this, whether you're traversing based on a camera, height query, or whatever functionality comes in the future...

Cesium3DTilesSelection/src/TerrainQuery.cpp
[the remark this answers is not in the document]
csciguy8: Fix committed

Cesium3DTilesSelection/src/TerrainQuery.cpp
csciguy8: This case should be handled.

The general flow is `TerrainQuery::findCandidateTiles` first to gather the potential tiles we need to intersect. If any of these aren't loaded, we wait for them to get loaded, then call `TerrainQuery::findCandidateTiles` again to get the final set of loaded nodes (I assume temporary nodes are no longer in here). Then all these nodes pass through `TerrainQuery::intersectVisibleTile`.

I suspect these two functions might continue to be modified in the future. There may be more edge cases where the traversal isn't right and we may want to support more progressive intersect traversals too, rather than just waiting until the most detailed tiles are completely loaded.

killjin: How to use in unity?

j9liu: Hi @killjin,

This PR is for adding the basic functionality in Cesium Native. Our next steps would include designing the API + user experience, to expose this to our runtimes.

I don't have a time estimate for when this will be completed, but you're always welcome to modify the plugin yourself to access these features early. :smile:

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -294,6 +310,127 @@ Tileset::updateViewOffline(const std::vector<ViewState>& frustums) {
...
+          this->_workerThreadLoadQueue.end(),
+          [pTile](const TileLoadTask& task) { return task.pTile == pTile; });
+
+      if (foundIt == this->_workerThreadLoadQueue.end())
+        this->_workerThreadLoadQueue.push_back(
+            {pTile, priorityGroup, priority});
+
+    } else if (loadState == TileLoadState::ContentLoaded) {
+
+      if (pTile->isRenderContent()) {
```
csciguy8, line 415: Performance question, do we absolutely need to transition this from `ContentLoaded` to `ContentDone`?

Can we perform the intersection tests in this state, to save waiting for all the things that are done in `finishLoading`.

kring:
I did some benchmarking of `tryCompleteHeightRequest`. My gut feel was that doing a tile traversal per height query per frame is exorbitantly expensive, but the Unreal performance tests in CesiumGS/cesium-unreal#1421 didn't seem too bad, so I wanted to check my assumptions.

I modified `FCesiumTerrainQuerySingleQuery` to query a grid of 50x50=2500 heights. The number of tiles they're spread across shouldn't matter _too_ much because I'm just trying to measure the per frame CPU time while waiting for the requests to complete. In any case, I left the spacing at the default.

On my system, the 2500 height queries took about 2-3ms per frame to run through the loop that calls `findCandidateTiles`. That's a pretty big chunk of time, but it's not so much that you'd notice it in the frame rate or interactivity unless your system is already struggling.

When I changed to 5000 height queries, it took about 4-5ms per frame. So it seems to scale roughly linearly with the number of height queries, as expected.

So overall I'd say this is slower than I'd like, but calling it "exorbitant" is unfair. It's a lot, though, considering this isn't the time to actually find the terrain intersection or load the tiles, it's basically just the overhead of waiting.

So that was with Cesium World Terrain, but then I tried Google Photorealistic 3D Tiles. Height queries against that tileset are _drastically_ slower. 5000 height queries require over 500ms of `findCandidateTiles` overhead each frame (especially later as more external tilesets are loaded), which obviously makes everything completely unusable. Why is it so much slower? Well, my guess is it's because GP3DT has more and deeper tiles, so the traversal time is much longer. I'm not totally sure. I expected it to be slower, but not this much slower.

kring:
I changed how `tryCompleteHeightRequest` works. Rather than doing a full search for relevant tiles for every point every frame, it just does a partial traversal starting with each candidate tile for each point every frame. Since the candidate tiles don't have any children the vast majority of the time (either because their children haven't been loaded yet, or because they're truly leaves), this is drastically faster.

With these changes, the slowest frame with GP3DT takes 106 ms to run through the `findCandidateTiles` loop for 2500 height queries. All but a very few take under 2ms. That's probably 100x faster on average. Still not exactly lightning fast, but it's a huge improvement.

CWT is much faster now too, taking well under 0.5ms on average.

This isn't going to work well  with additive-refined tilesets, though, because the candidate list will contain the root tile, so we'll end up doing the full traversal as before plus some additional pointless traversals. Still need to work on that.

killjin:
> I changed how `tryCompleteHeightRequest` works. Rather than doing a full search for relevant tiles for every point every frame, it just does a partial traversal starting with each candidate tile for each point every frame. Since the candidate tiles don't have any children the vast majority of the time (either because their children haven't been loaded yet, or because they're truly leaves), this is drastically faster.
> 
> With these changes, the slowest frame with GP3DT takes 106 ms to run through the `findCandidateTiles` loop for 2500 height queries. All but a very few take under 2ms. That's probably 100x faster on average. Still not exactly lightning fast, but it's a huge improvement.
> 
> CWT is much faster now too, taking well under 0.5ms on average.
> 
> This isn't going to work well with additive-refined tilesets, though, because the candidate list will contain the root tile, so we'll end up doing the full traversal as before plus some additional pointless traversals. Still need to work on that.

------
How to use it in Unity?

RoeDefense2:
They said earlier this is for Cesium Native. It's not yet integrated to
Unity.

Default

kring:
Some quick performance numbers, using the `SingleQuery` performance test in https://github.com/CesiumGS/cesium-unreal/tree/terrain-query-test, which queries heights for a 20x20 grid of points in the Denver Hills with 0.001 degrees between each row and column in the grid.

_Load_: Time to load all of the necessary tiles for the query, starting with a cold cache.
_Compute_: Time to find the intersections of the rays with the triangles in the loaded tiles.
 
| *Tileset* | *Load* | *Compute* |
|----------|---------|--------------|
| Cesium World Terrain | 2700ms | 30 ms |
| Google Photorealistic 3D Tiles | 9200 ms | 20 ms |

Currently the second part, finding the intersections, happens in the main thread, which isn't ideal. There's definitely some potential for frame rate hitches there.

I considered moving the computation to a work thread, but ultimately I think it's fast enough that we don't need to do that urgently. Moving it to a worker is mostly straightforward, except we need to make sure the glTF isn't unloaded or modified while the background computation is operating on it. And it's also likely to add some latency to the overall computation. Dispatching to the worker is pretty fast, but if we want to raise the "done" notification in the main thread, that has to wait for the next tick. Which means we might be adding 16ms of overhead to a computation that only takes 20ms (or less if you're not querying 400 positions). So I'm not sure it's justified when users should be able to mostly avoid frame rate hitches just by doing their queries in smaller batches, if necessary.

kring:
I believe this is now ready to merge. However, because I've now done a fair bit of work on it, it's probably a good idea to get someone else's eyes on it first.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/SampleHeightResult.h
```diff
@@ -0,0 +1,42 @@
...
+ */
+struct SampleHeightResult {
+  /**
+   * @brief The positions and sampled heights.
+   *
+   * The longitudes and latitudes will match the values at the same index in the
+   * original input positions. Each height will either be the height sampled
+   * from the tileset at that position, or the original input height if the
+   * height could not be sampled. To determine which, look at the value of
+   * {@link SampleHeightResult::heightSampled} at the same index.
```
j9liu, line 22:
```suggestion
   * @brief The positions and their sampled heights.
   *
   * For each resulting position, its longitude and latitude values will match
   * values from its input. Its height will either be the height sampled from
   * the tileset at that position, or the original input height if the sample
   * was unsuccessful. To determine which, look at the value of
   * {@link SampleHeightResult::sampleSuccessful} at the same index.
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/SampleHeightResult.h
```diff
@@ -0,0 +1,42 @@
...
+   */
+  std::vector<CesiumGeospatial::Cartographic> positions;
+
+  /**
+   * @brief Specifies whether the height for the position at this index was
+   * sampled successfully. If true, {@link SampleHeightResult::positions} has
+   * a valid height sampled from the tileset at this index. If false, the height
+   * could not be sampled for the position at this index, and so the height in
+   * {@link SampleHeightResult::positions} is unchanged from the original input
+   * height.
```
j9liu, line 32:
```suggestion
   * @brief The success of each sample.
   *
   * Each entry specifies whether the height for the position at the
   * corresponding index was successfully sampled. If true, then
   * {@link SampleHeightResult::positions} has a valid height sampled from the
   * tileset at this index. If false, the height could not be sampled, leaving the
   * height in {@link SampleHeightResult::positions} unchanged from the original
   * input height.
```

j9liu (commented):
@kring this looks good to me! I didn't look over the tests or the algorithm too closely, but I'm sure I can verify correctness when I look at the corresponding Unity / Unreal PRs.

Most of my comments are doc / understanding related. I'm pushing a commit that addresses some of these, but I wanted to leave some up for input. But none of these have to hold up the PR before the release!

Cesium3DTilesSelection/include/Cesium3DTilesSelection/SampleHeightResult.h
```diff
@@ -0,0 +1,42 @@
...
+   */
+  std::vector<bool> heightSampled;
```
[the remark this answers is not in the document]
j9liu: Actually I kind of prefer `sampleSuccess` more, since that's briefer.
Other suggestions I could think of are
- `successfulSample`
- `successfullySampled`
- `heightWasSampled`
- `wasHeightChanged` / `heightWasChanged`.

j9liu:
One thing I noticed while testing https://github.com/CesiumGS/cesium-unreal/pull/1421: if the tileset is invalid (e.g., invalid ion asset ID), it never completes the async action, and the `"Height requests could not complete because the tileset failed to load."` message never gets printed. It doesn't have to hold up this PR, since we're low on time, but I think we should address it shortly after, since it will be confusing to the user to not have feedback.

Cesium3DTilesSelection/src/TilesetHeightQuery.h
```diff
@@ -0,0 +1,172 @@
...
+   * tiles loaded for height queries stay loaded long enough to complete the
+   * query and no longer.
+   * @param tilesNeedingLoading Tiles that needs to be loaded before this height
+   * request can complete.
+   */
+  bool tryCompleteHeightRequest(
+      TilesetContentManager& contentManager,
+      const TilesetOptions& options,
+      Tile::LoadedLinkedList& loadedTiles,
+      std::set<Tile*>& tilesNeedingLoading);
```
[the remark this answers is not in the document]
kring: I don't think `tileLoadQueue` works for me, because it implies an order. I renamed it to `tileLoadSet`.

kring:
> if the tileset is invalid (e.g., invalid ion asset ID), it never completes the async action, and the "Height requests could not complete because the tileset failed to load." message never gets printed

I believe I fixed this in this commit, but cesium-unreal wasn't updated to use it yet:
https://github.com/CesiumGS/cesium-native/pull/783/commits/4850751a25c3cc95e8b9458a904b043d98bbd284

I'll write a test in cesium-native for this case, though.

kring:
Thanks for the review @j9liu, this should be ready for another look.

Cesium3DTilesSelection/src/TilesetHeightQuery.h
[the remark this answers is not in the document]
j9liu: I missed that it was a set, thanks for the catch!

j9liu:
Looks great, thanks @kring ! :smile:

PR #792, opened 2024-01-16
Allow implicit subtree buffers to be padded to 8 bytes.

kring:
The latest version of the 3D Tiles spec says that [binary implicit tiling subtrees](https://github.com/CesiumGS/3d-tiles/blob/main/specification/ImplicitTiling/README.adoc#subtree-binary-format) should have a binary buffer that is padded to 8 bytes. But cesium-native was only expecting 4-byte alignment, and considering the subtree to be invalid if it was too large. To make matters worse, the "buffer is not the expected size" error was lost and not displayed to the user. This PR fixes both problems.

As reported here:
https://community.cesium.com/t/re-catching-up-cesium-scaled-foundations/29226

### DIFF
--- a/Cesium3DTilesSelection/src/ImplicitOctreeLoader.cpp
+++ b/Cesium3DTilesSelection/src/ImplicitOctreeLoader.cpp
@@ -221,10 +221,13 @@ ImplicitOctreeLoader::loadTileContent(const TileLoadInput& loadInput) {
             this->addSubtreeAvailability(
                 subtreeID,
                 std::move(*subtreeAvailability));
-          }
 
-          // tell client to retry later
-          return TileLoadResult::createRetryLaterResult(nullptr);
+            // tell client to retry later
+            return TileLoadResult::createRetryLaterResult(nullptr);
+          } else {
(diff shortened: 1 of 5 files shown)


Cesium3DTilesSelection/src/ImplicitOctreeLoader.cpp
```diff
@@ -221,10 +221,13 @@ ImplicitOctreeLoader::loadTileContent(const TileLoadInput& loadInput) {
...
-          }
 
-          // tell client to retry later
-          return TileLoadResult::createRetryLaterResult(nullptr);
+            // tell client to retry later
+            return TileLoadResult::createRetryLaterResult(nullptr);
+          } else {
+            // Subtree load failed, so this tile fails, too.
+            return TileLoadResult::createFailedResult(nullptr);
+          }
```
kring, line 230: This change (and the similar one below) ensures that invalid subtrees are not loaded repeatedly, spamming an error to the log each time.

j9liu:
Thanks @kring !

PR #793, opened 2024-01-17
Change `FeatureIdTextureView` to account for sampler wrap values in `getFeatureID`

### LINKED ISSUE #741: Feature ID textures ignore sampler wrap values
Although feature ID textures should only use `NEAREST` filtering, they are allowed to use any wrap mode. However, it was assumed that they must also use `CLAMP_TO_EDGE` based on misleading CesiumJS code. In reality, the sampler wrap should be accounted for.

This is a rather simple task but due to time restraints I'll leave it on the backburner. It would be similar to how sampler wraps are used in `PropertyTexturePropertyView`.

- [ ] Move `applySamplerWrapS` and `applySamplerWrapT` to separate header and cpp files (for now, they are in `PropertyTexturePropertyView`
- [ ] Add a `_pSampler` member variable to `FeatureIdTextureView`
- [ ] Use `_pSampler`'s wrap mode in `get(U, V)`'s implementation (look at `PropertyTexturePropertyView` for reference)
- [ ] Add an error to `FeatureIdTextureViewStatus` for if the sampler is invalid
- [ ] Add unit tests (can just copy the ones in `TestPropertyTexturePropertyView`)

azrogers:
Fixes #741.

Previously, `FeatureIdTextureView` acted like the feature ID texture's sampler always had `CLAMP_TO_EDGE` specified as its wrap mode, even when this wasn't the case. This change makes `FeatureIdTextureView` apply the sampler's wrap mode before sampling the texture, using the same method as `PropertyTexturePropertyView`.


CesiumGltf/include/CesiumGltf/PropertyTexturePropertyView.h
j9liu, line 1: I think these lines can be removed now that `SamplerUtil.h` exists

https://github.com/CesiumGS/cesium-native/blob/df7bb9fa914c60ed082b5e562040f7eecfd9a81c/CesiumGltf/include/CesiumGltf/PropertyTexturePropertyView.h#L194-L195

CesiumGltf/include/CesiumGltf/SamplerUtil.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
```
j9liu, line 1: Can this be renamed to `SamplerUtility.h` to be stylistically consistent with the other files? Same for `SamplerUtil.cpp` -> `SamplerUtility.cpp`.

CesiumGltf/src/FeatureIdTextureView.cpp
```diff
@@ -63,11 +74,19 @@ FeatureIdTextureView::FeatureIdTextureView(
...
-int64_t FeatureIdTextureView::getFeatureID(double u, double v) const noexcept {
+int64_t FeatureIdTextureView::getFeatureID(double u, double v) noexcept {
   if (this->_status != FeatureIdTextureViewStatus::Valid) {
     return -1;
   }
 
+  if (this->_pSampler == nullptr) {
+    this->_status = FeatureIdTextureViewStatus::ErrorInvalidSampler;
+    return -1;
+  }
```
j9liu, line 85: When `pSampler` is `nullptr`, the status is set to `ErrorUninitialized` or `ErrorInvalidSampler` in the constructors, so I think this check is redundant. The view is meant to validate itself during construction before `getFeatureID` is called, such that `getFeatureID` shouldn't have to change the view's state.

So in other words, I think this check can be removed and `getFeatureID` can remain `const`

CesiumGltf/src/FeatureIdTextureView.cpp
```diff
@@ -35,6 +39,13 @@ FeatureIdTextureView::FeatureIdTextureView(
     return;
   }
 
+  if (static_cast<size_t>(texture.sampler) >= model.samplers.size()) {
```
j9liu, line 42: Just for extra safety:

```suggestion
  if (texture.sampler < 0 || static_cast<size_t>(texture.sampler) >= model.samplers.size()) {
```

CesiumGltf/src/FeatureIdTextureView.cpp
```diff
@@ -35,6 +39,13 @@ FeatureIdTextureView::FeatureIdTextureView(
     return;
```
j9liu, line 39: This comment should be removed since we are no longer ignoring the texture's sampler :D

https://github.com/CesiumGS/cesium-native/blob/df7bb9fa914c60ed082b5e562040f7eecfd9a81c/CesiumGltf/src/FeatureIdTextureView.cpp#L35

j9liu (commented):
Looks good @azrogers ! I left some comments but they're mostly about code cleanup + style consistency.

Also, can you add a note about this change to `CHANGES.md`?

azrogers:
@j9liu Incorporated the changes you recommended and updated CHANGES.md!

j9liu:
Thanks @azrogers ! I just readded the `const` to `getFeatureID`'s declaration. Merging now.

PR #794, opened 2024-01-18
Fix how batch tables choose sentinel values

j9liu:
This PR fixes how batch tables choose `noData` values when upgrading to `EXT_structural_metadata`. Specifically,

- For signed `int`s, the order of checks in `getSentinelValue` prevented -1 from being chosen as a sentinel value if 0 was taken. The order has been changed to account for this.
- Unsigned `int`s will be upgraded to the next biggest signed `int` (if possible) if 0 cannot act as a sentinel value. This allows -1 to be used as its sentinel value instead.
- If no sentinel value is available for a property that has `null`, explicitly force it to a string property.

Without these checks, there is some unpredictable behavior because some of the code assumes a valid sentinel value exists, leading to dereferenced `nullopt`s 😬

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,10 @@
 
 - Added conversions from `std::string` to other metadata types in `MetadataConversions`. This enables the same conversions as `std::string_view`, while allowing runtime engines to use `std::string` for convenience.
 
+##### Fixes :wrench:
+
+- Fixed a bug where upgraded batch table properties were not always assigned sentinel values, even when such values were available and required.
+
 ### v0.31.0 - 2023-12-14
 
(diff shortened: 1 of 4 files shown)


CesiumGltf/include/CesiumGltf/MetadataConversions.h
```diff
@@ -414,9 +414,12 @@ template <> struct MetadataConversions<float, std::string> {
       return std::nullopt;
     }
 
+    errno = 0;
```
j9liu, line 417: Not sure why this shows up in the diff? This was part of #778

kring:
Thanks @j9liu!

PR #795, opened 2024-01-18
Interpret offsets as array indices, not byte offsets

j9liu:
Depends on #794, so merge that first.

In `EXT_structural_metadata`, the `arrayOffsets` of a `PropertyTableProperty` represent array indices. However, Cesium Native treats them as byte offsets that are multiples of `sizeof(T)`, which is wrong. This PR corrects this behavior and edits the unit tests for the proper values.

Also, I arbitrarily found and replaced instances of `glm::i32mat2x2 -> glm::imat2x2`, so the diff looks bigger than it actually needs to be 😅


kring:
Thanks @j9liu!

PR #797, opened 2024-01-29
Add GlobeFlightPath helper class to handle math for fly-to components.

azrogers:
Required for CesiumGS/cesium-unity#395.

Both our Unreal and Unity implementations implement a fly-to class that animates a camera movement from one position on the globe to another. For one, the math between these two implementations is identical, and it would be nice to share code between the two implementations. But more importantly, Unity lacks double-precision implementations of the math functions we need to calculate this path, and trying to do the math with floats results in precision errors that cause jumping at the end of the flight (see CesiumGS/cesium-unity#390). By moving this implementation to the C++ side, we can use glm's double-precision math to solve these issues, while also eventually allowing us to unify the Unreal and Unity implementations.

This is implemented as a class that can take a normalized value representing position along the line and return ECEF coordinates for that point. This allows the path to be sampled continuously for each frame, while also allowing each implementation to choose how that normalized value is computed. This keeps things like the `_flyToProgressCurve` on the implementation side, letting us use the engine's curve classes with the side effect of allowing use-cases beyond just the fly-to controller. For example, you could use this class to place markers along the flight path, or to turn the curve into a set of vertices to display the line an object is moving along.


CesiumGeospatial/include/CesiumGeospatial/GlobeFlightPath.h
```diff
@@ -0,0 +1,100 @@
...
+      glm::dvec3 scaledDestinationEcef,
+      glm::dvec3 originalSourceEcef,
+      glm::dvec3 originalDestinationEcef);
+
+  double _totalAngle;
+  double _sourceHeight;
+  double _destinationHeight;
+  double _length;
+
+  glm::dvec3 _sourceDirection;
```
timoore, line 95: These fields should be const.

timoore, line 95: The private members should be const.

timoore:
This approach is an approximation to the geodesic (*cough* jargon *cough*), and the approximation has the problem that the velocity over the ellipsoid isn't constant as one changes the angle between the start and end vectors. If this is going into cesium-native, I wonder if it would be better to have real solution to the geodesic problem? Either follow the methods in the [wikipedia page](https://en.wikipedia.org/wiki/Geodesics_on_an_ellipsoid) and the linked reference https://doi.org/10.1007%2Fs00190-012-0578-z, or punt and use libPROJ.

CesiumGeospatial/include/CesiumGeospatial/GlobeFlightPath.h
lilleyse: Not sure if relevant to this class, but the C++ Core Guidelines advises against this for copyable / movable types: [C.12](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines)

csciguy8:
I'm not aware of any requirement for constant velocity for Unreal or Unity. The "fly to" animations have other aesthetic aspects, like height and progress curves, that make their use less than scientific.

I do like the simplicity and performance of this class. It's at least a good place to start if you're starting from scratch.

kring (commented):
This looks good, and I think it's very reasonable to a) use this class for our camera flights, at least for now, and b) include this class in cesium-native. I think we should give it a different name, though.

The trouble is it's a bit hard to describe what this curve actually _is_. It's not a geodesic. It's not a rhumb line. It's not even the curve described in 3D Engine Design for Virtual Globes in section 2.4, because that curve scales each point to the geocentric surface.

I don't think if there's a name for this curve. It's basically, you know, some random curve that mostly looks fine for camera flights and is easy to compute. 😆 

But now that I think about it, I think I made a mistake in CesiumGS/cesium-unreal#1333. We had a jump in the camera flight at the end. I realized that was caused by the start and end points being incorrect because taking the end direction, scaling it to the geodetic surface and adding the end height does not yield the original end point. I fixed that by computing the start and end directions after scaling to the geodetic surface. However, I should have gone in the geocentric direction instead. Leave the direction computation as-is, and instead change the step computation to scale to the geocentric surface before adding the geocentric height. This would make the math consistent with that section 2.4 in 3D Engine Design for Virtual Globes. And it would make all the points on the path lie in a plane. It still wouldn't actually be a geodesic, but it'd be pretty close. And it's a faster and simpler computation, too.

So, at the risk of drawing this task out further, let's do this:

1. Add `scaleToGeocentricSurface` to our `Ellipsoid` class. You can port the implementation from CesiumJS, which is only a few lines of code: https://github.com/CesiumGS/cesium/blob/50373e66f455a01b9428cb3320db229b552a6a36/packages/engine/Source/Core/Ellipsoid.js#L550
2. No need to scale the source and destination points to the geodetic surface before computing the rotation axis and angle. Effectively, revert my change in CesiumGS/cesium-unreal#1333.
3. The source and destination heights shouldn't be computed with a `cartesianToCartographic` anymore. Instead, they should be `magnitude(sourceEcef - scaledSourceEcef)` (and similar for destination).
4. At each step, scale to the _geocentric_ surface and then add the height. But don't add the height along the `geodeticUp` vector. Instead, add it along the geocentric up vector, which is simply `normalize(position)`.
5. Rename `GlobeFlightPath` to perhaps `SimplePlanarEllipsoidCurve`. The documentation should say that it produces points that lie in a plane going through the center of the Earth, and that their height above the ellipsoid will be a linear interpolation between the source and destination heights.

This turned into a lot of words. I hope it's not totally confusing. Let me know if anything doesn't make sense or if you disagree, because I'm not 100% sure I've thought through it completely.

The bigger picture here is still that CesiumJS's flight implementation is probably the better way to go. So if what I just described is a lot of work, it's probably not worth doing. And feel free to consider jumping right to the CesiumJS flight implementation instead if you prefer!

CesiumGeospatial/include/CesiumGeospatial/GlobeFlightPath.h
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * @brief Describes a flight path along the surface of a WGS84 ellipsoid
+ * interpolating between two points. The path is continuous and can be sampled,
+ * for example, to move a camera between two points on the ellipsoid.
+ */
+class CESIUMGEOSPATIAL_API GlobeFlightPath final {
+public:
+  /**
+   * @brief Creates a new instance of \ref GlobeFlightPath from a source and
```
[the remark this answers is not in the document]
azrogers: Is there a way to get Visual Studio to support the `@link` syntax? Intellisense seems to just stop parsing the rest of the comment when it encounters one.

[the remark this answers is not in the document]
kring: Not that I know of, but I haven't really looked. I'm not opposed to switching to different syntax if it has advantages for us. I think the main reason we used that syntax in the first place was to make it easy to copy/paste doc comments from CesiumJS when we were porting some code from there. It's only a small advantage, though. In any case, if we're going to switch, it'd be best to do it in a separate PR that switches everything at once.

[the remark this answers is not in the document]
azrogers: I'll change it to `{@link}` for now - however, as it turns out, this isn't Doxygen's syntax, this is JSDoc's, which is why Visual Studio can't parse it. I've made an issue for fixing this: #800.

azrogers:
@kring I've updated the PR based on your feedback, and implemented the geocentric calculation you recommended. As far as I can tell I still think the CesiumJS solution is just lerping between cartographic coordinates - see [here](https://github.com/CesiumGS/cesium/blob/1ece49aa34e97d90802e4ef54c1ce90970c50fbe/packages/engine/Source/Scene/CameraFlightPath.js#L305). I think our solution is probably more accurate.

CesiumGeospatial/include/CesiumGeospatial/SimplePlanarEllipsoidCurve.h
```diff
@@ -0,0 +1,93 @@
...
+#include <CesiumGeospatial/Ellipsoid.h>
+
+#include <glm/vec3.hpp>
+
+#include <optional>
+
+namespace CesiumGeospatial {
+
+/**
+ * @brief Produces points on an ellipse that lies on a plane that intersects the
```
csciguy8, line 14: Could add a note that the curve generated is specifically on a WGS 84 ellipse. (in case people think it's more generalized that in actually is)

azrogers: Well, the curve it produces isn't on a WGS84 ellipsoid, since it's floating above the surface. It's a section of an ellipse formed by points on a WGS84 ellipsoid. But I'll clarify that this curve is specific to a WGS84 ellipsoid. Though, actually, it may be better to let the end user specify an ellipsoid so that this supports other ellipsoids right out of the box...

CesiumGeospatial/test/TestSimplePlanarEllipsoidCurve.cpp
```diff
@@ -0,0 +1,87 @@
+#include <CesiumGeospatial/SimplePlanarEllipsoidCurve.h>
+
+#include <catch2/catch.hpp>
+
+using namespace CesiumGeospatial;
+using namespace CesiumUtility;
+
+TEST_CASE("SimplePlanarEllipsoidCurve") {
```
csciguy8, line 8: These tests look like a great start.

A suggestion for the future...

Currently it looks like each test section tests our assumptions of what a flight should be
  1) Is the start and end ok?
  2) Is it interpolating sensibly?
  3) Etc

We may want to flip this around so that each section tests a specific flight of interest, with each flight tested against our assumptions.
   1) Philly to Tokyo
   2) Tokyo To Philly
   3) Philly to Philly (same start / dest)
   4) Flight created from SimplePlanarEllipsoidCurve::fromEarthCenteredEarthFixedCoordinates
   5) Flight created from SimplePlanarEllipsoidCurve::fromLongitudeLatitudeHeight
   6) Flight from center of the earth
   7) etc.

CesiumGeospatial/include/CesiumGeospatial/SimplePlanarEllipsoidCurve.h
csciguy8: Good point. WGS84 ellipsoid inspired?

I did see mentions of WGS84 in the creation functions, so this is probably not a big deal.

azrogers: I've updated SimplePlanarEllipsoidCurve to take an Ellipsoid as input. It's a simple change but it means that it'll support, for example, generating curves between different points on the moon or on mars, if someone wants to do that.

azrogers: Unfortunately we can't currently support passing the ellipsoid in without some changes to how ellipsoids are represented on the Unity side, which is definitely beyond the scope here. So it'll have to be hardcoded to a WGS84 ellipsoid in Unity for now. I'll make a note of that.

CesiumGeospatial/test/TestSimplePlanarEllipsoidCurve.cpp
azrogers: @csciguy8 I've expanded the tests based on your list here. One note is that a flight from the center of the earth shouldn't work - it's the one position that can't be mapped to a geocentric surface, as it'll result in +Infinity when calculating `1.0 / sqrt(0.0)`. CesiumJS also returns NaN for this input to `scaleToGeocentricSurface`, so I've made this a condition to fail the construction of a `SimplePlanarEllipsoidCurve` and added a test for it.

CesiumGeospatial/src/SimplePlanarEllipsoidCurve.cpp
```diff
@@ -0,0 +1,102 @@
...
+}
+
+glm::dvec3 SimplePlanarEllipsoidCurve::getPosition(
+    double percentage,
+    double additionalHeight) const {
+  if (percentage >= 1.0) {
+    // We can shortcut our math here and just return the destination.
+    return this->_destinationEcef;
+  }
+
```
csciguy8, line 52: Can we could do a similar shortcut for percentage <= 0?

CesiumGeospatial/test/TestSimplePlanarEllipsoidCurve.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+using namespace CesiumGeospatial;
+using namespace CesiumUtility;
+
+const glm::dvec3
+    philadelphiaEcef(1253264.69280105, -4732469.91065521, 4075112.40412297);
+const glm::dvec3
+    tokyoEcef(-3960158.65587452, 3352568.87555906, 3697235.23506459);
+
+// Values obtained from the Unreal implementation with all curves disabled.
```
csciguy8, line 13: I'm generally not a fan of a unit test that checks itself based on an earlier implementation. What if we change the interpolation logic slightly? Would we update these values? How do we know the values are actually good?

From what I gather, we want to check "Is the flight path sensible in the middle part?". There are probably other ways to do this... ex. checking height never goes below 0 for a test with start / end with position heights. ex. check that the generated points actually lie on a plane (like the class suggest, etc...

What do you think?

csciguy8 (approved):
Looks great @azrogers! Reviewed the code and tested in cesium for unreal.

Have a couple comments that I could go either way on. 

Let me know what you think, then we can merge.

CesiumGeospatial/test/TestSimplePlanarEllipsoidCurve.cpp
azrogers: Removed the midpoint test, changed the "reverse path" test and "equivalent LLH coordinates" tests to check themselves against other curves instead of known values, and added tests for coplanarity and making sure the path doesn't cross through the earth.

csciguy8: Looks great!

PR #798, opened 2024-01-30
Replace zlib with zlib-ng

### LINKED ISSUE #707: Consider switching from zlib to zlib-ng
https://github.com/zlib-ng/zlib-ng

Some other internal projects at Cesium are considering making this switch. Perhaps we should, too.

azrogers:
Closes #707.

This change replaces the dependency on zlib with [zlib-ng](https://github.com/zlib-ng/zlib-ng). This is a fork of zlib with improved performance, including support for hardware acceleration. It also adds tests to the CesiumUtility Gunzip implementation.


jherico:
@kring zlib-ng is available in VCPKG as well.  Should I incorporate these changes into my VCPKG PR?

kring:
@jherico we haven't reviewed this yet, but I suspect it should be a pretty decent win with no downside, so I think yes, we'll want it in the vcpkg PR eventually. I don't want to create work for you when we haven't even given you the review we owe you yet, though!

kring:
Thanks @azrogers!

PR #799, opened 2024-01-31
Support texture transform extension on occlusion and normal textures.

kring:
These two texture types have custom classes, `MaterialOcclusionTextureInfo` and `MaterialNormalTextureInfo`, but the `KHR_texture_transform` extension was configured to only apply to `TextureInfo` classes. The custom classes are derived from `TextureInfo`, but the code generator isn't that clever.

The end result was that instances of `KHR_texture_transform` attached to an occlusion or normal texture showed up as an unknown extension (expressed as a `JsonValue`) rather than as `ExtensionKhrTextureTransform`.

### DIFF
--- a/CesiumGltfReader/generated/src/registerExtensions.cpp
+++ b/CesiumGltfReader/generated/src/registerExtensions.cpp
@@ -27,6 +27,8 @@
 #include <CesiumGltf/Buffer.h>
 #include <CesiumGltf/BufferView.h>
 #include <CesiumGltf/Material.h>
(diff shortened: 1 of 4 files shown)


j9liu:
Thanks @kring !

lilleyse:
I think this is also needed for `FeatureIdTexture` and `PropertyTextureProperty`, which also subclass `TextureInfo`.

j9liu:
Good catch @lilleyse. That probably means that `FeatureIdTextureView` and `PropertyTexturePropertyView` need to store the extension values, too, and apply them to the input UV coordinates.

Issue #801, opened 2024-01-31
Support `KHR_texture_transform` for feature ID and property textures


j9liu:
Noted from #799.

Support for `KHR_texture_transform` should also be added for `FeatureIdTexture` and `PropertyTextureProperty`, which also subclass `TextureInfo`. But I don't believe that `generate-classes` handles extensions of extensions, so there isn't currently a way to indicate this in `glTF.json`.

This probably means that FeatureIdTextureView and PropertyTexturePropertyView need to store the extension values, too, and apply them to the input UV coordinates.

PR #802, opened 2024-02-01
Attach KHR_texture_transform to other extensions

lilleyse:
Partially addresses https://github.com/CesiumGS/cesium-native/issues/801.

Updates the generator code to register `KHR_texture_transform` on `ExtensionExtFeatureMetadataFeatureIDTexture` and `PropertyTextureProperty` but does not update `FeatureIdTextureView` or `PropertyTexturePropertyView`.


tools/generate-classes/index.js
```diff
@@ -77,6 +79,16 @@ for (const schemaPath of argv.schemas) {
...
 const schemaCache = new SchemaCache(schemaBasePaths, argv.extensions);
 
+// Add extensions to schema cache so that extensions can extend objects in other extensions
+for (const extension of config.extensions) {
+  const extensionSchema = schemaCache.loadExtension(
+    extension.schema,
+    extension.extensionName
+  );
+  const { _, schemaBasePath } = splitSchemaPath(extensionSchema.sourcePath);
+  schemaCache.schemaPaths.push(schemaBasePath);
```
lilleyse, line 89: I don't really like how this is modifying `schemaPaths`, but this approach didn't require too many code changes.

j9liu:
@lilleyseshould this extend `FeatureIdTexture` for `EXT_mesh_features` too?  I see that this extends the legacy `EXT_feature_metadata` texture but not the newer one

lilleyse:
Oh yes it should. Thanks for catching that.

lilleyse:
Hm, so the problem is that `featureIdTexture.schema.json` is in both extensions and it ends up picking whichever one was added to the schema base paths first...

At this point maybe we can just drop support for the older extensions?

j9liu:
Hm that's probably fine! No one has complained about it not being supported in our clients (yet?)

lilleyse:
I opened a PR to remove EXT_feature_metadata: https://github.com/CesiumGS/cesium-native/pull/812. Once that's merged I'll update this PR.

tools/generate-classes/index.js
j9liu: @kring Do you have any concerns with this change? If not, happy to merge this. Just wanted to double-check

kring: No concerns, it seems reasonable enough to me.

kring:
Merging this in order to move on to #811. Thanks @lilleyse! And thanks @j9liu for all the review work.

Issue #803, opened 2024-02-03
All tiles excluded when a clipping CesiumPolygonRasterOverlay has both "Invert Selection" and "Exclude Selected Tiles"


corybarr:
To reproduce, use Google Photorealistic 3D Tiles and the coordinates provided in the `Cesium for Omniverse: Tileset Clipping` tutorial.

Default `CesiumPolygonRasterOverlay` settings look good:

![image](https://github.com/CesiumGS/cesium-native/assets/1040194/8d36fced-d826-4a16-97e7-252e9e27c3e8)

When "Invert Selection" is enabled, no tiles are rendered:

![image](https://github.com/CesiumGS/cesium-native/assets/1040194/29fa79ff-35c2-4dc8-93c1-c620ad010402)

If "Invert Selection" is enabled and "Exclude Selected Tiles" is disabled, the inverted clip looks good:

![image](https://github.com/CesiumGS/cesium-native/assets/1040194/84a8ffaa-3a4d-46e2-83e3-211029db0d95)

I'm seeing the same behavior while adding this capability to `CesiumPolygonRasterOverlays` in the Omniverse extension.

![image](https://github.com/CesiumGS/cesium-native/assets/1040194/f98d3c39-7c83-4ffa-846b-2823d68a3ab3)

with Cesium for Unreal 2.2.0

kring:
The problem seems to be that `estimateGlobeRectangle` (in `BoundingVolume.cpp`) does a very poor job of estimating the bounding region for an oriented bounding box that encompasses the entire globe. The root tile of Google Photorealistic 3D Tiles has an oriented bounding box located at the center of the Earth and with three axis-aligned half-axes of about 8 million meters each. So that definitely encompasses the entire Earth. But `estimateGlobeRectangle` comes up with a nonsense answer that only extends to +/- 135 degrees longitude and +/- 35 degrees latitude. If the polygon is located entirely outside that box, then the root tile will be excluded, and nothing will render at all.

lilleyse:
Could you check if the oriented bounding box contains the center point (0, 0, 0) and if so return a full rectangle?

Looking at the code a bit there will probably be other inaccuracies (after all it's called `estimateGlobeRectangle`) but maybe this is enough to unblock the issue with Google's data?

kring:
Yeah something like that would probably work as a quick fix for this specific issue.

PR #808, opened 2024-02-09
Fix `estimateGlobeRectangle` for boxes and spheres that encompass the entire globe

lilleyse:
Fixes https://github.com/CesiumGS/cesium-native/issues/803 by returning the maximum rectangle if the box or sphere contains `(0, 0, 0)`


kring:
Thanks @lilleyse, looks good! There's of course the possibility of large bounding sphere, such as covering a quarter of the globe, being incorrectly flagged as covering the whole globe just because it happens to contain the center. But generally it's better to return a rectangle that is too large than one that is too small, so this is a definite improvement.

PR #810, opened 2024-02-12
Remove Pull Request triggers for CI builds

csciguy8:
It's not obvious to me why this is needed.

It's also a little annoying (and wasteful) to have a PR created, push commits to it, and see two builds get triggered.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,5 @@
 name: cesium-native
-on: [push, pull_request]
+on: [push]
 jobs:
   QuickChecks:
     name: "Quick Checks"


kring:
The `pull_request` builds are useful when someone creates a PR from a fork. For example, here's a PR that only has pull_request builds:
https://github.com/CesiumGS/cesium-native/pull/786

I'm not sure if there's a way to avoid the duplicate-ish (they're not identical if main has changed since the PR's branch was created) build from in-repo commits.

csciguy8:
Makes sense. @kring says this type of problem goes back to the early days of CesiumJS too.

I don't think it's a big enough problem to spend any more time on. Closing this without merging.

PR #811, opened 2024-02-13
Account for `KHR_texture_transform` in feature ID texture + property texture property views

### LINKED ISSUE #801: Support `KHR_texture_transform` for feature ID and property textures

j9liu:
Depends on #802 so merge that first.

Closes #801.

This adds a new `KhrTextureTransform` utility class to parse a `KHR_texture_transform` extension. If successfully parsed, UVs can be transformed via the `applyTransform` function.

Both `FeatureIDTextureView` and `PropertyTexturePropertyView` will now account for `KHR_texture_transform` when their respective `get` functions are called.


CesiumGltfReader/src/applyKhrTextureTransform.cpp
```diff
@@ -11,49 +12,20 @@ namespace {
 void transformBufferView(
     const AccessorView<glm::vec2>& accessorView,
     std::vector<std::byte>& data,
-    const ExtensionKhrTextureTransform& textureTransform) {
-
-  if (textureTransform.offset.size() < 2 || textureTransform.scale.size() < 2) {
+    const ExtensionKhrTextureTransform& textureTransformExtension) {
+  KhrTextureTransform textureTransform(textureTransformExtension);
+  if (textureTransform.status() != KhrTextureTransformStatus::Valid) {
+    // TODO: should this report an error somehow?
```
kring, line 18: I think you can just use the return value to indicate whether it worked or not.

CesiumGltf/include/CesiumGltf/PropertyTexturePropertyView.h
```diff
@@ -331,6 +338,20 @@ class PropertyTexturePropertyView<ElementType, false>
...
+        property.getExtension<ExtensionKhrTextureTransform>();
+
+    if (pTextureTransform) {
+      this->_textureTransform = KhrTextureTransform(*pTextureTransform);
+
+      if (this->_textureTransform->status() ==
+              KhrTextureTransformStatus::Valid &&
+          pTextureTransform->texCoord) {
+        // Override with the extension's texcoord.
+        this->_texCoordSetIndex = *pTextureTransform->texCoord;
```
kring, line 352: A slight reservation here... By clobbering the original texture coordinate index, aren't we breaking clients that don't understand this extension? Cause those original texture coordinates can, in some cases, but usable without the extension, right? As a practical matter, it's probably not a big deal, but I'm curious what you think.

kring (commented):
Looks great! Just two comments.

CesiumGltf/include/CesiumGltf/PropertyTexturePropertyView.h
j9liu: That's a fair point! I'll have the `KhrTextureTransform` store the texcoord, and clients can decide whether or not to use it.

j9liu: Though, it feels awkward that even if clients don't understand this extension, the views will still transform the UVs anyway? So maybe I shouldn't transform the UVs, and just leave it to the clients to apply the transforms if supported?

j9liu:
@kring Following your comment, I decided to untangle the texture transform implementation from the texture views. If a client doesn't understand / implement `KHR_texture_transform`, we should probably respect that. The texture transform associated with the textures can still be retrieved; it's just up to clients to apply the transform themselves.

CesiumGltf/include/CesiumGltf/PropertyTexturePropertyView.h
j9liu: @kring Trying to re-explain: I agree that we shouldn't overwrite the original texcoord, in case the extension isn't supported / understood by the client. But should we apply the transform in `getFeatureId`? Because in that case, what if they're using the original UVs that aren't referenced by `KHR_texture_transform`?

kring: Ah ok I'm with you now. Yeah I think you're right that a client that doesn't understand `KHR_texture_transform` would be passing in UVs derived from the original `_texCoordSetIndex`, not the ones referenced in the extension.

On the other hand, a client that does understand that extension would have texture coordinates from the extension's set, and if they didn't realize that had to explicitly apply the transformation, then they'd have a bug.

I've long had this idea that clients should explicitly  "tell" cesium-native which glTF and 3D Tiles extensions they support. That would allow cesium-native to do smart things, like eliminate extensions when possible (decode a KTX texture to raw bytes, for example), and report an error if a required extension is not supported and cannot be eliminated by cesium-native. See CesiumGS/cesium-native#515.

That sort of thing could help here, but it's definitely outside the scope of this PR. So what do you think of making it an option (parameter) when you construct the `PropertyTexturePropertyView` / `FeatureIdTextureView`? Something like `useKhrTextureTransformExtension`?

kring: If the parameter isn't workable, or is just a lot of trouble for not much benefit, then I think a) not applying the transform and b) adding clear documentation to remind users that they need to do it themselves, is fine.

j9liu: I think the parameter is a good idea! I'll try that

j9liu:
@kring Made some changes based on our discussion. Let me know what you think!

kring:
Looks great, thanks @j9liu!

PR #816, opened 2024-02-22
Upgrade spdlog

josephbirkner:
This PR makes the following changes:

* Allow using an externally supplied spdlog (by checking if the target exists before add_subdirectory).
* Allow using a newer version of fmt (which is bundled in newer versions of spdlog).
* Upgraded the spdlog submodule reference from 1.10.0 to 1.13.0


csciguy8:
Thanks @josephbirkner !

CMake modification looks reasonable. Verified the spdlog module update is the official 1.13 release. Browsed the release notes from 1.10 -> 1.13.

Nice changes.

csciguy8 (approved):

csciguy8:
Merged this, then reverted when I encountered some errors in cesium-unreal.

Make a new PR to work with in the meantime. Not sure what the fix is quite yet.

PR #819, opened 2024-02-22
Replace zlib with zlib-ng (again)

kring:
I reverted #798 because we're having trouble with on macOS / ARM64. This PR reapplies that change and should be merged after we sort of our Mac problems.


kring:
We switched to zlib-ng when we started using vcpkg for dependency management. So this is OBE.

PR #821, opened 2024-02-26
Fix erroneous thread id assertion for classes derived from ReferenceCounted

csciguy8:
Noticed this when running Cesium for Unreal with the latest.

Fix bug where the client build system can define the order of members differently than cesium-native. Note the #ifndef NDEBUG statement.

The workaround is to just move the debug member to the end of the class, to preserve the non-debug order.


csciguy8:
Closing without merging. Will go another direction to solve this problem...

PR #825, opened 2024-03-04
Add AsyncSystem::waitInMainThread

kring:
Adds a new method, `waitInMainThread`, to `AsyncSystem`. 

Calling `asyncSystem.waitInMainThread(std::move(future))` is similar to calling `future.wait()`. Both wait for the future to resolve or reject, and then either return the value or throw an exception. `waitInMainThread` _also_ continues to dispatch main thread tasks while waiting, however. This means that it is safe to call from the main thread even if the future has `thenInMainThread` continuations. `future.wait()`, on the other hand, is very likely to deadlock in this situation.

### DIFF
--- a/CesiumAsync/include/CesiumAsync/Impl/QueuedScheduler.h
+++ b/CesiumAsync/include/CesiumAsync/Impl/QueuedScheduler.h
@@ -3,6 +3,8 @@
 #include "ImmediateScheduler.h"
 #include "cesium-async++.h"
 
+#include <atomic>
+
 namespace CesiumAsync {
 namespace CesiumImpl {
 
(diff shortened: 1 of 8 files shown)


csciguy8 (approved):
Looks like a nice addition to the async system @kring!

The goal makes sense to me, and it could be merged as is, but I have a couple of thoughts too...
1) General usability - Can `waitInMainThread` be in `Future` instead of `AsyncSystem`? That way you can use it in the same way that you use `wait`, you just need the Future.
2) `giveUpTimeSlice` - I understand abstracting this away from the user, but maybe it's not necessary? If I'm writing async code, I'm most likely going to want to know whether it's calling sleep(0) or yield at a glance, and maybe wouldn't even use this.
3) `waitInMainThread` - I could see having a sleep time in milliseconds as a parameter. Hard to predict future uses, but some might want to be even friendlier to other threads than just `sleep(0)`.

What do you think?

kring:
Alright Brian, you've successfully nerd sniped me here.

Your questions reminded me that the whole `sleep_for` thing was a cop out. There shouldn't be any spin waiting; it should use a mutex and a condition variable to suspend the thread until there's work to do.

Conceptually this is easy-ish, but it's hard because some details we need to fiddle with are buried in Async++. Basically, to do this, we need access to the std::mutex that Async++ locks when new work is added to the main thread queue. But it's not available through the public interface of `fifo_scheduler`. Ok, well, we already have a wrapper for `fifo_scheduler` (called `QueuedScheduler`). And `fifo_scheduler` is actually pretty simple, so let's copy/paste it into `QueuedScheduler`. Ok it's implemented with `fifo_queue`, which is internal, but we can use it with just a slight hack (`#include <async++/../../src/fifo_queue.h>`). Our `QueuedScheduler` now becomes the new `fifo_queue` rather than delegating to it. Now to build up a signaling mechanism based on a `std::condition_variable` that gets notified when something new is added to the main thread task queue. Add a function to dispatch the main thread queue, waiting on the condition variable when there's nothing else to do. And make sure the condition variable is notified when the task we're waiting for completes (just in case the future is not resolved in the main thread).

Yep, that's much better! 🎉 

Huh... where did the last 6 hours of my day go?

Well, at least I've addressed all your feedback. I also added support for waiting on `SharedFuture` as well and merged in main. This is ready for another look.

kring:
Don't merge this yet. The `Future resolving from a worker while main thread is waiting` test is consistently deadlocking on Windows (other platforms are either fine or are much less consistent about it). I need to figure out why.

kring:
I found and fixed the race condition causing the deadlock. This is ready for another review.

CesiumAsync/include/CesiumAsync/Impl/QueuedScheduler.h
```diff
@@ -3,19 +3,88 @@
...
+    // the unblockTask continuation has not actually finished running in the
+    // worker thread yet. The main thread starts waiting on the condition
+    // variable again.
+    // 4. The `unblockTask` continuation finally finishes, making
+    // `unblockTask.ready()` return true, but it's too late. The main thread is
+    // already waiting on the condition variable.
+    //
+    // By setting the atomic `isDone` flag before calling `unblock`, we ensure
+    // that the loop termination condition is satisfied before the main thread
+    // is awoken, avoiding the potential deadlock.
```
csciguy8, line 44: Nice catch

CesiumAsync/src/QueuedScheduler.cpp
```diff
@@ -1,17 +1,66 @@
...
+}
+
+bool QueuedScheduler::dispatchInternal(bool blockIfNoTasks) {
+  async::task_run_handle t;
+
+  {
+    std::unique_lock<std::mutex> guard(this->_pImpl->mutex);
+    t = this->_pImpl->queue.pop();
+    if (blockIfNoTasks && !t) {
+      this->_pImpl->conditionVariable.wait(guard);
```
csciguy8, line 48: This looks fine, but I'm trying to imagine a catastrophic situation where `::wait_for` would be preferable, and waiting with a timeout is better, but nothing is coming to mind.

PR #826, opened 2024-03-04
Add `pointInTriangle` for 3D triangles + output for barycentric coordinates

j9liu:
This adds `IntersectionTests::pointInTriangle`, a 3D version of `pointInTriangle2D`. There is an overload that outputs the barycentric coordinates of the point, assuming the intersection is valid.

There may be a more efficient way of computing the barycentric coordinates, but my linear algebra is really rusty and I didn't want to copy + paste random code from the internet. So I went with the method I learned in college 😃

### DIFF
--- a/CesiumGeometry/include/CesiumGeometry/IntersectionTests.h
+++ b/CesiumGeometry/include/CesiumGeometry/IntersectionTests.h
@@ -42,6 +42,38 @@ public:
       const glm::dvec2& triangleVertA,
       const glm::dvec2& triangleVertB,
       const glm::dvec2& triangleVertC) noexcept;
+
+  /**
+   * @brief Determines whether the point is completely inside the triangle.
+   *
+   * @param point The point to check.
+   * @param triangleVertA The first vertex of the triangle.
+   * @param triangleVertB The second vertex of the triangle.
+   * @param triangleVertC The third vertex of the triangle.
+   * @return Whether the point is within the triangle.
+   */
+  static bool pointInTriangle(
+      const glm::dvec3& point,
+      const glm::dvec3& triangleVertA,
+      const glm::dvec3& triangleVertB,
+      const glm::dvec3& triangleVertC) noexcept;
+
+  /**
+   * @brief Determines whether the point is completely inside the triangle and
+   * outputs the barycentric coordinates for the point.
+   *
+   * @param point The point to check.
+   * @param triangleVertA The first vertex of the triangle.
+   * @param triangleVertB The second vertex of the triangle.
+   * @param triangleVertC The third vertex of the triangle.
+   * @return Whether the point is within the triangle.
+   */
+  static bool pointInTriangle(
+      const glm::dvec3& point,
+      const glm::dvec3& triangleVertA,
+      const glm::dvec3& triangleVertB,
+      const glm::dvec3& triangleVertC,
+      glm::dvec3& barycentricCoordinates) noexcept;
 };
 
 } // namespace CesiumGeometry
--- a/CesiumGeometry/include/CesiumGeometry/IntersectionTests.h
+++ b/CesiumGeometry/include/CesiumGeometry/IntersectionTests.h
@@ -51,7 +51,7 @@ public:
    * defined by three 3D points.
    *
    * Returns false for degenerate triangles.
-   * 
+   *
    * @param point The point to check.
    * @param triangleVertA The first vertex of the triangle.
    * @param triangleVertB The second vertex of the triangle.
--- a/CesiumGeometry/include/CesiumGeometry/IntersectionTests.h
+++ b/CesiumGeometry/include/CesiumGeometry/IntersectionTests.h
@@ -32,6 +32,8 @@ public:
    * @brief Determines whether a given point is completely inside a triangle
    * defined by three 2D points.
    *
+   * Returns false for degenerate triangles.
+   *
    * @param point The point to check.
    * @param triangleVertA The first vertex of the triangle.
    * @param triangleVertB The second vertex of the triangle.
@@ -48,6 +50,8 @@ public:
    * @brief Determines whether a given point is completely inside a triangle
    * defined by three 3D points.
    *
+   * Returns false for degenerate triangles.
+   * 
    * @param point The point to check.
    * @param triangleVertA The first vertex of the triangle.
    * @param triangleVertB The second vertex of the triangle.
@@ -64,6 +68,8 @@ public:
    * @brief Determines whether the point is completely inside a triangle defined
    * by three 3D points. If the point is inside, this also outputs the
    * barycentric coordinates for the point.
+   *
+   * Returns false for degenerate triangles.
    *
    * @param point The point to check.
    * @param triangleVertA The first vertex of the triangle.
--- a/CesiumGeometry/include/CesiumGeometry/IntersectionTests.h
+++ b/CesiumGeometry/include/CesiumGeometry/IntersectionTests.h
@@ -29,7 +29,8 @@ public:
   rayPlane(const Ray& ray, const Plane& plane) noexcept;
 
   /**
-   * @brief Determines whether the point is completely inside the triangle.
+   * @brief Determines whether a given point is completely inside a triangle
+   * defined by three 2D points.
    *
    * @param point The point to check.
    * @param triangleVertA The first vertex of the triangle.
@@ -37,14 +38,15 @@ public:
    * @param triangleVertC The third vertex of the triangle.
    * @return Whether the point is within the triangle.
    */
-  static bool pointInTriangle2D(
+  static bool pointInTriangle(
       const glm::dvec2& point,
       const glm::dvec2& triangleVertA,
       const glm::dvec2& triangleVertB,
       const glm::dvec2& triangleVertC) noexcept;
 
   /**
-   * @brief Determines whether the point is completely inside the triangle.
+   * @brief Determines whether a given point is completely inside a triangle
+   * defined by three 3D points.
    *
    * @param point The point to check.
    * @param triangleVertA The first vertex of the triangle.
@@ -59,8 +61,9 @@ public:
       const glm::dvec3& triangleVertC) noexcept;
 
   /**
-   * @brief Determines whether the point is completely inside the triangle and
-   * outputs the barycentric coordinates for the point.
+   * @brief Determines whether the point is completely inside a triangle defined
+   * by three 3D points. If the point is inside, this also outputs the
+   * barycentric coordinates for the point.
    *
    * @param point The point to check.
    * @param triangleVertA The first vertex of the triangle.
--- a/CesiumGeometry/src/IntersectionTests.cpp
+++ b/CesiumGeometry/src/IntersectionTests.cpp
@@ -118,6 +118,10 @@ bool IntersectionTests::pointInTriangle(
   }
 
   const double triangleCAPRatio = 1.0 - triangleABPRatio - triangleBCPRatio;
+  if (triangleCAPRatio < 0 || triangleCAPRatio > 1) {
+    return false;
+  }
+
   barycentricCoordinates.x = triangleBCPRatio;
   barycentricCoordinates.y = triangleCAPRatio;
   barycentricCoordinates.z = triangleABPRatio;
--- a/CesiumGeometry/src/IntersectionTests.cpp
+++ b/CesiumGeometry/src/IntersectionTests.cpp
@@ -76,6 +76,8 @@ bool IntersectionTests::pointInTriangle(
     const glm::dvec3& triangleVertA,
     const glm::dvec3& triangleVertB,
     const glm::dvec3& triangleVertC) noexcept {
+  // PERFORMANCE_IDEA: Investigate if there is a faster algorithm that can do
+  // this test, but without needing to compute barycentric coordinates
   glm::dvec3 unused;
   return IntersectionTests::pointInTriangle(
       point,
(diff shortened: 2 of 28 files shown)


csciguy8 (changes requested):
Thanks @j9liu!

Appreciate the new functionality, tests, and tests to cover the existing `::pointInTriangle2D` function.

The main issue I see is performance in `::pointInTriangle`. I count 3 branches (if) and 3 square root operations. Imagine this function called for thousands or millions of triangles. The less operations, and less expensive operations, the better.

If you want to stick to your algorithm, but make it faster, that works. But there are lots of other references, [here](https://blackpawn.com/texts/pointinpoly/), [another here](https://totologic.blogspot.com/2014/01/accurate-point-in-triangle-test.html), and of course [stack overflow discussion](https://stackoverflow.com/questions/37545304/determine-if-point-is-inside-triangle-in-3d)

CesiumGeometry/include/CesiumGeometry/IntersectionTests.h
```diff
@@ -42,6 +42,38 @@ class CESIUMGEOMETRY_API IntersectionTests final {
...
+  /**
+   * @brief Determines whether the point is completely inside the triangle.
+   *
+   * @param point The point to check.
+   * @param triangleVertA The first vertex of the triangle.
+   * @param triangleVertB The second vertex of the triangle.
+   * @param triangleVertC The third vertex of the triangle.
+   * @return Whether the point is within the triangle.
+   */
+  static bool pointInTriangle(
```
csciguy8, line 55: This may be an opportunity to make naming more consistent?

Either... rename `pointInTriangle2d` to just `pointInTriangle`, since we are function overloading anyway.

Or... rename `pointInTriangle` to `pointInTriangle3d` to tag along with the precendent already set.

CesiumGeometry/test/TestIntersectionTests.cpp
```diff
@@ -35,3 +35,214 @@ TEST_CASE("IntersectionTests::rayPlane") {
...
+  bool result = IntersectionTests::pointInTriangle(
+      testCase.point,
+      testCase.triangleVert1,
+      testCase.triangleVert2,
+      testCase.triangleVert3,
+      barycentricCoordinates);
+
+  REQUIRE(result == testCase.expected);
+  if (result) {
+    CHECK(barycentricCoordinates == testCase.expectedCoordinates);
```
csciguy8, line 246: We should probably always check that the returned `barycentricCoordinates` is what we expect, even when the function returns `false`.

j9liu:
Thanks for the review @csciguy8. Based on offline conversation I'm pushing back at changing the algorithm, at the moment. We aren't using it for a use case that demands the optimization. But I left a `PERFORMANCE` comment to be sure we remember. :)

I addressed your other comments, let me know if there are any other issues.

CesiumGeometry/src/IntersectionTests.cpp
```diff
@@ -31,31 +32,97 @@ IntersectionTests::rayPlane(const Ray& ray, const Plane& plane) noexcept {
...
+  const glm::dvec3 triangleNormal = glm::cross(ab, bc);
+  const double lengthSquared = glm::length2(triangleNormal);
+  if (lengthSquared < CesiumUtility::Math::Epsilon8) {
+    // Don't handle invalid triangles.
+    return false;
+  }
+
+  // Technically this is the area of the parallelogram between the vectors,
+  // but triangles are 1/2 of the area, and the division will cancel out later.
+  const double triangleAreaInv = 1.0 / glm::sqrt(lengthSquared);
```
timoore, line 104: You don't need to calculate the total area of the triangle. The barycentric coordinates must sum to 1.0, so you can add up your computed area coordinates (lengths of the cross products below) and then normalize by that at the end.

CesiumGeometry/src/IntersectionTests.cpp
```diff
@@ -31,31 +32,97 @@ IntersectionTests::rayPlane(const Ray& ray, const Plane& plane) noexcept {
...
+  }
+
+  const glm::dvec3 bp = point - triangleVertB;
+  const double triangleBCPRatio =
+      glm::length(glm::cross(bc, bp)) * triangleAreaInv;
+  if (triangleBCPRatio < 0 || triangleBCPRatio > 1) {
+    return false;
+  }
+
+  const double triangleCAPRatio = 1.0 - triangleABPRatio - triangleBCPRatio;
```
timoore, line 120: I think you need to test the third side. The test for the area being greater than the total triangle area (what you've done above) isn't sufficient.

CesiumGeometry/src/IntersectionTests.cpp
j9liu: Sorry I'm not sure I follow. How would you normalize the sum of the areas without computing the total area of the triangle?

I'm using the first two area sizes to deduce the third one with subtraction. If I only have the first two, then I still would have to deduce the first one by doing `totalArea - firstTriangle - secondTriangle`, and then normalizing it.

timoore:
If you are going to calculate barycentric coordinates, which is a good approach to finding the intersection point, I highly recommend checking out the section on triangle - ray intersection in the Physically Based Rendering book, https://www.pbr-book.org/4ed/Shapes/Triangle_Meshes . They project everything into 2D, so the cross products give the double area directly and don't need to be normalized.

CesiumGeometry/src/IntersectionTests.cpp
j9liu: I saw your other comment and agreed with it, and I added a check for `triangleCAPRatio` being negative. Can I just clarify how you are suggesting I rewrite the code? This is what I think I'm deducing from your comment:

```
 // Technically this is the area of the parallelogram between the vectors,
  // but triangles are 1/2 of the area, and the division will cancel out later.
  const glm::dvec3 ap = point - triangleVertA;
  const double areaABP =
      glm::length(glm::cross(ab, ap));

  const glm::dvec3 bp = point - triangleVertB;
  const double areaBCP =
      glm::length(glm::cross(bc, bp));

  const glm::dvec3 cp = point - triangleVertC;
  const double areaCAP =
      glm::length(glm::cross(ca, cp));
      
  const double denominator = areaABP + areaBCP + areaCAP;
  const double triangleABPRatio = areaABP / denominator;
  // etc.
  
  if (triangleABPRatio < 0 || triangleBCPRatio < 0 || triangleCAPRatio < 0)
    return false;
```

CesiumGeometry/src/IntersectionTests.cpp
```diff
@@ -31,31 +32,101 @@ IntersectionTests::rayPlane(const Ray& ray, const Plane& plane) noexcept {
...
+  if (triangleCAPRatio < 0) {
+    return false;
+  }
```
timoore, line 123: I take it back; if the sum of the two coordinates you test is > 1, then the test obviously fails. Nevertheless, I think it's better for accuracy to explicitly calculate all three coordinates and then normalize using their sum.

CesiumGeometry/src/IntersectionTests.cpp
j9liu: Sorry just realizing my error with the above (at the bottom), the areas will all be positive so they're never going to be negative. So I feel like I still have to check for the total area because, normalized, the sum could be greater than 1?

CesiumGeometry/src/IntersectionTests.cpp
j9liu: Just saw this, ignore my previous response

CesiumGeometry/src/IntersectionTests.cpp
timoore: Ack! Of course the areas aren't negative, so feel free to ignore most of my comments. I was fixated on the 2D case where the result of the cross product can be negative.

csciguy8:
I think I have an idea to narrow the scope of this PR a bit...

[The PR that depends on this](https://github.com/CesiumGS/cesium-unity/pull/428/files) looks like it doesn't really need a fast `pointInTriangle` test, it needs to calculate barycentric coordinates.

Given this, you may be able to:
* Leave the new `pointInTriangle` overload that calculates barycentric coords and if an intersection happens. Could even rename it with `Barycentric` somewhere in the name for clarity. In the comments, you could mention that this is not the function you would choose if performance was the highest concern.
* Remove the new `pointInTriangle` 3d point version completely. We don't need to worry about performance for a function that doesn't exist. We could always implement this functionality later, and could base it on the "check sides with dot product" approach, which has good performance, and is similar to our `pointInTriangle` 2D point logic.

In the end, we don't have to change this algorithm, nor are we leaving a performance TODO for ourselves.

csciguy8:
Was poking around the test code and noticed a couple things:
* The 3d triangle tests are only testing tris on the XY plane (all z values are 0)
* The tests are only testing right triangles
* We should probably test [degenerate tris](https://en.wikipedia.org/wiki/Degeneracy_(mathematics)#Triangle)

Also, not sure the best thing to do when the input is garbage, like equal to [NaN](https://en.cppreference.com/w/cpp/types/numeric_limits/quiet_NaN) or [infinity](https://en.cppreference.com/w/cpp/types/numeric_limits/infinity).

j9liu:
> I think I have an idea to narrow the scope of this PR a bit...
> 
> [The PR that depends on this](https://github.com/CesiumGS/cesium-unity/pull/428/files) looks like it doesn't really need a fast `pointInTriangle` test, it needs to calculate barycentric coordinates.
> 
> Given this, you may be able to:
> 
> * Leave the new `pointInTriangle` overload that calculates barycentric coords and if an intersection happens. Could even rename it with `Barycentric` somewhere in the name for clarity. In the comments, you could mention that this is not the function you would choose if performance was the highest concern.
> * Remove the new `pointInTriangle` 3d point version completely. We don't need to worry about performance for a function that doesn't exist. We could always implement this functionality later, and could base it on the "check sides with dot product" approach, which has good performance, and is similar to our `pointInTriangle` 2D point logic.
> 
> In the end, we don't have to change this algorithm, nor are we leaving a performance TODO for ourselves.

Not trying to be contrarian on purpose.

It's common practice for us to leave `PERFORMANCE_IDEA` comments in our software so I think it's 100% acceptable for us to leave it as is. In no way am I defending my algorithm as 100% optimized or performant, I fully recognize that it is not.

But, as a personal admission, I would not be able to derive an algorithm from those external resources from a conceptual level.  I simply cannot vouch for the code I would write as being completely derived by myself with linear algebra skills, rather than copying or taking heavy inspiration from the code you shared. These sources either do not have a license explicitly specified, or use noncommercial Creative Commons licenses, which are not compatible with Apache. (including the PBR book code that Tim linked unfortunately)

Both the licensing issues with using external resources, and the timeframe necessary for just getting this function, out there, are my main forces of resistance. I'm MORE than happy to review and accept a follow-up PR for a more optimized, license-compatible (or self-derived) implementation. I'm also willing to relearn linear algebra and recompute the algorithm by hand. But I sense that this isn't necessary for the scope of this PR, nor enough of a priority that this follow-up needs to happen ASAP.
 
I'm in the middle of processing your comments about unit tests, which I agree can and should be expanded. But I would like to avoid other modifications unless they are related to correctness or other points of necessity.

timoore:
> > I think I have an idea to narrow the scope of this PR a bit...

> But, as a personal admission, I would not be able to derive an algorithm from those external resources from a conceptual level. I simply cannot vouch for the code I would write as being completely derived by myself with linear algebra skills, rather than copying or taking heavy inspiration from the code you shared. These sources either do not have a license explicitly specified, or use noncommercial Creative Commons licenses, which are not compatible with Apache. (including the PBR book code that Tim linked unfortunately)
> 
Just a note that algorithms aren't copyrightable. I hadn't thought about the license on the PBR code cuz I would not think to copy / paste that code, nor recommend it; it's an implementation for a raytracer and goes to a great deal of trouble to stay in fp32, which would be nice but perhaps not possible for our needs. Nevertheless, the algorithm is pretty clever. I'm pretty sure you could write the code after reading the algorithm; heavy inspiration is not at crime :smile:

j9liu:
@csciguy8 

> The 3d triangle tests are only testing tris on the XY plane (all z values are 0)
The tests are only testing right triangles
We should probably test [degenerate tris](https://en.wikipedia.org/wiki/Degeneracy_(mathematics)#Triangle)

I addressed the first two with the most recent commits.

For the third item, what do you mean by testing degenerate triangles? Testing that they work with this algorithm, or testing that they are rejected?

csciguy8:
> For the third item, what do you mean by testing degenerate triangles? Testing that they work with this algorithm, or testing that they are rejected?

Yes, and yes :)  I think it's up to you as to whether you think intersections on a degenerate triangle are valid or not. Is a triangle that has degenerated into a line essentially now doing a pointInLine test? Or is a "no area" triangle not a valid object in 3d space so it should always return false?

I think the important thing is that it's known what the behavior will be (and that it doesn't crash).

j9liu:
@csciguy8 This is ready for another look when you get the chance!

csciguy8 (approved):

csciguy8:
@j9liu Looks good!

I took the liberty in adding another performance comment, and also made the reverse winding test apply to all tests (so we don't need explicit tests for this).

@timoore, I saw you had one comment that wasn't resolved yet. Are you satisfied with the latest algorithm?

timoore:
> @timoore, I saw you had one comment that wasn't resolved yet. Are you satisfied with the latest algorithm?

That comment referred to an earlier version and isn't relevant anymore. Looks fine to me.

PR #827, opened 2024-03-04
Add `PositionAccessorType` to `AccessorUtility.h`

j9liu:
Depends on #826 so merge that first.

This adds the `PositionAccessorType` typedef to `AccessorUtility.h`, as well as the `getPositionAccessorView` method. Although there's only one possible type of position accessor, it's still helpful to put the construction logic here to avoid duplication in our runtimes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Additions :tada:
 
 - Added `IntersectionTests::pointInTriangle`, which tests if a point is inside the triangle specified with 3D coordinates. One overload includes a `barycentricCoordinates` parameter that outputs the barycentric coordinates at that point.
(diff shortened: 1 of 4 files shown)


csciguy8 (approved):
Looks like a nice addition @j9liu !  Straight forward new functionality and tests. 

The only comment I have would be to prefer alias declarations (C++11) over `typedef`. They are basically equivalent, with subtle differences, but I tend to prefer them because they are easier to read.

Ex.
`typedef AccessorView<AccessorTypes::VEC3<float>> PositionAccessorType;`
becomes
`using PositionAccessorType = AccessorView<AccessorTypes::VEC3<float>>;`

... but it looks like typedefs are the precedent set, so I'll leave this up to you later.

PR #828, opened 2024-03-06
Fix problems with QuantizedMeshLoader-generated glTFs

kring:
Before this PR, the glTFs generated from quantized-mesh tiles were technically invalid. Our game engine clients weren't too fussed about these problems, which is why we didn't notice.

Fixed the following problems:

* The `asset.version` field was not set.
* The `byteLength` of the buffers was not set. Most clients just used `cesium.data.size()` anyway.
* The `target` type of the bufferView used to store indices was incorrectly set to `ARRAY_BUFFER` instead of `ELEMENT_ARRAY_BUFFER`.
* The bufferView used to store indices had its `byteStride` property set, which is not allowed.
* The glTF did not have any scenes.

### DIFF
--- a/Cesium3DTilesContent/test/TestQuantizedMeshLoader.cpp
+++ b/Cesium3DTilesContent/test/TestQuantizedMeshLoader.cpp
@@ -678,6 +678,47 @@ static void checkGeneratedGridNormal(
   }
 }
 
+static void checkGltfSanity(const Model& model) {
+  CHECK(model.asset.version == "2.0");
+  REQUIRE(model.scene >= 0);
+  REQUIRE(size_t(model.scene) < model.scenes.size());
+  CHECK(!model.getSafe(model.scenes, model.scene).nodes.empty());
+
+  for (const Buffer& buffer : model.buffers) {
+    CHECK(buffer.byteLength > 0);
+  }
+
+  for (const BufferView& bufferView : model.bufferViews) {
+    CHECK(bufferView.byteLength > 0);
+  }
+
+  for (const Mesh& mesh : model.meshes) {
+    for (const MeshPrimitive& primitive : mesh.primitives) {
+      const Accessor* pIndicesAccessor =
+          model.getSafe(&model.accessors, primitive.indices);
+      REQUIRE(pIndicesAccessor);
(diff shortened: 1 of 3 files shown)


csciguy8:
Do we have any unit tests for `QuantizedMeshLoader::load`? (an unsurprising question from me)

If not, sounds like it would be useful. Could mimic what a typical native client would do, and have all our assertions about what a proper glTF should look like.

kring:
> Do we have any unit tests for QuantizedMeshLoader::load?

Yes we do, though it looks like it missed a rename somewhere along the way and was named `TestQuantizedMeshContent.cpp` instead of `TestQuantizedMeshLoader.cpp`. I've now renamed it in this branch.

> Could mimic what a typical native client would do, and have all our assertions about what a proper glTF should look like.

That would be a glTF validator. It would be great to have such a thing in cesium-native, but it's way outside the scope of this PR.

I've updated this PR to add some assertions about the things that were fixed, though. This is really just closing the barn door after the horse already bolted, and does nothing to prove the generated glTFs aren't incorrect in other ways we haven't already noticed. But it's easy to do and perhaps could catch regressions in the future, so why not?

Cesium3DTilesContent/test/TestQuantizedMeshLoader.cpp
```diff
@@ -678,6 +678,47 @@ static void checkGeneratedGridNormal(
...
 }
 
+static void checkGltfSanity(const Model& model) {
+  CHECK(model.asset.version == "2.0");
+  REQUIRE(model.scene >= 0);
+  REQUIRE(size_t(model.scene) < model.scenes.size());
+  CHECK(!model.getSafe(model.scenes, model.scene).nodes.empty());
+
+  for (const Buffer& buffer : model.buffers) {
+    CHECK(buffer.byteLength > 0);
```
j9liu, line 688: I'd like to add one more check for sanity:

```suggestion
    CHECK(buffer.byteLength > 0);
    CHECK(buffer.byteLength == static_cast<int64_t>(buffer.cesium.data.size()));
```

j9liu:
Looks good @kring ! I'll wait for CI to pass, then merge.

PR #829, opened 2024-03-11
Fix compiling in GCC 13

lilleyse:
A few changes so that cesium-native compiles in GCC 13 (I tested with 13.1.0). I confirmed that it still compiles with GCC 9 and GCC 11.

* Updated Draco to pick up this fix: https://github.com/google/draco/issues/966
* Ignore `-Wstringop-overflow` when including `picosha2.h`
* Ignore `-Wdangling-reference` which was causing false positives in test files
* Include `<cstdint>` in more places


timoore:
I would turn off those warnings globally at the CMake level rather than in just the files where they are causing false positives now. I believe that these have a pretty bad reputation.

lilleyse:
I moved `dangling-reference` to CMake but kept `stringop-overflow` in the code since there are already other pragmas surrounding `<picosha2.h>`

j9liu:
I don't have GCC set up to try this locally, but these changes look good + make sense to me. Thanks @lilleyse !

PR #830, opened 2024-03-12
Fix `SimplePlanarEllipsoidCurve` when one of the points has a negative height

azrogers:
The `SimplePlanarEllipsoidCurve` used for flight paths currently doesn't work for paths where the source point is below the surface of the ellipsoid. It currently causes the calculated coordinates to be inverted, giving points on the other side of the earth. This is because the math in `SimplePlanarEllipsoidCurve` first scales the source and destination points to the surface of the ellipsoid, and then assumes that these surface points must be lower in elevation than the original points for the rest of the calculations. This meant that height wasn't calculated correctly, and crucially, `_sourceDirection` was calculated pointing the wrong direction, resulting in the inverted coordinates. I've fixed the math so that it should work the same when points are below the ellipsoid.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+- Fixed a bug where coordinates returned from `SimplePlanarEllipsoidCurve` were inverted if one of the input points had a negative height.
+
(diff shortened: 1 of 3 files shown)


azrogers:
@csciguy8 Updated CHANGES.md!

PR #779, opened 2023-12-15
Request gap refactor

csciguy8:

PR #831, opened 2024-03-14
Fix Tileset::ComputeLoadProgress incorrectly reporting done, when main thread loading exists

csciguy8:
Fixes performance tests reporting completion times that are a little too early (1 second in some warm cache tests).

Could also fix any bugs in user applications where they use `::computeLoadProgress` to determine loading completion. 
Ex. waiting for loading before taking a screenshot, some parts of the screenshot might not have loaded to final detail

This fixes a bug in the mentioned function, where `this->_workerThreadLoadQueue.size()` is being used to calculate tile completion, but it's always 0, because it is cleared at the end of every `::update_view`. The fix is to use the saved value in `_updateResult`.

Discovered while working on PR #779.

Also relates to PR #733, which has a similar fix. I wouldn't be surprised if this was the original cause, but that PR just happened to fix it too. Regardless, the older PR is still valuable. Counting "tiles kicked" as current progress sounds reasonable.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,7 @@
 
 - Fixed a bug in `BoundingVolume::estimateGlobeRectangle` where it returned an incorrect rectangle for boxes and spheres that encompass the entire globe.
 - Fixed an incorrect computation of wrapped texture coordinates in `applySamplerWrapS` and `applySamplerWrapT`.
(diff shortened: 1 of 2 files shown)


j9liu:
Thanks @csciguy8 !

Issue #765, opened 2023-11-19
Error building in Arch Linux


samtux:
In the Platform: x86_64 Arch Linux, GCC v13.2.1, clang  v16.0.6, cmake v3.27.8.

I get the next error building in Arch Linux: 

```
[ 25%] Built target Async++
[ 26%] Built target sqlite3
[ 26%] Building CXX object CesiumAsync/CMakeFiles/CesiumAsync.dir/src/SqliteCache.cpp.o
In file included from /home/smesa/makepkg/cesium-native/cesium-native/CesiumAsync/include/CesiumAsync/ICacheDatabase.h:3,
                 from /home/smesa/makepkg/cesium-native/cesium-native/CesiumAsync/include/CesiumAsync/SqliteCache.h:3,
                 from /home/smesa/makepkg/cesium-native/cesium-native/CesiumAsync/src/SqliteCache.cpp:1:
/home/smesa/makepkg/cesium-native/cesium-native/CesiumAsync/include/CesiumAsync/CacheItem.h:27:15: error: expected ')' before 'cacheStatusCode'
   27 |       uint16_t cacheStatusCode,
      |               ^~~~~~~~~~~~~~~~
      |               )
/home/smesa/makepkg/cesium-native/cesium-native/CesiumAsync/include/CesiumAsync/CacheItem.h:26:16: note: to match this '('
   26 |   CacheResponse(
      |                ^
/home/smesa/makepkg/cesium-native/cesium-native/CesiumAsync/include/CesiumAsync/CacheItem.h:37:3: error: 'uint16_t' does not name a type
   37 |   uint16_t statusCode;
      |   ^~~~~~~~
/home/smesa/makepkg/cesium-native/cesium-native/CesiumAsync/include/CesiumAsync/CacheItem.h:12:1: note: 'uint16_t' is defined in header '<cstdint>'; did you forget to '#include <cstdint>'?
   11 | #include <vector>
  +++ |+#include <cstdint>
   12 | 
/home/smesa/makepkg/cesium-native/cesium-native/CesiumAsync/include/CesiumAsync/ICacheDatabase.h:52:7: error: 'uint16_t' has not been declared
   52 |       uint16_t statusCode,
      |       ^~~~~~~~
[...]

PR #833, opened 2024-03-19
Addressed Include Errors for Linux/GCC Build

drewtu2:
Adds the `#include <cstdint>` header to a handful of headers to build successfully on linux. This does not address a similar issue in with the Draco library which is mentioned in the linked issue. I will open a separate PR for that. 

Built with Ubuntu 18.04 w/ gcc 14.0, cmake 3.29

This issue is discussed in https://github.com/CesiumGS/cesium-native/issues/765

Additional note: I need to remove the `-werror` to get this to build successfully. It seems like there are additional issue with some of the sign conversions that throw warnings, however that is not addressed in this PR.

### DIFF
--- a/CesiumAsync/include/CesiumAsync/CacheItem.h
+++ b/CesiumAsync/include/CesiumAsync/CacheItem.h
@@ -5,6 +5,7 @@
 
 #include <gsl/span>
 
+#include <cstdint>
 #include <cstddef>
 #include <cstdint>
 #include <ctime>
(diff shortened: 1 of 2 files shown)


drewtu2:
This might be covered by https://github.com/CesiumGS/cesium-native/pull/829

drewtu2:
after syncing and comparing, it looks like #829 addressed these changes as well. Closing this PR since it's already been fixed.

PR #835, opened 2024-03-21
Implement a class to load and sample an EGM96 grid

### LINKED ISSUE #637: Add a function to convert between WGS84 and EGM96/EGM2008 heights
Users often have heights referenced to mean sea level (MSL), so this would allow them to convert to ellipsoidal heights that are required for placement on the globe.

azrogers:
Closes #637.

Users often have data with heights referenced to mean sea-level (MSL), but Cesium uses the WGS84 ellipsoid internally, which can differ from MSL by dozens of meters in some locations. Accurately determining MSL requires an external geopotential model. This change allows users to load the EGM96 model and sample heights from it, allowing for MSL heights to be converted to ellipsoid heights. Converting from MSL to the ellipsoid becomes as simple as:

```
const std::optional<EarthGravitationalModel96Grid> grid = EarthGravitationalModel96Grid::fromFile("WW15MGH.DAC");
assert(grid.has_value());
const Cartographic mappedPosition = Cartographic(
    position.longitude,
    position.latitude,
    position.height + grid->sampleHeight(position));
```


CesiumGeospatial/include/CesiumGeospatial/EarthGravitationalModel96Grid.h
```diff
@@ -0,0 +1,72 @@
...
+private:
+  EarthGravitationalModel96Grid(const std::vector<int16_t>& gridValues);
+
+  /**
+   * Returns the height of the given grid value, in meters.
+   */
+  double getHeightForIndices(const int vertical, const int horizontal) const;
+
+  /**
+   * Ensures latitude is in the range(-PI, PI) with an exclusive upper bound -
```
timoore, line 58: Latitude actually ranges from pi/2 (90°) to -pi/2 (-90°).

CesiumGeospatial/src/EarthGravitationalModel96Grid.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+  const std::vector<int16_t>::size_type index =
+      static_cast<std::vector<int16_t>::size_type>(
+          clampedVertical * NUM_COLUMNS + horizontal);
+  const double result = static_cast<double>(_gridValues[index]) / 100.0;
+
+  return result;
+}
+
+double EarthGravitationalModel96Grid::mapLatitude(double latitude) const {
```
timoore, line 116: As I said above, I think clamping to the correct bounds is better.

CesiumGeospatial/src/EarthGravitationalModel96Grid.cpp
```diff
@@ -0,0 +1,132 @@
...
+  return latitude - Math::TwoPi * floor((latitude + Math::OnePi) / Math::TwoPi);
+}
+
+double EarthGravitationalModel96Grid::mapLongitude(double longitude) const {
+  const double modTwoPi =
+      fmod(fmod(longitude, Math::TwoPi) + Math::TwoPi, Math::TwoPi);
```
timoore, line 122: I would recommend std::remainder(longitude - Math::Pi, Math::TwoPi) + Math::Pi. This avoids the kinda nasty double fmod.

CesiumGeospatial/src/EarthGravitationalModel96Grid.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+  if (Math::equalsEpsilon(modTwoPi, 0, Math::Epsilon14) &&
```
timoore, line 124: What's this about?

azrogers:
@timoore Updated review based on your comments. Latitude values are now clamped.

timoore:
I'm ready to merge this, but I notice that the WW15MGH.DAC file is only test data. What's the plan to make this more official? Egm96 height will very useful to users and should be easy to use from cesium-native. Perhaps the source data file should be transformed into C++ source by a python script or something and linked in?

azrogers:
We could definitely include WW15MGH.DAC in the binary, by generating a header file either [manually](https://notisrac.github.io/FileToCArray/) or as part of the build process. However, I wonder if it doesn't just make more sense to have users bring their own WW15MGH.DAC? After all, most users probably won't touch the EGM96 grid, so it would save on 2MB of extra dependencies for people who don't. We'll definitely have to do this if we decide to add other models (the EGM08 grid is 128MB) and it makes sense to me to keep it consistent, even if 2MB isn't *that* much.

timoore:
> We could definitely include WW15MGH.DAC in the binary, by generating a header file either [manually](https://notisrac.github.io/FileToCArray/) or as part of the build process. However, I wonder if it doesn't just make more sense to have users bring their own WW15MGH.DAC? After all, most users probably won't touch the EGM96 grid, so it would save on 2MB of extra dependencies for people who don't. We'll definitely have to do this if we decide to add other models (the EGM08 grid is 128MB) and it makes sense to me to keep it consistent, even if 2MB isn't _that_ much.

I see a lot of forum posts that boil down to ellipsoid vs geoid confusion, so I think this functionality is useful; on the other hand, I don't see users bringing their own WW15MGH.DAC file; where would they even get it? In other words, the set of users helped by this new functionality is a lot larger than the set than know how to help themselves.

I don't care much if the data is a built-in array or read from a file. Do we have any external data files in cesium-native that need to be installed and used by the clients? The clients e.g. cesium-omniverse already have external data dependencies.

kring:
For our integrations (Unreal, Unity, Omniverse), we should include the WW15MGH.DAC file with the distribution so users don't have to think about it.

For cesium-native, users should supply the file or its contents when constructing an instance of the type, as is already done in this PR. But we should make it easy for them to do so by telling people in the reference docs (doxygen comments) where to get it. It used to be really easy to get this file directly from NGA, but that doesn't seem to be true anymore. So we should probably also include it in a `data` directory (but maybe not under `test`?)._

CesiumGeospatial/include/CesiumGeospatial/EarthGravitationalModel96Grid.h
```diff
@@ -0,0 +1,60 @@
...
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace CesiumGeospatial {
+
+/**
+ * @brief Allows elevation data in the format provided by the Earth
+ * Gravitational Model 1996 to be loaded and queried.
```
kring, line 18: Active voice would be better:

```suggestion
 * @brief Loads and queries heights from the Earth Gravity Model 1996 (EGM96).
```

Just call me the MS Word grammar checker. 😎 

Also elevation -> height. The word elevation tends to imply a height above sea level, so we avoid it.

CesiumGeospatial/include/CesiumGeospatial/EarthGravitationalModel96Grid.h
```diff
@@ -0,0 +1,60 @@
...
+  fromBuffer(const gsl::span<const std::byte>& buffer);
+
+  /**
+   * Samples the height at the given position.
+   *
+   * @param position The position to sample.
+   * @returns The height representing the difference in meters of mean sea-level
+   * from the surface of a WGS84 ellipsoid.
+   */
+  double sampleHeight(Cartographic& position) const;
```
kring, line 47:
```suggestion
  double sampleHeight(const Cartographic& position) const;
```

CesiumGeospatial/include/CesiumGeospatial/EarthGravitationalModel96Grid.h
```diff
@@ -0,0 +1,60 @@
...
+   */
+  static std::optional<EarthGravitationalModel96Grid>
+  fromBuffer(const gsl::span<const std::byte>& buffer);
+
+  /**
+   * Samples the height at the given position.
+   *
+   * @param position The position to sample.
+   * @returns The height representing the difference in meters of mean sea-level
+   * from the surface of a WGS84 ellipsoid.
```
kring, line 45: Can we state this a little more clearly? The current wording doesn't make it clear to me whether the returned value is positive or negative when the EGM96 surface is above the ellipsoid. Something like:

"The height (in meters) of the EGM96 surface above the WGS84 ellipsoid. A positive value indicates that the EGM96 surface is above the ellipsoid, while a negative value indicates that the MSL surface is below the ellipsoid."

(Please double check I have that right! I can never remember if it's that or the opposite. 😆)

CesiumGeospatial/include/CesiumGeospatial/EarthGravitationalModel96Grid.h
```diff
@@ -0,0 +1,60 @@
...
+   */
+  double sampleHeight(Cartographic& position) const;
+
+private:
+  EarthGravitationalModel96Grid(const std::vector<int16_t>& gridValues);
```
kring, line 50: Save a copy!

```suggestion
  EarthGravitationalModel96Grid(std::vector<int16_t>&& gridValues);
```

CesiumGeospatial/src/EarthGravitationalModel96Grid.cpp
```diff
@@ -0,0 +1,119 @@
...
+    // WW15MGH.DAC is in big endian, so we swap the bytes
+    const std::byte msb = buffer.data()[i];
+    const std::byte lsb = buffer.data()[i + 1];
+    const int16_t gridValue =
+        static_cast<int16_t>(lsb) |
+        static_cast<int16_t>(static_cast<int16_t>(msb) << 8);
+    gridValues.push_back(gridValue);
+  }
+
+  return EarthGravitationalModel96Grid(gridValues);
```
kring, line 60:
```suggestion
  return EarthGravitationalModel96Grid(std::move(gridValues));
```

CesiumGeospatial/src/EarthGravitationalModel96Grid.cpp
```diff
@@ -0,0 +1,119 @@
...
+    const gsl::span<const std::byte>& buffer) {
+  const int expectedBytes = NUM_ROWS * NUM_COLUMNS * sizeof(int16_t);
+
+  // Not enough data - is this a valid WW15MGH.DAC?
+  if (buffer.size_bytes() < expectedBytes) {
+    return std::nullopt;
+  }
+
+  std::vector<int16_t> gridValues;
+  const int size = static_cast<int>(buffer.size_bytes());
```
kring, line 48: Use `size_t` instead of `int`, here and below.

CesiumGeospatial/src/EarthGravitationalModel96Grid.cpp
```diff
@@ -0,0 +1,119 @@
...
+CesiumGeospatial::EarthGravitationalModel96Grid::fromBuffer(
+    const gsl::span<const std::byte>& buffer) {
+  const int expectedBytes = NUM_ROWS * NUM_COLUMNS * sizeof(int16_t);
+
+  // Not enough data - is this a valid WW15MGH.DAC?
+  if (buffer.size_bytes() < expectedBytes) {
+    return std::nullopt;
+  }
+
+  std::vector<int16_t> gridValues;
```
kring, line 47: Use `gridValues.reserve` to avoid reallocating the array multiple times while adding values.

CesiumGeospatial/src/EarthGravitationalModel96Grid.cpp
```diff
@@ -0,0 +1,119 @@
...
+  const int size = static_cast<int>(buffer.size_bytes());
+
+  for (int i = 0; i < size; i += 2) {
+    // WW15MGH.DAC is in big endian, so we swap the bytes
+    const std::byte msb = buffer.data()[i];
+    const std::byte lsb = buffer.data()[i + 1];
+    const int16_t gridValue =
+        static_cast<int16_t>(lsb) |
+        static_cast<int16_t>(static_cast<int16_t>(msb) << 8);
```
kring, line 56: This might be fine, but I'm not confident because I'm unsure of the sign extension behavior. I'd be more confident in a simple byte swap. Something like:

```cpp
assert((buffer.size() % 2) == 0);
std::vector<int16_t> gridValues;
gridValues.resize(buffer.size() / 2);
const std::byte* pRead = buffer.data();
std::byte* pWrite = reinterpret_cast<std::byte*>(gridValues.data());
for (size_t i = 0; i < buffer.size_bytes(); i += 2) {
  // WW15MGH.DAC is in big endian, so we swap the bytes
  pWrite[i] = pRead[i + 1];
  pWrite[i + 1] = pRead[i];
}
```

CesiumGeospatial/src/EarthGravitationalModel96Grid.cpp
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  return EarthGravitationalModel96Grid(gridValues);
+}
+
+double
+EarthGravitationalModel96Grid::sampleHeight(Cartographic& position) const {
+  const double longitude =
+      std::remainder(position.longitude - Math::OnePi, Math::TwoPi) +
+      Math::OnePi;
```
kring, line 67: Use `CesiumUtility::Math::zeroToTwoPi`.

CesiumGeospatial/src/EarthGravitationalModel96Grid.cpp
```diff
@@ -0,0 +1,119 @@
...
+  const double latitude =
+      std::clamp(position.latitude, -Math::PiOverTwo, Math::PiOverTwo);
```
kring, line 69:
```suggestion
  const double latitude =
      Math::clamp(position.latitude, -Math::PiOverTwo, Math::PiOverTwo);
```

CesiumGeospatial/src/EarthGravitationalModel96Grid.cpp
```diff
@@ -0,0 +1,119 @@
...
+double EarthGravitationalModel96Grid::getHeightForIndices(
+    const int horizontal,
+    const int vertical) const {
+
+  int clampedVertical = vertical;
+  if (vertical > NUM_ROWS - 1) {
+    clampedVertical = NUM_ROWS - 1;
+  }
+
+  const std::vector<int16_t>::size_type index =
```
kring, line 111: You can just use `size_t` instead of `std::vector<int16_t>::size_type`.

CesiumGeospatial/src/EarthGravitationalModel96Grid.cpp
```diff
@@ -0,0 +1,119 @@
...
+      static_cast<std::vector<int16_t>::size_type>(
+          clampedVertical * NUM_COLUMNS + horizontal);
+  const double result = static_cast<double>(_gridValues[index]) / 100.0;
```
kring, line 114: Is this static_cast needed? I think it shouldn't be.

CesiumGeospatial/test/TestEarthGravitationalModel96Grid.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+// Long, Lat values calculated randomly and paired with expected results from
+// https://www.unavco.org/software/geodetic-utilities/geoid-height-calculator/geoid-height-calculator.html
+const Egm96TestCase testCases[] = {
+    Egm96TestCase(
+        Cartographic::fromDegrees(135.89012584487307, 11.046411138991914, 0),
+        57.79),
+    Egm96TestCase(
+        Cartographic::fromDegrees(179.78766535213848, -66.77911223257036, 0),
+        -57.37),
```
kring, line 28: You're testing both positive and negative values, so my fears about sign extension are unfounded. Up to you if you like the code I suggested better anyway.

kring (commented):
Lots of little comments, but it looks great overall @azrogers! Also please update CHANGES.md.

kring:
@azrogers bump! Not sure if you saw that I reviewed this.

CesiumGeospatial/include/CesiumGeospatial/EarthGravitationalModel96Grid.h
azrogers: It's surprisingly hard to find out which one it is, but it seems like other tools loading EGM96 treat positive as above the ellipsoid.

kring:
Sorry for the delay on this @azrogers, somehow I didn't realize you had addressed the feedback in the review! Looking at it again today, there were a few more changes I wanted to see. But since it's been so long since you worked on it, I decided to just make those few changes myself. Then it turned into a few more changes. Here's the full list, just FYI (nothing too major though, this looks great overall):

1. Renamed the class to `EarthGravitationalModel1996Grid`. 1996 instead of 96.
2. Removed some unnecessary includes where a forward declaration would suffice, especially in headers.
3. Added a bit more detail to some of the doc comments.
4. Eliminated a few more uses of `int` (replaced with more specific integer types).
5. Eliminated use of `const` with non-reference types in the `getHeightForIndices` declaration. This is required by our [style guide](https://github.com/CesiumGS/cesium-native/blob/main/doc/style-guide.md#-const-by-value-parameters).
6. Made the `NUM_ROWS` and `NUM_COLUMNS` constants constexpr and put them in the anonymous namespace. No reason to add things like this to the symbol table unnecessarily.
7. Fixed up the handling of unicode in `EarthGravitationalModel1996Grid::fromFile`. Previously, non-ascii characters probably wouldn't have worked right on Windows.
8. Only read the required number of bytes for a valid grid from the file, rather than reading the entire file no matter how large it is.
9. Moved the `WW15MGH.DAC` file into a top-level `data` directory.
10. Fixed reads past the end of a buffer in the tests caused by constructs like `gsl::span<const std::byte>(&zeroByte, 3000001)`. Actually reading the bytes from that span will result in undefined behavior, because only one byte actually exists at the pointer and the span refers to 3 million and one bytes. Usually that undefined behavior will actually end up being "it works just fine but you get random bytes", which is what the tests expect. But crashing wouldn't be unlikely and technically anything at all is allowed to happen in this scenario according to the C++ standard.
11. Changed the `equalsEpsilon` height checks in the tests to use an absolute epsilon, which is I think what was intended.
12. Removed the "Fails on odd-length buffer" test, as well as the implementation code that checks for this case. If we're ignoring extra bytes past the expected grid, then it really doesn't matter if there are an odd or even number of those extra bytes.
13. Moved all of the implementation of `OwnedTempFile` into the .cpp.

Thanks again @azrogers. Merging!

PR #837, opened 2024-03-22
Allow externally supplied SPDLOG

csciguy8:
Based off original community contribution in [this PR](https://github.com/CesiumGS/cesium-native/pull/816). 

This PR makes the following changes:

* Allow using an externally supplied spdlog (by checking if the target exists before add_subdirectory).
* ~Allow using a newer version of fmt (which is bundled in newer versions of spdlog).~
* ~Upgraded the spdlog submodule reference from 1.10.0 to 1.13.0~

Investigates updating spdlog to 1.13.0, but this proved to be incompatible with Unreal 5.1 at this time. I think we can update once UE 5.3 becomes our minimum supported version.

### DIFF
--- a/Cesium3DTilesContent/src/B3dmToGltfConverter.cpp
+++ b/Cesium3DTilesContent/src/B3dmToGltfConverter.cpp
@@ -153,7 +153,7 @@ rapidjson::Document parseFeatureTableJsonData(
     result.errors.emplaceError(fmt::format(
         "Error when parsing feature table JSON, error code {} at byte offset "
         "{}",
-        static_cast<int32_t>(document.GetParseError()),
+        document.GetParseError(),
         document.GetErrorOffset()));
     return document;
   }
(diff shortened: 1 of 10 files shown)


csciguy8:
Looks like the Unreal 5.3 build system is having a hard time with this code...
image

Compilation is fine with spdlog 1.10, but any later version shows this compile error.

csciguy8:
So the cesium-unreal build problems go away if updating build settings to V4. This is probably something we want to do anyway.

Not really seeing any other way of getting spdlog to compile. `has_format_as` isn't driven by any other compile settings.

image

csciguy8:
This PR is ready for a look.

In the end, decided to wait on updating spdlog until cesium-unreal can support this upgrade. The notes in this PR will be valuable for that. 

Also, allowing an externally supplied spdlog is still a valuable feature, even if spdlog itself isn't upgraded.

kring:
Thanks @csciguy8!

PR #838, opened 2024-03-26
Fix Typo in param names

germanmarques:
Fixes typos in the param names in comments in the LocalHorizontalCoordinateSystem header file


csciguy8:
Looks good, thanks for the catch @germanmarques !

PR #840, opened 2024-03-29
Create AccessorWriter from AccessorView

lilleyse:
Adds an `AccessorWriter` constructor that takes an `AccessorView`.

This is helpful when you have an `AccessorView` but don't have the underlying `Accessor` or index, like if you're using some of the helper functions in `AccessorUtility`.


kring:
Thanks @lilleyse!

PR #841, opened 2024-04-02
Support for Cesium ion servers running in single-user authentication mode.

azrogers:
Out of the box, Cesium ion Self-Hosted is set-up in the single-user authentication mode, which means that it has no accounts or tokens - any request that can reach the server is considered valid. This is a problem for cesium-native and its integrations, which expect Cesium ion to always require a token. 

This change is a necessary part of adding this support. It changes cesium-native to allow requesting the active authentication mode from the `/appData` endpoint, and allows Unity and Unreal integrations to access the current authentication mode. It also provides dummy responses for the `/me` and `/tokens` endpoints when running in single-user mode, as these endpoints don't exist in the single-user configuration.


CesiumIonClient/include/CesiumIonClient/ApplicationData.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include <cstdint>
+#include <string>
+
+namespace CesiumIonClient {
+
+/**
+ *  @brief An enumeration representing the values of {@link ApplicationData::applicationMode}.
+ */
+enum ApplicationMode { CesiumIon, Saml, SingleUser };
```
timoore, line 11: I had to go look up "Saml." If it's a term of art that every programmer should know these days, then I'll try to get with the program, but I think a comment would be useful; at least, what's different about the three different modes.

timoore, line 11: General comment about "Application": you're describing characteristics of the ion server, and that really doesn't have anything to do with an "application." I think you could come up with a more descriptive name i.e. "IonServerState".

CesiumIonClient/include/CesiumIonClient/Connection.h
```diff
@@ -103,9 +104,20 @@ class CESIUMASYNC_API Connection {
...
+      const CesiumIonClient::ApplicationData& appData,
       const std::string& ionApiUrl = "https://api.cesium.com/",
       const std::string& ionAuthorizeUrl = "https://ion.cesium.com/oauth");
 
+  /**
+   * @brief Retrieves information about the ion API server.
+   *
+   * @return A future that resolves to the application information.
+   */
+  static CesiumAsync::Future<Response<ApplicationData>> appData(
```
timoore, line 116: Kinda think the verb form "getAppData" would be more descriptive.

CesiumIonClient/src/Connection.cpp
```diff
@@ -364,6 +395,60 @@ CesiumAsync::Future<Response<Profile>> Connection::me() const {
...
+CesiumIonClient::Connection::appData(
+    const CesiumAsync::AsyncSystem& asyncSystem,
+    const std::shared_ptr<CesiumAsync::IAssetAccessor>& pAssetAccessor,
+    const std::string& apiUrl) {
+  return pAssetAccessor
+      ->get(
+          asyncSystem,
+          CesiumUtility::Uri::resolve(apiUrl, "/appData"),
+          {{"Accept", "application/json"}})
+      .thenInMainThread([](std::shared_ptr<CesiumAsync::IAssetRequest>&&
```
timoore, line 408: Why in the main thread? I totally accept the answer "because every other function in this file resolves the future in the main thread" and I would have written it the same way for that reason, but it would be nice to have a comment somewhere explaining why this has to be done in the main thread.

csciguy8: I'm not seeing anything in this lambda that's obviously thread-unsafe, or can even be accessed by other functions. It's just building a result for the caller.

I would choose `::thenImmediately`. Odds are this isn't a performance critical path though, difference is probably academic.

CesiumIonClient/include/CesiumIonClient/Connection.h
azrogers: This would make sense, but I wanted to keep the naming consistent with the rest of the methods. In this class, all the get* methods return private members of the class, whereas all the methods calling endpoints are named with that endpoint name (`me`, `assets`, `defaults`, etc).

CesiumIonClient/src/Connection.cpp
```diff
@@ -304,6 +307,34 @@ bool parseJsonObject(const IAssetResponse* pResponse, rapidjson::Document& d) {
...
+    profile.email = "none@example.com";
+    profile.emailVerified = true;
+    profile.scopes = {
+        "assets:read",
+        "assets:list",
+        "assets:write",
+        "profile:read",
+        "tokens:read",
+        "tokens:write"};
+    profile.avatar = "https://www.gravatar.com/avatar/"
```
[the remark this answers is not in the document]
azrogers: ion accounts use Gravatar for avatars - it's a service where you can upload an image that gets associated with your email, and other websites can look up your avatar by email and use it. If you don't have an image set it will just give you the default Gravatar avatar, which is the same as you get just requesting `https://www.gravatar.com/avatar/`.

CesiumIonClient/include/CesiumIonClient/ApplicationData.h
```diff
@@ -0,0 +1,57 @@
...
+   * permissions to use its endpoints. In this mode, some endpoints (like /me
+   * and /tokens) are unavailable.
+   */
+  SingleUser
+};
+
+/**
+ * @brief Data about the Cesium ion server itself.
+ */
+struct ApplicationData {
```
j9liu, line 35: @azrogers I'm realizing that I forgot to leave a comment on this :sob:

I think a better name would be `ServerData` instead? At least for me, `ApplicationData` implies data about the application that the server is being used for, and not the server itself.

I don't want to trouble you for another find + replace change, so I'll push a commit for it but please let me know if you disagree.

j9liu: eh actually on second thought, Cesium ion _does_ call it appData so I guess we should keep consistent. I'll just add documentation tweaks so it's a little more obvious

j9liu:
Thanks @azrogers !

PR #842, opened 2024-04-04
Add new CesiumQuantizedMeshTerrain library/namespace

kring:
This is a PR into #828 so merge that first.

The main motivation for moving the quantized-mesh and layer.json functionality it to its own library is that this makes it easier to use the same code generator we use for glTF and 3D Tiles to generate classes for reading/writing layer.json.

So, this new library / namespace has:

* `QuantizedMeshLoader` (previously in `Cesium3DTilesContent`)
* Generated code for representing, reading, and writing layer.json.
* Overridden `Layer` class adding some helpers for working with this format to the generated layer.json class.

Other related bonus changes:

* Added support for arrays of arrays in generated classes. This is needed for layer.json availability rectangles.
* Fixed some name conflicts between the generated Reader and Writer classes, so that both can be generated into a single library / namespace.
* Fixed a bug in resolving schemas during code generation when the schema files are stored on disk rather than on the web.
* Changed the generator to avoid generating properties named `extensions`, `extras`, or `unknownProperties`, because these will conflict with built-in fields of the generated classes.


csciguy8:
> The term "legacy terrain" here refers to quantized-mesh and layer.json.

Kind of a pre-review question, and maybe naïve, but any reason we need to call this format "legacy"?

For example:
* The `CesiumLegacyTerrain` library could just be `CesiumTerrain`
* The class `LegacyTerrainUtilities` could be `QuantizedMeshUtilities`

Previous experiences have made me a little biased against using the word "legacy" in code bases, but am open to it if it has use.

kring:
Well, I think `CesiumTerrain` is way too general. Terrain is increasingly represented using 3D Tiles. IMO it's very much fair to call quantized-mesh and layer.json a legacy format at this point, though that doesn't mean it will go away tomorrow. Anyone who has a strong opinion on this, please chime in. If we don't like the term Legacy, I guess maybe I'd call it `CesiumQuantizedMeshTerrain`? Though that's not quite right either because this library would also include the even more legacy `heightmap-1.0` format if we ever feel the need to support it.

csciguy8:
> Anyone who has a strong opinion on this, please chime in.

I don't have a strong enough opinion to push too hard. More like, if I can convince you with the reasons, great. If not, let's move on.

(take these musings with a grain of salt, just a personal take)
- Naming things with "legacy" is naming them based on the plan for them, instead of what they actually do
- The plan for a legacy thing can change. It may not go away. Worst case, it actually becomes more prominent
- You still have to get past the name, and figure out what the thing actually is

pjcozzi:
@kring @csciguy8 I would suggest to avoid "legacy" in the name as Cesium is still actively supporting and producing data in quantized-mesh.

kring:
> I'm getting an error generating the gltf classes.

This was caused by a change in the glTF schemas. It was fixed in this PR:
https://github.com/CesiumGS/cesium-native/pull/851

So we just need to merge in main.

> I would suggest to avoid "legacy" in the name as Cesium is still actively supporting and producing data in quantized-mesh.

Thanks @pjcozzi.

Maybe `CesiumOGTerrain` then? 😆 In all seriousness, I think `CesiumQuantizedMeshTerain` is the best alternative, so I'll rename it to that.

pjcozzi:
> Maybe CesiumOGTerrain then? 😆 In all seriousness, I think CesiumQuantizedMeshTerain is the best alternative, so I'll rename it to that.

@kring lol. 👍

kring:
@csciguy8 I've merged in main and renamed the library / namespace. This is ready for another look.

CHANGES.md
```diff
@@ -2,10 +2,15 @@
 
 ### ? - ?
 
+##### Breaking Changes :mega:
+
+- Moved `QuantizedMeshLoader` from `Cesium3DTilesContent` to `CesiumQuantizedMeshTerrain`. If experiencing related linker errors, add `CesiumQuantizedMeshTerrain` to the libraries you link against.
```
csciguy8, line 7: I added a little more migration help here after testing these changes out in cesium-unreal.

CesiumJsonReader/include/CesiumJsonReader/ArrayJsonHandler.h
```diff
@@ -453,4 +453,112 @@ class CESIUMJSONREADER_API ArrayJsonHandler<std::string, StringJsonHandler>
...
+    this->_arrayIsOpen = false;
+    this->_elementHandler.reset(this->_handlerFactory());
+  }
+
+  virtual IJsonHandler* readNull() override {
+    return this->invalid("A null")->readNull();
+  }
+
+  virtual IJsonHandler* readBool(bool b) override {
+    return this->invalid("A bool")->readBool(b);
```
csciguy8, line 483: This was originally "A null", which I assumed was a typo. I've changed it, and another occurrence in this file.

csciguy8 (approved):
Looks great! @kring 

I've built and run the generator command locally as well as tried cesium-native with cesium-unreal. These appear to be nice fixes and reorganization.

Can you quickly sanity check the last 3 commits I snuck in? If all is well, let's merge this.

kring:
Your changes look good to me. Thanks @csciguy8!

PR #844, opened 2024-04-04
Add TileTransform::setTransform

kring:
To match the `getTransform` we already had. And add tests and doc for all.


j9liu (commented):
@kring looks good! Just had 2 comments that I'll go ahead and do myself to save you time. (But let me know if you disagree with anything)

PR #846, opened 2024-04-05
Add options to upsampleGltfForRasterOverlays to specify the texture coordinate base name and whether to invert the V texture coordinate

kring:
Previously, `upsampleGltfForRasterOverlays` assumed that raster overlay texture coordinates were found in attributes named `_CESIUMOVERLAY_n`, and that the origin of the V coordinate was at the Southern end of the tile. Both of these assumptions are fine in our game engine integrations where a custom shader does the overlay rendering. But they're problematic if we want to produce a standard glTF.

So this PR makes the base name of the texture coordinates a parameter, and defaults it to the standard glTF texture coordinate name (`TEXCOORD_n`). It also adds a `hasInvertedVCoordinate` parameter indicating that the model's V texture coordinate starts is 0.0 at the Northern end of the tile rather than the Southern end, consistent with a right-side-up map image.

Some other improvements in this PR:

* Images are now copied to the upsampled glTF. Previously they were lost.
* Fixed the `target` property of the created bufferviews for indices. It was incorrectly set to `ARRAY_BUFFER` instead of `ELEMENT_ARRAY_BUFFER`.
* Set the `byteLength` property of the created vertex and index buffers.


kring:
I added a tiny unrelated commit to this PR that improves a comment in `RasterOverlayUtilities.cpp`, for lack of a better place to put it.

Cesium3DTilesSelection/src/RasterOverlayUpsampler.cpp
```diff
@@ -72,6 +72,8 @@ RasterOverlayUpsampler::loadTileContent(const TileLoadInput& loadInput) {
         auto model = upsampleGltfForRasterOverlays(
             parentModel,
             TileID,
+            false,
+            "_CESIUMOVERLAY_",
```
csciguy8, line 76: Maintainability comment, I'm seeing a lot of hits (25) of the `_CESIUMOVERLAY_` string literal in the code.

Any chance we can move this to a central place (like `RasterOverlayUtilities.h`), or reduce the number of occurrences?

Cesium3DTilesContent/src/upsampleGltfForRasterOverlays.cpp
```diff
@@ -1134,6 +1158,9 @@ static void addSkirts(
       edgeIndices.north.end(),
       sortEdgeIndices.begin(),
       [](const EdgeVertex& v) { return v.index; });
+  if (hasInvertedVCoordinate) {
+    std::reverse(sortEdgeIndices.begin(), sortEdgeIndices.end());
```
csciguy8, line 1162: Is it possible to eliminate the `std::reverse` operation by just swapping the iterators in `std::transform` or inverting the comparison in the `std::sort`?

Cesium3DTilesSelection/src/RasterOverlayUpsampler.cpp
kring: Ok, I added a constant to RasterOverlayUtilities as you suggested and replaced _most_ of the occurrences of `_CESIUMOVERLAY_` with it. There are some remaining in documentation, as well as remaining occurrences of `"_CESIUMOVERLAY_0"` in the tests because I thought a bit of repitition was better than something like `fmt::format("{}_{}", RasterOverlayUtilities::DEFAULT_TEXTURE_COORDINATE_BASE_NAME, 0)`.

Cesium3DTilesContent/src/upsampleGltfForRasterOverlays.cpp
```diff
@@ -1076,6 +1097,9 @@ static void addSkirts(
       edgeIndices.south.end(),
       sortEdgeIndices.begin(),
       [](const EdgeVertex& v) { return v.index; });
+  if (hasInvertedVCoordinate) {
+    std::reverse(sortEdgeIndices.begin(), sortEdgeIndices.end());
```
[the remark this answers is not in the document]
kring: Yep, I incorporated it into the std::transforms.

kring:
This is ready for another look. In addition to your suggestions, I moved `upsampleGltfForRasterOverlays` into `RasterOverlayUtilities`. I think it makes a lot more sense that way than as a global function in `Cesium3DTilesContent`.

Cesium3DTilesSelection/src/RasterOverlayUpsampler.cpp
csciguy8: That's fair. And I suspected tedious code might be the result of that suggestion, so thank you for being sensible.

CesiumRasterOverlays/src/RasterOverlayUtilities.cpp
```diff
@@ -339,4 +517,1138 @@ RasterOverlayUtilities::createRasterOverlayTextureCoordinates(
...
+  const size_t indexBufferViewIndex = model.bufferViews.size();
+  model.bufferViews.emplace_back();
+
+  BufferView& vertexBufferView = model.bufferViews[vertexBufferViewIndex];
+  vertexBufferView.buffer = static_cast<int>(vertexBufferIndex);
+  vertexBufferView.target = BufferView::Target::ARRAY_BUFFER;
+
+  BufferView& indexBufferView = model.bufferViews[indexBufferViewIndex];
+  indexBufferView.buffer = static_cast<int>(indexBufferIndex);
+  indexBufferView.target = BufferView::Target::ELEMENT_ARRAY_BUFFER;
```
csciguy8, line 766: This looks like the fix mentioned in PR description.
(marking for easy lookup later)

CesiumRasterOverlays/src/RasterOverlayUtilities.cpp
```diff
@@ -339,4 +517,1138 @@ RasterOverlayUtilities::createRasterOverlayTextureCoordinates(
...
+  newIndicesAccessor.count = int64_t(indices.size());
+  newIndicesAccessor.componentType = Accessor::ComponentType::UNSIGNED_INT;
+  newIndicesAccessor.type = Accessor::Type::SCALAR;
+
+  // Populate the buffers
+  Buffer& vertexBuffer = model.buffers[vertexBufferIndex];
+  vertexBuffer.cesium.data.resize(newVertexFloats.size() * sizeof(float));
+  float* pAsFloats = reinterpret_cast<float*>(vertexBuffer.cesium.data.data());
+  std::copy(newVertexFloats.begin(), newVertexFloats.end(), pAsFloats);
+  vertexBuffer.byteLength = vertexBufferView.byteLength =
```
csciguy8, line 1064: This looks like the fix mentioned in PR description.
(marking for easy lookup later)

CesiumRasterOverlays/src/RasterOverlayUtilities.cpp
```diff
@@ -339,4 +517,1138 @@ RasterOverlayUtilities::createRasterOverlayTextureCoordinates(
...
+      int64_t(vertexBuffer.cesium.data.size());
+  vertexBufferView.byteStride = vertexSizeFloats * int64_t(sizeof(float));
+
+  Buffer& indexBuffer = model.buffers[indexBufferIndex];
+  indexBuffer.cesium.data.resize(indices.size() * sizeof(uint32_t));
+  uint32_t* pAsUint32s =
+      reinterpret_cast<uint32_t*>(indexBuffer.cesium.data.data());
+  std::copy(indices.begin(), indices.end(), pAsUint32s);
+  indexBuffer.byteLength = indexBufferView.byteLength =
```
csciguy8, line 1073: This looks like the fix mentioned in PR description.
(marking for easy lookup later)

CesiumRasterOverlays/src/RasterOverlayUtilities.cpp
```diff
@@ -277,6 +284,177 @@ RasterOverlayUtilities::createRasterOverlayTextureCoordinates(
...
+  result.extensionsRequired = parentModel.extensionsRequired;
+  result.asset = parentModel.asset;
+  result.extras = parentModel.extras;
+
+  // TODO: check if this is enough, not enough, or overkill
+  result.extensions = parentModel.extensions;
+  // result.extras_json_string = parentModel.extras_json_string;
+  // result.extensions_json_string = parentModel.extensions_json_string;
+
+  copyImages(parentModel, result);
```
csciguy8, line 408: This added line was mentioned as a fix in the PR.

@kring , do you have any suspicion that this might cause a performance problem?

More a sanity check question. I didn't see problems in my testing. But in theory it's copying image data, where it didn't copy it before.

kring: Yeah it could cause a small performance hit in certain cases. For example, when draping a raster overlay over a textured tileset like Google Photorealistic 3D Tiles. Previously, the decoded images were copied when upsampling, because these are stored in `gltf.images[n].cesium`. That's still true. But the image buffers prior to decoding (i.e., the original JPG or whatever) were not copied, and now they are. Instead, the image would previously end up referring to an invalid bufferView.

Given that upsampling isn't particularly performance-sensitive, I think the correctness is worth the cost here. But a nice improvement across the board - not just for upsampling - would be to _remove_ bufferViews (and sections of buffers) that hold the compressed images after we decompress them, in order to save memory. It's outside the scope of this PR though, I think.

PR #847, opened 2024-04-05
Add `BoundingRegionBuilder::toGlobeRectangle`

kring:
A small PR that allows the rectangle of a `BoundingRegionBuilder` to be accessed directly, without needing to construct a `BoundingRegion`, which is a much more time-consuming thing to do.


CesiumGeospatial/test/TestBoundingRegionBuilder.cpp
```diff
@@ -17,6 +19,10 @@ TEST_CASE("BoundingRegionBuilder") {
     CHECK(!rectangle.contains(Cartographic(0.0, 1.0, 0.0)));
     CHECK(!rectangle.contains(Cartographic(0.0, -1.0, 0.0)));
 
+    // The rectangle returned by toGlobeRectangle should be identical.
+    GlobeRectangle rectangle2 = builder.toGlobeRectangle();
+    CHECK(std::memcmp(&rectangle, &rectangle2, sizeof(GlobeRectangle)) == 0);
```
j9liu, line 24: Super nitpicky, but would you object to explicitly writing each edge check instead?

```
CHECK(rectangle.getWest() == rectangle2.getWest());
```
For me, these checks would make more sense from a glance. And if you're okay with it, I don't mind replacing the lines on your behalf :smile:

(Didn't realize you could do an equality test with `std::memcmp` , that's neat!)

kring: I'm not opposed, just lazy. Even better than putting the individual checks in the tests would be to provide an `equals` / `equalsEpsilon` method on `GlobeRectangle` itself. `memcmp` works great for something with 4 doubles like GlobeRectangle, but isn't a general solution. Even just padding between fields of different sizes could throw it off.

j9liu:
@kring I went ahead and added `GlobeRectangle::equals` and `GlobeRectangle::equalsEpsilon`. Since my # changed lines is probably larger than yours now, do you want to look over my changes and merge? 😅

kring:
Looks great @j9liu, thanks for adding the extra GlobeRectangle tests, too!

PR #849, opened 2024-04-05
Add `getPath` and `setPath` methods to the `Uri` class, fix a bug in `joinToString`

kring:

PR #850, opened 2024-04-05
Add TMS trailing slash at end of path, rather than end of url.

kring:
This is PR into #849, so merge that first.

`TileMapServiceRasterOverlay` attempts to add a `/` to the end of the URL it's given when that URL doesn't end in `.xml` and it doesn't already end in a slash. But the code was looking at the last character of the entire URL, rather than the last character of the path. So if the URL had query parameters, it didn't work at all.


csciguy8 (commented):
Thanks @kring , looks sensible, but I do have a question.

My test case was in cesium-unreal, using this url
https://sandcastle.cesium.com/CesiumUnminified/Assets/Textures/NaturalEarthII/tilemapresource.xml?some=parameter
image

Running through the debugger (with this PR), I did see that the url built in line 438 no longer has an extra slash on the end.

But user facing, I didn't see any difference. The url seemed to produce a valid map with no console errors, with or without these PR changes.

Was there a specific case that you were trying to fix here?

(btw, this kind of logic looks like a good candidate for a unit test, but I can write an issue for that later)

kring:
As we discussed offline, `https://sandcastle.cesium.com/CesiumUnminified/Assets/Textures/NaturalEarthII/tilemapresource.xml?some=parameter` should demonstrate the bug in that it will become `https://sandcastle.cesium.com/CesiumUnminified/Assets/Textures/NaturalEarthII/tilemapresource.xml?some=parameter/` before this PR (note the trailing slash at the end). But because the sandcastle.cesium.com web server ignores that query parameter, it doesn't matter if there's an extra slash on the end of it or not, so nothing breaks. In my case the query parameter was an API key, so the extra slash made it invalid.

kring:
I'll add a quick test for it, one moment... (mentioning it so you don't write that issue)

csciguy8 (approved):
Looks great @kring !

Will merge once CI finishes...

PR #852, opened 2024-04-08
Add support for `CESIUM_primitive_outline` extension.

kring:
This is a PR into #851 so merge that first.

Adds support for the `CESIUM_primitive_outline` glTF extension to CesiumGltf / CesiumGltfReader / CesiumGltfWriter. It doesn't do anything to actually render these outlines.

This PR only changes one file by hand: `tools/generate-classes/glTF.json`. The rest is generated code.


lilleyse:
Looks good

Issue #458, opened 2022-03-08
Support .i3dm tile payloads


argallegos:
Add support for [Instanced 3D Models](https://github.com/CesiumGS/3d-tiles/tree/main/specification/TileFormats/Instanced3DModel).

Once this is completed, some work will need to be done in Cesium Native integrations (e.g. Cesium for Unreal, Cesium for O3DE) before the feature is fully operational in each engine.

argallegos:
Pull request addressing this issue: https://github.com/CesiumGS/cesium-native/pull/290

argallegos:
Cesium for Unreal issue: https://github.com/CesiumGS/cesium-unreal/issues/671

Cesium for O3DE issue: https://github.com/CesiumGS/cesium-o3de/issues/32

lilleyse:
Useful diagram from https://github.com/CesiumGS/cesium/pull/10302 showing the expected transform order between i3dm and gltf + EXT_mesh_gpu_instancing

![image](https://user-images.githubusercontent.com/8422414/163865444-c92f5940-f10b-474f-a72b-ef6f03e73a0f.png)

lilleyse:
See https://github.com/CesiumGS/3d-tiles-tools/pull/52 for ideas for upgrading i3dm to EXT_mesh_gpu_instancing. It's not trivial because of transformation order differences (see diagram above).

CC @javagl

javagl:
One of the insight from the linked PR was that it is sometimes plainly (mathematically) not possible to upgrade I3DM to `EXT_mesh_gpu_instancing`. An I3DM could contain an animated GLB that was _non-uniformly scaled_ via the instancing transform . This can not be represented as a standard glTF animation.

It could be possible to lay out the map of "what is and what isn't possible" more strictly and systematically. But for the main case of I3DMs that contain "standard (non-animated) geometry", the approach from https://github.com/CesiumGS/3d-tiles-tools/pull/52#issuecomment-1679433253 turned out to be relatively easy, and could probably be applied in `cesium-native` as well, when there is a function to "bake" the transform of glTF nodes into the attached meshes.

lilleyse:
It could make sense to implement this in three phases:

* Add support for [`EXT_mesh_gpu_instancing`](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Vendor/EXT_mesh_gpu_instancing/README.md) in cesium-native and runtimes
[...]

PR #854, opened 2024-04-08
Support 3D Tile I3dm legacy tile format

timoore:
Initial commit for discussion. Some implementation remains, such as implementation of the ENU rotation option.

I3dm strains the GltfConverters interface because it can require a "sub read" of external content: the instanced glb file can be specified by a URI. I created a "ConverterSubprocessor" subclass (pretty bad name) to pass, the async system asset Accessor, base URI, etc. GltfConverters::convert returns a complete model, not a future, so I do a wait() on the accessor future that returns the .glb file. I don't know if that's bad form or if there's a better alternative.

This addresses #458.


Cesium3DTilesSelection/src/ImplicitOctreeLoader.cpp
```diff
@@ -147,7 +147,8 @@ CesiumAsync::Future<TileLoadResult> requestTileContent(
           CesiumGltfReader::GltfReaderOptions gltfOptions;
           gltfOptions.ktx2TranscodeTargets = ktx2TranscodeTargets;
           gltfOptions.applyTextureTransform = applyTextureTransform;
-          GltfConverterResult result = converter(responseData, gltfOptions);
+          GltfConverterResult result =
+              converter(responseData, gltfOptions, nullptr); // XXX
```
csciguy8, line 151: Did you leave the XXX here as a reminder for something?

Cesium3DTilesSelection/src/ImplicitQuadtreeLoader.cpp
```diff
@@ -157,7 +157,8 @@ CesiumAsync::Future<TileLoadResult> requestTileContent(
           CesiumGltfReader::GltfReaderOptions gltfOptions;
           gltfOptions.ktx2TranscodeTargets = ktx2TranscodeTargets;
           gltfOptions.applyTextureTransform = applyTextureTransform;
-          GltfConverterResult result = converter(responseData, gltfOptions);
+          GltfConverterResult result =
+              converter(responseData, gltfOptions, nullptr); // XXX
```
csciguy8, line 161: Did you leave the XXX here as a reminder for something?

csciguy8:
Haven't dug in real deep yet, but noticed there are no new unit tests .

Might be nice to add a few, especially if you know of typical use cases, or cases that should fail for some reason.

Cesium3DTilesSelection/src/ImplicitOctreeLoader.cpp
timoore: Yes, and will add comment. I wasn't sure how to calculate the tile transform for the implicit loaders.

Cesium3DTilesSelection/src/ImplicitQuadtreeLoader.cpp
timoore: See comment about ImplicitOctreeLoader.

kring:
> GltfConverters::convert returns a complete model, not a future, so I do a wait() on the accessor future that returns the .glb file. I don't know if that's bad form or if there's a better alternative.

Yeah adding a `wait` is pretty bad. In Unreal, for example, that'll block up Unreal's task graph on a network request, which in extreme cases could completely pause rendering.

Based on a quick look, I don't think it will be too big a change to make `GltfConverters::convert` return a `Future`. If it is, though, another possibility is to convert the i3dm to glTF synchronously and include references to the external resources in that glTF (using a private extension if needed). Then do the async portion later. If you happen to be able to hook into the existing `GltfReader::resolveExternalData`, the extra async step will just happen automatically. If not, you'll need to add another processing step near where `resolveExternalData` is located in order to do this extra async work.

csciguy8:
> GltfConverters::convert returns a complete model, not a future, so I do a wait() on the accessor future that returns the .glb file. I don't know if that's bad form or if there's a better alternative.

Agree with @kring , hidden waits aren't great when trying to improve performance.

Will also chime in that hidden `IAssetAccessor::get`s can become problematic too, especially when trying to manage simultaneous connections to a server, or decouple their execution from other futures that are just CPU processing work. This is the bulk of the work in this [Staged Loading PR](https://github.com/CesiumGS/cesium-native/pull/779), and it's not a trivial refactor.

timoore:
> > GltfConverters::convert returns a complete model, not a future, so I do a wait() on the accessor future that returns the .glb file. I don't know if that's bad form or if there's a better alternative.
> 
> Agree with @kring , hidden waits aren't great when trying to improve performance.
> 

I'm rewriting the converter code to return a future.

> Will also chime in that hidden `IAssetAccessor::get`s can become problematic too, especially when trying to manage simultaneous connections to a server, or decouple their execution from other futures that are just CPU processing work. This is the bulk of the work in this [Staged Loading PR](https://github.com/CesiumGS/cesium-native/pull/779), and it's not a trivial refactor.

What makes an IAssetAccessor::get call hidden? Or, put another way, what's the alternative when another get() needs to be made in the while processing the content of a previous read? I guess I should look at #779 to see what you're doing there.

csciguy8:
> What makes an IAssetAccessor::get call hidden? Or, put another way, what's the alternative when another get() needs to be made in the while processing the content of a previous read? I guess I should look at #779 to see what you're doing there.

That's a fair question. In #779, a good example of the approach is [here ](https://github.com/CesiumGS/cesium-native/blob/a16c3d5fa6ef974b2faede193f7cca80cba1e741/Cesium3DTilesSelection/src/ImplicitOctreeLoader.h#L43 )in `ImplicitOctreeLoader`
* The caller would typically call `::loadTileContent`, which returns a future, and somewhere in the chain is a `IAssetAccessor::get`
* This PR reworks the approach to have the caller use `::getLoadWork`, which returns the needed request and the CPU processing work separately
* This pain is in the rework of the code in the .cpp, and not break anything

Admittedly, this is "a" solution to separating asset request work from other work. This PR hasn't been reviewed or merged, so there may be a better solution still.

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,738 @@
...
+*/
+
+CesiumAsync::Future<ConvertResult> convertInstancesContent(
+    const gsl::span<const std::byte>& instancesBinary,
+    const InstancesHeader& header,
+    uint32_t headerLength,
+    const CesiumGltfReader::GltfReaderOptions& options,
+    const ConverterSubprocessor& subprocessor,
+    GltfConverterResult& result) {
+  ConvertResult subResult;
```
j9liu, line 225: If you rename `ConvertResult` -> `ConvertedI3dm`, then could you also rename this?

```suggestion
  ConvertedI3dm convertedI3dm;
```

It's more intuitive in my head to see `convertedI3dm.gltfResult` than `subResult.gltfResult`

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,738 @@
...
+  DecodedInstances& decodedInstances = subResult.decodedInstances;
+  subResult.gltfResult = result;
+  auto finishEarly = [&]() {
```
j9liu, line 228: No change here, but I like this lambda define and may steal this idea from you for the future :smile:

Cesium3DTilesContent/include/Cesium3DTilesContent/LegacyUtilities.h
```diff
@@ -0,0 +1,133 @@
...
+struct ByteResult {
+  std::vector<std::byte> bytes;
+  CesiumUtility::ErrorList errorList;
+};
+
+CesiumAsync::Future<ByteResult>
+get(const ConverterSubprocessor& subprocessor, const std::string& relativeUrl);
+
+namespace LegacyUtilities {
+std::optional<uint32_t> parseOffset(
```
j9liu, line 33:
```suggestion
std::optional<uint32_t> parseByteOffsetForSemantic(
```

Cesium3DTilesContent/src/LegacyUtilities.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+std::optional<glm::dvec3>
+parseArrayValueDVec3(const rapidjson::Value& arrayValue) {
+  if (validateJsonArrayValues(arrayValue, 3, &rapidjson::Value::IsNumber)) {
+    return std::make_optional(glm::dvec3(
+        arrayValue[0].GetDouble(),
+        arrayValue[1].GetDouble(),
+        arrayValue[2].GetDouble()));
+  }
+  return {};
```
j9liu, line 69: This probably resolves to a `std::nullopt`, but do you mind writing it out for a slow and easily-confused coworker like me? 😃 

```suggestion
  return std::nullopt;
```

Cesium3DTilesContent/src/LegacyUtilities.cpp
```diff
@@ -0,0 +1,181 @@
...
+}
+
+std::optional<glm::dvec3>
+parseArrayValueDVec3(const rapidjson::Document& document, const char* name) {
+  const auto arrayIt = document.FindMember(name);
+  if (arrayIt != document.MemberEnd()) {
+    return parseArrayValueDVec3(arrayIt->value);
+  }
+  return {};
```
j9liu, line 78:
```suggestion
  return std::nullopt;
```

Cesium3DTilesContent/include/Cesium3DTilesContent/LegacyUtilities.h
```diff
@@ -0,0 +1,133 @@
...
+
+std::optional<glm::dvec3>
+parseArrayValueDVec3(const rapidjson::Value& arrayValue);
+
+std::optional<glm::dvec3>
+parseArrayValueDVec3(const rapidjson::Document& document, const char* name);
+
+int32_t createBufferInGltf(CesiumGltf::Model& gltf);
+int32_t
+createBufferInGltf(CesiumGltf::Model& gltf, std::vector<std::byte>&& buffer);
```
j9liu, line 89: These are named the same function but do slightly different things; can they be named differently for clarity?
```suggestion
int32_t createEmptyBufferInGltf(CesiumGltf::Model& gltf);

int32_t
createBufferInGltf(CesiumGltf::Model& gltf, std::vector<std::byte>&& buffer);
```

Cesium3DTilesContent/include/Cesium3DTilesContent/LegacyUtilities.h
```diff
@@ -0,0 +1,133 @@
...
+int32_t createAccessorInGltf(
+    CesiumGltf::Model& gltf,
+    const int32_t bufferViewId,
+    const int32_t componentType,
+    const int64_t count,
+    const std::string type);
+
+void applyRTC(CesiumGltf::Model& gltf, const glm::dvec3& rtc);
+
+template <typename GLMType, typename GLTFType>
```
j9liu, line 106: The capitalization rules I previously referenced would also apply here (and to the other abbreviations in this file):

```suggestion
template <typename GlmType, typename GltfType>
```

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,738 @@
...
+      return finishEarly();
+    }
+  }
+  if (parsedContent.positionQuantized) {
+    parsedContent.quantizedVolumeOffset =
+        parseArrayValueDVec3(featureTableJson, "QUANTIZED_VOLUME_OFFSET");
+    if (!parsedContent.quantizedVolumeOffset) {
+      subResult.gltfResult.errors.emplaceError(
+          "Error parsing I3DM feature table, No valid "
+          "QUANTIZED_VOLUME_OFFSET property");
```
j9liu, line 291: Just a bit more explicit 

```suggestion
          "Error parsing I3DM feature table, the I3dm uses quantized positions "
          "but has no valid QUANTIZED_VOLUME_OFFSET property");
```

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,738 @@
...
+      return finishEarly();
+    }
+    parsedContent.quantizedVolumeScale =
+        parseArrayValueDVec3(featureTableJson, "QUANTIZED_VOLUME_SCALE");
+    if (!parsedContent.quantizedVolumeScale) {
+      subResult.gltfResult.errors.emplaceError(
+          "Error parsing I3DM feature table, No valid "
+          "QUANTIZED_VOLUME_SCALE property");
```
j9liu, line 299:
```suggestion
          "Error parsing I3DM feature table, the I3dm uses quantized positions "
          "but has no valid QUANTIZED_VOLUME_SCALE property");
```

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,738 @@
...
+    EXT_mesh_gpu_instancing extension for that node. It would be nice to avoid a
+    lot of duplicate data for mesh nodes with the same chain of transforms to
+    the tile root. One way to do this would be to store the accessors in a hash
+    table, hashed by mesh transform.
+  + Add the instance transforms to the glTF buffers, buffer views, and
+    accessors.
+  + Metadata / feature id?
+*/
+
+CesiumAsync::Future<ConvertResult> convertInstancesContent(
```
j9liu, line 218: If you're willing, this function could benefit from the parsing logic being broken up into helper functions. Like in `PntsToGltfConverter`, there's `parsePositionsFromFeatureTableJson` and `parsePositionsFromFeatureTableBinary`, so that the larger function as a whole reads more neatly.

Even though you're writing to a `DecodedInstances`, I still think you can pass it in and do `decodePositionsFromFeatureTableBinary`

Cesium3DTilesContent/include/Cesium3DTilesContent/GltfConverters.h
```diff
@@ -12,6 +14,30 @@
 #include <string_view>
 
 namespace Cesium3DTilesContent {
+
+/**
+ * Data required to make a recursive request to fetch an asset, mostly for the
+ * benefit of I3dm files.
+ */
+struct CESIUM3DTILESCONTENT_API ConverterSubprocessor {
```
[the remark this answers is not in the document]
j9liu: Or, an alternative is to move some of the "process" / "get" logic here. For example, in `LegacyUtilities` the static `get` function could be made a function of this struct instead.

Cesium3DTilesContent/include/Cesium3DTilesContent/LegacyUtilities.h
```diff
@@ -0,0 +1,133 @@
...
+
+struct ByteResult {
+  std::vector<std::byte> bytes;
+  CesiumUtility::ErrorList errorList;
+};
+
+CesiumAsync::Future<ByteResult>
+get(const ConverterSubprocessor& subprocessor, const std::string& relativeUrl);
+
+namespace LegacyUtilities {
```
j9liu, line 32: If I may nitpick... the name `LegacyUtilities` has the possibility of sounding like the "utilities" themselves are legacy, instead of being utility for legacy content.

But I also think it's okay to forgo the "legacy" entirely and just say

```suggestion
namespace GltfConverterUtility {
```

or something similar.

Cesium3DTilesContent/include/Cesium3DTilesContent/LegacyUtilities.h
```diff
@@ -0,0 +1,133 @@
...
+    const int64_t byteStride);
+
+int32_t createAccessorInGltf(
+    CesiumGltf::Model& gltf,
+    const int32_t bufferViewId,
+    const int32_t componentType,
+    const int64_t count,
+    const std::string type);
+
+void applyRTC(CesiumGltf::Model& gltf, const glm::dvec3& rtc);
```
j9liu, line 104: When I first saw this, I was confused why it was needed because we could just inject a `CesiumRtc` extension into the glTF. But looking at the body, I see it's a lot more complicated. So maybe this can be renamed for clarity (and potentially documented with a comment)

```suggestion
void applyRTCToNodeTransforms(CesiumGltf::Model& gltf, const glm::dvec3& rtc);
```

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,738 @@
...
+  }
+  if (!decodedInstances.scales.empty()) {
+    result = scale(result, glm::dvec3(decodedInstances.scales[i]));
+  }
+  return result;
+}
+
+std::vector<glm::dmat4> meshGpuInstances(
+    GltfConverterResult& result,
+    const ExtensionExtMeshGpuInstancing& gpuExt) {
```
j9liu, line 480: This conveys intent more quickly for me?
```suggestion
std::vector<glm::dmat4> getExtMeshGpuInstances(
    const ExtensionExtMeshGpuInstancing& gpuExt,
    GltfConverterResult& result) {
```

j9liu (commented):
@timoore I did a first pass through everything, and so far it's looking good! I definitely want to look more closely at the tile loading changes / actual matrix math, when I have some more uninterrupted time.

I tried to cut down on bikeshedding, but most of my comments ended up being about clarity. 😅 It can get confusing when both I3DM instances and `ext_mesh_gpu_instancing` are operating in the same file. So I apologize in advance for the quantity and nitpickiness, feel free to push back if something is off

CesiumGltf/include/CesiumGltf/AccessorUtility.h
```diff
@@ -308,4 +308,21 @@ struct TexCoordFromAccessor {
   int64_t index;
 };
 
+/**
+ * Type definition for quaternion accessors, as defined for ExtMeshGpuInstancing
+ * and animation samplers.
```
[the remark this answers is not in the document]
timoore: I decided to go with "rotations."

Cesium3DTilesContent/include/Cesium3DTilesContent/GltfConverters.h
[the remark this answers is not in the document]
timoore: I decided to go with your last suggestion: change the name to AssetFetcher and move the get() function into the class.

Cesium3DTilesContent/include/Cesium3DTilesContent/GltfConverters.h
```diff
@@ -12,6 +14,30 @@
...
+  ConverterSubprocessor(
+      const CesiumAsync::AsyncSystem& asyncSystem_,
+      const std::shared_ptr<CesiumAsync::IAssetAccessor>& pAssetAccessor_,
+      const std::string& baseUrl_,
+      const glm::dmat4 tileTransform_,
+      const std::vector<CesiumAsync::IAssetAccessor::THeader>& requestHeaders_)
```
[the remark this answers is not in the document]
timoore: My experience with gcc is that omitting the underscores produces a warning about shadowing, which then causes an error if all warnings are errors.

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,738 @@
...
+#include <optional>
+#include <set>
+
+using namespace CesiumGltf;
+
+namespace Cesium3DTilesContent {
+using namespace LegacyUtilities;
+
+namespace {
+struct InstancesHeader {
```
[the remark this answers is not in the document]
timoore: No, I agree with you

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,738 @@
...
+  std::optional<uint32_t> position;
+  std::optional<uint32_t> positionQuantized;
+  std::optional<uint32_t> normalUp;
+  std::optional<uint32_t> normalRight;
+  std::optional<uint32_t> normalUpOct32p;
+  std::optional<uint32_t> normalRightOct32p;
+  std::optional<uint32_t> scale;
+  std::optional<uint32_t> scaleNonUniform;
+  std::optional<uint32_t> batchId;
+  // batch table format?
```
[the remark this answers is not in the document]
timoore: I don't remember! In any event, batch table decoding is out-of-scope for this task.

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,738 @@
...
+        auto transVec = toGlm<glm::dvec3>(translationAccessor[i]);
+        instances[i] = glm::translate(instances[i], transVec);
+      }
+    } else {
+      return errorOut("invalid accessor for instance translations");
+    }
+  }
+  if (rotations) {
+    auto quatAccessorView = getQuaternionAccessorView(model, rotations);
+    std::visit(
```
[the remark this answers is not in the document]
timoore: Perhaps... I do wonder how much of the i3dm code and what I wrote in LegacyUtilities (a lot of it copied from your pnts converter code) should be used in refactoring the rest of this 3D Tiles 1.0 reader.

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,738 @@
...
+  // conj(q). But here we are doing an inverse rotation of the right vector into
+  // the "up frame."
+  glm::quat temp =
+      conjugate(upRot) * glm::quat(0.0f, right.x, right.y, right.z) * upRot;
+  glm::vec3 innerRight(temp.x, temp.y, temp.z);
+  glm::quat rightRot = rotation(glm::vec3(1.0f, 0.0f, 0.0f), innerRight);
+  return upRot * rightRot;
+}
+
+struct ConvertResult {
```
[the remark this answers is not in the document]
timoore: Good call.

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
timoore: Art is theft. -- Pablo Picasso

Cesium3DTilesContent/include/Cesium3DTilesContent/GltfConverters.h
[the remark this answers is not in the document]
j9liu: Makes sense, you can ignore this suggestion then!

Cesium3DTilesContent/src/LegacyUtilities.cpp
timoore: I personally use "return {};" as a kind of 21st century "return NULL;". It works in a lot of different contexts as a null value: smart pointers, std::optional, even pointers though I do use nullptr for that. I'll gladly follow the Cesium native coding conventions, but {} is concise and, I think, pretty useful.

Cesium3DTilesContent/include/Cesium3DTilesContent/LegacyUtilities.h
timoore: I would have said that it is one function with an optional argument, but it has to be written as two because the optional argument is an rvalue. Actually, I think I see what can be done.

Apropos of nothing, this is one of the many functions I copied from PntsToGltfConverter.cpp. Do you think it would be worthwhile in this PR to refactor PntsToGltfConverter to use the common versions?

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
timoore: I've split the JSON parsing out into a helper function, but I'm going to wait on the rest until I tackle the batch table and metadata.

Cesium3DTilesContent/include/Cesium3DTilesContent/LegacyUtilities.h
timoore: I was thinking of Tiles10Utility.

j9liu:
@timoore haven't forgotten about this PR, I'll plan to take another look tomorrow :pray:

Cesium3DTilesContent/src/LegacyUtilities.cpp
j9liu: That's fair, and I'm not sure if this goes against an actual item in the coding guide. It's a personal preference, but feel free to skip

Cesium3DTilesContent/include/Cesium3DTilesContent/LegacyUtilities.h
j9liu: Yeah, if you're willing to refactor the existing converters, go ahead! :pray:

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,777 @@
...
+      glm::dmat4x4(1.0));
+  result.model->forEachPrimitiveInScene(
+      -1,
+      [&](Model& gltf,
+          Node& node,
+          Mesh&,
+          MeshPrimitive&,
+          const glm::dmat4& transform) {
+        auto [nodeItr, notSeen] = meshNodes.insert(&node);
+        if (!notSeen) {
```
j9liu, line 656: Could this be renamed to a variable without "not" in the name? It is a little confusing to read logic that resolves to "if not not ___"

Maybe something like...
```suggestion
        auto [nodeItr, firstTimeSeen] = meshNodes.insert(&node);
        if (!firstTimeSeen) {
```

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,777 @@
...
+  }
+  return instances;
+}
+
+const size_t rotOffset = sizeof(glm::vec3);
+const size_t scaleOffset = rotOffset + sizeof(glm::quat);
+const size_t totalStride =
+    sizeof(glm::vec3) + sizeof(glm::quat) + sizeof(glm::vec3);
+
+void copyToBuffer(
```
j9liu, line 607: Just to be more explicit

```suggestion
void copyInstanceToBuffer(
```

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,777 @@
...
+    std::byte* bufferLoc) {
+  glm::vec3 fposition(position);
+  std::memcpy(bufferLoc, &fposition, sizeof(fposition));
+  glm::quat frotation(rotation);
+  std::memcpy(bufferLoc + rotOffset, &frotation, sizeof(frotation));
+  glm::vec3 fscale(scale);
+  std::memcpy(bufferLoc + scaleOffset, &fscale, sizeof(fscale));
+}
+
+void copyToBuffer(
```
j9liu, line 620:
```suggestion
void copyInstanceToBuffer(
```

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,777 @@
...
+        instanceBuffer.cesium.data.resize(dataBaseOffset + instanceDataSize);
+        // Transform instance transform into local glTF coordinate system.
+        const auto toTile = upToZ * transform;
+        const auto toTileInv = inverse(toTile);
+        size_t destInstanceIndx = 0;
+        for (unsigned i = 0; i < numInstances; ++i) {
+          auto instMat = composeInstanceTransform(i, decodedInstances);
+          instMat = toTileInv * instMat * toTile;
+          for (const auto& innerMat : existingInstances) {
+            auto finalMat = instMat * innerMat;
```
j9liu, line 694: I think some name tweaks would help with readability:

```suggestion
        const glm::dmat4 toTile = upToZ * transform;
        const glm::dmat4 toTileInv = inverse(toTile);
        size_t destInstanceIndx = 0;
        for (unsigned i = 0; i < numInstances; ++i) {
          glm::dmat4 i3dmTransform = composeInstanceTransform(i, decodedInstances);
          glm::dmat4 i3dmToGltf = toTileInv * i3dmTransform* toTile;
          for (const auto& gltfTransform : existingInstances) {
            glm::dmat4 finalTransform = i3dmToGltf * gltfTransform;
```

Not sure if `i3dmToGltf` is the exact correct name, but something similar would help

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,777 @@
...
+                destInstanceIndx++);
+          }
+        }
+        auto posAccessorId = createAccessorInGltf(
+            gltf,
+            instanceBufferViewId,
+            Accessor::ComponentType::FLOAT,
+            numNewInstances,
+            Accessor::Type::VEC3);
+        auto& posAcessor = gltf.accessors[static_cast<uint32_t>(posAccessorId)];
```
j9liu, line 713:
```suggestion
        auto& posAccessor = gltf.accessors[static_cast<uint32_t>(posAccessorId)];
```

Cesium3DTilesContent/include/Cesium3DTilesContent/LegacyUtilities.h
```diff
@@ -0,0 +1,124 @@
...
+    const int64_t byteStride);
+
+int32_t createAccessorInGltf(
+    CesiumGltf::Model& gltf,
+    const int32_t bufferViewId,
+    const int32_t componentType,
+    const int64_t count,
+    const std::string type);
+
+void applyRtcToNodes(CesiumGltf::Model& gltf, const glm::dvec3& rtc);
```
j9liu, line 95:
```suggestion
/**
 * Applies the given relative-to-center (RTC) translation to the transforms of
 * all nodes in the glTF. This is used for i3dm, where the RTC translation must
 * be applied to the instances in EXT_mesh_gpu_instancing *before* they are
 * affected by other transforms.
 */
void applyRtcToNodes(CesiumGltf::Model& gltf, const glm::dvec3& rtc);
```

j9liu (changes requested):
@timoore Thanks for addressing my previous feedback, and for your patience with the second review.

I haven't verified the math yet in Unreal, but compared with the notes in #458 , the computation looks good. And I assume that Kevin already did some testing. :smile:

CesiumNativeTests/include/CesiumNativeTests/FileAccessor.h
j9liu, line 1: Forgot to ask, do you plan to add unit tests for i3dm loading?

CesiumUtility/test/TestMath.cpp
```diff
@@ -144,3 +145,15 @@ TEST_CASE("Math::mod") {
   CHECK(Math::mod(-1.0, -1.0) == -0.0);
   CHECK(Math::equalsEpsilon(Math::mod(-1.1, -1.0), -0.1, Math::Epsilon15));
 }
+
+TEST_CASE("Math::perpVec") {
```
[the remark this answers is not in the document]
timoore: I added a test and found a bug! :open_mouth:

Cesium3DTilesContent/src/CmptToGltfConverter.cpp
```diff
@@ -88,26 +90,26 @@ GltfConverterResult CmptToGltfConverter::convert(
...
-
-    result.errors.merge(innerTiles[i].errors);
+    return assetFetcher.asyncSystem.createResolvedFuture(std::move(result));
   }
 
-  return result;
+  return assetFetcher.asyncSystem.all(std::move(innerTiles))
+      .thenImmediately([](std::vector<GltfConverterResult>&& innerResults) {
+        if (innerResults.size() == 1) {
+          return innerResults[0];
```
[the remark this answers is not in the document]
timoore: I believe I had the std:move() in there and got a warning about it inhibiting optimization. The code is different than before because innerResults is an rvalue reference and not a local variable. I couldn't tell you if that actually makes a difference :smile:

Cesium3DTilesSelection/src/ImplicitOctreeLoader.cpp
```diff
@@ -130,8 +135,8 @@ CesiumAsync::Future<TileLoadResult> requestTileContent(
               "Received status code {} for tile content {}",
               statusCode,
               tileUrl);
-          return TileLoadResult::createFailedResult(
-              std::move(pCompletedRequest));
+          return asyncSystem.createResolvedFuture(
+              TileLoadResult::createFailedResult(std::move(pCompletedRequest)));
```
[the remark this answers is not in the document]
timoore: I've changed it. the loadTileContent() function for both ImplicitOctreeLoader and ImplicitQuadtreeLoader could use the same treatment, but I'm going to postpone that. Glad you like it though!

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,777 @@
...
+    parsedContent.quantizedVolumeScale =
+        parseArrayValueDVec3(featureTableJson, "QUANTIZED_VOLUME_SCALE");
+    if (!parsedContent.quantizedVolumeScale.has_value()) {
+      errors.emplaceError(
+          "Error parsing I3DM feature table, the I3dm uses quatized positions "
+          "but has no valid QUANTIZED_VOLUME_SCALE property");
+      return {};
+    }
+  }
+  if (auto optENU = getValue<bool>(featureTableJson, "EAST_NORTH_UP")) {
```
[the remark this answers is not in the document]
timoore: I decided to keep the variable definition inside the if statement with its explicit type. I like limiting the scope of variables where possible.

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,777 @@
...
+CesiumAsync::Future<ConvertedI3dm> convertI3dmContent(
+    const gsl::span<const std::byte>& instancesBinary,
+    const I3dmHeader& header,
+    uint32_t headerLength,
+    const CesiumGltfReader::GltfReaderOptions& options,
+    const AssetFetcher& assetFetcher,
+    GltfConverterResult& result) {
+  ConvertedI3dm convertedI3dm;
+  DecodedInstances& decodedInstances = convertedI3dm.decodedInstances;
+  convertedI3dm.gltfResult = result;
```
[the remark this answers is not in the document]
timoore: There was some vestigial code left over from my first efforts at copying PntsToGltfConverter. I've done this in a slightly different way now.

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,777 @@
...
+  std::optional<I3dmContent> parsedJsonResult =
+      parseI3dmJson(featureTableJsonData, convertedI3dm.gltfResult.errors);
+  if (!parsedJsonResult) {
+    finishEarly();
+  }
+  const I3dmContent& parsedContent = *parsedJsonResult;
+  decodedInstances.rtcCenter = parsedContent.rtcCenter;
+  decodedInstances.rotationENU = parsedContent.eastNorthUp;
+
+  auto featureTableBinaryData = instancesBinary.subspan(
```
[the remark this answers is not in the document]
timoore: There are all the other span variables in the function as well. I'm going to plead that `const gsl::span<const std::byte>`  is "painful" to write every time, which is one of the justifications of using `auto` as I understand it.

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,777 @@
...
+      headerLength + header.featureTableJsonByteLength,
+      header.featureTableBinaryByteLength);
+  auto binaryData = featureTableBinaryData.data();
```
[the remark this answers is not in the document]
timoore: OK, but I don't like that kind of "Hungarian" notation very much and find that it adds little information.

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,777 @@
...
+    return finishEarly();
+  }
+  const uint32_t gltfStart = headerLength + header.featureTableJsonByteLength +
+                             header.featureTableBinaryByteLength +
+                             header.batchTableJsonByteLength +
+                             header.batchTableBinaryByteLength;
+  const uint32_t gltfEnd = header.byteLength;
+  auto gltfData = instancesBinary.subspan(gltfStart, gltfEnd - gltfStart);
+  std::optional<CesiumAsync::Future<ByteResult>> assetFuture;
+  auto featureTableJsonData =
```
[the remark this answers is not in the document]
timoore: See my comment above about pleading hardship...

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -0,0 +1,777 @@
...
+  const uint32_t numInstances = parsedContent.instancesLength;
+  decodedInstances.positions.resize(numInstances, glm::vec3(0.0f, 0.0f, 0.0f));
+  if (parsedContent.position.has_value()) {
+    gsl::span<const glm::vec3> rawPositions(
+        reinterpret_cast<const glm::vec3*>(
+            binaryData + *parsedContent.position),
+        numInstances);
+    decodedInstances.positions.assign(rawPositions.begin(), rawPositions.end());
+  } else {
+    gsl::span<const uint16_t[3]> rawQPositions(
```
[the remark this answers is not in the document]
timoore: Not really, other than I was fixated on the values in the i3dm being "raw bytes," which isn't consistent with using glm::vec3 everywhere. I'll take another look at this.

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
timoore: I changed `notSeen` to `inserted`. This matches the semantics of the result of `insert`, and I think the code reader can understand the intent.

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
timoore: I've changed the names and reworked this a bit.

CesiumNativeTests/include/CesiumNativeTests/FileAccessor.h
timoore: Yes, probably a good idea

j9liu:
Thanks for addressing all the feedback @timoore ! Sorry I didn't see this earlier, you're welcome to @ me after you push changes to get my attention :smile:

Merging now!

PR #858, opened 2024-04-12
Remove `attachTo` for schema files that already have correct prefix

lilleyse:
Small follow up to https://github.com/CesiumGS/cesium-native/pull/851. Removed `attachTo` for schema files that now have the correct prefix.


kring:
Thanks @lilleyse!

Issue #813, opened 2024-02-15
Linker error with `CesiumGltfReader`


jaadelgren:
`CesiumGltfReader` makes use of `stb_image` image resizing functionality (specifically, `stbir_resize_uint8`, see [GltfReader.cpp](https://github.com/CesiumGS/cesium-native/blob/09c1eb76b4f3313ac8517b7c9dea70af6b5c8344/CesiumGltfReader/src/GltfReader.cpp#L938), but does not include the implementation by defining `STB_IMAGE_RESIZE_IMPLEMENTATION` before the `stb_image_resize.h` include. Because the implementation is never built, linking to `CesiumGltfReader` will result in an error, assuming clients don't separately build `stb_image_resize` or link against `CesiumGltfContent` (see below).

It looks like `CesiumGltfContent` _does_ define the macro (in [ImageManipulation.cpp](https://github.com/CesiumGS/cesium-native/blob/09c1eb76b4f3313ac8517b7c9dea70af6b5c8344/CesiumGltfContent/src/ImageManipulation.cpp#L7), and therefore does have the necessary `stbir_resize_XX` functions implemented. However, `CesiumGltfReader` does not link against `CesiumGltfContent`.

Of course, there may be duplicate symbols issues if both libraries build the implementations. It's unclear to me whether it would be appropriate for `CesiumGltfReader` to simply link to `CesiumGltfContent`, or if perhaps another, separate library (e.g. `CesiumtImageResize`) is a better route.

appybara13:
I've encountered the same linker error, and sure enough I had to link `CesiumGltfReader` against `CesiumGltfContent` to avoid issues with duplicate symbols.

I'd say a separate library for all stb implementations would make the most sense, to avoid similar problems in the future.

csciguy8:
It sounds like the output libraries from cesium-native might need to be looked at again, but from a usage perspective. 

For example, if `CesiumGltfReader` isn't meant to be used by itself, but also with `CesiumGltfContent`, maybe it shouldn't even be compiled into a separate lib? Could be all the Gltf related logic just goes into a single `CesiumGltf` lib.

This inspired me to write up #855. This relates to the general question of "what are the ways I can use cesium-native?". If cherry picking libs is something we intend users to do, we should have some reference for this, or at least document it.

PR #860, opened 2024-04-14
Improvements to the build process

kring:
* Upgrade KTX-Software to v4.3.2. See https://github.com/CesiumGS/KTX-Software/pull/1.
* Use the `ktx` library instead of `ktx_read`. While cesium-native currently only needs the read functionality, by depending on `ktx_read` cesium-native prevents anything that uses cesium-native from bringing in the write functionality. This is because there's no `ktx_write`. The only options are `ktx_read` which is read-only, and `ktx` which includes both. If you try to link against both `ktx` and `ktx_read`, you'll end up with multiply-defined symbols.
* Move the STB image resize implementation code into CesiumGltfReader (instead of CesiumGltfContent). This fixes a weird situation where applications that only depend on `CesiumGltfReader` would get linker errors.
* Disable compilation of the KTX CLI by setting `KTX_FEATURE_TOOLS` to off. This saves a bit of time, and also stops KTX from bringing in its only implementation of cxxopts.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,11 +203,11 @@ endif()
 install(TARGETS GSL)
 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/extern/GSL/include/gsl TYPE INCLUDE)
 
-# ktx_read's PUBLIC_HEADER has paths relative to its own directory, so the install line below will fail.
+# ktx's PUBLIC_HEADER has paths relative to its own directory, so the install line below will fail.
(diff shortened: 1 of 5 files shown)


lilleyse:
> * Move the STB image resize implementation code into CesiumGltfReader (instead of CesiumGltfContent). This fixes a weird situation where applications that only depend on `CesiumGltfReader` would get linker errors.

Does this fix https://github.com/CesiumGS/cesium-native/issues/813?

lilleyse:
Looks good!

PR #863, opened 2024-04-17
Don't create an invalid GLB when the buffer is too large.

kring:
Because the GLB format includes its own length as a uint32 field inside itself, it's not possible to create a valid GLB file that is 4GiB or larger. Previously GltfWriter would try, and end up producing an invalid GLB; the length fields would end up being the actual length modulo 2^32. With this PR, GLB creation now fails with an error.


lilleyse:
Looks good

PR #864, opened 2024-04-19
Fix parsing of protocol-relative URIs

azrogers:
As noted [on the forums](https://community.cesium.com/t/unable-to-use-bing-maps-labels-only-raster-overlay/31497) and documented in CesiumGS/cesium-unity#455, some Ion imagery assets provide [protocol-relative URLs](https://en.wikipedia.org/wiki/Wikipedia:Protocol-relative_URL) for their content. While this is technically supported by the RFC 3986 standard, our URI parsing library chokes on them. This change makes cesium-native prepend https to these URLs, meaning this imagery can be loaded in Unreal and Unity again. Technically, the protocol should be the same as the URL the content was loaded from - but that's a more complex change, and the only place we've seen this problem crop up is over ion, where all requests would be https anyways.


CesiumUtility/src/Uri.cpp
```diff
@@ -10,20 +10,35 @@
...
 namespace CesiumUtility {
+const char* HTTPS_PREFIX = "https";
+const char* HTTPS_PREFIX_COLON = "https:";
+
+std::string cesiumConformUrl(const std::string& url) {
+  if (url.rfind("//", 0) == 0) {
+    return std::string(HTTPS_PREFIX_COLON).append(url);
+  } else if (url.find("://", 0) == 0) {
+    return std::string(HTTPS_PREFIX).append(url);
+  }
```
csciguy8, line 21: Could probably save some performance here by checking if the char at index 1 is not a '/', just return the url. As it stands, a non-protocol-relative url executes both finds when passed into here.

Come to think of it, you could probably just check the chars at indices 0, 1, 2 directly without doing any finds at all.

azrogers: Very true! This was actually supposed to be `rfind` at index 0, which would make a single comparison with the start of the string, but I accidentally rewrote it to `find` some point while working on it. Though, I've also realized we don't need to check the `://` case, since despite what I remembered, this isn't actually a valid form of a protocol-relative URI.

CesiumUtility/src/Uri.cpp
```diff
@@ -10,20 +10,35 @@
...
+  return url;
+}
+
 std::string Uri::resolve(
     const std::string& base,
     const std::string& relative,
     bool useBaseQuery) {
+  const std::string conformedBase = cesiumConformUrl(base);
```
csciguy8, line 29: I noticed we assume the current protocol is HTTPS in `cesiumConformUrl`.

Probably safe, but another option would be to have the caller pass in the current protocol (maybe as an optional 4th arg?) 

Then we can support HTTP too, avoid any complicated 'what protocol should we default to' logic, and let the caller make its own assumptions.

csciguy8 (commented):
Looks like a good addition, @azrogers. 

Sanity checked these with cesium-unreal-samples, and everything looks fine.

Left a few code suggestions for you, let me know what you think!

csciguy8 (approved):
Looks great @azrogers !

Will merge once CI passes

PR #868, opened 2024-04-24
Handle GLB BIN padding bytes

javagl:
In a binary glTF, the first `buffer` refers to the BIN chunk data. According to the [glTF specification](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#glb-stored-buffer) ... 

> The byte length of the BIN chunk MAY be up to 3 bytes bigger than JSON-defined buffer.byteLength value to satisfy GLB padding requirements.

This was checked a bit too strictly in [the `GltfReader`](https://github.com/CesiumGS/cesium-native/blob/a739037f9eb133c44bf5e5f54860af7cd13c6c20/CesiumGltfReader/src/GltfReader.cpp#L210): It reported any BIN chunk that was _more_ than 3 bytes larger than the buffer as an ERROR, and did not return a proper model.

This PR changes the behavior to

- Report an ERROR when the buffer size is _larger_ than the BIN chunk size
- Report a WARNING when the buffer size is more than 3 bytes smaller than the BIN chunk size, but still return a valid model
- Report nothing otherwise (because the buffer being up to 3 bytes smaller than the BIN is valid according to the specification)

A test case has been added with a GLB that contains a `buffer` with a length of 44 bytes and a BIN chunk with a length of 48. This could not be loaded before this PR. Now it (only) causes a warning

> The size of the first buffer in the JSON chunk is 44, which is more than 3 bytes smaller than the size of the GLB binary chunk (48)

but still returns a valid model. 

---

A bit more context:

When GLB data is part of a B3DM, then there are stricter alignment requirements. And until [a recent clarification](https://github.com/CesiumGS/3d-tiles/pull/759), the specification required the GLB to _end_ at an 8-byte boundary, even though this requirement does not apply to _the GLB itself_, but to _the tile data_. The previous wording had caused several tools to just append padding bytes to the GLB BIN chunk, leading to many B3DMs containing GLB that caused the validation message

> `BUFFER_GLB_CHUNK_TOO_BIG`: "GLB-stored BIN chunk contains 4 extra padding byte(s)."

and that could _not_ be read by `cesium-native` due to the strict size check.


CesiumGltfReader/test/data/TriangleWithPaddingInGlbBin/README.md
```diff
@@ -0,0 +1,6 @@
+A a GLB of the 'Triangle' glTF sample model (CC0), in a form that
```
lilleyse, line 1:
```suggestion
A GLB of the 'Triangle' glTF sample model (CC0), in a form that 
```

CesiumGltfReader/test/TestGltfReader.cpp
```diff
@@ -246,6 +246,15 @@ TEST_CASE("Read TriangleWithoutIndices") {
   CHECK(position[2] == glm::vec3(0.0, 1.0, 0.0));
 }
 
+TEST_CASE("Read TriangleWithPaddingInGlbBin") {
+  std::filesystem::path glbFile = CesiumGltfReader_TEST_DATA_DIR;
+  glbFile /= "TriangleWithPaddingInGlbBin/TriangleWithPaddingInGlbBin.glb";
+  std::vector<std::byte> data = readFile(glbFile);
+  GltfReader reader;
+  GltfReaderResult result = reader.readGltf(data);
+  REQUIRE(result.model);
```
lilleyse, line 255: Should this test check for the warning?

lilleyse:
@javagl can you update `CHANGES.md`? Then this should be good to merge.

lilleyse:
Thanks @javagl

PR #869, opened 2024-04-27
Fix cmake issue with KTX and iOS

csciguy8:
Showed up in cesium-unreal as this cmake error...

`CMake Error at cesium-native/CMakeLists.txt:210 (install): install TARGETS given no FRAMEWORK DESTINATION for static library FRAMEWORK target "ktx".`
([link to CI run](https://github.com/CesiumGS/cesium-unreal/actions/runs/8838079866))

Setting FRAMEWORK DESTINATION explicitly seems to fix it.

Also remove previous line that set `PROPERTIES PUBLIC_HEADER` to fix the install. This doesn't seem to be needed anymore as the cmake generation works without it

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,7 +203,9 @@ endif()
 install(TARGETS GSL)
 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/extern/GSL/include/gsl TYPE INCLUDE)
 
-install(TARGETS ktx)
+install(TARGETS ktx
+		FRAMEWORK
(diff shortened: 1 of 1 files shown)


j9liu:
Thanks so much @csciguy8. I can confirm that removing the `PROPERTIES PUBLIC_HEADER` check still results in valid compilation in our runtimes, but I'll tag @kring for a gut check since he was the one who added support for it. (I imagine he put that in as necessary for his own environment.)

j9liu:
Thanks @csciguy8 :pray: Happy to merge once CI completes.

PR #870, opened 2024-04-29
Add `getComponentCountFromPropertyType`

lilleyse:
Added `getComponentCountFromPropertyType`. It's a bit similar to `getDimensionsFromPropertyType`, but more convenient in certain situations.


j9liu:
Thanks @lilleyse !

PR #873, opened 2024-05-03
Preserve alignment in GltfUtilities::compactBuffer.

### LINKED ISSUE #862: compactBuffer does not preserve glTF alignment requirements
As explained in detail by @javagl here:
https://github.com/CesiumGS/cesium-native/pull/857#issuecomment-2059773267

kring:
Fixes #862 

This is pretty much the simplest possible approach. When deleting bytes from the buffer, it only deletes multiples of 8 bytes.


lilleyse:
Looks good

PR #874, opened 2024-05-03
Add ability to force assertions

csciguy8:
This adds the ability for cesium-native asserts to execute in configurations where they are usually turned off.

A good example is with cesium-unreal. Unreal Engine lets their plugins run in Debug or Release, but the engine is built in Release (with NDEBUG present). Assertions get compiled out either way. In this situation, we can use the new `CESIUM_FORCE_ASSERTIONS` define with our debug build, to see debug assertions like we expect. This addresses [cesium-unreal #1366](https://github.com/CesiumGS/cesium-unreal/issues/1366).

To use, the native app needs to build cesium-native with the `CESIUM_FORCE_ASSERTIONS` preprocessor present. If defining this for a configuration that would normally have assertions anyway, it is ignored.

A side benefit is that we can now force assertions in Release builds. This could be useful in some situations, like when trying to diagnose bugs that happen in Release only.


CesiumUtility/include/CesiumUtility/Assert.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+//
+// Define our own assertion, so that users can forcebly turn them on if needed
```
csciguy8, line 4: This is a good place to start a review

kring:
This looks good to me @csciguy8. The only small reservation I have is the name `ASSERT`. Macros can't be namespaced, so we're basically claiming this name globally. There's significant risk of a conflict. So even though it's more verbose, we should should probably namespace the name itself a bit. `CESIUM_ASSERT`, perhaps?

There's also a slightly different approach used by another project internal to Cesium that might be worth considering. I'll send you a link to that separately.

csciguy8:
> So even though it's more verbose, we should should probably namespace the name itself a bit. `CESIUM_ASSERT`, perhaps?

Sounds reasonable. Namespace conflicts looks to be something that our external deps also consider...

```
#    define FMT_ASSERT(condition, message)              // spdlog
#  define Assert(cond,msg) {if(!(cond)) z_error(msg);} // zlib
#  define LIBASYNC_ASSERT(pred, except, message)  //asyncplusplus
#define DRACO_DCHECK(x) (assert(x));                       // draco
```

We could do something more sophisticated, like make a new assert-like function in the `Cesium` namespace, but I'm leaning toward a renaming.  It's simple and fits the previous patterns too.

So what to rename it too? `CESIUM_ASSERT` looks fine, but a little long, like you said.

`CASSERT`? That looks too much like we're wrapping a compile time assert
`CN_ASSERT`? As in cesium-native assert, hmm, we don't really have this convention anywhere
`CESIUM_CHECK`? Saves just one character, nah

`CESIUM_ASSERT` it is!

tools/generate-classes/generate.js
```diff
@@ -336,7 +336,7 @@ function generate(options, schema, writers) {
         }
 
         CesiumJsonReader::IJsonHandler* ${name}JsonHandler::readObjectKey(const std::string_view& str) {
-          assert(this->_pObject);
+          ASSERT(this->_pObject);
```
kring, line 339:
```suggestion
          CESIUM_ASSERT(this->_pObject);
```

kring (commented):
Just noticed one small problem, so I'll fix it myself and merge this. Thanks @csciguy8!

PR #875, opened 2024-05-06
Various glTF fixes, mostly meshopt related

kring:
`EXT_meshopt_compression` is fun because it is, as far as I'm aware, the only thing other than a regular `bufferView` that can reference a `buffer`. This wasn't being accounted for in a lot of the glTF manipulation code.

From the changelog:

- Added support for the following glTF extensions to `Model::merge`. Previously these extensions could end up broken after merging.
  -  ...
  - `EXT_meshopt_compression`
- `GltfUtilities::collapseToSingleBuffer` now works correctly even if some of the buffers in the model have a `uri` property and the data at that URI has not yet been loaded. Such buffers are left unmodified.
- `GltfUtilities::collapseToSingleBuffer` now works correctly with bufferViews that have the `EXT_meshopt_compression` extension.
- `GltfUtilities::compactBuffer` now accounts for bufferViews with the `EXT_meshopt_compression` when determining unused buffer ranges.
- When `GltfReader` decodes buffers with data URLs, and the size of the data in the URL does not match the buffer's `byteLength`, the `byteLength` is now updated and a warning is raised. Previously, the mismatch was ignored and would cause problems later when trying to use these buffers.
- `EXT_meshopt_compression` and `KHR_mesh_quantization` are now removed from `extensionsUsed` and `extensionsRequired` after they are decoded by `GltfReader`.

Draft because I need to write tests for a lot of this, though I have empirical evidence in another project that these changes are necessary and that they are working well.

### DIFF
--- a/CesiumGltfContent/src/GltfUtilities.cpp
+++ b/CesiumGltfContent/src/GltfUtilities.cpp
@@ -2,6 +2,7 @@
 #include <CesiumGeometry/Transforms.h>
 #include <CesiumGeospatial/BoundingRegionBuilder.h>
 #include <CesiumGltf/AccessorView.h>
+#include <CesiumGltf/ExtensionBufferViewExtMeshoptCompression.h>
 #include <CesiumGltf/ExtensionCesiumPrimitiveOutline.h>
 #include <CesiumGltf/ExtensionCesiumRTC.h>
 #include <CesiumGltf/ExtensionCesiumTileEdges.h>
@@ -25,6 +26,10 @@
 using namespace CesiumGltf;
 
 namespace CesiumGltfContent {
+
+namespace {
+std::vector<int32_t> getIndexMap(const std::vector<bool>& usedIndices);
+}
 
 /*static*/ std::optional<glm::dmat4x4>
 GltfUtilities::getNodeTransform(const CesiumGltf::Node& node) {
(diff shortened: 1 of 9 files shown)


CesiumGltfContent/src/GltfUtilities.cpp
```diff
@@ -269,13 +274,55 @@ GltfUtilities::parseGltfCopyright(const CesiumGltf::Model& gltf) {
...
+  keepBuffer[0] = true;
+
   for (size_t i = 1; i < gltf.buffers.size(); ++i) {
     Buffer& sourceBuffer = gltf.buffers[i];
+
+    // Leave intact any buffers that have a URI and no data.
+    if (sourceBuffer.uri && sourceBuffer.cesium.data.empty()) {
+      keepBuffer[i] = true;
+      continue;
+    }
```
lilleyse, line 287: Meshopt fallback buffers should also be left intact:

https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Vendor/EXT_meshopt_compression/README.md#fallback-buffers

kring:
I've addressed @lilleyse's comment and added tests, so this is now ready to review and merge.

lilleyse:
The macOS runners are taking a while to launch. But everything else passed and the code looks good to me, so I'm going to merge.

Issue #739, opened 2023-10-06
Memory leak when using external tilesets.


ZackOfAllTrades:
While working with the Google dataset I've been running out of memory. Looking at the call stack 
![image](https://github.com/CesiumGS/cesium-native/assets/13579475/f8b69ebd-27e6-4686-a8f2-744d549e438a)
It is clear it has to do with external tilesets and indeed if the content is external it is not unloaded as part of the _unloadCachedTiles update.
![image](https://github.com/CesiumGS/cesium-native/assets/13579475/c040222d-b5d6-4295-8e88-42bdf88dd95e)

In practice even when I set maximumCachedBytes to 0 the _loadedTiles will continue to grow indefinitely 
![image](https://github.com/CesiumGS/cesium-native/assets/13579475/e939b32e-81d7-4d6f-a778-ce060d414bad)

I don't have enough context on how external tilesets are used so I could be way off but it seems like there needs to be some kind of tree-trimming functionality to unload parts of the explicit json tile tree that are no longer in use.

csciguy8:
Can you provide a particular location you were viewing? Also, how long of a session were you in before it crashed?

Ex. Flying the camera around Manhattan for 10 mins straight.

Just trying to establish an easy scenario, so it can be reproduced quickly.

ZackOfAllTrades:
Hi thanks for the response, the location does not matter and on my pc I too can almost endlessly zoom around but the memory does increase. I've only crashed on hololens2 where I have very little working memory.

I'm sure it has to do with never evicting the explicit external tileset json files, these can be quite large. 
![image](https://github.com/CesiumGS/cesium-native/assets/13579475/b9dce407-6357-4eef-a766-b05beaacd899)
![image](https://github.com/CesiumGS/cesium-native/assets/13579475/2b910ee4-5492-4393-b16d-d79bb9321b9a)

kring:
Indeed the external tileset data isn't unloaded. Generally that's not a major problem, but in the case of Google Photorealistic 3D Tiles it adds up pretty quickly. A couple of minutes flying around can easily end up with a gigabyte of Tile instances in memory.

It's not easy to unload this data, because we need to make sure it's not being used when it's unloaded. For example, one of the Tile instances in the external tileset could be in the process of being loaded in a worker thread at the moment we decide the root Tile of the external tileset can be unloaded. If we actually went through with the unloading, we'd end up deleting a Tile in one thread while it's in use in another.

[...]

PR #876, opened 2024-05-06
Unload content of external tilesets to save memory usage

azrogers:
As described in #739, external tilesets are currently never unloaded by cesium-native. This results in memory accumulating as tiles are loaded, eventually causing the device to run out of memory. This change allows tiles loaded from external sources to be unloaded when they're no longer used, as long as all of their child tiles are also ready to unload.


Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -1461,7 +1463,30 @@ void Tileset::_processMainThreadLoadQueue() {
...
+    this->_externalTilesPendingClear.remove(*pPendingExternalTile);
+    // We need to remove children recursively, as children of this tile might
+    // also be in the _externalTilesPendingClear list
+    this->_unloadPendingChildren(*pPendingExternalTile);
+    pPendingExternalTile->setState(TileLoadState::Unloaded);
+    pPendingExternalTile->clearChildren();
+  }
+
+  // Clear list of pending external tiles
+  this->_externalTilesPendingClear = Tile::LoadedLinkedList();
```
kring, line 1488: This won't do what you want. In fact, we should probably disable the assignment operator on this class. Take a look at how this LoadedLinkedList works. It's an "intrusive" linked list, where the links between nodes are actually stored in the data (Tiles) themselves. This means that a tile can only be in one list at a time. And clearing the tiles from a list requires traversing the entire list and setting the previous and next pointers in each tile to nullptr.

Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -1050,46 +1075,50 @@ bool TilesetContentManager::unloadTileContent(Tile& tile) {
...
-      // tile's content from another thread via lambda capture. We can't unload
-      // it right now. So mark the tile as in the process of unloading and stop
-      // here.
-      tile.setState(TileLoadState::Unloading);
-      return false;
+  if (isReadyToUnload) {
+    // Make sure we unload all children
+    gsl::span<Cesium3DTilesSelection::Tile> children = tile.getChildren();
+    for (Tile& child : children) {
+      this->unloadTileContent(child);
```
kring, line 1098: This raises some performance red flags, though I don't know exactly how to tell you to fix it. When we decide to unload a tile, we'll traverse _all_ of its descendants in `isReadyToUnload` to see if it can unloaded. And if it can, we'll then unload its children (here), which will involve calling `isReadyToUnload` _again_ on each child, which will again traverse the entire tree. A tile subtree can have thousands of tiles, and this is an `n^2` operation in the number of tiles, so 😱.

Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -1050,46 +1075,50 @@ bool TilesetContentManager::unloadTileContent(Tile& tile) {
...
   // If we make it this far, the tile's content will be fully unloaded.
   notifyTileUnloading(&tile);
   content.setContentKind(TileUnknownContent{});
-  tile.setState(TileLoadState::Unloaded);
+  if (isExternalContent) {
+    // We don't want to set external tilesets as Unloaded quite yet, because
+    // then they might get reloaded before we've had a chance to clear their
+    // children and cause an error. They'll get their children cleared and their
+    // state set to Unloaded before next clean up
+    tile.setState(TileLoadState::Done);
```
kring, line 1110: I don't think this is safe. Setting it to `Done` will give the selection algorithm free reign to render this tile. Which will leads to holes. Perhaps `Unloading`?

Cesium3DTilesSelection/src/TilesetContentManager.h
```diff
@@ -65,6 +68,7 @@ class TilesetContentManager
   void updateTileContent(Tile& tile, const TilesetOptions& tilesetOptions);
 
   bool unloadTileContent(Tile& tile);
+  bool handleUpsampledTileChildren(Tile& tile);
```
kring, line 71: Please use more descriptive names. "Handle" how?

Cesium3DTilesSelection/src/TilesetContentManager.h
```diff
@@ -145,6 +149,7 @@ class TilesetContentManager
   int32_t _tileLoadsInProgress;
   int32_t _loadedTilesCount;
   int64_t _tilesDataUsed;
+  Tile::LoadedLinkedList& _loadedTiles;
```
kring, line 152: This is dangerous. The `TilesetContentManager` can and sometimes does outlive the `Tileset`. So by having this reference here, we need to be careful that we'll never try to use it after the `Tileset` has been destroyed.

kring (commented):
I took a look and have a few comments below. I noticed that in this branch, tile content never seems to be getting unloaded at all. The immediate cause is that in `Tileset::_unloadCachedTiles`, the head of the `_loadedTiles` linked list is nullptr. The tail isn't nullptr, however, which is an invalid and nonsensical state. So the linked list is getting corrupted, possibly by the attempt to clear the `_externalTilesPendingClear` list by assignment (see below).

azrogers:
After consulting with the rest of the team, we're going to put this PR on hold for now as it's taking an unexpectedly large amount of effort to get working. In its current state the code works fine with @javagl's example above, and does load and unload Google P3DT when tested with CesiumGS/cesium-unreal#1415, but currently only loads a limited amount of tiles closest to the camera. In theory, any tile that's being visited as part of building the list of tiles to render should never be unloaded, but for some reason this only applies to the closest tiles and not to ones farther away. This likely has something to do with `unloadTileContent` and `loadTileContent` - for whatever reason, tiles outside of the immediate circle around the camera remain "Unloaded". Further investigation is needed, if anyone reading this wants to take a crack at it!

javagl:
> Further investigation is needed, if anyone reading this wants to take a crack at it!

I haven't beem following the details and evolution of  `cesium-native`, and wouldn't claim to be able to even _guess_ the reason for certain issues. But if it is possible to describe "a minimal tileset that is sufficient for testing (and debugging) this", I could try to create such a test tileset.

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -1509,6 +1541,15 @@ void Tileset::_unloadCachedTiles(double timeBudget) noexcept {
...
   this->_loadedTiles.insertAtTail(tile);
+  // Don't clear the children of this tile next frame
+  this->_externalTilesPendingClear.remove(&tile);
+  if (tile.getState() == TileLoadState::Unloaded &&
+      !tile.getChildren().empty()) {
+    // We were going to clear this tile's children, but it's still in use, so we
+    // should restore it to Done instead.
+    tile.setState(TileLoadState::Done);
+    tile.setContentShouldContinueUpdating(false);
+  }
```
kring, line 1552: This is the cause of the missing tiles. `tile.getState() == TileLoadState::Unloaded && !tile.getChildren().empty()` will be true for a very large number of tiles before the first time they're loaded. Setting them to `Done` will ensure they never get loaded. Do you need to check that it's also in `_externalTilesPendingClear` maybe?

kring:
What I've called `doNotUnloadCount` above is really more like `doNotUnloadSubtreeCount`. It may be safe to unload renderable content when this count is non-zero (subject to the rules around this that we already have), but it's not safe to unload an external tileset when this count is greater than 0.

javagl:
I'm watching this, from a huge distance (from on an island in Indonesia, so to speak), and see quite some "reference counting" going on in various forms. And I'm wondering how much code (and how many bugs) could be avoided by just saying: Let's handle each `Tile` as a `std::shared_ptr`. Yes, this has to do this ~"expensive reference counting", and "... performance...", but ... apparently, reference counting **is** necessary, and doing it manually (with different places that increase/decrease the 'reference count', and different _threads_ being involved) is error prone. 

(I'm just wondering. I know, it won't happen. So 🤞 that you can sort this out in another way. But don't forget to add comments accordingly, e.g. in `ViewUpdateResult`, saying things like: "Thou shalt not store these `Tile*` pointers, because the ownership and lifetime of these is not specified")

kring:
Even putting aside costs, using a std::shared_ptr wouldn't work here. We can't destroy individual tiles, we only want to destroy entire subtrees loaded from external tilesets. So it is reference counting, but even though Tile stores the count, it's not really the individual Tile instances we're controlling the lifetime of. With std::shared_ptr, the thing counted and the thing whose lifetime is controlled are always one and the same.

timoore:
> Even putting aside costs, using a std::shared_ptr wouldn't work here. We can't destroy individual tiles, we only want to destroy entire subtrees loaded from external tilesets. So it is reference counting, but even though Tile stores the count, it's not really the individual Tile instances we're controlling the lifetime of. With std::shared_ptr, the thing counted and the thing whose lifetime is controlled are always one and the same.

I don't understand external tilesets well, but if the problem is that the lifetime of the external tileset structure is different than the individual tiles in it, then it seems that a shared_ptr would work. You don't really care about the precise reference count, just whether or not any references to a tile are held.

As an old Lisp guy, I get a bit jaded about performance and shared_ptr, putting complaints in the same category as complaints about garbage collection performance. If you need to track object lifetimes, then you need to pay for it somewhere. Was shared_ptr deemed to be too expensive at some point in Cesium Native's history? If so, I'd look at nearby solutions like std::make_shared and intrusive pointers before tossing the whole idea.

javagl:
Some of this may be too unrelated to this PR (and a too large topic in general) to be discussed in this PR. I just threw it in here, because of the difficulties of the whole "lifetime and ownership" question that _seems_ to be at the core here, and the attempt to solve it with _manual_ reference counting is bound to involve intricate handling of the reference count, and will require deeeep knowledge about _what_ is happening _where_ on _which_ thread. 

(An aside: I also doubt that replacing manual reference counting with a `shared_ptr` would have a prohibitively large performance impact. In fact, I'm pretty sure that it would not have any measurable impact at all. But there's no point in arguing about this without benchmarking. So right now, I cannot contribute to a profound discussion. For that, I'd have to start with basic things, like ... figuring out what the difference between a `ReferenceCounted` and a shared pointer is to begin with...)

A few abstraction levels higher: I think that the "lifetime and ownership" question that is hidden in the point about 'individual tiles vs. subtrees' is crucial. But I assume that the state space, state transitions, lifetimes, and ownerships will raise enough questions to be documented incrementally. One overly specific example is that I assume that `ViewUpdateResult` _will_ eventually contain a hint saying something like "The `Tile*` pointers in the `tilesToRenderThisFrame` may become invalid at time X under condition Y. Clients may not store these pointers"...

(The question about "tile-vs-subtree" may, at the bottom, involve seemingly trivial aspects like the `std::vector<Tile> _children;` not being a `std::vector<std::shared_ptr<Tile>> _children;`, but if the question of "Where to use which kind of pointer" is supposed to be sorted out more holistically, it should be done in a dedicated issue...)

kring:
> I don't understand external tilesets well, but if the problem is that the lifetime of the external tileset structure is different than the individual tiles in it, then it seems that a shared_ptr would work.

There is no external tileset structure. Well, there's TileExternalContent, but there are no Tiles in it (just metadata).

It would be interesting to consider moving to a model where there's an explicit representation of an external tileset, and it owns all of the Tile instances within it. Perhaps `TilesetJsonLoader` is a start toward that.

But tackling that as part of this PR definitely risks turning a relatively focused (if slightly tricky and possibly error-prone) task into a major refactoring effort.

> Was shared_ptr deemed to be too expensive at some point in Cesium Native's history?

No, we use shared_ptr. But not for tiles. Not so much because of the cost of shared_ptr, but because of the cost of heap allocating every Tile individually.

I'm aware of `std::make_shared` and we always use it. It cuts the heap allocations in half, but that's still a huge number of heap allocations in the case of Tiles.

We also have `CesiumUtility::IntrusivePointer` for things that are always reference counted, don't need weak_ptr, and benefit from a little bit cheaper reference counting.

kring:
> the attempt to solve it with manual reference counting is bound to involve intricate handling of the reference count, and will require deeeep knowledge about what is happening where on which thread.

cesium-native's rule is that Tile instances can only ever be accessed from the main thread (or the thread that created them if they haven't been passed to the main thread yet, i.e., during loading). So there's no tricky threading issue here, at least.

A non-owning smart pointer class might be a nice elegant way to solve the "manual" part of the reference counting, though. On construction with a Tile*, it increments `doNotUnloadSubtreeCount` for the tile and all its ancestors. On destruction, it decrements the same. Then anything that holds a Tile pointer outside the normal flow and needs to make sure the TIle doesn't disappear (such as TileLoadInput) just has to use this smart pointer class.

> I also doubt that replacing manual reference counting with a shared_ptr would have a prohibitively large performance impact. In fact, I'm pretty sure that it would not have any measurable impact at all. But there's no point in arguing about this without benchmarking.

Yeah and I agree with you on this point anyway. As I mentioned above, it's not the reference counting part that's costly, it's the memory allocation. I don't think I need to do a benchmark to be convinced that heap-allocating every Tile would be a mistake, though. In fact, I bet we'd see a small but measurable performance improvement if we could avoid heap-allocating (via a std::vector) each block of child Tiles, as we're currently doing. Avoiding unnecessary heap allocations is just good practice, particularly in code where performance matters. It's even in the [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r5-prefer-scoped-objects-dont-heap-allocate-unnecessarily). I'm not saying we have to obsess over it of course, sometimes dynamic allocation is the best solution.

> I'd have to start with basic things, like ... figuring out what the difference between a ReferenceCounted and a shared pointer is to begin with...)

[ReferenceCounted](https://github.com/CesiumGS/cesium-native/blob/main/CesiumUtility/include/CesiumUtility/ReferenceCounted.h) is an optional, convenience base class that can be inherited from in order to easily enable a particular type for use with `CesiumUtility::IntrusivePointer`.

I could try to explain the tradeoffs of IntrusivePointer versus shared_ptr, but I'll just refer you to Boost's documentation on their intrusive pointer to start (not exactly the same, but same idea and same tradeoffs):
https://www.boost.org/doc/libs/1_85_0/libs/smart_ptr/doc/html/smart_ptr.html#intrusive_ptr

> But I assume that the state space, state transitions, lifetimes, and ownerships will raise enough questions to be documented incrementally.

Yes, no question, we need to be clear about the lifetime rules. We've gotten away with this with Tiles until now because Tiles, once created, were never destroyed, other than when the entire Tileset was destroyed.

javagl:
> cesium-native's rule is that Tile instances can only ever be accessed from the main thread (or the thread that created them if they haven't been passed to the main thread yet, i.e., during loading). So there's no tricky threading issue here, at least.

That _seems_ to be focussed on the usage side (i.e. what is supposed to be visible for clients). One of my concerns about manual reference counting referred to the guts of the implementation. I _think_ that it could easily happen that someone is doing _something_ on _some_ thread and this affects the reference counter. This does _not_ refer to something explicit and obvious like
```
something.thenInWorkertThread([...] {
    someTileThatSomeoneGotFromSomewhere.decreaseReferenceCount();
}
```
that could easily spotted during a review. It rather referred to the "implicit" changes, in scenarios like
- an issue is caused: Some resource or memory is not properly cleaned up
- it has to be fixed: When a certain function is called, then something else has to happen, to "clean up/release" something
  - (This will affect some reference counter, down the call chain, which may not be obvious)
- it is not clear or obvious _which_ thread is calling the function, and/or which thread is allowed to touch the reference counter

With very hazy memories about some details, but roughly looking at your list above, there may be something like
```
Tile::upsample() {
    parent.increaseReferenceCount();
    ....
}
```
<sup>Which thread is calling 'upsample'? You probably know that. You could claim that "this sort of operation is always done by thread X". But it's probably not documented (and even less _enforced_) explicitly.</sup>

Now, there has to be a matching `parent.DEcreaseReferenceCount();` call somewhere. This could be part of some `finishedUpsampling()` method or so. But where it's not clear or obvous which thread is calling _this_ method either.

(All this could probably be summarized with: "See `ReferenceCountedThreadSafe` vs `ReferenceCountedNonThreadSafe`" 😉 )

---

The Boost documentation does not go into details about _reasons_ for using `intrusive_ptr` (but mentions that one should start with `shared_ptr` by default). And the reasons that I found by 5 minutes of websearching always seem to refer to "performance". I know that there are general rules and best practices, and some design decisions on a high level (~"preferring stack over heap") that have to be carefully considered in view of their implications on performance and (IMHO more importantly:) lifetime of objects. 

But as a baseline: I doubt *any* performance claim that is made without a benchmark. And when there is a benchmark, then I doubt that benchmark 😈 . _Quickly_ looking at the first, naive, shallow search results like https://stackoverflow.com/questions/14837829/intrusive-ptr-shared-ptr-performance-tests or https://groups.google.com/g/benchmark-discuss/c/uENrnU4J3AM :
- they are sometimes pretty old (comparing Boost `intrusive_ptr` to Boost `shared_ptr` might have made sense 10 years ago, but now there is `std::shared_ptr`, which may have wildly different performance characteristics...)
- they are varying wildly in terms of the performance/timing claims that they make
- important: they seem to be run _single-threaded_, meaning that they may be "distorted" when it comes to the impact of thread safety on overall performance of a multi-threaded application

(Whether or not `intrusive_ptr` is thread safe to begin with? I don't know. It can probably be derived from https://www.boost.org/doc/libs/1_85_0/libs/smart_ptr/doc/html/smart_ptr.html#intrusive_ref_counter , but I didn't look into the details....)

One reason why I didn't look into all the details: Imagine there is a `Tile` that has to be created, and this can be a `shared_ptr` and be allocated on the heap and require atomic access for reference counting. All this is expensive. But then this Tile is filled with life, by doing a network request and donwloading 5 MB of data that has to be processed and uploaded to the GPU. I'm convinced that even the most naive approach for managing tiles will (in a real application, i.e. outside the scope of a microbenchmark) change the "average FPS over 100 frames" from 42.123 to something like 41.987 or so. 

Proving or disproving this with a benchmark could be difficult: One can not just introduce some `#define USE_APPROACH_X true` somewhere and run both versions. The _structural_ changes all over the codebase (implied by the change from "manual reference counting and intrusive pointers" to "naive, blanket `std::shared_ptr` everywhere") would be so large that the results could hardly be compared objectively. But even in a very narrow sope: If there was a measurable performance difference between `Tile::_children` being a `vector<Tile>` and a `vector<shared_ptr<Tile>>`, then I would at least consider this as "evidence". But I'm not sure whether it's worth trying that out.

kring:
> That seems to be focussed on the usage side (i.e. what is supposed to be visible for clients). One of my concerns about manual reference counting referred to the guts of the implementation. I think that it could easily happen that someone is doing something on some thread and this affects the reference counter.

I'm talking about implementation, too. What you describe can't happen as long as we follow our rule of never passing a `Tile` to a worker thread. That's why we have that rule.

> Which thread is calling 'upsample'?

Upsampling is done in a worker, but it does not require access to a `Tile`.

The glTF, however, is made available to the upsample process, which is ugly and dangerous but necessary.

> The Boost documentation does not go into details about reasons for using intrusive_ptr

It lists three reasons:

> * Some existing frameworks or OSes provide objects with embedded reference counts;
> * The memory footprint of intrusive_ptr is the same as the corresponding raw pointer;
> * intrusive_ptr<T> can be constructed from an arbitrary raw pointer of type T*.

shared_ptr is quite complicated, by necessity. Primarily because of two design decisions: the count being stored separately from the object being controlled, and the availability of weak_ptr. An intrusive pointer is simpler. That makes it smaller and faster.

> One reason why I didn't look into all the details: Imagine there is a Tile that has to be created, and this can be a shared_ptr and be allocated on the heap and require atomic access for reference counting. All this is expensive. But then this Tile is filled with life, by doing a network request and donwloading 5 MB of data that has to be processed and uploaded to the GPU. I'm convinced that even the most naive approach for managing tiles will (in a real application, i.e. outside the scope of a microbenchmark) change the "average FPS over 100 frames" from 42.123 to something like 41.987 or so.

You missed a critical detail. `Tile` instances are created in bulk, sometimes many thousands at a time, and only a small fraction of those, sometime later, may have content loaded for them.

As a completely unsurprising (at least to me) point of reference, at one point I tried to switch the Tile loading code from its current approach, which reads the tile hierarchy from a tileset.json into a RapidJSON Document and then creates Cesium3DTilesSelection::Tile instances for each tile represented in the Document, to a slightly different approach that reads a Cesium3DTiles::Tileset and then creates Cesium3DTilesSelection::Tile instances from those. It took somewhere around twice as long to load the tile hierarchy, adding as much as a full second to the initial load time of a Tile-rich tileset like Melbourne Photogrammetry. Why? The JSON reader for `Cesium3DTiles::Tileset` should be extremely efficient. However, Cesium3DTiles::Tileset uses vectors of Tiles (just like Cesium3DTilesSelection::Tile), which requires a _lot_ of heap allocations. Heap allocations are expensive.

Believe it or not, I'm not some performance zealot trying to make your life hard by imposing questionable rules without evidence. 😀

javagl:
> Believe it or not, I'm not some performance zealot trying to make your life hard by imposing questionable rules without evidence. 😀

I assume that you did tests and comparisons (beyond what you already _know_), and the point about "many tiles being created at once (and sometimes only few of them being filled later)" is certainly a valid one: It sounds like the two factors here are "multiplicative": "Heap-vs-Stack" (with the first one being slower) and "Individual-vs-Bulk" (with the first one being slower). 

I'm lacking some context, but 

> Cesium3DTiles::Tileset uses vectors of Tiles ...

I don't see a `vector<Tile>` there, and if this referred to the `Tile::children`, I'd have to think through why this makes a difference. A guess: Could _collecting_ these `Cesium3DTIles::Tile` instances (in one huge vector) and then _bulk-creating_ the matching `Cesium3DTilesSelection::Tile` instances from them have been a reasonable attempt for alleviating the problem?

And an aside: I assume that this experiment was in the context (or at least related to) the consideration to replace the whole third-party-JSON parsing with the auto-generated `cesium-native` JSON handling. A devil's advocate question could be: If the auto-generated structures do not allow an implementation that is as efficient as manual JSON twiddling, what would have to be _changed_ in order to _allow_ such an efficient implementation?

---

In my naive, convenient Java world, there is no real "stack", and nothing that is _not_ a "shared pointer", so to speak. And that may be a reason for me to (always keep performance in mind, but) lean towards the "safe and simple" approach, Often, time is wasted in unexpected places. For example, for that JSON parsing layer: I've heard rumors about ~20 MB `tileset.json` files, where questions like "SAX-vs-DOM" and maybe even "indented-vs-non-indented" may come into play, and the overarching question would be "Couldn't you split that into '1MB + (19 x 1MB)' tilesets?" ... with the latter being _external_ ones... leading back to the purpose of this PR (sorry for derailing that a bit...).

kring:
> if this referred to the Tile::children

Yep, that's the vector of tiles I'm referring to.

> A guess: Could collecting these Cesium3DTIles::Tile instances (in one huge vector) and then bulk-creating the matching Cesium3DTilesSelection::Tile instances from them have been a reasonable attempt for alleviating the problem?

Yes, that would help significantly. Switching instead to a `std::vector<any_kind_of_pointer<Tile>>` would be a significant step in the wrong direction, even if we also moved to a single vector to hold all the tiles.

In theory it's possible to use a custom allocator to get `std::vector<any_kind_of_pointer<Tile>>` semantics while keeping close to `std::vector<Tile>` performance, but it's tricky. Particularly if the reason you're doing that is so that you can allocate/free individual Tiles whenever you like.

> I assume that this experiment was in the context (or at least related to) the consideration to replace the whole third-party-JSON parsing with the auto-generated cesium-native JSON handling.

Yes that's right.

> If the auto-generated structures do not allow an implementation that is as efficient as manual JSON twiddling, what would have to be changed in order to allow such an efficient implementation?

It's not that "manual JSON twiddling" is faster. It's that RapidJSON Document is obsessively focused on avoiding heap allocations. Basically, it's fast in all the ways the naive implementation is not. And that makes it win _even though_ it requires some manual JSON twiddling.

> In my naive, convenient Java world, there is no real "stack", and nothing that is not a "shared pointer", so to speak.

Interestingly, one of the advantages of having a garbage collector - especially a compacting, generational one - is that heap allocations are _much_ faster.

javagl:
> Interestingly, one of the advantages of having a garbage collector - especially a compacting, generational one - is that heap allocations are much faster.

The other "advantage" is that as a developer, you don't have to give a ... ... care about _why_ this is the case 🙂 
Even though it might be a disadvantage _now_ - I suspect this is related to fragmentation and the bookkeeping that is necessary for non-GC'ed heap allocation. As a "Java fanboi" (looking at my name here...) I did dig _far_ deeper into the details of the language and JVM than most other developers, but eventually, the memory management and GC _has to be_ treated as a "black box" (and its details may change over time). Things like https://softwareengineering.stackexchange.com/a/208676 seem to be reasonable summaries, though...

---

It's surprisingly hard to find concise, reliable information about the performance differences of heap-vs-stack in C++. 

(By "hard" I mean that there doesn't seem to be much in the first 10 search results. By "concise" I mean that there _are_ things like stack overflow answers that talk about this, and _why_ one is faster than the other, but not _how much_. And by "reliable" I mean that https://publicwork.wordpress.com/2019/06/27/stack-allocation-vs-heap-allocation-performance-benchmark/ is easy to find and concise, but too narrowly focused on a too artifical microbenchmark). 

One approach for avoiding the heap costs could be some sort of "pool" - that's probably what you referred to with..

> In theory it's possible to use a custom allocator to get `std::vector<any_kind_of_pointer<Tile>>` semantics while keeping close to `std::vector<Tile>` performance, but it's tricky. Particularly if the reason you're doing that is so that you can allocate/free individual Tiles whenever you like.

And the trickiness obviously is: What can you do _manually_ that is faster than what the `new/delete` operators are doing? 
The first thought would be: Not much, probably. Except for very narrow, special cases of allocation/deallocation patterns. *But...* I could imagine that for `Tile`, we **do** have such a narrow case: Tiles form a hierarchy, and I could imagine that the constraints and patterns that are implied by having a _tree structure_ could allow for specialized/optimized implementations of such a "pool". (Not trivial by any means, and ... even more unrelated to this PR than much of the previous disucssion, but a thought...)

kring:
> And the trickiness obviously is: What can you do manually that is faster than what the new/delete operators are doing?

Yep. An allocator for a single type of object is a bit simpler than a general one, so that helps. It's also possible to do the sort of tricks that RapidJSON does: https://rapidjson.org/md_doc_internals.html#MemoryPoolAllocator
So something like: give each tileset its own pool for allocating Tile instances, and once memory is allocated from that pool, never return it. The pool itself can be freed by unloading the entire tileset. (by tileset I also mean _external_ tileset, so external tileset subtrees can be unloaded without unloading the entire top-level tileset)

Actually now that I'm looking at the stackoverflow link you posted above, it actually kind of says what I just wrote above better than I did.

By the way, you don't have to convince me that good garbage collectors are awesome. I completely agree.

javagl:
From a naive, high-level point, this now sounds like something that could actutally be ... "less non-trivial" ... than many other forms of pooling: Storing the `tile.children` as a `std::vector<Pointer<Tile>>`, where these are just pointing into one huge Tile array that is maintained _in the tileset_. But one would certainly have to think about some details here. For example, conditions for "compacting/defragmenting" this array. Unloading some external tileset might be one of them, but depending on how this "mapping" of "pointers to array entries" is established and maintained, the challenge would then be do do this _without_ causing the usual "<strike>GC</strike> compacing pauses" that can easily ruin _every_ interactive experience...

azrogers:
Closing in favor of #1107.

PR #879, opened 2024-05-16
Add new functions to Uri class

kring:
Adds several new functions to the `Uri` class:

* `unescape`
* `unixPathToUriPath`
* `windowsPathToUriPath`
* `nativePathToUriPath`
* `uriPathToUnixPath`
* `uriPathToWindowsPath`
* `uriPathToNativePath`

### DIFF
--- a/CesiumUtility/include/CesiumUtility/Uri.h
+++ b/CesiumUtility/include/CesiumUtility/Uri.h
@@ -24,7 +24,104 @@ public:
       const std::string& templateUri,
       const std::function<SubstitutionCallbackSignature>& substitutionCallback);
 
+  /**
+   * @brief Escapes a portion of a URI, percent-encoding disallowed characters.
+   *
+   * @param s The string to escape.
+   * @return The escaped string.
+   */
   static std::string escape(const std::string& s);
+
+  /**
+   * @brief Unescapes a portion of a URI, decoding any percent-encoded
+   * characters.
+   *
+   * @param s The string to unescape.
+   * @return The unescaped string.
+   */
+  static std::string unescape(const std::string& s);
+
+  /**
+   * @brief Converts a Unix file system path to a string suitable for use as the
+   * path portion of a URI. Characters that are not allowed in the path portion
+   * of a URI are percent-encoded as necessary.
+   *
+   * If the path is absolute (it starts with a slash), then the URI will start
+   * with a slash as well.
+   *
+   * @param unixPath The file system path.
+   * @return The URI path.
+   */
+  static std::string unixPathToUriPath(const std::string& unixPath);
+
+  /**
+   * @brief Converts a Windows file system path to a string suitable for use as
+   * the path portion of a URI. Characters that are not allowed in the path
+   * portion of a URI are percent-encoded as necessary.
+   *
+   * Either `/` or `\` may be used as a path separator on input, but the output
+   * will contain only `/`.
+   *
+   * If the path is absolute (it starts with a slash or with C:\ or similar),
+   * then the URI will start with a slash well.
+   *
+   * @param windowsPath The file system path.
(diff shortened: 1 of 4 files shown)


CesiumUtility/src/Uri.cpp
```diff
@@ -200,6 +201,131 @@ std::string Uri::escape(const std::string& s) {
...
+    return uriPath;
+  } else {
+    // Truncate at first null character
+    result.resize(std::strlen(result.data()));
+    return result;
+  }
+}
+
+std::string Uri::uriPathToNativePath(const std::string& nativePath) {
+  if constexpr (std::filesystem::path::preferred_separator == '\\') {
```
csciguy8, line 322: If the goal is to detect whether we are on windows or not, can we just use `#ifdef _WIN32`?

CesiumUtility/test/TestUri.cpp
```diff
@@ -96,3 +96,65 @@ TEST_CASE("Uri::resolve") {
           false,
           false) == "http://www.example.com/page/test");
 }
+
+TEST_CASE("Uri::escape") {
+  CHECK(Uri::escape("foo") == "foo");
+  CHECK(Uri::escape("foo/bar") == "foo%2Fbar");
+  CHECK(Uri::escape("🤞") == "%F0%9F%A4%9E");
```
csciguy8, line 103: I had no idea our code could show emojis, very cool! (or is it?)

CesiumUtility/include/CesiumUtility/Uri.h
```diff
@@ -24,8 +24,105 @@ class Uri final {
...
+   * path portion of a URI. Characters that are not allowed in the path portion
+   * of a URI are percent-encoded as necessary.
+   *
+   * If the path is absolute (it starts with a slash), then the URI will start
+   * with a slash as well.
+   *
+   * @param unixPath The file system path.
+   * @return The URI path.
+   */
+  static std::string unixPathToUriPath(const std::string& unixPath);
```
csciguy8, line 55: I noticed these 3 functions are basically the same, other than the file system they target. (same for the next group of 3 that does the reverse).

Might be able to simplify the API a bit by just making the file system an argument...
```
enum class FileSystem {
  Unix,
  Windows,
  AutoDetect
};
static std::string fileSystemPathToUriPath(const std::string& path, FileSystem fileSystem = AutoDetect);
```

csciguy8 (commented):
Thanks @kring !

Looks like nice new functionality and tests. Had a couple comments that are optional if you'd like to act on them or not.

Otherwise, looks good to go.

CesiumUtility/src/Uri.cpp
kring: I had that first, but I think this is better. The main difference between the unix and windows functions in UriParser is their handling of backslashes, so I think choosing based on the platform path separator is slightly better.

CesiumUtility/test/TestUri.cpp
kring: Isn't UTF-8 great?

CesiumUtility/include/CesiumUtility/Uri.h
kring: Hmm it's an interesting idea, but it would add a tiny runtime cost and I'm not sure what we'd really be gaining.

csciguy8: > not sure what we'd really be gaining

Not all that much. Just one function declared in the .h, rather than three (they could just live in the .cpp).

CesiumUtility/src/Uri.cpp
csciguy8: Ok and you've well documented the behavior, so all good.

PR #882, opened 2024-05-22
Faster `collapseToSingleBuffer`

lilleyse:
I'm working with a model that has tens of thousands of buffers and buffer views and `collapseToSingleBuffer` started to become a bottleneck. I changed the code so that there's only a single pass over buffer views which dramatically speeds things up.

For a model with 34680 buffers and 34680 buffer views it goes from 49796 ms to 47 ms.

### DIFF
--- a/CesiumGltfContent/src/GltfUtilities.cpp
+++ b/CesiumGltfContent/src/GltfUtilities.cpp
@@ -278,6 +278,8 @@ GltfUtilities::parseGltfCopyright(const CesiumGltf::Model& gltf) {
   std::vector<bool> keepBuffer(gltf.buffers.size(), false);
   keepBuffer[0] = true;
 
+  std::vector<int64_t> bufferStarts(gltf.buffers.size(), 0);
+
   for (size_t i = 1; i < gltf.buffers.size(); ++i) {
     Buffer& sourceBuffer = gltf.buffers[i];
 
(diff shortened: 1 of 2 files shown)


CesiumGltfContent/src/GltfUtilities.cpp
```diff
@@ -292,17 +294,45 @@ GltfUtilities::parseGltfCopyright(const CesiumGltf::Model& gltf) {
...
+        destinationBuffer.cesium.data.data() + start,
+        sourceBuffer.cesium.data.data(),
+        sourceBuffer.cesium.data.size());
+
+    sourceBuffer.byteLength = 0;
+    sourceBuffer.cesium.data.clear();
+    sourceBuffer.cesium.data.shrink_to_fit();
+
+    destinationBuffer.byteLength =
+        int64_t(destinationBuffer.cesium.data.size());
```
lilleyse, line 323: This section is copied directly from `moveBufferContent`. I couldn't think of a clean way to continue to use `moveBufferContent` while still being performant. Not sure how much use `moveBufferContent` is anymore...

kring:
Good stuff, thanks @lilleyse! I just factored out the duplicate code in `moveBufferContent` and `collapseToSingleBuffer` into a local helper function (`moveBufferContentWithoutRenumbering`). And updated CHANGES.md.

PR #884, opened 2024-05-22
Added `forEachNodeInScene`

lilleyse:
I found myself needing a function like `forEachPrimitiveInScene` but without drilling into meshes / primitives and just traversing over nodes.

### DIFF
--- a/CesiumGltf/src/Model.cpp
+++ b/CesiumGltf/src/Model.cpp
@@ -802,21 +802,21 @@ void generateSmoothNormals(
 
   const size_t normalBufferId = gltf.buffers.size();
   Buffer& normalBuffer = gltf.buffers.emplace_back();
-  normalBuffer.byteLength = int64_t(normalBufferSize);
+  normalBuffer.byteLength = static_cast<int64_t>(normalBufferSize);
   normalBuffer.cesium.data = std::move(normalByteBuffer);
 
   const size_t normalBufferViewId = gltf.bufferViews.size();
   BufferView& normalBufferView = gltf.bufferViews.emplace_back();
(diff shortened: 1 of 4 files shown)


CesiumGltf/src/Model.cpp
```diff
@@ -402,76 +402,84 @@ void forEachPrimitiveInMeshObject(
...
+      return std::nullopt;
     }
 
-    nodeTransform =
-        nodeTransform * translation * glm::dmat4(rotationQuat) * scale;
+    return translation * glm::dmat4x4(rotationQuat) * scale;
   }
 
+  return glm::dmat4x4(1.0);
+}
```
lilleyse, line 468: This looks like a lot of changes, but really all I did was duplicate `GltfUtilities::getNodeTransform` into here. The previous code did the same thing but with some small differences.

Also, this code used to be inside `forEachPrimitiveInNodeObject` but now it's factored out into a helper function that `forEachNode` can also use.

kring:
Thanks @lilleyse! Some of this code is pretty old, written before we had really settled on best practices for integer conversions. So I snuck in some more changes:
* Replaced `int` with `int32_t`.
* Use constructor notation for very safe integer conversions (where the range is checked) rather than static_cast.
* Use a better form of index checking. For example, rather than:

```cpp
if (meshId >= 0 && meshId < static_cast<int32_t>(model.meshes.size()))
```

Do this:

```cpp
if (meshId >= 0 && size_t(meshId) < model.meshes.size())
```

The former could reject valid indices in the (admittedly very unlikely) event that a model has more than 2^31 - 1 meshes.

PR #886, opened 2024-05-22
Don't hang in waitForMainThread waiting for a future that rejects

### LINKED ISSUE #881: waitInMainThread may fail to terminate if the future waited on is rejected
If you call `waitInMainThread` on a Future, and then that Future rejects, the `waitInMainThread` may never return. The problem is that `QueuedScheduler::dispatchUntilTaskCompletes` registers for task completion with a `.then` continuation, which isn't called when the task rejects. Fixing this may be as simple as using the overload of `.then` that takes a `async::task` instead of the resolved value.

kring:
Fixes #881


csciguy8 (approved):
Looks great @kring , thanks!

I snuck in one more test for a throw inside the lambda.

kring:
Your test looks good, thanks @csciguy8! Merging.

PR #888, opened 2024-05-22
Fixed crash when writing statically-typed glTF extensions that aren't registered

lilleyse:
I'm using some statically-typed glTF extensions that are custom to my application. Previously the code assumed that if an extension wasn't registered it must be a `JsonValue`, which would cause a crash due to a bad `any_cast`. Now statically-typed extensions that haven't been registered are ignored.


kring:
How do you feel about raising a warning when the extension isn't registered, instead of silently ignoring it? I ask because I've run into the crash before, and actually found it useful ("oh yeah, I forgot to remove that extension"). I think it's just a matter of adding a method to `JsonWriter` to report (and store) warnings, calling it from `writeJsonExtensions` when `createExtensionHandler` returns nullptr, and then copying those warnings to e.g., `TilesetWriterResult` in `TilesetWriter::writeTileset`. The reader side has a similar warning mechanism already.

lilleyse:
@kring this is ready for another look

kring:
Thanks @lilleyse!

PR #889, opened 2024-05-23
Regenerate glTF schemas for metadata extensions

lilleyse:
There have been a few corrections to the `EXT_mesh_features` and `EXT_structural_metadata` schemas which affect the generated code.

* https://github.com/CesiumGS/glTF/pull/67
* https://github.com/CesiumGS/glTF/pull/68


kring:
Thanks @lilleyse!

PR #890, opened 2024-05-24
Fully support specifying a custom ellipsoid

azrogers:
cesium-native already allows specifying a custom ellipsoid with most of its operations. This change removes the few places where the WGS84 ellipsoid was hardcoded, allowing all of cesium-native to be used with other ellipsoids. This is necessary to support CesiumGS/cesium-unreal#366.


Cesium3DTilesSelection/src/BoundingVolume.cpp
```diff
@@ -189,7 +194,7 @@ estimateGlobeRectangle(const BoundingVolume& boundingVolume) {
     }
   };
 
-  return std::visit(Operation{}, boundingVolume);
+  return std::visit(Operation(ellipsoid), boundingVolume);
```
kring, line 197: Change this to:

```suggestion
  return std::visit(Operation{ellipsoid}, boundingVolume);
```

And then you don't need the explicit constructor defined on `Operation`.

Cesium3DTilesSelection/src/BoundingVolume.cpp
```diff
@@ -72,9 +72,14 @@ glm::dvec3 getBoundingVolumeCenter(const BoundingVolume& boundingVolume) {
...
 
-std::optional<GlobeRectangle>
-estimateGlobeRectangle(const BoundingVolume& boundingVolume) {
+std::optional<GlobeRectangle> estimateGlobeRectangle(
+    const BoundingVolume& boundingVolume,
+    const CesiumGeospatial::Ellipsoid& ellipsoid) {
   struct Operation {
+    const CesiumGeospatial::Ellipsoid& ellipsoid;
+
+    Operation(const Ellipsoid& ellipsoid_) : ellipsoid(ellipsoid_) {}
```
kring, line 81: This constructor can be removed if you use brace initialization. See below.

Cesium3DTilesSelection/src/LayerJsonTerrainLoader.cpp
```diff
@@ -554,28 +561,37 @@ LayerJsonTerrainLoader::createLoader(
...
+      .thenInMainThread([ellipsoid](LoadLayersResult&& loadLayersResult) {
+        return convertToTilesetContentLoaderResult(
+            ellipsoid,
+            std::move(loadLayersResult));
       });
 }
 
 CesiumAsync::Future<TilesetContentLoaderResult<LayerJsonTerrainLoader>>
 Cesium3DTilesSelection::LayerJsonTerrainLoader::createLoader(
+    const TilesetExternals& externals,
```
kring, line 573: I'd suggest just passing in the Ellipsoid itself, rather than the entire TilesetExternals. That way the TilesetContentManager doesn't need to lambda-capture the entire `TilesetExternals` instance.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetOptions.h
```diff
@@ -318,6 +319,13 @@ struct CESIUM3DTILESSELECTION_API TilesetOptions {
    * so it must be inexpensive to copy.
    */
   std::any rendererOptions;
+
+  /**
+   * @brief The ellipsoid to use for this tileset.
+   *
+   * If no ellipsoid is set, Ellipsoid::WGS84 will be used by default.
+   */
+  std::shared_ptr<CesiumGeospatial::Ellipsoid> ellipsoid = nullptr;
```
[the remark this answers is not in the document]
kring: Why is an Ellipsoid needed in both `TilesetOptions` and `TilesetExternals`? I think just one of these should be sufficient (probably `TilesetOptions`).

I agree with Brian's suggestion to avoid the shared_ptr, too. An embedded Ellipsoid value in the TilesetOptions should be just fine, and copy it where it's needed. I wouldn't suggest holding persistent references to this, though, just copy instead. For example, passing a reference to the ellipsoid in the TilesetOptions when calling some functions is 👍, but if that function then needs to keep the ellipsoid for later, it should do so by copy, not by storing the reference passed into the function. Ellipsoid is ~104 bytes, which isn't tiny, but not too prohibitive, either, as long as we're not storing it in every `Tile` or anything like that.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
```diff
@@ -494,6 +494,10 @@ class CESIUM3DTILESSELECTION_API Tileset final {
   // scratch variable so that it can allocate only when growing bigger.
   std::vector<const TileOcclusionRendererProxy*> _childOcclusionProxies;
 
+  std::shared_ptr<CesiumGeospatial::Ellipsoid> _ellipsoid;
+
+  const CesiumGeospatial::Ellipsoid& GetEllipsoid() const;
```
kring, line 499: To match our naming conventions:

```suggestion
  const CesiumGeospatial::Ellipsoid& getEllipsoid() const;
```

kring (commented):
Looks great @azrogers, just a few comments below.

csciguy8 (commented):
Thanks for the changes @azrogers !

From what I can tell, all review comments have been resolved. I'll give @kring another chance to chime in though...

One clarifying question from me, do we still need `CESIUM_DISABLE_DEFAULT_ELLIPSOID`? If not, can we remove it? And if so, is it useful enough to put in our release notes so that other developers can use it?

(Btw, I appreciate the sheer number of files you needed to change for this PR, quite impressive!)

csciguy8 (approved):

kring:
> One clarifying question from me, do we still need CESIUM_DISABLE_DEFAULT_ELLIPSOID? If not, can we remove it? And if so, is it useful enough to put in our release notes so that other developers can use it?

I asked Ashley to add this because I think it's very useful, such as when adding alternate ellipsoid support to a new product that uses cesium-native. Adding it to the release notes seems like a good idea.

Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -462,7 +470,8 @@ void calcFittestBoundingRegionForLooseTile(
       // We need to compute an accurate bounding region
       result.updatedBoundingVolume = GltfUtilities::computeBoundingRegion(
           model,
-          tileLoadInfo.tileTransform);
+          tileLoadInfo.tileTransform,
+          *result.ellipsoid);
```
kring, line 474: It's dangerous to dereference this optional here without checking that it's valid. Maybe the field doesn't actually need to be an optional? Or if it does, change to `result.ellipsoid ? *result.ellipsoid : Ellipsoid::WGS84`.

kring: Or a third possibility: we're sure the optional has a value in this code path. In which case add an assertion for it. And a comment explaining why it's the case would be helpful would be helpful as well.

CesiumGeospatial/include/CesiumGeospatial/BoundingRegion.h
```diff
@@ -135,6 +141,7 @@ class CESIUMGEOSPATIAL_API BoundingRegion final {
       const Ellipsoid& ellipsoid);
 
   GlobeRectangle _rectangle;
+  Ellipsoid _ellipsoid;
```
kring, line 144: This significantly increases the size of a bounding region, which significantly increases the size of a `Tile`, which significantly increases our memory usage. Is it really needed? It looks like the ellipsoid is passed in to the functions that need it anyway.

CesiumGeospatial/include/CesiumGeospatial/Projection.h
```diff
@@ -12,6 +12,8 @@
 
 namespace CesiumGeospatial {
 
+class Ellipsoid;
```
kring, line 15: This forward declaration isn't sufficient in the usual case that `CESIUM_DEFAULT_ELLIPSOID` is set to ` = CesiumGeospatial::Ellipsoid::WGS84`. So we can definitely just remove it, and we may want to add the actual header. We're probably getting away without the include because one of the other headers includes Ellipsoid.h.

CesiumGeospatial/include/CesiumGeospatial/S2CellBoundingVolume.h
```diff
@@ -103,6 +108,8 @@ class CESIUMGEOSPATIAL_API S2CellBoundingVolume final {
   BoundingRegion computeBoundingRegion() const noexcept;
 
 private:
+  Ellipsoid _ellipsoid;
```
kring, line 111: Same thing that I mentioned in BoundingRegion here: adding this field this drastically increases the size of this type, so we should try really hard to avoid it.

CesiumGeospatial/src/BoundingRegion.cpp
```diff
@@ -223,7 +224,8 @@ BoundingRegion::computeUnion(const BoundingRegion& other) const noexcept {
   return BoundingRegion(
       this->_rectangle.computeUnion(other._rectangle),
       glm::min(this->_minimumHeight, other._minimumHeight),
-      glm::max(this->_maximumHeight, other._maximumHeight));
+      glm::max(this->_maximumHeight, other._maximumHeight),
+      this->_ellipsoid);
```
kring, line 228: Add the ellipsoid as a parameter to this function rather than using the field, so that we can eliminate the field.

CesiumGeospatial/src/S2CellBoundingVolume.cpp
```diff
@@ -427,5 +428,6 @@ BoundingRegion S2CellBoundingVolume::computeBoundingRegion() const noexcept {
   return BoundingRegion(
       this->_cellID.computeBoundingRectangle(),
       this->_minimumHeight,
-      this->_maximumHeight);
+      this->_maximumHeight,
+      this->_ellipsoid);
```
kring, line 432: Same here, make the ellipsoid a parameter to this function.

CesiumRasterOverlays/include/CesiumRasterOverlays/RasterOverlayUtilities.h
```diff
@@ -116,7 +116,9 @@ struct CESIUMRASTEROVERLAYS_API RasterOverlayUtilities {
       bool hasInvertedVCoordinate = false,
       const std::string_view& textureCoordinateAttributeBaseName =
           DEFAULT_TEXTURE_COORDINATE_BASE_NAME,
-      int32_t textureCoordinateIndex = 0);
+      int32_t textureCoordinateIndex = 0,
+      const CesiumGeospatial::Ellipsoid& ellipsoid =
+          CesiumGeospatial::Ellipsoid::WGS84);
```
kring, line 121: Use `CESIUM_DEFAULT_ELLIPSOID` here.

kring (changes requested):
Getting close now!

CesiumRasterOverlays/include/CesiumRasterOverlays/RasterOverlayUtilities.h
azrogers: Can't, it would cause a compile error because of the other default parameters before it.

azrogers:
@kring Updated the PR based on the review. BoundingRegion, S2CellBoundingVolume and Tileset no longer store copies of the ellipsoid, which required a good number of changes.

CMakeLists.txt
```diff
@@ -35,11 +35,16 @@ option(CESIUM_TRACING_ENABLED "Whether to enable the Cesium performance tracing
...
 option(CESIUM_GLM_STRICT_ENABLED "Whether to force strict GLM compile definitions." ON)
+option(CESIUM_DISABLE_DEFAULT_ELLIPSOID "Whether to disable the WGS84 default value for ellipsoid parameters across cesium-native." OFF)
 
 if (CESIUM_TRACING_ENABLED)
     add_compile_definitions(CESIUM_TRACING_ENABLED=1)
 endif()
 
+if(CESIUM_DISABLE_DEFAULT_ELLIPSOID)
+  add_compile_definitions(CESIUM_DISABLE_DEFAULT_ELLIPSOID=1)
+endif()
```
kring, line 46: It would be better to move this to `CesiumGeospatial/CMakeLists.txt` and write it as:

```
if(CESIUM_DISABLE_DEFAULT_ELLIPSOID)
  target_compile_definitions(CesiumGeospatial PUBLIC CESIUM_DISABLE_DEFAULT_ELLIPSOID=1)
endif()
```

`add_compile_definitions` is an old-school function with global effect and is best avoided.

Cesium3DTilesSelection/src/ImplicitQuadtreeLoader.cpp
```diff
@@ -115,9 +129,11 @@ CesiumAsync::Future<TileLoadResult> requestTileContent(
     const std::vector<CesiumAsync::IAssetAccessor::THeader>& requestHeaders,
     CesiumGltf::Ktx2TranscodeTargets ktx2TranscodeTargets,
     bool applyTextureTransform,
-    const glm::dmat4& tileTransform) {
+    const glm::dmat4& tileTransform,
+    const CesiumGeospatial::Ellipsoid& ellipsoid) {
   return pAssetAccessor->get(asyncSystem, tileUrl, requestHeaders)
-      .thenInWorkerThread([pLogger,
+      .thenInWorkerThread([&ellipsoid,
```
kring, line 135: This capture of the ellipsoid by reference looks unsafe. What guarantees do we have that the ellipsoid passed to requestTileContent won't be destroyed before the request completes and the worker thread lambda runs?

Cesium3DTilesSelection/src/ImplicitQuadtreeLoader.cpp
```diff
@@ -171,25 +187,27 @@ CesiumAsync::Future<TileLoadResult> requestTileContent(
...
-                    CesiumGeometry::Axis::Y,
-                    std::nullopt,
-                    std::nullopt,
-                    std::nullopt,
-                    std::move(pCompletedRequest),
-                    {},
-                    TileLoadResultState::Success};
-              });
+              .thenImmediately(
+                  [&ellipsoid, pLogger, tileUrl, pCompletedRequest](
```
kring, line 191: Same here.

Cesium3DTilesSelection/src/LayerJsonTerrainLoader.cpp
```diff
@@ -640,41 +659,46 @@ Future<QuantizedMeshLoadResult> requestTileContent(
...
-            const IAssetResponse* pResponse = pRequest->response();
-            if (!pResponse) {
-              QuantizedMeshLoadResult result;
-              result.errors.emplaceError(fmt::format(
-                  "Did not receive a valid response for tile content {}",
-                  pRequest->url()));
-              result.pRequest = std::move(pRequest);
-              return result;
-            }
+      .thenInWorkerThread([&ellipsoid,
```
kring, line 666: Capture by reference looks unsafe here, too.

Cesium3DTilesSelection/src/RasterOverlayUpsampler.cpp
```diff
@@ -57,14 +57,19 @@ RasterOverlayUpsampler::loadTileContent(const TileLoadInput& loadInput) {
...
   }
 
+  const CesiumGeospatial::Projection& projection = parentProjections.at(index);
+  const CesiumGeospatial::Ellipsoid& ellipsoid =
+      getProjectionEllipsoid(projection);
+
   const CesiumGltf::Model& parentModel = pParentRenderContent->getModel();
   return loadInput.asyncSystem.runInWorkerThread(
       [&parentModel,
+       &ellipsoid,
```
kring, line 72: This one _might_ be safe to capture by reference, but unless you're very sure it's better to just capture it by value.

kring (changes requested):
A few more small changes, I think this should be the last of it!

kring:
Thanks @azrogers!

PR #892, opened 2024-05-27
Move Abseil code in s2geometry to an isolated namespace.

kring:
This builds on #891 so merge that first.

This PR moves the Abseil functionality embedded in the s2geometry library into an isolated namespace (`cesium_s2geometry_absl`). This avoids linker errors when packaging an Unreal Engine 5.4 game that includes both the Cesium for Unreal and Pixel Streaming plugins, because the Pixel Streaming plugin also embeds some Abseil code.

Fixes CesiumGS/cesium-unreal#1420

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,6 +31,7 @@
 - When `GltfReader` decodes buffers with data URLs, and the size of the data in the URL does not match the buffer's `byteLength`, the `byteLength` is now updated and a warning is raised. Previously, the mismatch was ignored and would cause problems later when trying to use these buffers.
 - `EXT_meshopt_compression` and `KHR_mesh_quantization` are now removed from `extensionsUsed` and `extensionsRequired` after they are decoded by `GltfReader`.
 - Fixed a bug in the `waitInMainThread` method on `Future` and `SharedFuture` that could cause it to never return if the waited-for future rejected.
+- Moved the small amount of Abseil code embedded into the s2geometry library from the `absl` namespace to the `cesium_s2geometry_absl` namespace, in order to avoid linker errors when linking against both cesium-native and the full Abseil library.
 
 ### v0.35.0 - 2024-05-01
 
--- a/extern/CMakeLists.txt
+++ b/extern/CMakeLists.txt
@@ -112,6 +112,10 @@ if (MSVC)
   target_compile_definitions(s2geometry PRIVATE NOMINMAX _USE_MATH_DEFINES)
 endif()
 
+# S2 embeds Abseil, which may be used elsewhere as well.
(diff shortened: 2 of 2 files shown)


Immersiv-1:
Hi @kring, with the update. Is there a way we can integrate or access? Thanks

kring:
Yep, just grab a build from this PR:
https://github.com/CesiumGS/cesium-unreal/pull/1431

Instructions here:
https://github.com/CesiumGS/cesium-unreal/blob/main/Documentation/using-prerelease-packages.md

csciguy8 (approved):
Thanks @kring !

In general, this seems ok, especially if we want a quick fix for users encountering this problem. I tested out with cesium-unreal, and all seems fine.

It does feel a little hacky to force `absl` to a different value using a compile definition. My only worry would be potential build problems in other situations that aren't obvious (although I can't think of any). Ideally, `s2geometry` would have some kind of official configuration to solve problems like this. 

That said, I did notice that our `s2geometry` version is a little old (from March 2019). And there have been some later PRs related to Abseil that might help this issue (ex. [this one](https://github.com/google/s2geometry/pull/169), or [this one](https://github.com/google/s2geometry/pull/285)). Could be worth updating, now or in the future.

kring:
I agree it could be worth upgrading @csciguy8, but I don't think either of those s2geometry PRs will help us. They would allow us to use an external Abseil, but I don't think that helps.

We can't supply our own Abseil, because Unreal will still have its own, and we'd still have the conflict.

We can't use Unreal's Abseil (even if using libraries from Unreal in cesium-native were easy, which it's not) because Unreal doesn't appear to include headers and an independent library for it. It's just baked into the WebRTC library. In theory we could figure out what version of Abseil Unreal uses, and use the headers for the same version. But we'll still have challenges at link time. In a packaged build where WebRTC and Cesium for Unreal are baked into a single executable, webrtc.lib includes Abseil, and so as long as we're compiled against the right version, everything is fine. However, in a regular Editor build, WebRTC and Cesium for Unreal are in separate DLLs, and the Abseil symbols are almost certainly not exported from the WebRTC DLL. So even if we made Cesium for Unreal depend on WebRTC (which we don't want to do), we'd still need to somehow link against Abseil in a non-packaged build in order to avoid undefined symbol linker errors.

So I think the approach in this PR is the best available. Using the preprocessor to rename a namespace is a little hacky, sure, but it's simple and effective, and tightly scoped to just s2geometry (the `#define absl` doesn't exist except while we're building s2geometry). Theoretical problems we might have with it in the future (such as an unrelated symbol named `absl` getting inadvertently renamed) are likely to manifest as compiler or linker errors, so it should be relatively safe as well.

PR #895, opened 2024-06-03
Add i3dm tests and fix some bugs.

kring:
Fixes the glTF 1.0 crash I reported here:
https://github.com/CesiumGS/cesium-unreal/pull/1371#issuecomment-2138534687

Before this PR, any errors while loading a glTF would have resulted in a similar crash, and any warnings would have been silently dropped.

I also noticed and fixed two other bugs:
* `convertI3dmContent` was accidentally not returning `finishEarly` in one case, so it wasn't finishing early. This would result dereferencing an empty optional if it ever actually happened.
* `EXT_mesh_gpu_instancing` was not added to the glTF's `extensionsUsed` and `extensionsRequired` lists.

I also wrote a few basic tests of the i3dm converter, because we had none previously.

I'm going to self merge this because it's needed for today's release.


timoore:
Thanks for getting on this. I started to write similar tests that found the same issue with extensionsUsed et al., but I'll build on yours for additional i3dm tests.

PR #899, opened 2024-06-04
Support `AsyncSystem::all` on void Futures.

### LINKED ISSUE #898: `AsyncSystem::all()` does not support void futures
Calling `AsyncSystem::all()` on a vector of futures, which are of the type `Future<void>`, results in a compile-time error. It looks like the function attempts to store results in a vector, which obviously can't happen for `void`.

kring:
Fixes #898 

CC @jaadelgren


jaadelgren:
That was quick, thanks!

kring:
Moving this out of tomorrow's release because our capacity for reviews before then is low, and because it's not user-visible anyway.

PR #901, opened 2024-06-06
update rayEllipsoid Param inverseRadii to radii

jiangheng90:


jiangheng90:
the param `radii` may be better than `inverseRadii`

j9liu:
Thank you so much @jiangheng90 ! :smile:

PR #902, opened 2024-06-07
Add check for ellipsoid radii = 0 in `IntersectionTests`

j9liu:
After merging #901, I realized we needed to account for the possibility that radii = 0, leading to a division by 0 error. Completely forgot to think of it, so I've opened this PR with a fix. And some doc tweaks as well

### DIFF
--- a/CesiumGeometry/src/IntersectionTests.cpp
+++ b/CesiumGeometry/src/IntersectionTests.cpp
@@ -35,6 +35,10 @@ IntersectionTests::rayPlane(const Ray& ray, const Plane& plane) noexcept {
 std::optional<glm::dvec2> IntersectionTests::rayEllipsoid(
     const Ray& ray,
     const glm::dvec3& radii) noexcept {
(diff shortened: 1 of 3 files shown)


CesiumGeometry/src/IntersectionTests.cpp
```diff
@@ -35,6 +35,10 @@ IntersectionTests::rayPlane(const Ray& ray, const Plane& plane) noexcept {
 std::optional<glm::dvec2> IntersectionTests::rayEllipsoid(
     const Ray& ray,
     const glm::dvec3& radii) noexcept {
+  if (radii.x == 0 || radii.y == 0 || radii.z == 0) {
```
kring, line 38:
```suggestion
  if (radii.x == 0.0 || radii.y == 0.0 || radii.z == 0.0) {
```

kring (commented):
Just one trivial change, so I'll make it myself and merge this. Thanks @j9liu!

PR #908, opened 2024-06-12
Add an ellipsoid tileset loader for flat terrain generation

jqntn:
Hi,

This is a proposal to implement an ellipsoid tileset loader. It leverages the custom loader API and allows the `Cesium3DTileset` to generate terrain by creating an ellipsoidal surface.

I believe this is equivalent to [EllipsoidTerrainProvider](https://cesium.com/learn/ion-sdk/ref-doc/EllipsoidTerrainProvider.html) in [CesiumJS](https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Core/EllipsoidTerrainProvider.js).

In my opinion, this can be very useful, especially on low-end devices, or if you simply don't have a tileset at your disposal.

Here is a Unity example using the raster overlay, which you can try on [this branch](https://github.com/CesiumGS/cesium-unity-samples/pull/70) using [this version](https://github.com/CesiumGS/cesium-unity/pull/467) of the cesium-unity package:

image
image

CC @carlopiersanti


azrogers:
Hi @jqntn - thanks for the pull request! Haven't dug through it yet, I'd like to point you to PR #890, which is almost in, that adds custom ellipsoid support to cesium-native. For cesium-native, this means instead of passing the ellipsoid as a separate parameter, you can access it from `TilesetOptions::ellipsoid`. This will make sure it matches the ellipsoid that the georeference is using. 

Alternatively, if you see a circumstance where you'd want to use a different ellipsoid for terrain generation than you use for georeferencing, I'll point you to CesiumGS/cesium-unity#465. With that PR, you can change "From WGS84 Ellipsoid" to "From Ellipsoid Asset" or something of the sort on Cesium3DTileset in Unity, and add a property to it that takes a CesiumEllipsoid asset (a ScriptableObject that contains the ellipsoid's radii).

Regardless, this looks good on a quick skim - please @ me when you're ready to have it reviewed!

jqntn:
Hi @azrogers, thanks for pointing me to #890. I'm ready to make appropriate changes on my end as soon as your PR is in👍

azrogers:
@jqntn Looks like #890 just made it in!

Cesium3DTilesSelection/src/EllipsoidTilesetLoader.cpp
```diff
@@ -0,0 +1,279 @@
+#include <Cesium3DTilesContent/ImplicitTilingUtilities.h>
+#include <Cesium3DTilesSelection/EllipsoidTilesetLoader.h>
+#include <Cesium3DTilesSelection/Tile.h>
+#include <Cesium3DTilesSelection/TileContent.h>
+#include <Cesium3DTilesSelection/TileLoadResult.h>
+#include <Cesium3DTilesSelection/TileRefine.h>
+#include <Cesium3DTilesSelection/Tileset.h>
+#include <Cesium3DTilesSelection/TilesetContentLoader.h>
+#include <Cesium3DTilesSelection/TilesetExternals.h>
+#include <Cesium3DTilesSelection/TilesetOptions.h>
```
kring, line 10: There's a lot of unnecessary duplication between the headers included here and in the .h file. Some general rules:

* Always forward declare instead of including a header, if you can.
* Always make sure everything that is needed is included. Don't rely on someone else including some other header before they include yours.
* In general, a source file should include what it uses. But if a .h file includes a particular header because it has a field of that type, for example, it's not necessary to also include it in the .cpp file where that field is used.

We're a bit particular about this because it helps keep compile times as low as possible.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/EllipsoidTilesetLoader.h
```diff
@@ -0,0 +1,63 @@
...
+#include <vector>
+
+using namespace CesiumGltf;
+using namespace CesiumAsync;
+using namespace CesiumUtility;
+using namespace CesiumGeometry;
+using namespace CesiumGeospatial;
+
+namespace Cesium3DTilesSelection {
+class EllipsoidTilesetLoader : public TilesetContentLoader {
```
kring, line 30: It would be good to have some doc comments on this class and its members.

Cesium3DTilesSelection/src/EllipsoidTilesetLoader.cpp
```diff
@@ -0,0 +1,279 @@
...
+      TileLoadResultState::Success});
+}
+
+TileChildrenResult EllipsoidTilesetLoader::createTileChildren(
+    const Tile& tile,
+    const CesiumGeospatial::Ellipsoid& ellipsoid) {
+  // We don't use the ellipsoid parameter, since we already have an ellipsoid
+  // specified, but if we don't use the variable we'll get an error. This line
+  // counts as using it.
+  (void)ellipsoid;
```
kring, line 105: The more common pattern here is to declare the parameter as either:

```cpp
const CesiumGeospatial::Ellipsoid& /*ellipsoid*/
```

If the parameter has no name, the compiler won't flag it as unused. Or, you do this:

```cpp
[[maybe_unused]] const CesiumGeospatial::Ellipsoid& ellipsoid
```

This is the better option when the parameter is sometimes used. For example, when it's only used in an assertion which gets removed in release builds.

Cesium3DTilesSelection/src/EllipsoidTilesetLoader.cpp
```diff
@@ -0,0 +1,274 @@
...
+Model EllipsoidTilesetLoader::createModel(const Geometry& geometry) const {
+  const std::vector<uint16_t>& indices = geometry.indices;
+  const std::vector<glm::vec3>& vertices = geometry.vertices;
+  const std::vector<glm::vec3>& normals = geometry.normals;
+
+  size_t indicesSize = indices.size() * sizeof(uint16_t);
+  size_t verticesSize = vertices.size() * sizeof(glm::vec3);
+  size_t normalsSize = verticesSize;
+
+  Model model;
```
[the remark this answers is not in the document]
kring: Also, you need to add:

```
model.extras["gltfUpAxis"] = JsonValue(std::underlying_type_t<CesiumGeometry::Axis>(CesiumGeometry::Axis::Z));
```

Without that, the glTF is assumed to use a coordinate system in which Y is considered Up, and cesium-native will add an additional transformation to the model to make Z up. It might be hard to tell this is happening with a smooth ellipsoid, but it means the radii at various points on Earth may not be correct.

kring (changes requested):
Thanks @jqntn! I have a few suggestions for changes below. Let us know if anything is unclear, or if you want some help with any of it. We really appreciate your work on it so far!

kring:
I'm moving this out of the July release for now, to give time to make a few more changes.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/EllipsoidTilesetLoader.h
jqntn: I've added doc comments for public members, as I've usually seen done. Let me know if you meant to have these for private members as well.

kring: Thanks! No need for doc comments on private members.

jqntn:
Hi @kring! Any update on this?

j9liu:
Hi @jqntn, sorry for the delay. Thank you for implementing @kring 's suggestions! We'll get a pair of eyes on the changes as soon as we can, thank you for your patience :pray:

kring:
Really sorry, but we have to move this out of this month's release again.

azrogers:
Thank you for your patience on this, @jqntn. This PR is looking good, just one tiny change - can you move the additions you've made to CHANGES.md up to the "Not Released Yet" temporary section? It's currently under 0.37.0. I think that's the last bit needed to merge this in.

azrogers:
Scratch that, I've made that change myself. Once CI completes I'll make the merge!

jqntn:
Hi @azrogers, thank you for keeping me updated 🙂

PR #909, opened 2024-06-18
Change `TilesetJsonLoader::createLoader` overload to return `Future`

j9liu:
Previously, one of the `createLoader` overloads in `TilesetJsonLoader` returned the `TilesetContentLoaderResult` normally. This PR changes the method so it now returns a `Future` to the `TilesetContentLoaderResult`.

This is useful in cases where async tasks must be queued after a `tileset.json` has finished parsing.


Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -708,9 +709,16 @@ TilesetContentManager::TilesetContentManager(
...
+                return TilesetJsonLoader::createLoader(
+                           asyncSystem,
+                           pAssetAccessor,
+                           pLogger,
+                           url,
+                           requestHeaders,
+                           tilesetJson)
+                    .thenImmediately(
+                        [](TilesetContentLoaderResult<TilesetContentLoader>&&
+                               result) { return std::move(result); });
```
kring, line 721: This `thenImmediately` isn't doing anything AFAICT. It can be removed completely.

Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -663,6 +663,7 @@ TilesetContentManager::TilesetContentManager(
             [pLogger = externals.pLogger,
              asyncSystem = externals.asyncSystem,
              pAssetAccessor = externals.pAssetAccessor,
+             &requestHeaders = _requestHeaders,
```
kring, line 666: You can get the request headers from the `IAssetRequest` instead of capturing them here.

That's important because capturing them by reference here is a potentially dangerous pattern. We end up sharing the request headers between the thenInWorkerThread lambda and the TilesetContentManager running in the main thread. That's fine as long as neither thread modifies the request headers. If the headers can ever be modified, though, then behavior is undefined, which probably means in this case "an extremely rare and hard to debug crash." I think that can't actually happen currently, but it's a precarious situation as the code base evolves.

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -746,7 +746,10 @@ TilesetJsonLoader::createLoader(
     const std::vector<CesiumAsync::IAssetAccessor::THeader>& requestHeaders) {
   return externals.pAssetAccessor
       ->get(externals.asyncSystem, tilesetJsonUrl, requestHeaders)
-      .thenInWorkerThread([pLogger = externals.pLogger](
+      .thenInWorkerThread([asyncSystem = externals.asyncSystem,
+                           pAssetAccessor = externals.pAssetAccessor,
+                           pLogger = externals.pLogger,
+                           &requestHeaders](
```
kring, line 752: Same here, use the headers from the request.

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -788,15 +791,23 @@ TilesetJsonLoader::createLoader(
...
-            pLogger,
-            pCompletedRequest->url(),
-            tilesetJson);
+                   asyncSystem,
+                   pAssetAccessor,
+                   pLogger,
+                   pCompletedRequest->url(),
+                   requestHeaders,
+                   tilesetJson)
+            .wait();
```
kring, line 800: We should never call `wait`. Instead, return the Future directly, and change the other places that are returning a value to instead return `asyncSystem.createResolvedFuture(result)`.

Waiting in a `thenInWorkerThread` will monopolize a thread pool / task graph thread, preventing it from doing useful work until the network request completes.

kring (changes requested):
Looks good, just a few things to change below.

Cesium3DTilesSelection/src/TilesetContentManager.cpp
j9liu: Unfortunately if I don't include this, I get these build errors:
```
3>C:\Dev\cesium-native\Cesium3DTilesSelection\src\TilesetContentManager.cpp(718,1): error C3487: 'CesiumAsync::Future<Cesium3DTilesSelection::TilesetContentLoaderResult<Cesium3DTilesSelection::TilesetJsonLoader>>': all return expressions must deduce to the same type: previously it was 'CesiumAsync::Future<Cesium3DTilesSelection::TilesetContentLoaderResult<Cesium3DTilesSelection::TilesetContentLoader>>'
3>C:\Dev\cesium-native\Cesium3DTilesSelection\src\TilesetContentManager.cpp(712,17): error C2440: 'return': cannot convert from 'CesiumAsync::Future<Cesium3DTilesSelection::TilesetContentLoaderResult<Cesium3DTilesSelection::TilesetJsonLoader>>' to 'CesiumAsync::Future<Cesium3DTilesSelection::TilesetContentLoaderResult<Cesium3DTilesSelection::TilesetContentLoader>>'
3>C:\Dev\cesium-native\Cesium3DTilesSelection\src\TilesetContentManager.cpp(712,17): message : No constructor could take the source type, or constructor overload resolution was ambiguous
```

I saw something similar being done for `LayerJsonTerrainLoader` several lines down, so I thought this was just a necessary workaround 😅

j9liu:
Thanks @kring for the review! I think I addressed all your comments, let me know if I missed anything else

Cesium3DTilesSelection/src/TilesetContentManager.cpp
kring: Yikes it's relying an implicit conversion of `TilesetContentLoaderResult<TilesetJsonLoader>` to `TilesetContentLoaderResult<TilesetContentLoader>`. The implicit conversion happens when calling the `thenImmediately` lambda. That's really not obvious from the code!

Anyway, as you mentioned, it's done elsewhere. And it's not very obvious what we could do that would be better, so :shipit:.

kring:
Thanks @j9liu!

PR #910, opened 2024-06-19
Fixed build warning - I3dmToGltfConverter.cpp:634:14: error: ‘positio…

Ulvesson:
Found a compiler warning that might be problematic. See
I3dmToGltfConverter.cpp:634:14: error: ‘position.glm::vec<3, double, glm::packed_highp>::x’ may be used uninitialized [-Werror=maybe-uninitialized]

copyInstanceToBuffer might return data based on uninitialized variables.

Fixed by adding a check and an error message. The error message might not be good enough but the compiler is happy.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,10 @@
 - Added `removeUnusedMeshes` and `removeUnusedMaterials` to `GltfUtilities`.
 - Added `rayEllipsoid` static method to `CesiumGeometry::IntersectionTests`.
 
+##### Fixes :wrench:
+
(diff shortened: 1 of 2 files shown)


kring:
Thanks for the PR @Ulvesson! Before we can review and merge this, can I please bug you to sign our Contributor License Agreement?
https://github.com/CesiumGS/cesium/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla

Ulvesson:
I have signed the CLA now.

timoore:
This looks good to me. decompose() can return without setting the result variables... which shouldn't happen, but you never know.

kring (changes requested):
Looks good @Ulvesson, and thanks again for the PR! I just have one requested change, if you don't mind.

Ulvesson:
I am not 100% sure about the default values but I think I have fixed the code as requested.

kring:
Thanks again for the PR @Ulvesson!

PR #913, opened 2024-06-24
Fix uninitialized value and missing include

jaadelgren:
Minor fix for uninitialized values, which will produce an error if the build is set up to treat certain warnings as errors, and a fix for a missing include.

### DIFF
--- a/Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
+++ b/Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
@@ -631,9 +631,9 @@ void copyInstanceToBuffer(
     const glm::dmat4& instanceTransform,
     std::vector<std::byte>& bufferData,
     size_t i) {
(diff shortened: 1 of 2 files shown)


timoore:
The i3dm fix is in #910 .

jaadelgren:
> The i3dm fix is in https://github.com/CesiumGS/cesium-native/pull/910 .

Gotcha, I didn't see that. Let me know if you'd like me to just revert those changes. I was also getting an error with the missing include, so I could just leave that portion.

timoore:
> > The i3dm fix is in #910 .
> 
> Gotcha, I didn't see that. Let me know if you'd like me to just revert those changes. I was also getting an error with the missing include, so I could just leave that portion.

I think that's best. The contributor of #910 also handled some rare error conditions in decompose.

kring (approved):
This looks good to me. I noticed it's a PR into `3dtiles-voxel-content` though. Feel free to hit merge if that's where you want it. Or to rebase it onto main.

PR #919, opened 2024-06-29
Add a test for raster overlay upsampled bounding volumes, and fix a bug with upsampling and gltfUpAxis

kring:
This is a follow on to #917 that adds a test for one of the bug fixes in that PR:

> Fixed a bug while upsampling tiles for raster overlays that could cause them to have an incorrect bounding box, which in some cases would lead to the raster overlay being missing entirely from the upsampled tile.

However, in the course of writing this test, I found an unrelated bug where raster overlay upsampling was not correctly accounting for the non-standard but relatively widely used `gltfUpAxis` property. So this PR fixes that as well.


kring:
I'm tempted to merge this myself to include in tomorrow's release, since it's a useful fix and low risk. But I'll move it out of the release instead so that it can have a proper review.

Cesium3DTilesSelection/test/TestTilesetContentManager.cpp
```diff
@@ -1153,6 +1267,283 @@ TEST_CASE("Test the tileset content manager's post processing for gltf") {
...
+        new TilesetContentManager{
+            externals,
+            {},
+            std::move(rasterOverlayCollection),
+            {},
+            std::move(pMockedLoader),
+            std::move(pRootTile)};
+
+    SECTION(
+        "Generate raster overlay details when tile don't have loose region") {
```
j9liu, line 1380:
```suggestion
        "Generate raster overlay details when tile doesn't have loose region") {
```

j9liu (changes requested):
@kring looks like there's a failing test across CI (the same one): 

```
The following tests FAILED:
	 72 - Test the tileset content manager's post processing for gltf (SEGFAULT)
```

When I run the tests locally, I'm getting an error out here: 

image
image

Do you mind fixing this? Once it's fixed I'm happy to merge :smile:

j9liu:
Thanks @kring! Merging now.

PR #921, opened 2024-07-09
Add method to convert integer `AccessorSpec::ComponentType` to `PropertyComponentType`

j9liu:
This is useful for creating class definitions for `PropertyAttribute` from glTF accessors.

`ClassProperty` uses strings to convey component type, but `AccessorSpec::ComponentType` is an integer. But you can use the new method with `convertPropertyComponentTypeToString` to get the corresponding component type string.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Additions :tada:
+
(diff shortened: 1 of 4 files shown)


csciguy8 (approved):
Look like a nice addition, thanks @j9liu !

PR #922, opened 2024-07-11
Add missing include for `CesiumUtility/Assert.h`

j9liu:
A lot of the files that used `CESIUM_ASSERT` (as of #874) were missing the include for the file that actually contains the macro. This would sometimes result in compilation errors.


csciguy8 (approved):
Looks great @j9liu !

Thanks for helping out with our header build problems

PR #923, opened 2024-07-16
Changed `AccessorWriter` constructor to take `std::byte*` instead of `uint8_t*`

lilleyse:
Changed `AccessorWriter` constructor to take `std::byte*` instead of `uint8_t*` which matches the `AccessorView` constructor and prevents a compilation error.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ### ? - ?
 
+##### Breaking Changes :mega:
+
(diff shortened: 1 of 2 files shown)


csciguy8 (approved):
Looks good to me, but what compile error did you observe that led you to this change?

lilleyse:
I don't have the error handy right now, but it was about disallowed implicit conversion from `uint8_t*` to `std::byte*`

csciguy8:
Makes sense. Thanks!

Issue #497, opened 2022-05-03
Cache and reuse render resources for external images


nithinp7:
@sanjeetsuhag brought to my attention that CesiumJS saw some significant performance improvements in tilesets outputted by the CDB to 3D Tiles converter by introducing an in-memory cache for external images. Apparently the converter makes extensive use of external images and often shares the same external images amongst several tiles. I have particularly noticed frequent freezes in Cesium for Unreal / Cesium Native that can last multiple seconds when viewing the Aden, Yemen tileset (an OGC CDB dataset converted to 3D Tiles Next), with excessive texture memory being the almost guaranteed culprit. We should modify our image API to make it possible to cache external image renderer resources.

nithinp7:
This came up in a recent f[orum post](https://community.cesium.com/t/texture-loading-optimization-for-buildings/20826) and I thought some more about the details: 
> Currently, resource request urls (e.g., external image urls) are looked-up in the cache before making a network request - this means when there are multiple tiles using the same image, they should currently only be making a single network request and later reusing the response data from the cache. We could go further by adding a “resource table lookup” before checking the cache, which checks if the image is actually already loaded and resident in GPU memory. If so, we can give the new tile a shared reference to the already loaded image resource. When there are no tiles referencing the shared image, the reference count should go to 0 and it should be removed from the resource table and evicted from GPU memory.

>There are a lot of implementation details to sort out with this though. Resource ownership has to be decoupled from the lifetime of individual tiles. We need to restructure the image representations in Cesium Native to accommodate the fact that they can be external and “shared” (currently we assume they are embedded in the gltf after loading). There might be tricky corner cases like tiles referencing the same image but disagreeing on sampler filter modes, wrapping, etc, it might be unclear what the shared image should look like in such a situation.

And I'll mention again that CesiumJS has something like this, so it's worth taking inspiration from their approach.

kring:
> when there are multiple tiles using the same image, they should currently only be making a single network request and later reusing the response data from the cache

I suspect this won't actually work when two tiles request the same image at close to the same time. They'll both do a cache lookup, both get a cache miss, and both start a real network request. It's possible Unreal's HttpManager will be smart enough to avoid the redundant request, but I wouldn't count on it.

nithinp7:
>They'll both do a cache lookup, both get a cache miss, and both start a real network request

[...]

Issue #727, opened 2023-09-14
Support external metadata schemas


j9liu:
To my knowledge, the `schemaUri` property of `EXT_structural_metadata` isn't handled, and probably should be. Support was not implemented when `EXT_feature_metadata` support was added, so it was not addressed in #654.

lilleyse:
In case it helps, here's an example that uses an external schema file: https://github.com/CesiumGS/3d-tiles-samples/tree/main/glTF/EXT_structural_metadata/PropertyAttributesPointCloud

j9liu:
Closed by #969

PR #926, opened 2024-07-23
Cache images across glTFs to avoid duplication

azrogers:
As described in #497, sometimes tilesets contain multiple tiles that point to the same image resources. At the moment, these images are loaded once for every tile that uses them, meaning a tileset with 1,000 tiles that use the same image will load that image 1,000 times (and the runtimes will allocate video memory for each of them). This change adds a `SharedAssetDepot` class that stores images across tiles, and a `SharedAsset` smart pointer type for tracking and cleaning up the images. 

SharedAssetDepot is set up so that in the future, if we need, we can extend this feature to handle glTF buffers as well. 

There's a few TODO items currently (besides the runtime implementations):
- Every copy of the image is still counted towards memory used, making the metric inaccurate.
- Currently, the `SharedAsset` contains either a pointer to a stored asset and counter, or it contains the asset itself if no `SharedAssetDepot` was provided. This means that the "smart pointer" is at least 96 bytes (the size of ImageCesium), even if it does contain a pointer, which is unacceptable. I'm not sure what to do about this, short of just requiring the use of a `SharedAssetDepot` - but this would increase the number of necessary code changes to other projects that use CesiumGltf besides the native runtimes.
- The way I've laid out namespaces and usings in `SharedAssetDepot.h` is almost certainly not compatible with the style guide (I started it before the meeting) so I need to fix that.


lilleyse:
@azrogers do you think it will be straightforward to extend this approach to also support external schemas (#727) in a follow-up PR?

azrogers:
@lilleyse I've designed the SharedAssetDepot to support multiple kinds of assets, though only images are implemented in this PR. It should support extending it to also handle loading external schemas.

kring (commented):
Here are some first comments for you @azrogers. I mostly focused on SharedAssetDepot.h because I think that's the meat of the design, so this isn't a complete review yet. It's looking good! Just a few comments to tighten up the design a bit. In general, I think it would be valuable to really focus on how different things are passed and stored. I think there are some cases where things are copied unnecessarily, and for an ImageCesium that's a very expensive problem. I tried to flag some of these issues, but didn't look exhaustively. If you haven't read it before, Effective Modern C++ has lots of good info about how to decide between r-value references, pass-by-value, etc. (even though it only covers up to C++14).

CesiumGltf/include/CesiumGltf/SharedAssetDepot.h
```diff
@@ -0,0 +1,363 @@
...
+  friend class SharedAssetDepot;
+  friend class AssetContainer<AssetType>;
+  friend class SingleAssetDepot<AssetType>;
+};
+
+/**
+ * Contains one or more assets of the given type.
+ * @tparam AssetType The type of asset stored in this depot.
+ */
+template <typename AssetType> class SingleAssetDepot {
```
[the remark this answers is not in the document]
azrogers: It's the class that stores all assets of a single type, is what I'm trying to get across. The idea here is that we'll have multiple types of assets eventually in this depot (external metadata schemas being next on the agenda), and so the tileset will have one asset depot that can contain multiple types of assets. It's a little clunky but I was trying to avoid ending up with separate `getImagesSharedAssetDepot`, `getMetadataSchemasSharedAssetDepot`, `getBuffersSharedAssetDepot` etc methods on the TilesetContentManager.

CesiumGltf/include/CesiumGltf/SharedAssetDepot.h
```diff
@@ -0,0 +1,363 @@
...
+
+/**
+ * An asset that may or may not be stored in the SharedAssetDepot and shared
+ * across multiple tiles. This can either be an ImageCesium itself or a
+ * reference-counted pointer to an entry in the SharedAssetDepot. You should
+ * always treat it as if it was an ImageCesium, not a smart pointer.
+ * @tparam AssetType The type of the asset that we're getting a reference to.
+ */
+template <typename AssetType> class SharedAsset {
+  typedef std::variant<AssetContainer<AssetType>*, AssetType> AssetContents;
```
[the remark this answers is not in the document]
azrogers: I ended up going with the solution you suggested in the first comment here, as refactoring ImageCesium to be reference counted was proving to be a more significant refactor than I hoped for. I can go for that solution if you think it's worth the time.

CesiumGltf/include/CesiumGltf/SharedAssetDepot.h
```diff
@@ -0,0 +1,356 @@
...
+ * reference-counted pointer to an entry in the SharedAssetDepot. You should
+ * always treat it as if it was an ImageCesium, not a smart pointer.
+ * @tparam AssetType The type of the asset that we're getting a reference to.
+ */
+template <typename AssetType> class SharedAsset {
+public:
+  SharedAsset(AssetType&& asset)
+      : contents(new AssetContainer<AssetType>(
+            std::string(),
+            std::forward<AssetType>(asset),
```
kring, line 64: I noticed this while reviewing CesiumGS/cesium-unreal#1521 and wanted to mention it before I forget.

This `std::forward` doesn't make sense (nor does the one below). `std::forward` should only be used with forwarding references, which look like `T&&` where T is a template parameter _on that particular function_. Because your `AssetType` in this case is a parameter of the class, not the function/constructor, it's not a forwarding reference. It's just a regular r-value reference, and so `std::move` should be used instead of `std::forward`. The behavior is the same with either option, but the use of std::forward is misleading.

The constructor below is more interesting. It's really unusual to use `std::forward` with something declared `AssetType&`, and I wasn't sure offhand what it would do. But apparently what it does in this case is cast to an r-value reference. So the user calls a constructor, passing it an l-value, and the constructor moves it away as if it were an r-value. Really confusing.

I think what you want here is two constructors. One takes `AssetType&&` and uses `std::move`. The other takes `const AssetType&`, effectively promising not to modify what was passed in. But then how do you give it to `AssetContainer`, which only accepts the asset as an r-value reference? You can create an r-value from a const l-value by copying it:

```cpp
new AssetContainer<AssetType>(
  std::string(),
  AssetType(asset),
  nullptr)
```

Or, alternatively, you can add an `AssetContainer` constructor taking a const reference, too.

It's also possible to use one constructor with a forwarding reference instead of two. That would look like this:

```cpp
template <typename T>
SharedAsset(T&& asset)
    : contents(new AssetContainer<AssetType>(
          std::string(),
          std::forward<T>(asset),
          nullptr)) {}
```

Notice the use of `T` instead of `AssetType`. The downside to this is the compiler errors will be fairly horrific when there's a problem, such as when `T` and `AssetType` are not the same type.

javagl:
> Every copy of the image is still counted towards memory used, making the metric inaccurate.

Even though it is not really "_relevant_", I cannot help but draw the connection to https://github.com/CesiumGS/cesium/issues/11897 
(Sooo... at least we have _consistency_ 😶 )

kring:
I'm thinking about how the shared image cache should work. As implemented so far, it's time based. When a potentially-shared image becomes unused, it's added to a list of deletion candidates. The deletion candidates are allowed to age for 60 seconds, and then they're deleted. (Aside: I don't see any code currently to remove a shared image from the `deletionCandidates` if it becomes used again before it's evicted entirely.)

This requires a bit of bookkeeping every frame to age the deletion candidates, but the bigger problem is that 60 seconds is a long time. On a fast connection (or when loading from the request cache), we could end up loading a massive amount of data in that time, and keeping it resident unnecessarily. This probably isn't likely to be a huge problem in the normal use-cases for which this feature was implemented, but consider a pathological tileset where every glTF has a separate image loaded from a separate external URL. Or if we start using this same system for raster overlay quadtree sub-tiles as mentioned in #958. We could end up with some very high memory usage.

I think it would be better to limit the cache by size instead. When an image becomes unused, it gets added to the end of the deletion queue and the total size of the deletion queue is increased accordingly. When the total size of the unused images is above a user-configurable limit, the images at the front of the queue are deleted until we're below the limit again. If an image becomes used again while it's already in the deletion queue, be sure to remove it from the deletion queue.

This pattern should require a bit less bookkeeping, and should keep memory usage more stable.

I also would really like to see if it's possible to remove the AssetContainer and SharedAsset types, and have the AssetDepot hold reference-counted ImageCesium instances directly. I find the relationship between these types quite confusing (even after I tried to draw myself a diagram). For one thing, I can't quite understand why `AssetContainer` has two reference counts: one in its base class `CesiumUtility::ReferenceCounted`, and one in a field called `counter`. I think there might be a lot more complexity here than there needs to be. It would be fine to add an `assetId` to `ImageCesium`

I'm happy to take a crack at these two if that sounds good to you, @azrogers. Or if you want to work on it, perhaps after we've had a chance to discuss, then that's fine too!

javagl:
[This comment is **dark** 😈 ](https://github.com/CesiumGS/cesium-native/blob/91f5b5fc653dbb97339264367bfdc6c6fab1d95e/CesiumGltf/include/CesiumGltf/SharedAssetDepot.h#L263)

But more seriously, and without having looked at the details of the implementation: The "eviction strategy" - i.e. whether it is time-based or size-based - might lend itself to be a `virtual class EvictionStrategy`. There could then be two implementations (size and time), or a "combined" strategy. And I'd **strongly** suggest to at least _consider_ the common, well-known eviction strategies (Least Recently Used, Least Frequently Used, ... or maybe something like "evict largest entries first"). Each of them may have pros and cons for the respective combination of "tileset structure" and "usage pattern". But it sounds like an important point to "hook in" some abstraction, to quickly try out different strategies.

CesiumGltf/include/CesiumGltf/ImageCesium.h
```diff
@@ -110,5 +111,7 @@ struct CESIUMGLTF_API ImageCesium final {
    * this image.
    */
   int64_t sizeBytes = -1;
+
+  int64_t getSizeBytes() const override { return this->sizeBytes; }
```
kring, line 115: No need to add an accessor for a public field.

CesiumGltf/include/CesiumGltf/SharedAsset.h
```diff
@@ -0,0 +1,122 @@
...
+   * (GLB) buffer. An example of a shareable asset is an image referenced in a
+   * glTF by URI.
+   */
+  bool isShareable() const { return this->_pDepot != nullptr; }
+
+  /**
+   * The number of bytes of memory usage that this asset takes up.
+   * This is used for deletion logic by the {@link SharedAssetDepot}.
+   */
+  virtual int64_t getSizeBytes() const = 0;
```
kring, line 109: It doesn't make sense to have a virtual method in this class, because it's not designed for runtime polymorphism. `SharedAsset` uses the curiously-recurring template pattern ([CRTP](https://www.fluentcpp.com/2017/05/12/curiously-recurring-template-pattern/)), which is a technique for static / compile-time polymorphism. It's not possible to have a `SharedAsset*` and call `getSizeBytes` on it, because `SharedAsset` requires a template argument. The type would have to actually be `SharedAsset<ImageCesium>*`, which by construction is also an `ImageCesium` (every time!) and so there's no need for virtual dispatch.

So, `SharedAssetDepot` can simply call `getSizeBytes` on its instances, without the need for that method to be declared anywhere. If someone tries to create a new asset type but doesn't add a `getSizeBytes` method, it will be a compile error. Unfortunately, the error message won't be super helpful. C++20 Concepts are the solution to this, but sadly we're still on C++17 for the moment. For now we just live with it.

This is causing a compiler warning in cesium-unreal:

```
warning C4265: 'CesiumGltf::SharedAsset<CesiumGltf::ImageCesium>': class has virtual functions, but its non-trivial destructor is not virtual; instances of this class may not be destructed correctly
```

(Aside: It's interesting that this warning doesn't show up when compiling cesium-native itself. I think that's because we're incorrectly declaring the cesium-native headers as "system" headers. Yikes. Anyway, that's not something that needs to be fixed for this feature.)

CesiumGltf/include/CesiumGltf/SharedAsset.h
```diff
@@ -0,0 +1,122 @@
...
+
+private:
+  const std::string& getUniqueAssetId() const { return this->_uniqueAssetId; }
```
kring, line 112: I'm definitely not opposed to adding this accessor, rather than letting `SharedAssetDepot` access the fields directly. But we should probably do so with the other fields, too.

CesiumGltf/include/CesiumGltf/SharedAssetDepot.h
```diff
@@ -0,0 +1,260 @@
...
+   *
+   * When cached assets are no longer needed, they're marked as
+   * candidates for deletion. However, this deletion doesn't actually occur
+   * until the total byte usage of deletion candidates exceeds this threshold.
+   * At that point, assets are cleaned up in the order that they were marked for
+   * deletion until the total dips below this threshold again.
+   *
+   * Default is 100MB.
+   */
+  int64_t staleAssetSizeLimit = 1000 * 1000 * 100;
```
kring, line 44: This default seems a little high to me (especially on mobile devices), given that it's memory used by assets that are theoretically shareable between models, but that currently aren't used by any of them. We may have to get more experience with different tilesets to really decide, but in the meantime I think it's better to err on the side of "too small" (which can only force us to load an image from the request cache again) rather than "too large" (which can waste GPU memory in users' applications). The similar "sub-tile cache" in raster overlays defaults to 16 MiB (`16 * 1024 * 1024`), and I think that would be a reasonable default here, too.

CesiumGltf/include/CesiumGltf/SharedAssetDepot.h
```diff
@@ -0,0 +1,260 @@
...
+      total += age;
+      count++;
+    }
+
+    outAverageAge = count > 0 ? total / count : 0;
+    outCount = count;
+  }
+
+private:
+  // Disable copy
```
kring, line 191: Can probably just remove this comment, it's not adding anything that isn't obvious.

CesiumGltf/include/CesiumGltf/SharedAssetDepot.h
```diff
@@ -0,0 +1,260 @@
...
+      pendingAssets;
+  // Mutex for checking or editing the pendingAssets map
+  std::mutex assetsMutex;
+
+  // List of assets that are being considered for deletion, in the order that
+  // they were added.
+  std::list<CesiumUtility::IntrusivePointer<AssetType>> deletionCandidates;
+  // The total amount of memory used by all assets in the deletionCandidates
+  // list.
+  int64_t totalDeletionCandidateMemoryUsage;
```
kring, line 253: I believe this needs to be a `std::atomic` for thread safety.

CesiumGltf/include/CesiumGltf/SharedAssetDepot.h
```diff
@@ -0,0 +1,260 @@
...
+  std::unordered_map<
+      std::string,
+      CesiumAsync::SharedFuture<CesiumUtility::IntrusivePointer<AssetType>>>
+      pendingAssets;
+  // Mutex for checking or editing the pendingAssets map
+  std::mutex assetsMutex;
+
+  // List of assets that are being considered for deletion, in the order that
+  // they were added.
+  std::list<CesiumUtility::IntrusivePointer<AssetType>> deletionCandidates;
```
kring, line 250: It would be better to use our `DoublyLinkedList` class instead of `std::list`. Because with `DoublyLinkedList` the data item itself (e.g., `ImageCesium`) is the linked list node, there would be no need to search for it. So `unmarkDeletionCandidate` would become O(1) instead of O(n) (where n is the number of deletion candidates). It also avoids the need to new/delete linked list node instances. The down side is that the previous/next pointers need to be embedded in the asset class, but I think that's a reasonable tradeoff here. See [Tile](https://github.com/CesiumGS/cesium-native/blob/v0.40.1/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tile.h#L557) for an example of how to set this up.

kring (commented):
Deleting logic is looking good! I did a pass over just those changes and made some comments.

azrogers:
Trying to slim down the number of open PRs we have on this. Two remaining bullet points from @kring to cover:

- [ ] Construct SharedAssetDepot with a Factory, instead of passing one in on getOrFetch call.
- [ ] Kind of vague, but I'm not happy with GltfSharedAssetSystem just yet.

azrogers:
Made the two changes I mentioned in standup, passing the `AssetFactory` to the `SharedAssetDepot` through the constructor instead of through the `getOrFetch` call, and changing the `deletionCandidates` list to a `DoublyLinkedList`. Going to investigate getting it building with Unity next.

azrogers:
Changes to get Unity building again are now in CesiumGS/cesium-unity#518. These changes don't implement the shared asset depot, they only get it back to the current level of functionality before the cesium-native shared asset changes.

azrogers:
@kring Reviewed your changes to cesium-native and cesium-unreal. Both look good, only found a minor typo to fix (which I did myself). Looks like getting all the CI checks to pass is the next step here - anything else remaining before we can get this in?

kring:
Thanks @azrogers, I plan to merge this today!

kring:
This looks great, merging!

PR #931, opened 2024-08-03
Add `removeExtensionUsed` and `removeExtensionRequired`

lilleyse:
Adds helper functions for removing extensions from `extensionsUsed` and `extensionsRequired` and fixes a bug where `KHR_texture_transform` wasn't removed in `applyKhrTextureTransform`.

### DIFF
--- a/CesiumGltfReader/src/applyKhrTextureTransform.cpp
+++ b/CesiumGltfReader/src/applyKhrTextureTransform.cpp
@@ -118,5 +118,7 @@ void applyKhrTextureTransform(Model& model) {
       }
     }
   }
+
+  model.removeExtensionRequired(ExtensionKhrTextureTransform::ExtensionName);
 }
 } // namespace CesiumGltfReader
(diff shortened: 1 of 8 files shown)


CesiumGltfReader/src/applyKhrTextureTransform.cpp
```diff
@@ -118,5 +118,7 @@ void applyKhrTextureTransform(Model& model) {
       }
     }
   }
+
```
j9liu, line 121: Remove extra space?

j9liu (commented):
Looks good @lilleyse ! Noticed one extra line break, but I'll just push a fix and merge after.

CesiumGltfReader/src/applyKhrTextureTransform.cpp
lilleyse: I think it's ok to keep the extra space, it's consistent with the the files below

j9liu: Sorry, I had split view enabled on Github so it looked like there were two line breaks in a row. 😬 This is fine!

PR #932, opened 2024-08-03
Add `resolveExternalImages` flag to `GltfReaderOptions`

lilleyse:
Adds a new option `resolveExternalImages` to `GltfReaderOptions` so that you can load a glTF without resolving and decoding external images. In my case, I'm using `GltfReader` in a pipeline that doesn't need access to texture data.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ### ? - ?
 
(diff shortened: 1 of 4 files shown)


kring:
Thanks @lilleyse!

PR #933, opened 2024-08-03
Add `CesiumGltfWriter::SchemaWriter`

lilleyse:
Adds a serializer for `CesiumGltf::Schema` objects.

This is essentially the same as `Cesium3DTilesWriter::SchemaWriter` but I wasn't sure how to unify them in a simple way...


kring:
I don't know of a simple way to unify them either. For now we're just living with the duplication. Thanks @lilleyse!

PR #937, opened 2024-08-14
Updates to generate updated HTML documentation

corybarr:
A few of the more recent namespaces weren't being generated in the documentation:

image


azrogers:
All looks good to me. Thank you @corybarr!

PR #938, opened 2024-08-14
Use vcpkg instead of submodules for cesium-native's third party dependencies

kring:
Most of this is @jherico's work from #820.


cmake/defaults.cmake
```diff
@@ -0,0 +1,16 @@
+# Default values for all projects that can be set before the project() command
+# No `if()` statements allowed in here, particularly anything that's predecated
+# on platform or compiler detection
+
+set_property(GLOBAL PROPERTY USE_FOLDERS ON)
+set_property(GLOBAL PROPERTY PREDEFINED_TARGETS_FOLDER "CMakeTargets")
+
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+set(CMAKE_CXX_STANDARD 17)
```
j9liu, line 9: Just to confirm, cesium-native intends to stay on C++ 17 for now? (as opposed to C++ 20)

CesiumGltf/test/TestMetadataConversions.cpp
```diff
@@ -2,6 +2,13 @@
 
 #include <catch2/catch.hpp>
 
+#ifdef _MSC_VER
+// Some tests intentiontionally use a double value that overflows a float.
```
j9liu, line 6:
```suggestion
// Some tests intentionally use a double value that overflows a float.
```

CesiumAsync/src/QueuedScheduler.cpp
```diff
@@ -3,17 +3,59 @@
...
 
-// Hackily use Async++'s internal fifo_queue. We could copy it instead - it's
-// not much code - but why create the duplication? However, we are assuming that
-// Async++'s source (not just headers) are available while building
-// cesium-native. If that's a problem in some context, we'll need to do that
-// duplication of fifo_queue after all.
-#include <async++/../../src/fifo_queue.h>
+namespace {
+
+// This class is copied from async++'s fifo_queue.h:
```
j9liu, line 8: Maybe add a note about Async++'s source not being available when building `cesium-native`? (As the old comment suggests)

CesiumGltf/include/CesiumGltf/PropertyArrayView.h
```diff
@@ -34,49 +36,91 @@ template <typename ElementType> class PropertyArrayView {
...
+private:
+  gsl::span<const ElementType> _values;
+};
+
+/**
+ * @brief A view on an array element of a {@link PropertyTableProperty}
+ * or {@link PropertyTextureProperty}.
+ *
+ * Provides utility to retrieve the data stored in the array of
+ * elements via the array index operator.
```
j9liu, line 59:
```suggestion
 * @brief A copy of an array element of a {@link PropertyTableProperty}
 * or {@link PropertyTextureProperty}.
 *
 * Whereas {@link PropertyArrayView} is a pointer to data stored in a
 * separate place, a PropertyArrayCopy owns the data that it's viewing. 
 *
 * Provides utility to retrieve the data stored in the array of
 * elements via the array index operator.
```

CesiumGltf/include/CesiumGltf/PropertyArrayView.h
```diff
@@ -34,49 +36,91 @@ template <typename ElementType> class PropertyArrayView {
...
+ */
+template <typename ElementType> class PropertyArrayCopy {
```
j9liu, line 61: This is definitely helpful, but I'm curious what scenario it was solving (just so I know where I broke things :sweat_smile:)

CesiumGltf/include/CesiumGltf/PropertyTablePropertyView.h
```diff
@@ -304,30 +304,27 @@ class PropertyTablePropertyView<ElementType, false>
    * @return The value of the element, or std::nullopt if it matches the "no
    * data" value
    */
-  std::optional<ElementType> get(int64_t index) const noexcept {
+  std::optional<PropertyValueViewToCopy<ElementType>>
```
j9liu, line 307: Do you need to change the `getRaw` functions to return `PropertyValueViewToCopy`, as well?

j9liu (changes requested):
I'm admittedly not familiar with this depth of build stuff, but this looks good to me @kring! 
Just had a few questions / comments.

I'm building on my local machine as a sanity check, but given the amount of testing you're doing I'm sure it'll be fine.

CesiumGltf/include/CesiumGltf/PropertyTablePropertyView.h
j9liu: Hm, maybe not. I got thrown off when I saw `PropertyValueViewToCopy<ElementType>` in `PropertyTexturePropertyView`, but we're actually constructing the values there. Here we don't have to do that.

cmake/defaults.cmake
kring: Yes, for now that's intentional.

CesiumGltf/include/CesiumGltf/PropertyArrayView.h
kring: Nah you didn't do anything wrong! The reason for this comes down to this note inn the [swl-variant](https://github.com/groundswellaudio/swl-variant) readme:

> Clang 12/13 will only work for trivially destructible types as their implementation of C++20 is incomplete.

We're not using `swl::variant` anywhere in cesium-native (yet), but we're using it in both Unreal and Unity in the vcpkg branches in order to drastically improve build times. But anywhere we replace `std::variant` with `swl::variant`, we need to watch out for this limitation that all of the types in the variant must be trivially destructible (roughly: they don't have an explicit destructor, and neither do any of their fields).

`std::vector` isn't trivially destructible because it needs to free the heap-allocated memory it uses to hold the elements, and so neither is anything that includes `std::vector`. The old implementation of `PropertyArrayView` wasn't trivially destructible either because it held a vector. Now, PropertyArrayView is trivially destructible, and that's what we use in the `swl::variant`. PropertyArrayCopy exists for the cases where we need ownership semantics.

Then we do dirty tricks, like this, from CesiumGS/cesium-unreal#1447:

```cpp
  template <typename ArrayType>
  explicit FCesiumMetadataValue(CesiumGltf::PropertyArrayCopy<ArrayType>&& Copy)
      : _value(), _valueType(), _storage() {
    this->_value = std::move(Copy).toViewAndExternalBuffer(this->_storage);

    ECesiumMetadataType type =
        ECesiumMetadataType(CesiumGltf::TypeToPropertyType<ArrayType>::value);
    ECesiumMetadataComponentType componentType = ECesiumMetadataComponentType(
        CesiumGltf::TypeToPropertyType<ArrayType>::component);
    bool isArray = true;
    this->_valueType = {type, componentType, isArray};
  }
```

When we construct a metadata value from a _copy_ we convert it to a _view_ plus external storage, and store each separately. That way metadata value's variant can be a swl::variant`.

That's a lot of words and still not very clear, so let me know if it makes any sense.

CesiumGltf/include/CesiumGltf/PropertyTablePropertyView.h
kring: Yeah the value returned by `PropertyTablePropertyView::getRaw` is never transformed, so we can always provide it as a view. Where `PropertyTexturePropertyView` transforms even the raw values is returns.

kring:
I think I've addressed everything, this is ready for another look.

CesiumGltf/include/CesiumGltf/PropertyArrayView.h
j9liu: I think I get it! I did notice the "hot potato" you were playing with the memory, but it all makes sense now 😛

j9liu:
Thanks @kring ! Just waiting on CI to complete, then I'll merge.

PR #940, opened 2024-08-23
Support for vcpkg manifest mode

keyboardspecialist:
This adds support for including cesium-native into builds using vcpkg in manifest mode by guarding the usage of ezvcpkg when the VCPKG_MANIFEST_MODE flag is set.


kring:
Looks good, thanks @keyboardspecialist!

PR #944, opened 2024-08-30
Remove EarthGravitationalModel1996Grid::fromFile

kring:
This function is - perhaps surprisingly - the only thing in cesium-native that does file I/O (other than via an IAssetAccessor, and other than in the tests). By removing it, we're forcing users to read the WW15MGH.DAC file themselves and provide the  contents as a buffer. It's a slight burden, but not too bad. 

The immediate motivation here is that iOS prior to version 13 doesn't support `ifstream`, so the code removed by this PR won't compile there. Unity supports these older iOS versions, so cesium-unity does as well. This is fixable, but it's tricky, especially if we also want to support UTF-8 filenames on Windows, too. We could use fopen/fread on iOS, but on Windows if we pass a `char*` to those functions, it's treated as the current codepage instead of UTF-8. So we'd have to convert it to UTF-16 instead and then call `_wfopen`. Or something. It's just not worth the trouble, when any application using cesium-native probably has its own way to read files.

This is not a breaking change because we haven't shipped a version of cesium-native with this class yet.

### DIFF
--- a/CesiumGeospatial/include/CesiumGeospatial/EarthGravitationalModel1996Grid.h
+++ b/CesiumGeospatial/include/CesiumGeospatial/EarthGravitationalModel1996Grid.h
@@ -24,19 +24,6 @@ class Cartographic;
 class CESIUMGEOSPATIAL_API EarthGravitationalModel1996Grid final {
 public:
   /**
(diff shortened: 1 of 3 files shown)


j9liu:
Thanks Kevin! Merging

PR #945, opened 2024-09-06
Improve raster overlay error reporting and handling

kring:
Previously, if a raster overlay tile provider ran into an error while loading overlay images (for example, if a WMS server returned an error because it didn't support EPSG:4236), those errors were ignored, and the tileset wouldn't show up at all. This was pretty unhelpful for users, who had no way to learn what was happening except by using the debugger. With this PR:
1. Errors loading raster overlays are logged.
2. Raster overlay failures will no longer stop the geometry tile they're attached to from rendering. Instead, once the image load fails, the geometry tile will simply be rendered without any raster overlay.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,16 @@
 # Change Log
 
+### Not Released Yet
+
+##### Breaking Changes :mega:
+
+- `LoadedRasterOverlayImage` now has a single `errorList` property instead of separate `errors` and `warnings` properties.
+
+##### Fixes :wrench:
+
+- Errors while loading raster overlays are now logged. Previously, they were silently ignored in many cases.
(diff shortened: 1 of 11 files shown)


j9liu:
@kring This looks good to me! I wanted to verify the behavior where tilesets can still load with erroneous overlays, but if I don't get to it in time it's okay to merge.

j9liu:
@kring I tested this in Unreal with a Cesium ion raster overlay draped on CWT, and set it to a non-imagery asset ID (like 1). Unfortunately this did prevent CWT from rendering. But maybe this case is special?

In any case I also tested this with an invalid WMS base URL, and found that it still prevented CWT from rendering... I swear I'm using this branch and that I rebuilt the extern folder. :')

image

Do you see something different?

kring:
That should be working now @j9liu. I was testing with raster overlays that could successfully create a tile provider, but then all of the tile loads failed. In your (perfectly valid) test, the tile provider creation itself was failing. That's handled now, too.

PR #946, opened 2024-09-10
Allow vcpkg overlays to be specified as CMake variables

lilleyse:
Currently projects using cesium-native set `VCPKG_OVERLAY_PORTS` and `VCPKG_OVERLAY_TRIPLETS` as environment variables. This PR allows you to instead set them as regular CMake variables, like what we do for `VCPKG_TRIPLET`.

@jherico would this be a welcome change to bring back into https://github.com/jherico/ezvcpkg?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,22 @@ endif()
 
 message(STATUS "VCPKG_TRIPLET ${VCPKG_TRIPLET}")
 
+if (NOT VCPKG_OVERLAY_PORTS)
+    if (DEFINED ENV{VCPKG_OVERLAY_PORTS})
+        set(VCPKG_OVERLAY_PORTS "$ENV{VCPKG_OVERLAY_PORTS}")
+    endif()
+endif()
+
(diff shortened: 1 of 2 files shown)


jherico:
@lilleyse I'll investigate.  I haven't actually done anything overlay specific with ezvcpkg or vcpkg in general yet, so I have little experience.  This change seems straightforward enough though.  If you want to create a PR I'll likely merge it quickly.

kring:
@azrogers the compiler argument code in `compiler.cmake` shouldn't be uncommented, because compiler options added in that way will apply globally, but just to cesium-native libraries. This isn't so much a problem in cesium-native, because everything outside cesium-native is built by vcpkg, which will not be affected. But other projects that use cesium-native can and do have third-party libraries, and we don't want to mess with their compiler flags.

The cmake code that sets the warnings for _just_ the cesium-native libraries is here:
https://github.com/CesiumGS/cesium-native/blob/f646271fa40ace3ecb0392ec42c01b1c2b4ee4ff/cmake/macros/configure_cesium_library.cmake#L3

PR #950, opened 2024-09-26
Add `CesiumUtility::gzip`

lilleyse:
Renamed `CesiumUtility/Gunzip.h` to `CesiumUtility/Gzip.h` and added a `gzip` function. Originally I wrote this for another project but thought it would be useful to include in native.

A note on the test files: they're intentionally larger than the chunk size (65536 bytes) so that it tests the while loop properly.


azrogers:
Thanks for the change, @lilleyse!

PR #952, opened 2024-09-28
Support for mutable lambdas in `catch` functions.

kring:
I noticed while working on CesiumGS/cesium-unreal#1526 that a bug made it impossible (compiler error) to pass a mutable lambda to a `catchImmediately` or `catchInMainThread` function. Mutable lambdas are necessary when you want to `std::move` a captured value elsewhere, because in a non-mutable lambda (the default), all captured parameters are const.

This PR fixes it by adding the missing `mutable` to the wrapper lambdas.


j9liu:
Seems good to me, thanks @kring !

PR #954, opened 2024-09-30
When tileset fails to load, report all heights not sampled

kring:
Previously, we rejected the future. Now, we resolve it, and report all heights as having failed to sample. This should be easier for users to deal with because they don't need to handle two different types of errors in their height queries.


kring:
I'm merging this for today's release.

Issue #953, opened 2024-09-30
Missing `ellipsoid` parameter


lilleyse:
An ellipsoid should get passed to `BoundingRegion` here:

https://github.com/CesiumGS/cesium-native/blob/31e93ac2992eeac7600357710dc50b7c70d54b85/Cesium3DTilesSelection/src/TilesetContentManager.cpp#L154-L157

Caught while compiling with `CESIUM_DISABLE_DEFAULT_ELLIPSOID`

azrogers:
Good catch! I'll make a PR to fix this. Perhaps we should be building for CI with `CESIUM_DISABLE_DEFAULT_ELLIPSOID` to catch bugs like this?

j9liu:
Closing since #956 was merged.

PR #956, opened 2024-10-01
Fix missing ellipsoid parameters

azrogers:
As Sean noted in #953, there's a few places in the codebase we're currently missing an ellipsoid parameter. These compile fine unless `CESIUM_DISABLE_DEFAULT_ELLIPSOID` is enabled, but they will cause incorrect results if not fixed. We should consider enabling `CESIUM_DISABLE_DEFAULT_ELLIPSOID` on CI to catch these issues in the future.

### DIFF
--- a/Cesium3DTilesSelection/src/TilesetContentManager.cpp
+++ b/Cesium3DTilesSelection/src/TilesetContentManager.cpp
@@ -154,7 +154,8 @@ getTileBoundingRegionForUpsampling(const Tile& parent) {
           CesiumGeospatial::BoundingRegion(
               globeRectangle,
               details.boundingRegion.getMinimumHeight(),
-              details.boundingRegion.getMaximumHeight()),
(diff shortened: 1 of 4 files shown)


j9liu:
@kring could I trouble you to review this? I've been distracted juggling other things, and might not get to this as soon as I want to

kring:
Looks good to me @azrogers! Can you please just update CHANGES.md, and then I'll merge this.

kring:
> Updated CHANGES!

Thanks! But isn't it true that these changes represent actual cases where the chosen ellipsoid wasn't considered? So the changelog can point to incorrect results when using a non-WGS84 ellipsoid, not just compiler errors in a debug mode?

azrogers:
@kring Good catch! Updated the wording to be more accurate.

kring:
Thanks @azrogers!

PR #959, opened 2024-10-03
Tweaks to asset sharing to support Unreal implementation

kring:
This is a PR into #926. It goes with CesiumGS/cesium-unreal#1527.

### DIFF
--- a/CesiumGltf/include/CesiumGltf/AttributeSemantics.h
+++ b/CesiumGltf/include/CesiumGltf/AttributeSemantics.h
@@ -1,0 +1,51 @@
+#pragma once
+
+#include <array>
+#include <string>
+
+namespace CesiumGltf {
+
+/**
+ * @brief The standard glTF attribute semantics from the specification.
+ */
+struct AttributeSemantics {
+  /**
+   * @brief Unitless XYZ vertex positions.
+   */
+  static const std::string POSITION;
+
+  /**
+   * @brief Normalized XYZ vertex normals.
+   */
+  static const std::string NORMAL;
+
+  /**
+   * @brief XYZW vertex tangents where the XYZ portion is normalized, and the W
+   * component is a sign value (-1 or +1) indicating handedness of the tangent
+   * basis.
+   */
+  static const std::string TANGENT;
+
+  /**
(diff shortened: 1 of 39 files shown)


CesiumGltf/include/CesiumGltf/AttributeSemantics.h
```diff
@@ -0,0 +1,51 @@
...
+#include <string>
+
+namespace CesiumGltf {
+
+/**
+ * @brief The standard glTF attribute semantics from the specification.
+ */
+struct AttributeSemantics {
+  /**
+   * @brief Unitless XYZ vertex positions.
```
j9liu, line 13: Just a drive by comment, I think the glTF spec says the [linear distances](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#coordinate-system-and-units) are in meters? 😅 

image

kring: I copied "Unitless XYZ vertex positions" right out of the glTF specification. See the table in this section:
https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#meshes-overview

I think the documentation is not actually contradicting itself. The POSITION semantic is unitless, but after it's transformed by the Node hierarchy it's supposed to be meters.

lilleyse:
Couple questions since I have a similar enum/struct thing in another project and would like to use this eventually

* Should it be called `VertexAttributeSemantics` in case we add `InstanceAttributeSemantics` in the future?
* Is there the possibility of allowing custom semantics like `_FEATURE_ID_n`?

kring:
> Should it be called VertexAttributeSemantics in case we add InstanceAttributeSemantics in the future?

Good idea, I'll rename it.

> Is there the possibility of allowing custom semantics like _FEATURE_ID_n?

Do you have an idea what that would look like? It's not obvious to me what value we could provide for custom attributes in a class like this, but I could be missing something.

kring:
Oh, ok, I don't have any problem with adding extra commonly-used attributes from extensions to this class. I thought you were suggesting we make it externally-extensible somehow.

PR #962, opened 2024-10-07
Install debug vcpkg libraries from `debug/lib`

lilleyse:
Debug vcpkg libraries should be installed from the `debug/lib` directory, not the `lib` directory.

image


kring:
@lilleyse I'm going to move this out of this week's release just to avoid adding any more risk to the Cesium for Unreal release, which is already a pretty risky one. Let me know if there's any reason I'm not aware of that it does need to ship this month.

lilleyse:
@kring that's fine

kring:
I just tried this out in cesium-unreal, and (perhaps unsurprisingly) it causes linker errors due to mismatched debug / release CRT. This is frustrating because our overlay triplet specifies:

```
set(VCPKG_CXX_FLAGS "/MD /Zp8")
set(VCPKG_C_FLAGS "${VCPKG_CXX_FLAGS}")
```

But it seems that something - perhaps individual libraries - are ignoring or overriding this.

kring:
I believe I have the Unreal issues fixed in https://github.com/CesiumGS/cesium-unreal/pull/1543. Once CI confirms that, I'll merge this. Thanks @lilleyse!

PR #968, opened 2024-10-22
Add support for EXT_accessor_additional_types

lilleyse:
[EXT_accessor_additional_types](https://github.com/KhronosGroup/glTF/pull/2395) defines additional accessor component types `INT`, `INT64`, `UNSIGNED_INT64`, and `DOUBLE`.

We're using this branch for some internal pipelines where it's helpful to have higher precision types when manipulating glTF data. The changes really only matter when constructing `AccessorViews` directly and shouldn't affect the runtimes.

The extension also defines `HALF_FLOAT` but since `std::float16_t` is only available in C++23 it's not included here.


javagl:
My gut feeling is that the spin-off extension at https://github.com/KhronosGroup/glTF/pull/2397 is more likely to become "official" - exactly because it is more fine-grained and avoids the potential hassle of "half" often not being present. 

But on the level of the implementation, this should be a true subset of what is now supported here (so... this comment may just be nitpicking/noise 😶 )

kring:
Thanks @lilleyse!

PR #969, opened 2024-10-22
Support `schemaUri` property of EXT_structural_metadata

azrogers:
As noted in #727, though `EXT_structural_metadata` can specify a schemaUri that contains the URI of a JSON file to load instead of the embedded schema, we don't currently support this property. This change adds that support. 

As part of the change, I gave the generate-classes tool some idea of what a "shared asset" is, so it can now generate a shared asset property directly rather than needing to mess with derived classes like in the case of `Image`/`ImageSpec`. The largest change, by lines of code, is that a `getSizeBytes` method is now generated for every type. This is necessary for the `SharedAssetDepot` to know when it's "full." 

Code needs cleanup, especially the JavaScript code which I don't have a formatter set up for. Also need to test in each of the runtimes - if there's any changes needed, they'll be small, but I haven't checked yet.


CesiumGltf/generated/include/CesiumGltf/AccessorSparse.h
```diff
@@ -36,5 +36,13 @@ struct CESIUMGLTF_API AccessorSparse final
    * accessor values.
    */
   CesiumGltf::AccessorSparseValues values;
+
+  int64_t getSizeBytes() const {
+    int64_t accum = 0;
+    accum += sizeof(this->count);
+    accum += this->indices.getSizeBytes();
+    accum += this->values.getSizeBytes();
+    return accum;
```
kring, line 45: The size should include the size of the ExtensibleObject base class as well. Extensions and extras could be significant in some cases.

Also, counting the size of each individual field like this won't account for padding added into the structure. It's better to get the size of the entire structure, for example, `sizeof(AccessorSparse)`. Then, as you're doing, you need to add in the size of any additional data that not stored directly in a field, and that's where it gets tricky, because you need to avoid double counting. I think the most accurate implementation looks something like this:

```suggestion
    int64_t accum = 0;
    accum += sizeof(AccessorSparse);
    accum += ExtensibleObject::getSizeBytes() - sizeof(ExtensibleObject);
    accum += this->indices.getSizeBytes() - sizeof(AccessorSparseIndices);
    accum += this->values.getSizeBytes() - sizeof(AccessorSparseValues);
    return accum;
```

Basically, when you're counting the size of `indices`, you need to avoid double-counting the `AccessorSparseIndices` structure itself, because it's included in both `sizeof(AccessorSparse)` and in `this->indices.getSizeBytes()`.

CesiumGltf/generated/include/CesiumGltf/AccessorSpec.h
```diff
@@ -146,6 +154,25 @@ struct CESIUMGLTF_API AccessorSpec : public CesiumGltf::NamedObject {
...
+    int64_t accum = 0;
+    accum += sizeof(this->bufferView);
+    accum += sizeof(this->byteOffset);
+    accum += sizeof(this->componentType);
+    accum += sizeof(this->normalized);
+    accum += sizeof(this->count);
+    accum += sizeof(this->type);
+    for (const double& value : this->max) {
+      accum += sizeof(value);
+    }
```
kring, line 167: Can we add some custom logic for arrays of primitives, so that this is generated as `accum += sizeof(double) * this->max.capacity()`?

CesiumGltf/generated/include/CesiumGltf/AccessorSpec.h
```diff
@@ -146,6 +154,25 @@ struct CESIUMGLTF_API AccessorSpec : public CesiumGltf::NamedObject {
...
+    for (const double& value : this->min) {
+      accum += sizeof(value);
+    }
+    accum += sizeof(this->sparse) +
+             (this->sparse.has_value() ? this->sparse->getSizeBytes() : 0);
```
kring, line 172: With `std::optional` it should look like this I believe:

```suggestion
    if (this->sparse) {
      accum += this->sparse->getSizeBytes() - sizeof(CesiumGltf::AccessorSparse);
    }
```

(I also removed the `.has_value()` call because it's not needed)

kring (commented):
@azrogers this looks great! There are just some details in the generated `getSizeBytes` implementations that need to be worked through.

CesiumGltf/generated/include/CesiumGltf/Class.h
```diff
@@ -34,5 +34,21 @@ struct CESIUMGLTF_API Class final : public CesiumUtility::ExtensibleObject {
...
+    if (this->name) {
+      accum += this->name->size();
+    }
+    if (this->description) {
+      accum += this->description->size();
+    }
+    for (auto& [k, v] : this->properties) {
+      accum += k.size();
+      accum += v.getSizeBytes() - sizeof(CesiumGltf::ClassProperty);
+    }
```
kring, line 50: `std::unordered_map`'s API doesn't seem to give us a clear way to understand its size. But something like this should be pretty good, and definitely close enough for our purposes:

```suggestion
    accum += this->properties.bucket_count() * (sizeof(std::string) + sizeof(CesiumGltf::ClassProperty));
    for (const auto& [k, v] : this->properties) {
      accum += k.capacity() * sizeof(char) - sizeof(std::string);
      accum += v.getSizeBytes() - sizeof(CesiumGltf::ClassProperty);
    }
```

CesiumGltf/generated/include/CesiumGltf/ModelSpec.h
```diff
@@ -137,6 +137,54 @@ struct CESIUMGLTF_API ModelSpec : public CesiumUtility::ExtensibleObject {
    */
   std::vector<CesiumGltf::Texture> textures;
 
+  int64_t getSizeBytes() const {
```
kring, line 140: Can we add a generated comment to these getSizeBytes methods explaining what it does, and also mentioning that it requires a traversal of the entire structure, which may be slow?

CesiumUtility/include/CesiumUtility/JsonValue.h
```diff
@@ -620,6 +620,48 @@ class CESIUMUTILITY_API JsonValue final {
...
+      int64_t operator()(const Array& value) {
+        int64_t accum = 0;
+        for (const JsonValue& v : value) {
+          accum += v.getSizeBytes();
+        }
+        return accum;
+      }
+    };
+
+    return std::visit(Operation{}, this->value);
```
kring, line 662: The size should always be `sizeof(JsonValue)` plus some extra externally-allocated memory for strings, arrays, and objects. The logic for the extra should be the same as for fields of those types in regular objects.

kring (commented):
Looking good, just a little more `getSizeBytes` nitpicking! Also, don't forget that all of the implementations of that method need to chain to their base class (usually `ExtensibleObject`) implementation, too!

azrogers:
@kring Just letting you know this should be good for another look!

kring (commented):
Nearly there! Just one tricky `getSizeBytes` situation left that I'm going to take a quick look at myself.

CesiumUtility/include/CesiumUtility/ExtensibleObject.h
```diff
@@ -118,5 +118,37 @@ struct CESIUMUTILITY_API ExtensibleObject {
...
+   */
+  int64_t getSizeBytes() const {
+    int64_t accum = 0;
+    accum += sizeof(ExtensibleObject);
+    accum += this->extensions.bucket_count() *
+             (sizeof(std::string) + sizeof(JsonValue));
+    for (auto& [k, v] : this->extensions) {
+      accum += k.capacity() * sizeof(char) - sizeof(std::string);
+      const JsonValue* pValue = std::any_cast<JsonValue>(&v);
+      accum += pValue->getSizeBytes() - sizeof(JsonValue);
```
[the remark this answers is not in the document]
kring: Well, this is a pain. A generic dispatcher like the one I envisioned when I wrote the above is difficult. We'd need to generate a dispatch function that does an `any_cast` to each of the possible extensions in turn, and then calls the provided callback for the first cast that succeeds. Not terribly efficient.

Or we do what the "writer" system does and look up classes based on the name of the extension and the name of the class extended (because a single logical glTF extension can have a different C++ type for extending different glTF classes).

Either way, it's a little tricky to figure out where to put this. It can't go in `ExtensibleObject` because that's in CesiumUtility, so we can't generate glTF-specific code there. So we'd have to pass in some kind of (generated) extension context to ExtensibleObject's `getSizeBytes` I guess. `ExtensibleObject::getSizeBytes` would have to be a template on the type of the extension context.

I think what we should do for now is:
1. Completely remove the extension size tracking from `ExtensibleObject::getSizeBytes`.
2. Update the generate doc for `getSizeBytes` to say it doesn't include the size of any extensions.

And I think that will be good enough. If we find that extension size tracking is actually important in the future, we can invest the time to do something like the above.

azrogers:
@kring Updated the doc comments and added const to those `for(auto& ...` loops!

kring:
All changes look good! But I tried regenerating the 3D Tiles (`npm run generate-3d-tiles`) and quantized-mesh (`npm run generate-quantized-mesh-terrain`), and the generated code has compiler errors in both cases.

azrogers:
Looks like it was a combination of a missing `#include` and an issue with nested loops in LayerSpec. Fixed those issues and regenerated those classes, as well as updating CHANGES.

kring:
I fixed some clang/gcc warnings caused by overzealous integer conversion checking (no, clang, sizeof(T) is never going to overflow an `int64_t`, but thanks for your concern).

This looks great, merging!

PR #977, opened 2024-11-06
Remove use of the GSL library

kring:
This builds on #976, because it switches `gsl::span` to `std::span`, which was added in C++20.

This is necessary to use cesium-native in UE 5.5 (Preview) after it introduced a broken copy of GSL.


j9liu:
Thanks @kring ! I didn't check every file, but I spot checked around 100 files and found no surprises. I'll merge once CI passes!

PR #978, opened 2024-11-06
Add `default` constructor to `ImageAsset`

j9liu:
This is a small change to make the changes from #926 compatible with C++ 20. I previously got errors while trying to instantiate an empty `ImageAsset`, because its underlying `SharedAsset` constructor is `protected`. I'm not sure why this wasn't flagged in C++17,  but oh well :shrug:


kring:
Thanks @j9liu!

Issue #958, opened 2024-10-03
QuadtreeRasterOverlay sub-tile cache size accounting is wrong


kring:
`QuadtreeRasterOverlay` has a small cache that is intended to briefly store decoded raster overlay tile images that span multiple geometry tiles, so we don't need to decode them multiple times (or download them multiple times in the hopefully-unlikely event that the requests aren't cached).

But it doesn't appear to be working correctly.

`QuadtreeRasterOverlayTileProvider` has a `_cachedBytes` field. When a new overlay image is loaded, this field is immediately incremented accordingly. The code that later _decrements_ this field looks like this:

```cpp
this->_tileLookup.erase(it->tileID);
it = this->_tilesOldToRecent.erase(it);

// If this is the last use of this data, it will be freed when the shared
// pointer goes out of scope, so reduce the cachedBytes accordingly.
if (pImage.use_count() == 1) {
  if (pImage->image) {
    this->_cachedBytes -= int64_t(pImage->image->pixelData.size());
    CESIUM_ASSERT(this->_cachedBytes >= 0);
  }
}
```

This snippet is executed when we're removing this tile from the cache because the cache is full and this tile is the one that is least-recently-used. It is _not_ necessarily true that this image is unused when this happens. And because of that `use_count() == 1` check, if this tile happens to be in use elsewhere when it's evicted from the cache, we'll never subtract its bytes.

As a result, over time (and it doesn't take very long), we'll think the cache is completely full at all times and each new image will be evicted immediately, making the cache pretty close to useless.

I think the solution is simply to remove that `use_count() == 1` check. When we evict an image from the cache, we should always reduce the number of bytes in the cache accordingly. Based on the comment, when I wrote this, I seemed to be hung up on the question of whether the image would actually be freed. But the freeing doesn't matter; all that matters is whether this image is kept in the set of tiles temporarily kept around even though they're potentially unused.

kring:
This entire sub-tile cache should probably just be replaced by the external image cache in #926!

kring:
Fixed in #992.

PR #987, opened 2024-11-08
Replace RasterOverlayTileProvider cache with SharedAssetDepot

azrogers:
There is currently a cache in the `QuadtreeRasterOverlayTileProvider` to hopefully prevent the same images from being downloaded multiple times when they're used by multiple different geometry tiles. There is currently a bug that will result in the overlay provider believing that the cache is full when it's not, making the cache eventually all but useless. As mentioned in #958, the issue for this bug, now that we have the `SharedAsset` system, it makes sense to just get rid of that old cache code and use a `SharedAssetDepot` instead. That's what this PR is for.

While this PR currently implements the cache on the RasterOverlayTileProvider base class itself, it will only actually be used when the `loadTileImageFromUrl` method is called, and only `QuadtreeRasterOverlayTileProvider`-derived classes seem to call this method.

While this PR is in good working order, at least from my testing, it feels like it loads raster overlays slower than it did previously. This isn't a particularly surprising issue to occur after these changes, but it does need to be fixed. I'll try to get to the bottom of this on Monday.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,10 @@
 - Added support for `EXT_accessor_additional_types` in `AccessorView`.
 - Added `EllipsoidTilesetLoader` that will generate a tileset by tesselating the surface of an ellipsoid, producing a simple globe tileset without any terrain features.
 
+##### Fixes :wrench:
+
+- Updated the CMake install process to install the vcpkg-built Debug binaries in Debug builds. Previously the Release binaries were installed instead.
+
 ### v0.41.0 - 2024-11-01
 
 ##### Breaking Changes :mega:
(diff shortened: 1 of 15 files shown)


kring:
@azrogers I took a look at this, and I'm not sure where the performance regression might be coming from. But there's a lot more complexity than I expected, and more than I _think_ is necessary (though it's always possible I'm missing something!).

I think it should be possible to replace `_tilesOldToRecent`, `_tileLookup`, and `_cachedBytes` in `QuadtreeRasterOverlayTileProvider` with a `SharedAssetDepot<LoadedQuadtreeImage, QuadtreeTileID>`. You'll have to make `LoadedQuadtreeImage` into a shared asset (instead of `LoadedRasterOverlayImage`). But not much else should have to change. In particular, there shouldn't be any breaking API changes for people that developed their own raster overlay types (and our own shouldn't have to change, either).

The trick (maybe!) is to think of `getQuadtreeTile` as the factory that the SharedAssetDepot uses to turn a QuadtreeTileID into a LoadedQuadtreeImage. `mapRasterTilesToGeometryTile` calls the asset depot's `getOrCreate` instead of calling `getQuadtreeTile` directly.

Let me know if you want to discuss. Or if I'm crazy and have missed something big.

azrogers:
Not sure why the most recent merge into main is now listed in this PR... but I don't think it'll cause issues. PR should be ready to review again @kring

PR #990, opened 2024-11-13
Remove old `CESIUM_NATIVE_` CMake variables

lilleyse:
I don't think these are needed anymore after the switch to vcpkg


kring:
Nice cleanup, thanks @lilleyse!

PR #995, opened 2024-11-15
Avoid writing empty extensions object

lilleyse:
Fixed a bug where an empty `extensions` object would get written if an `ExtensibleObject` only had unregistered extensions.

Before:

image

After:

image


kring:
Thanks @lilleyse!

PR #998, opened 2024-11-19
Run build on macos-15 instead of macos-12.

kring:
The latter is no longer supported by GitHub Actions.


kring:
I'm merging this to unbreak main.

PR #999, opened 2024-11-19
Schema is a struct not a class.

kring:
This showed up as a compiler error when building Cesium for Unreal against the latest version of cesium-native in main.


j9liu:
Indeed it is -- thanks @kring !

PR #1001, opened 2024-11-19
Add checks for `nullptr` credit system in raster overlays

j9liu:
Part of the fix for https://github.com/CesiumGS/cesium-unreal/issues/1549. This adds checks for a null `CreditSystem` to the various overlay types.

TMS is the only raster overlay type with unit tests, so that was the only one I was able to verify via test. But the change is similar for the other raster overlay types, so hopefully it's sufficient.


kring:
Thanks @j9liu!

PR #1002, opened 2024-11-20
Eliminate variable shadowing in JsonValue

kring:
Because Clang gets upset about this in the cesium-unreal build.

### DIFF
--- a/CesiumUtility/include/CesiumUtility/JsonValue.h
+++ b/CesiumUtility/include/CesiumUtility/JsonValue.h
@@ -625,32 +625,34 @@ public:
    */
   int64_t getSizeBytes() const noexcept {
     struct Operation {
-      int64_t operator()([[maybe_unused]] const Null& value) { return 0; }
-      int64_t operator()([[maybe_unused]] const double& value) { return 0; }
-      int64_t operator()([[maybe_unused]] const std::uint64_t& value) {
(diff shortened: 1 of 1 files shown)


CesiumUtility/include/CesiumUtility/JsonValue.h
```diff
@@ -625,32 +625,34 @@ class CESIUMUTILITY_API JsonValue final {
    */
   int64_t getSizeBytes() const noexcept {
     struct Operation {
-      int64_t operator()([[maybe_unused]] const Null& value) { return 0; }
-      int64_t operator()([[maybe_unused]] const double& value) { return 0; }
-      int64_t operator()([[maybe_unused]] const std::uint64_t& value) {
+      int64_t operator()([[maybe_unused]] const Null& /*inside*/) { return 0; }
+      int64_t operator()([[maybe_unused]] const double& /*inside*/) {
         return 0;
```
j9liu, line 630: Not in scope for this PR, just a question. What's the reason this doesn't return something like `sizeof(double)`?

j9liu (approved):
Thanks @kring! I had an unrelated question that's out of scope for this PR, so I'm going to go ahead and merge.

CesiumUtility/include/CesiumUtility/JsonValue.h
kring: Because the size of the double is already included in the size of the variant. We only need to count extra data external to the variant.

PR #1003, opened 2024-11-20
Clean up raster overlay handling of ellipsoids

kring:
This is a PR into #1002 so merge that first.

Fixes CesiumGS/cesium-unreal#1551

* BingMapsRasterOverlay no longer has a separate `ellipsoid` constructor parameter and field. Instead, it uses the one in `RasterOverlayOptions`.
* The `ellipsoid` field in `RasterOverlayOptions` is no longer optional. Instead, it defaults to WGS84.
* `WebMapServiceRasterOverlayOptions` and `TileMapServiceRasterOverlayOptions` no longer have an `ellipsoid` field. It's redundant with the one in `RasterOverlayOptions`.
* `DebugColorizeTilesRasterOverlay` no longer stores its ellipsoid redundantly.
* `IonRasterOverlay` now passes its `RasterOverlayOptions` through to the Bing or TMS overlay it creates.

This all helps fix CesiumGS/cesium-unreal#1551 by making sure we're using Unreal's WGS84 ellipsoid everywhere, rather than using that one in some places and the "real" WGS84 in other places. They're different by 1e-6 meters (1 micrometer) in the semi-minor axis, which shouldn't make any practical difference, but they don't compare as exactly equal.

### DIFF
--- a/CesiumRasterOverlays/include/CesiumRasterOverlays/BingMapsRasterOverlay.h
+++ b/CesiumRasterOverlays/include/CesiumRasterOverlays/BingMapsRasterOverlay.h
@@ -96,8 +96,6 @@ public:
    * all cultures are supported. See
    * http://msdn.microsoft.com/en-us/library/hh441729.aspx for information on
    * the supported cultures.
-   * @param ellipsoid The ellipsoid. Default value:
-   * {@link CesiumGeospatial::Ellipsoid::WGS84}.
    * @param overlayOptions The {@link RasterOverlayOptions} for this instance.
    */
   BingMapsRasterOverlay(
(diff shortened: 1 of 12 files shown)


azrogers:
No comments from looking over the PR, but [as per my comment](https://github.com/CesiumGS/cesium-unreal/pull/1554#issuecomment-2494650787) on CesiumGS/cesium-unreal#1554, I haven't been able to test this yet.

azrogers:
Was able to resolve my build issues and get this tested in Unreal. Together with CesiumGS/cesium-unreal#1554 this does seem to fix CesiumGS/cesium-unreal#1551. Thanks for the change!

PR #1004, opened 2024-11-20
Add clang-tidy targets and CI check

### LINKED ISSUE #460: Fix warnings in public headers

So we should:
* Fix the warnings

azrogers:
- Added clang-tidy targets to CMake when building for a non-MSVC target.  Currently clang-tidy is only set to warn, not raise errors, as there are too many issues to fix for one PR and it's not necessarily worth our time to do them all right now.
- Added the option CESIUM_ENABLE_CLANG_TIDY_ON_BUILD, which will run clang-tidy alongside every build command CMake runs. This slows down the build significantly, so it's disabled by default. However, this allows clang-tidy to be run even with an MSVC target. VSWhere is used on Windows to locate Visual Studio's copy of clang-tidy.
- Addressed part of #460, allowing warnings to be reported in public headers when cesium-native isn't being included in another CMake project. This has created a lot of errors around our use of int64_t versus size_t in PropertyTable-related classes, which has ended up requiring some pretty thorough changes to resolve.
- Added a CI action to run clang-tidy for each new commit.


azrogers:
After discussion with @kring, we're going to put aside completely resolving #460 for the moment, as adding a ton of casts to PropertyTable headers doesn't seem like it's necessarily the best use of resources right now. This means that CesiumGltf public headers will, for the moment, remain as `SYSTEM` and we won't receive build warnings or clang-tidy results for them. For future reference, making this change for CesiumGltf will require changing line 43 of CesiumGltf/CMakeLists.txt to the following:
```cmake
cesium_target_include_directories(
    TARGET
        CesiumGltf
    PUBLIC
        ${CMAKE_CURRENT_LIST_DIR}/include/
        ${CMAKE_CURRENT_LIST_DIR}/generated/include
    PRIVATE
        ${CMAKE_CURRENT_LIST_DIR}/src
        ${CMAKE_CURRENT_LIST_DIR}/generated/src
)
```

With that out of the way, this should be good for review.

azrogers:
Also, thank you to @lilleyse for the help here, especially with the changes to the code generation which were cherry-picked from his clang-tidy-3 branch.

kring (commented):
Looks good! I looked through all the non-generated changes, and only had a couple of small comments.

I also tried to build and run clang-tidy locally, though, and ran into some trouble.

First, cesium-native no longer builds on my system with MSVC 14.38:

> [build] F:\cesium\cesium-unreal-samples\Plugins\cesium-unreal\extern\cesium-native\CesiumGltfReader\generated\src\GeneratedJsonHandlers.cpp(66,39): error C2220: the following warning is treated as an error [F:\cesium\cesium-unreal-samples\Plugins\cesium-unreal\extern\build\cesium-native\CesiumGltfReader\CesiumGltfReader.vcxproj]
[build] F:\cesium\cesium-unreal-samples\Plugins\cesium-unreal\extern\cesium-native\CesiumGltfReader\generated\src\GeneratedJsonHandlers.cpp(66,39): warning C4455: 'operator ""s': literal suffix identifiers that do not start with an underscore are reserved [F:\cesium\cesium-unreal-samples\Plugins\cesium-unreal\extern\build\cesium-native\CesiumGltfReader\CesiumGltfReader.vcxproj]

This seems like it might be a bug with this older version of the compiler (https://developercommunity.visualstudio.com/t/warning-c4455-issued-when-using-standardized-liter/270349), but sadly Unreal requires us to use this version (or an even older one in some cases, 14.34). It's not obvious to me what changed in this branch to trigger this warning, though. Any ideas?

I also had trouble when compiling Cesium for Unreal's `extern` directory against this branch:

> [cmake] CMake Error at cesium-native/CMakeLists.txt:110 (cmake_dependent_option):
> [cmake]   Unknown CMake command "cmake_dependent_option".

I was able to fix this by adding this line to cesium-native's CMakeList.txt, which I think is a reasonable (required?) thing to do

```
include(CMakeDependentOption)
```

But I don't know why this only shows up when building cesium-native for Unreal, not when building cesium-native itself. 🤷

kring:
I also looked very briefly through the list of new clang-tidy warnings (read: I look at the first couple), and saw this:

```
/usr/bin/clang-tidy-18 -p=/home/runner/work/cesium-native/cesium-native/build /home/runner/work/cesium-native/cesium-native/CesiumGeospatial/test/TestGlobeRectangle.cpp
/home/runner/work/cesium-native/cesium-native/CesiumGeospatial/test/TestGlobeRectangle.cpp:4:1: warning: included header catch.hpp is not used directly [misc-include-cleaner]
    4 | #include <catch2/catch.hpp>
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~
    5 | 
/home/runner/work/cesium-native/cesium-native/CesiumGeospatial/test/TestGlobeRectangle.cpp:9:1: warning: no header providing "TEST_CASE" is directly included [misc-include-cleaner]
    5 | 
    6 | using namespace CesiumGeospatial;
    7 | using namespace CesiumUtility;
    8 | 
    9 | TEST_CASE("GlobeRectangle::fromDegrees example") {
      | ^
```

Which isn't a good look, because `catch2/catch.hpp` _is_ the header that provides `TEST_CASE`. Any idea why clang-tidy is confused here? Maybe we should be including individual catch2 headers?

lilleyse:
> Maybe we should be including individual catch2 headers?

Yeah, in this case you would need to include `<catch2/catch_test_macros.hpp>` to silence the warning.

Cesium3DTilesContent/test/TestUpgradeBatchTableToExtStructuralMetadata.cpp
```diff
@@ -103,16 +103,16 @@ static void checkArrayProperty(
     PropertyArrayView<PropertyViewType> value =
         propertyView.getRaw(static_cast<int64_t>(i));
     if (expectedCount > 0) {
-      REQUIRE(value.size() == expectedCount);
+      REQUIRE(static_cast<int64_t>(value.size()) == expectedCount);
```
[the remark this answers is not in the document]
azrogers: Actually, this is a change from my attempts to fix the PropertyTable-related headers in CesiumGltf that I didn't catch when reverting. It shouldn't be necessary at the moment.

azrogers:
The Catch2 GitHub README and tutorial show using `<catch2/catch_test_macros.hpp>` instead of `<catch2/catch.hpp>` which we currently do, so I think that's the right move.

azrogers:
> First, cesium-native no longer builds on my system with MSVC 14.38:
> 
> > [build] F:\cesium\cesium-unreal-samples\Plugins\cesium-unreal\extern\cesium-native\CesiumGltfReader\generated\src\GeneratedJsonHandlers.cpp(66,39): error C2220: the following warning is treated as an error [F:\cesium\cesium-unreal-samples\Plugins\cesium-unreal\extern\build\cesium-native\CesiumGltfReader\CesiumGltfReader.vcxproj]
> > [build] F:\cesium\cesium-unreal-samples\Plugins\cesium-unreal\extern\cesium-native\CesiumGltfReader\generated\src\GeneratedJsonHandlers.cpp(66,39): warning C4455: 'operator ""s': literal suffix identifiers that do not start with an underscore are reserved [F:\cesium\cesium-unreal-samples\Plugins\cesium-unreal\extern\build\cesium-native\CesiumGltfReader\CesiumGltfReader.vcxproj]
> 
> This seems like it might be a bug with this older version of the compiler (https://developercommunity.visualstudio.com/t/warning-c4455-issued-when-using-standardized-liter/270349), but sadly Unreal requires us to use this version (or an even older one in some cases, 14.34). It's not obvious to me what changed in this branch to trigger this warning, though. Any ideas?

Looks like this comes down to a change made in the code generator from `using namespace std::string_literals;` in the GenreatedJsonHandlers to `using std::string_literals::operator""s;`. The latter is definitely more correct, but for whatever reason MSVC seems to have an issue with it and not with the former. I've reverted the change, which seems to have fixed this error.

azrogers:
Unless I missed one, I believe I've taken care of all the review comments at the moment, whenever you get a chance to take another look @kring

azrogers:
The fact that clang-tidy produces a 21k line output for cesium-native is less than ideal. I think it's because each clang-tidy run for each file is independent, so the same headers end up getting checked and producing warnings over and over again. I feel like it would be nice to at least get all this outputted into a log file artifact to download and peruse, rather than almost crashing the browser tab trying to open it on GitHub, but I'm not sure whether that would really be all that useful.

tools/generate-classes/generate.js
```diff
@@ -315,35 +315,47 @@ function generate(options, schema, writers) {
   );
 
   const jsonHandlerHeadersImpl = lodash.uniq([
+    NameFormatters.getJsonHandlerIncludeFromName(base, readerNamespace),
     ...lodash.flatten(properties.map((property) => property.readerHeadersImpl)),
   ]);
+
+  if (thisConfig.extensionName) {
+    jsonHandlerHeadersImpl.push("<CesiumUtility/ExtensibleObject.h>");
+    jsonHandlerHeadersImpl.push("<any>");
```
kring, line 324: This shouldn't be needed, because ExtensibleObject.h already includes it.

tools/generate-classes/generate.js
```diff
@@ -315,35 +315,47 @@ function generate(options, schema, writers) {
...
   function generateJsonHandlerOptionsInitializerList(properties, varName) {
     const initializerList = properties
       .filter((p) => p.readerType.toLowerCase().indexOf("jsonhandler") != -1)
-      .map(
-        (p) =>
-          `_${p.cppSafeName}(${p.schemas && p.schemas.length > 0 ? varName : ""
-          })`
-      )
+      .filter((p) => p.schemas && p.schemas.length > 0)
+      .map((p) => `_${p.cppSafeName}(${varName})`)
```
kring, line 333: What's the reason for this change? It appears to result in not all of the fields being initialized in this constructor.

tools/generate-classes/generate.js
```diff
@@ -576,7 +588,9 @@ function formatProperty(property) {
   result += `${property.type} ${property.cppSafeName}`;
 
   if (property.defaultValue !== undefined) {
-    result += " = " + property.defaultValue;
+    if (property.defaultValue != '""') {
```
kring, line 591: I think this would be better, though it doesn't change the generated code at all.
```suggestion
  if (property.defaultValue !== undefined && property.defaultValue != '""') {
```

tools/generate-classes/generate.js
lilleyse: This was likely added because `misc-include-cleaner` doesn't allow indirect dependencies.

tools/generate-classes/generate.js
lilleyse: I believe it was for this check: https://clang.llvm.org/extra/clang-tidy/checks/readability/redundant-member-init.html

azrogers:
It seems like the CMAKE_CXX_CLANG_TIDY value doesn't get mapped over to anything in the generated Visual Studio project, if that's the generator selected. Apparently Visual Studio's CMake support uses Ninja, so setting the CMAKE_CXX_CLANG_TIDY value does work from there. I've updated the checks to depend on selected generator rather than compiler, and also fixed an issue when using CESIUM_ENABLE_CLANG_TIDY_ON_BUILD with GCC.

tools/generate-classes/generate.js
kring: Oh right, of course, sorry for the noise!

tools/generate-classes/generate.js
kring: Oh. Huh. That's a strange rule. I was under the impression the C++ init rules are too complicated for mere mortals to understand and therefore you should always init all fields if you init any of them, and in the order they're declared. Is there new wisdom here that I'm unaware of?

azrogers: We can always just disable the check.

kring: Based on a quick search, I think I got that from Effective C++. It could easily be outdated practice at this point, but it's not obvious to me why it would be.

kring: Ok sounds good @azrogers! I'm still interested in hearing why clang-tidy even has such a rule, if anyone knows. Presumably there's a reason for it.

kring:
I was able to run clang-tidy on Windows. First I had to edit `setup_clang_tidy.cmake` to explicitly invoke `run-clang-tidy` via Python, because this doesn't happen automatically otherwise.

```
add_custom_target(
    clang-tidy COMMAND python ${CLANG_TIDY_RUNNER_PATH} -clang-tidy-binary ${CLANG_TIDY_PATH} -p
    ${_PROJECT_BUILD_DIRECTORY} # path that contains a compile_commands.json
)
add_custom_target(
    clang-tidy-fix COMMAND python ${CLANG_TIDY_RUNNER_PATH} -fix -clang-tidy-binary ${CLANG_TIDY_PATH} -p
    ${_PROJECT_BUILD_DIRECTORY} # path that contains a compile_commands.json
)
```

Then I was able to invoke it by doing the following from a cmd prompt (PowerShell will _not_ work with vcvarsall.bat, which is annoying):

```
"c:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Auxiliary\Build\vcvarsall.bat" amd64
cmake -B build -S . -G Ninja
cmake --build build --target clang-tidy
```

I think before we "fully adopt" clang-tidy, we'll need to make this much easier somehow. Ideally, clang-tidy warnings would just automatically appear in VS Code, just like compiler errors do. I have no idea how realistic that is, but surely we're not the only ones who would want something like that, right?

But this is a great start! Merging!

PR #1006, opened 2024-11-21
Add support for Mermaid charts to Doxygen

### LINKED ISSUE #997: Include Mermaid diagrams in Doxygen
[Mermaid](https://mermaid.js.org/) is a JavaScript-based diagramming tool that would be very useful for us to have in our documentation. It would make it easier to create and maintain flow charts, sequence diagrams, and any other sorts of diagrams and charts we need. Implementing Mermaid in Doxygen is very possible, but not trivial. [This repository](https://github.com/tttapa/doxygen-mermaid) contains an example of one approach to implementing Mermaid, but it might not be the best. We should investigate our options here and implement a way to generate and include these charts in our documentation.

azrogers:
As per #997, this PR adds support for Mermaid charts in Doxygen. It allows Mermaid charts to be added to the `doc/diagrams` folder and included in comments using the `@mermaid{filename}` alias. You can test it out by grabbing some Mermaid code, for example from [this bit of their documentation](https://mermaid.js.org/intro/getting-started.html#features), putting it in `doc/diagrams/test.mmd`, and adding a `@mermaid{test}` command to any Doxygen comment. After generating the docs, you should see the chart show up where the command was placed.

I've put this into a draft PR for the moment because somehow these changes have resulted in the sidebar disappearing. I believe this is a result of modifying Doxygen's footer.html, but I can't tell exactly why it would have this effect. Needs to be fiddled with.

Fixes #997


azrogers:
For whatever reason, looks like reordering the footer.html and Doxygen settings made the treeview show up again. Black magic.

azrogers:
For reference, the icon pack I chose here is [Mingcute](https://icon-sets.iconify.design/mingcute/), chosen because it's Apache licensed, meaning we can use it without attribution, and it has a good mix of having a lot of icons to choose from, and not being as big as some (the material-symbols icon set is almost 7MB!). If anyone has any preferences for other icon sets to choose, there's no reason we can't swap them out.

kring:
I made a couple of small changes...

1. CMakeLists.txt was clobbering the previous value of `DOXYGEN_EXAMPLE_PATH`, which broke code samples added with `@snippet`. I just made it append the diagrams path instead.
2. The icons didn't work when loaded from a `file:///` URL, because browser cross-origin rules forbid fetching file from the local file system with `fetch`. I fixed this with an old school trick: renamed the .json to .json.js and loaded it via a script tag instead.

Otherwise, looks great, merging!

Issue #983, opened 2024-11-08
Fix Documentation CI task


j9liu:
Currently the Documentation CI task takes a while because it builds all the vcpkg dependencies unnecessarily. Fixing this will make the steps below faster.

j9liu:
Forgot to include: we should also review Doxygen log and fix problems it reports.

Additionally, we should change CI Documentation task to fail on Doxygen errors so that we avoid them going forward.

kring:
I wanted to mention a cool feature while I'm thinking about it, because I almost forgot about it myself. We have the ability to pull examples from tests and include them in the generated docs. Which is pretty nice because it allows us to make sure the examples compile and even function correctly. See the docs for the `Plane` constructors for an example:
https://cesium.com/learn/cesium-native/ref-doc/classCesiumGeometry_1_1Plane.html#a80cb61ea891fe39d6468eca8dbb67291

azrogers:
Split off fixing issues in the Doxygen log into issue #1010 to better keep track.

kring:
Fixed in #1011.

PR #1011, opened 2024-11-25
Use vcpkg cache for Doxygen CI task

azrogers:
Currently, the Doxygen CI task doesn't cache the vcpkg artifacts that are built when configuring the CMake project. This means it rebuilds every single one every time the task is run! We can significantly reduce the time it takes to build documentation by simply adopting the same vcpkg cache that the other CI tasks use. This brings the time it takes to run the Documentation CI task down from [12m 17s](https://github.com/CesiumGS/cesium-native/actions/runs/12005187635/job/33461379227) to [only 32s](https://github.com/CesiumGS/cesium-native/actions/runs/12019704527/job/33507514252)!

In theory, as mentioned in #983, we shouldn't need to build the vcpkg dependencies at all to generate the documentation. But this is an easy first step to reduce the time it takes to run CI without needing to mess around with our CMakeLists.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,6 +17,10 @@ jobs:
       - name: Install Doxygen
         run: |
           sudo apt install -y doxygen
+      - name: Check out repository code
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
       - name: ccache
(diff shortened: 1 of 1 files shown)


kring:
Nice, thanks @azrogers!

PR #1012, opened 2024-11-26
Add documentation for AsyncSystem and for our multi-threading policy

### LINKED ISSUE #329: Write documentation for cesium-native's `AsyncSystem`
Once the [clipping PR](#326) goes into main, the `./doc/threading.md` document will have been deleted. Was it out-of-date or was the removal otherwise necessary for some reason? For reference here is a permalink to the original `threading.md`: https://github.com/CesiumGS/cesium-native/blob/580beec2190755b9e5871d3d70d6d9f2b71f79e8/doc/threading.md

kring:
Fixes #329

There are three new pages:

* Multithreading
  * Thread Safety Rules
  * AsyncSystem

I also added the C++ Style Guide to the generated doc.

Other changes:

* Switched CI to use the latest version of Doxygen (v1.12.0) instead of whatever Ubuntu includes (1.9ish I think).
* Fixed a bug in `thenPassThrough` that caused a compiler error when given a value by r-value refrence. (I happened to run into this while writing examples for the doc)


doc/async-system.md
```diff
@@ -0,0 +1,209 @@
...
+
+Unfortunately, there is inevitably at least a little bit of tile loading work that can't be done in a background thread, though. Game engines, for example, usually have strict rules against creating game objects anywhere other than the main thread. We can't get anything onto the screen without creating game objects, so after the background work is complete, we then need to continue processing this tile in the main thread in order to do the final preparation to render it.
+
+`AsyncSystem` gives us an elegant way to express this kind of sequential process involving a series of asynchronous steps.
+
+## Creating an AsyncSystem {#creating-an-asyncsystem}
+
+Most applications have a single `AsyncSystem` that is used throughout. It is constructed from an [ITaskProcessor](@ref CesiumAsync::ITaskProcessor) instance, which is a simple interface used to perform some work in a background thread. The simplest possible implementation looks like this:
+
+\snippet{trimleft} ExamplesAsyncSystem.cpp simplest-task-processor
```
azrogers, line 28: This is really cool! I had no idea you could do this.

doc/async-system.md
```diff
@@ -0,0 +1,209 @@
...
+
+\snippet{trimleft} ExamplesAsyncSystem.cpp capture-by-value
+
+However, it is essential that the last `AsyncSystem` instance be destroyed only after everything is done using it. So a common pattern is to create and store it as a static local in an accessor function:
+
+\snippet{trimleft} ExamplesAsyncSystem.cpp async-system-singleton
+
+## Future<T> {#future}
+
+[Future<T>](@ref CesiumAsync::Future) represents an asynchronous operation that will eventually complete and produce a value of type `T`. If you're familiar with a JavaScript [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise), the concept is very much the same. We commonly see `Future<T>` as the return value of a method that starts an asynchronous operation. For example, we can use [IAssetAccessor](@ref CesiumAsync::IAssetAccessor) to asynchronously download the content of a web URL:
```
azrogers, line 46: I was going to leave a comment on how it might be useful to bring up the comparison to Promises, and then I read a bit more 😅

doc/async-system.md
```diff
@@ -0,0 +1,209 @@
...
+* _Resolved_: The operation has completed successfully and the result value is available.
+* _Rejected_: The operation failed, and an exception describing the error is available.
+
+If we want to block the current thread and wait for the request to complete, we can call [waitInMainThread](@ref CesiumAsync::Future::waitInMainThread):
+
+\snippet{trimleft} ExamplesAsyncSystem.cpp wait-in-main-thread
+
+`waitInMainThread` blocks until the `Future` completes. If it is resolved, `waitInMainThread` returns its value. If it is rejected, `waitInMainThread` throws an exception representing the error.
+
+> [!note]
```
azrogers, line 61: I don't think you need the arrow bracket for a Doxygen `\note`. [According to the documentation](https://www.doxygen.nl/manual/commands.html#cmdpar), you should be able to write this as:
```\note
The `std::move` is needed because `Future<T>` is meant to be consumed just once. This allows for best efficiency because the value resulting from the asynchronous operation can be provided to the caller without ever making a copy of it.
```

As it is, this renders as a blockquote that's prefixed with the text `[!note]`, which I don't think is the intention.

azrogers:
Looks good @kring! The AsyncSystem documentation will be especially helpful. Left a few minor comments when you have a chance to look at them.

doc/async-system.md
```diff
@@ -0,0 +1,209 @@
...
+1. Without ever blocking any threads.
+2. Without writing any complicated thread synchronization code, such as with `std::mutex`.
+
+## Catch {#catch}
+
+The `then...` family of functions register continuations that are invoked when a `Future` resolves (completes successfully). What if the `Future` rejects (completes with an exception) instead?
+
+We can register continuations for the rejection case by calling the [catchInMainThread](@ref CesiumAsync::Future::catchInMainThread) and [catchImmediately](@ref CesiumAsync::Future::catchImmediately) methods instead.
+
+> [!note]
```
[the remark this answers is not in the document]
kring: I think you might be using an old version of Doxygen. It should look like this:
image

doc/thread-safety-rules.md
```diff
@@ -0,0 +1,68 @@
...
+
+You must also consider the thread safety of any arguments passed to the static method. If any of them are shared between threads, then that usage must also be safe according to these rules.
+
+##### Implementation Notes
+
+A static method often operates only on the arguments passed to it. As long as those arguments are not shared between threads (or are safe to use in this way), the static method itself will be safe to use from multiple threads simultaneously without any particular effort from the implementer.
+
+When the static method accesses static fields, global variables, or singletons, more care must be taken by the implementer in order to satisfy this rule. If there is any possibility of the shared data changing (i.e., it's not completely immutable), a synchronization primitive such as a `std::mutex` must be used to serialize access to the shared data.
+
+### A single instance may be used concurrently as long as _all_ accesses are `const`.
```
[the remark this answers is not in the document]
kring: Another old version problem?
image

doc/CMakeLists.txt
```diff
@@ -72,6 +72,10 @@ if(DOXYGEN_FOUND)
     doxygen_add_docs(
         cesium-native-docs
         "../README.md"
+        "./style-guide.md"
+        "./multithreading.md"
+        "./thread-safety-rules.md"
+        "./async-system.md"
```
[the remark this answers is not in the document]
kring: Good idea. Maybe `topics`?

kring:
Thanks for the review @azrogers! I think I've addressed all of your comments.

azrogers:
You were right about the Doxygen version - upon updating, all the remaining comments I had are resolved! However, that Doxygen update seems to be [causing an issue on CI](https://github.com/CesiumGS/cesium-native/actions/runs/12126654314/job/33809205106?pr=1012). It looks like it's not always fetching the complete archive from Doxygen's website. Maybe there's some sort of rate limiting going on? I've changed the URL to the GitHub release instead, hopefully this resolves the issue.

doc/topics/async-system.md
```diff
@@ -0,0 +1,209 @@
...
+
+The `AsyncSystem` can then be created as follows:
+
+\snippet{trimleft} ExamplesAsyncSystem.cpp create-async-system
+
+`AsyncSystem` instances can be safely and efficiently stored and copied by value. This makes it easy to make them available wherever they're needed, including in lambda captures. You can think of an `AsyncSystem` instance as acting a bit like a smart pointer.
+
+\snippet{trimleft} ExamplesAsyncSystem.cpp capture-by-value
+
+However, it is essential that the last `AsyncSystem` instance be destroyed only after everything is done using it. So a common pattern is to create and store it as a static local in an accessor function:
```
timoore, line 40: "using it." Using what? Any instance of the AsyncSystem, or a particular one?

doc/topics/async-system.md
```diff
@@ -0,0 +1,209 @@
...
+
+\snippet{trimleft} ExamplesAsyncSystem.cpp create-request-future
+
+This [get](@ref CesiumAsync::IAssetAccessor::get) function will start the HTTP request, but it will return quickly, without waiting for the request to complete. Initially, the `Future` is in the _Pending_ state, meaning that the asynchronous operation has not yet completed. Once it does complete, the `Future` will be in one of these two states:
+
+* _Resolved_: The operation has completed successfully and the result value is available.
+* _Rejected_: The operation failed, and an exception describing the error is available.
+
+If we want to block the current thread and wait for the request to complete, we can call [waitInMainThread](@ref CesiumAsync::Future::waitInMainThread):
```
timoore, line 55: Do you mean `wait()` instead of `waitInMainThread()`? The concept of the main thread has only been touched on so far, and it seems you're talking here about the simplest way to wait on a `Future`'s results. And then there's the confusing subject of `waitInMainThread()` meaning that it can only be called in the main thread... :smiling_imp:

doc/topics/async-system.md
```diff
@@ -0,0 +1,209 @@
...
+
+* [thenInMainThread](@ref CesiumAsync::Future::thenInMainThread): The continuation function is invoked in the "main" thread. Many applications have a clear idea of which thread they consider the main one, but in Cesium Native, "main thread" very specifically means the thread that called `waitInMainThread` or [dispatchMainThreadTasks](@ref CesiumAsync::AsyncSystem::dispatchMainThreadTasks) on the `AsyncSystem`. This need not be the same as the thread your application considers to be the main one.
+* [thenInWorkerThread](@ref CesiumAsync::Future::thenInWorkerThread): The continuation function is invoked in a background worker thread by calling [startTask](@ref CesiumAsync::ITaskProcessor::startTask) on the `ITaskProcessor` instance with which the `AsyncSystem` was constructed.
+* [thenInThreadPool](@ref CesiumAsync::Future::thenInThreadPool): The continuation function is invoked in a specified [ThreadPool](@ref CesiumAsync::ThreadPool).
```
timoore, line 72: While it's good to be complete, thread pools have not been mentioned yet. A sentence or two about thread pools would be good in the section on `AsyncSystem`.

doc/topics/async-system.md
```diff
@@ -0,0 +1,209 @@
...
+* [thenImmediately](@ref CesiumAsync::Future::thenImmediately): The continuation function is invoked immediately in whatever thread happened to resolve the `Future`. This is only appropriate for work that completes very quickly and is safe to run in any thread.
```
timoore, line 73: "This is only appropriate for work that completes very quickly..." is that actually true if the future is known to be resolved in a worker thread e.g. we've written `thenInWorkerThread()` in the continuation chain?

doc/topics/async-system.md
```diff
@@ -0,0 +1,209 @@
...
+
+Continuing with our web request example above, we can register a continuation that receives the completed request and does something with the downloaded data in the main thread:
+
+\snippet{trimleft} ExamplesAsyncSystem.cpp continuation
+
+So far, this all looks like a glorified callback system. The real power of `AsyncSystem` becomes apparent when we chain continuations together. Here's what that might look like:
+
+\snippet{trimleft} ExamplesAsyncSystem.cpp chaining
+
+In this example, we call `get` to start a network request, as before. We immediately call `thenInWorkerThread` on the returned `Future` in order to register a continuation for when the request completes. This continuation runs in a background thread and calls a made-up function for this example, `processDownloadedContent`. We can imagine that it's doing some intensive processing, and we wouldn't want to block the main thread while it's doing it.
```
timoore, line 83: I don't think you need to call out functions as "made-up" as all the code is a big example.

doc/topics/async-system.md
```diff
@@ -0,0 +1,209 @@
...
+
+We could change `myComputeSomethingSlowlyWrapper` to catch the possible exception and turn it into a call to `reject` on the `Promise<T>`, and that would work fine. But there's an easier way. We can use the [createFuture](@ref CesiumAsync::AsyncSystem::createFuture) method on `AsyncSystem`:
+
+\snippet{trimleft} ExamplesAsyncSystem.cpp compute-something-slowly-wrapper-handle-exception
+
+The `createFuture` method takes a function as its only parameter, and it _immediately_ invokes that function (that's why it's safe to capture everything by reference `&` in this case). The function receives a parameter of type `Promise<T>`, which is identical to the one that would be created by `createPromise`. Just like before, we call the library function that we're wrapping and resolve the `Promise<T>` when its callback is invoked.
```
timoore, line 187: If I understand correctly, the function called by `createFuture()` need to explicitly schedule work with e.g. `runInWorkerThread` if it could block. If that's true, then I think it needs to be mentioned here.

timoore (commented):
Just comments on async-system.md.

doc/topics/async-system.md
```diff
@@ -0,0 +1,209 @@
...
+
+\snippet{trimleft} ExamplesAsyncSystem.cpp simplest-task-processor
+
+This implementation will work well, but it isn't very efficient because a brand new thread is created for each task. Most applications will implement this interface using a thread pool, task graph, or similar functionality that their application already contains.
+
+The `AsyncSystem` can then be created as follows:
+
+\snippet{trimleft} ExamplesAsyncSystem.cpp create-async-system
+
+`AsyncSystem` instances can be safely and efficiently stored and copied by value. This makes it easy to make them available wherever they're needed, including in lambda captures. You can think of an `AsyncSystem` instance as acting a bit like a smart pointer.
```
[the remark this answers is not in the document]
kring: I don't think "facade" is an accurate description of what's happening here. It's not like AsyncSystem is providing an alternate/simplified interface to a complex system. It's that an _instance_ of `AsyncSystem` conceptually acts like a _reference_. If I do this:

```cpp
AsyncSystem asyncSystem = getAsyncSystem();
AsyncSystem copy = asyncSystem;
```

I don't have two independent async systems now. I have one system and two references to it. It's exactly the same as doing this:

```cpp
std::shared_ptr<IAssetAccessor> pAssetAccessor = getAssetAccessor();
std::shared_ptr<IAssetAccessor> pCopy = pAssetAccessor;
```

And that's why I tried to describe it as being smart-pointer-like.

Probably the difficulty in explaining this comes from the fact that it's a bad design to begin with. We should probably just pass AsyncSystem around with shared_ptr just like we do with the other things in TilesetExternals. It would be a bit more verbose, but easier to explain. And less in need of explanation because it would behave the way people expect.

I'm not sure I can justify making that change now, though. So I'll just try to make the explanation here clearer.

doc/topics/async-system.md
kring: I've improved the wording here as well in the part about that previously talked about smart pointers. See what you think.

doc/topics/async-system.md
kring: Good idea. I changed to `wait`, but also added a brief note about its potential for deadlocking and the fact that you should use `waitInMainThread` if you need to wait in the main thread. By the way, it's not so much that `waitInMainThread` must be called in the main thread, more that it _defines_ the main thread, and that definition better not conflict with any other (such as another thread calling `dispatchMainThreadTasks`) or you'll have a problem.

doc/topics/async-system.md
kring: I added some more explanation to this dot point. Thread pools are actually very rarely used in Cesium Native, and almost not worth mentioning. But it also feels wrong to remove this completely. So I instead tried to briefly explanation the somewhat unusual situation where they are useful.

doc/topics/async-system.md
kring: Yes, even then. `thenInWorkerThread(...).thenImmediately(...)` is identical to `thenInWorkerThread(...).thenInWorkerThread(...)` in terms of where the second continuation runs. But the latter is clearer and won't accidentally switch to running in a different thread if someone inserts a `thenInMainThread` in between the two.

doc/topics/async-system.md
kring: I thought it was important to distinguish real parts of the Cesium Native API from functions made up for the purpose of the example. I didn't have this initially, but upon re-reading it with fresh eyes I thought it wasn't very obvious, so I tried to make it more so. Do you feel that referring to this function as "made up" detracts from the readability significantly? If not, I'm inclined to leave it.

doc/topics/async-system.md
kring: Well... yes and no. `createFuture` calls its callback immediately in the same thread (I updated the text to say this), so if you plan to do heavy CPU work, you certainly shouldn't do it there. In that scenario, you should call `runInWorkerThread` _instead_, no need for `createFuture` at all.

However, if you're planning to block (rather than just do heavy CPU work)... well, maybe don't do that? Blocking is bad, even if it's a worker thread you're blocking. It's usually possible to avoid. But if you truly must block, then that's another case for using `runInWorkerThread` and there's no need to use `createFuture`.

The real purpose of `createFuture`, though, is to interface with async APIs that use a different mechanism for notifying of completion, usually a callback. That's the example shown here. In that scenario, the async API probably already avoids doing heavy work in the calling thread, so there's no reason to start a worker thread for it.

doc/topics/async-system.md
```diff
@@ -0,0 +1,209 @@
...
+
+The important difference, however, is that if `computeSomethingSlowly` throws an exception, `createFuture` will automatically catch that exception and turn it into a rejection of the `Future<T>`.
+
+## SharedFuture<T> {#shared-future}
+
+A `Future<T>` may have at most one continuation. It's not possible to arrange for two continuation functions to be called when a given `Future<T>` resolves. This policy makes `Future<T>` more efficient by minimizing bookkeeping. In the majority of cases, this is sufficient. But not always.
+
+In those cases where we do want to be able to attach multiple continuations, we can convert a `Future<T>` into a [SharedFuture<T>](@ref CesiumAsync::SharedFuture) by calling the [share](@ref CesiumAsync::Future::share) method on it. The returned `SharedFuture<T>` instance works a bit like a smart pointer. It's safe to copy, and all copies are logically identical to the original. Just like `Future<T>`, we can call `then...` and `catch...` methods to attach continuations. The difference is that we can do this multiple times, and all attached continuations will be invoked when the future completes.
```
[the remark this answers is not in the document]
kring: It can be called multiple times on one copy. I updated the text to make this a bit clearer.

kring:
Thanks for the review, @timoore! I think I've addressed all your comments. Unfortunately, I rather incompetently forgot that I was working in main because this branch was already merged, and so I pushed my changes directly to main. Since I only changed documentation and examples, I think I'll share the diff rather than reverting:
https://github.com/CesiumGS/cesium-native/compare/624f5b29c15d66777dd8100c10b4db4d2ca3d93e...7fd670979011dcfc7d09223294cc3febd3152dbe

doc/topics/async-system.md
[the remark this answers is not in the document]
timoore: Why is the distinction important to contributors to Cesium Native, as opposed to "integrators" or implementors of ports? Even calling it a "smart reference" seems like to much detail at this level, and the detail that the underlying implementation object doesn't get destroyed until the last reference is gone belongs in an implementor's guide, not here IMHO.

doc/topics/async-system.md
timoore: Do you mean "the former is clearer?"  If these two forms are identical, then what's the way to schedule new, expensive work?

doc/topics/async-system.md
timoore: The real functions have Doxygen links, or they should, right? Even if not, it seems quite clear that `processDownloadedContent` comes from the example, and I find tagging it as "made-up" distracting.

doc/topics/async-system.md
timoore: > Well... yes and no. `createFuture` calls its callback immediately in the same thread (I updated the text to say this), so if you plan to do heavy CPU work, you certainly shouldn't do it there. In that scenario, you should call `runInWorkerThread` _instead_, no need for `createFuture` at all.
> 
> However, if you're planning to block (rather than just do heavy CPU work)... well, maybe don't do that? Blocking is bad, even if it's a worker thread you're blocking. It's usually possible to avoid. But if you truly must block, then that's another case for using `runInWorkerThread` and there's no need to use `createFuture`.
> 

Too bad I didn't have this guide when I wrote UrlAccessor :sunglasses:  I understand now that you can just return the result of `runInWorkerThread` without wrapping that call in `createFuture`.

doc/topics/async-system.md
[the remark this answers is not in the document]
kring: Sorry I'm not quite sure what you're getting at here Tim. IMO this doc is useful for both integrators (people "using" cesium-native) and for contributors (people "developing" cesium-native). It even says so in the second sentence of this doc. 😄 

I would also assert that the copy semantics of the AsyncSystem class are essential to understand for both audiences. Passing around AsyncSystems comes up a _lot_ both within cesium-native and when using it. If you don't understand the copy semantics, you might think it's ok to, for example, capture it by reference, when that is dangerous and mostly unnecessary. Or you might think you need to wrap it up in a `std::make_shared` or something like that, when there's no need at all.

It's true that the note about "the underlying implementation object doesn't get destroyed until the last reference is gone" is mostly only important for people integrating cesium-native into their own applications. Well... except maybe it comes up a bit when writing tests within cesium-native, too. But in any case, this doc doesn't need to be _exclusively_ for contributors even if contributors are our focus right now, right?

doc/topics/async-system.md
kring: No, I meant that `thenInWorkerThread(...).thenInWorkerThread(...)` is the clearer of the two, because it makes it clear that both lambdas are run in a worker thread (and not, for example, taking up time on the main thread).

> If these two forms are identical, then what's the way to schedule new, expensive work?

This _is_ the way to schedule new, expensive work, despite the fact that they're identical. Maybe I'm not understanding your concern?

Ok, so let me try to explain the nuance here... When a Future resolves in a worker thread (for example, the continuation passed to the first `thenInWorkerThread` in our current example returns), and its continuation is registered with `thenInWorkerThread`, that continuation will run _immediately_ in the same worker thread that just resolved the Future. Exactly the same as if that continuation were registered with `thenImmediately`.

However, if the Future resolved in some other thread (not a worker thread), and its continuation is registered with `thenInWorkerThread`, then the continuation will be _scheduled_ to run in a worker thread. It will _not_ run immediately.

This is an optimization, to avoid the overhead of repeatedly scheduling work that should run in the same context without requiring us to use explicit `thenImmediately` calls.

`thenImmediately` is error prone because we don't know for sure which thread resolves a future in all cases, so it's dangerous to make assumptions about it. For example, if I call `IAssetAccessor::get`, what thread resolves that Future? In Unreal, it happens to be in the main thread. In other applications, it might be a worker or an HTTP thread or something like that. We can't know for sure. But by simply using the correct `then..` for the context I want my continuation to run, I get the most efficient performance if that place happens to be the same context the Future was resolved. If I truly don't care where it runs - because it's quick and thread agnostic - and want to minimize scheduling overhead in all cases, _then_ it is appropriate to use `thenImmediately`.

The only downside I can think of to this has to do with prioritization. If I have `thenInWorkerThread(...).thenInWorkerThread(...)`, should other work that was scheduled earlier get to run in between these two worker thread lambdas? With the current implementation, it never will. I don't think this is a big deal. An early version of AsyncSystem gave users control over this by calling something like `thenInWorkerThread` versus `thenImmediatelyInWorkerThread`, but it was just a lot of unnecessary complexity IMO and that's why it's gone now.

doc/topics/async-system.md
kring: The pattern I've been following - inherited from standard practice at AGI 15 years ago! - is to make the first mention of a type a link, and successive mentions just `monospaced`. So the lack of a link doesn't clearly identify fake types and functions. As long as I'm in an AGI state of mind, the approach we used to use for made-up functions in documentation, as suggested by the late, great Ralph Patrick, was to name them things like `startAnyOldOperationThatMightFail` or `performAnyOldProcessingOnDownloadedContent`.
https://help.agi.com/STKComponents/html/AccessQueries.htm

I guess you're right that the potential for confusion isn't too high. So I'll remove "made-up" bits. The first time a user asks why they can't find the `startOperationThatMightFail` or `findReferenceImageUrl` function, though, I'm blaming you. 😜

doc/topics/async-system.md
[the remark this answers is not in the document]
timoore: I'll withdraw my concerns about "Contributors" vs "Implementors"; you're totally correct that this document should be useful to both. But I think this section could still be clearer, and while I think there's a place in this document for a sample implementation, putting it right in the introduction to AsyncSystem creates more questions than perhaps it answers:
- Why copy `AsyncSystem` objects around when they can just be created? Well, Cesium Native can't create one; it's passed in on startup more-or-less and then passed around.
- Why does the creation function return a reference if `AsyncSystem` has copy semantics?

Instead of sending endless criticisms your way, I think I need to actually do my Editor job and write what I think would  flow better here.

doc/topics/async-system.md
timoore: I think I understand now. I was fixated on a chain of continuation functions where you know explicitly what kind of thread will resolve the continuation because... you just wrote `thenInWorkerThread` or `thenInMainThread`. Like you say, if the future is returned from a function, or if the surrounding code changes in the future, you just wouldn't know.

doc/topics/async-system.md
timoore: Yes, you can send them my email 😎

PR #1013, opened 2024-11-26
Fix (some) warnings in generated documentation

### LINKED ISSUE #1010: Fix warnings reported by Doxygen
Split from #983. Doxygen currently produces a bunch of warnings when we generate the documentation. Some of these are warnings about undocumented classes and members, which we will likely tackle gradually. But some of these warnings are for issues that will result in broken documentation:
```
CesiumGltf/include/CesiumGltf/PropertyView.h:824: warning: @copybrief or @copydoc target 'PropertyView::scale' not found
CesiumRasterOverlays/include/CesiumRasterOverlays/RasterOverlay.h:215: warning: unable to resolve link to 'CreditSystem' for \link command
CesiumUtility/include/CesiumUtility/ErrorList.h:62: warning: argument 'error' of command @param is not found in the argument list of CesiumUtility::ErrorList::emplaceWarning(WarningStr &&warning)
```

Fixing these issues will improve the usability of our documentation.

azrogers:
When generating our documentation, Doxygen produces a whole host of warning messages. As noted in #1010, fixing these warnings will improve the usability of our documentation. Some highlights:

- All API macros are now added to the Doxygen `PREDEFINED` list, resolving issues where Doxygen wouldn't be able to parse a type because the API macro was undefined. `CESIUM_DEFAULT_ELLIPSOID` has also been properly defined, meaning the default parameter will correctly show in the documentation.
- CesiumGltfContent is now included in the documentation.
- All previously undocumented parameters are now documented.
- Almost all "unable to resolve link" warnings have been fixed. Unaddressed here, besides the PropertyTable warnings mentioned below, are four warnings: two in ApplicationData.h and RasterOverlayLoadFailureDetails.h relating to declaration order, one in RasterOverlayTile.h because Cesium3DTilesSelection isn't included by CesiumRasterOverlays, and one in RasterOverlayUtilities.h that references a `RasterOverlay::getTile` method that no longer exists, and that I'm not sure of the current equivalent.

Besides warnings about undocumented members, most of the warnings remaining are related to `PropertyView`, `PropertyArrayView`, and `PropertyTablePropertyView`. The heavy use of templates here seems to be confusing Doxygen and many of its attempts to resolve links here. Doxygen's docs recommend using the `CLANG_ASSISTED_PARSING` option to deal with heavily templated code. However, I haven't gone down this path as it would significantly slow down our documentation generation and \it would mean that the documentation can only be correctly generated when using a CMake generator that produces a compile_commands.json. 

Considering the number of remaining warnings, this doesn't resolve #1010 itself. But it's a step in the right direction.

### DIFF
--- a/Cesium3DTiles/include/Cesium3DTiles/MetadataQuery.h
+++ b/Cesium3DTiles/include/Cesium3DTiles/MetadataQuery.h
@@ -38,7 +38,7 @@ struct CESIUM3DTILES_API FoundMetadataProperty {
 
   /**
    * @brief A reference to the {@link ClassProperty} describing the found
-   * property within the {@lnik Schema}.
+   * property within the {@link Schema}.
    */
   const ClassProperty& propertyDefinition;
 
--- a/Cesium3DTilesReader/include/Cesium3DTilesReader/SubtreeFileReader.h
+++ b/Cesium3DTilesReader/include/Cesium3DTilesReader/SubtreeFileReader.h
@@ -11,7 +11,7 @@ namespace Cesium3DTilesReader {
 /**
  * @brief Reads 3D Tiles subtrees from a binary or JSON subtree file.
  *
- * While {@link SubtreeReader} can parse a {@link Subtree} from a binary buffer
+ * While {@link SubtreeReader} can parse a {@link Cesium3DTiles::Subtree} from a binary buffer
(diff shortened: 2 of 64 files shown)


kring:
I noticed the generated include directories weren't in the `LIB_DIRS` list, so I added those.
Nice improvement here, thanks @azrogers!

PR #1017, opened 2024-11-30
Switch to macos-13 runner, target macOS 10.15 for vcpkg dependencies

kring:
This fixes the extremely long macOS build times (when ccache doesn't eliminate most of the compiling) by switching from the macos-15 to the macos-13 runner. The older runner has double the memory, which is why it is so much faster. We were running out of memory and hitting swap on the macos-15 runner.

This PR also configures vcpkg to target macOS 10.15, which is what we target in Cesium for Unreal and Cesium for Unity. Previously, we were targetting 10.15 for cesium-native itself, but all of the vcpkg dependencies targeted whatever the host system was running (15 or, now, 13). The discrepancy caused a large number of linker warnings.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -126,13 +126,13 @@ jobs:
       fail-fast: false
       matrix:
         compiler: [ gcc, clang ]
-        platform: [ ubuntu-22.04, ubuntu-20.04, macos-15 ]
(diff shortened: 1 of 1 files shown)


kring:
I'm going to merge this because it's just a straightforward CI change and it cuts our macOS CI time in half or better.

Issue #974, opened 2024-10-31
Provide a means to navigate to a location on the globe and get its coordinates


LisaBosCesium:
Many AEC users of Cesium Native plugins have workflows that start with finding the lat long for the location of their scene, and then pasting the coordinates into plugin fields in the editor. They usually get the coordinates in an external application like Google Earth or a CesiumJS sandcastle. 

How can we make this easier and possible with our plugins?

kring:
Related issues:
* #70
* CesiumGS/cesium-unreal#99

Issue #70, opened 2020-12-08
Cesium ion geocoding API


kring:
Add a simple class that can be used to query the Cesium ion geocoder.

azrogers:
Completed in #1019.

PR #1019, opened 2024-12-02
Add Cesium ion geocoder to CesiumIonClient::Connection

azrogers:
As described in #974, we want to make it easier for users to find the coordinates of locations of interest in our integrations, as an alternative to looking up the coordinates through some other service and entering them manually. The first step of this is completing #70, an implementation of the Cesium ion geocoder API, which is what this PR is for. It adds a `geocode` method to `CesiumIonClient::Connection` modeled on the CesiumJS geocoder API. This isn't a complete geocoder implementation supporting multiple services, but it will allow us to implement CesiumGS/cesium-unreal#99 to improve Cesium for Unreal's ease of use.


CesiumIonClient/include/CesiumIonClient/Geocoder.h
```diff
@@ -0,0 +1,131 @@
...
+
+  /**
+   * @brief The region on the globe for this feature.
+   */
+  std::variant<CesiumGeospatial::GlobeRectangle, CesiumGeospatial::Cartographic>
+      destination;
+
+  GeocoderFeature(
+      std::string& displayName_,
+      CesiumGeospatial::GlobeRectangle& destination_)
```
kring, line 71:
```suggestion
      const std::string& displayName_,
      const CesiumGeospatial::GlobeRectangle& destination_)
```

CesiumIonClient/include/CesiumIonClient/Geocoder.h
```diff
@@ -0,0 +1,131 @@
...
+      : displayName(displayName_), destination(destination_) {}
+  GeocoderFeature(
+      std::string& displayName_,
+      CesiumGeospatial::Cartographic& destination_)
```
kring, line 75:
```suggestion
      const std::string& displayName_,
      const CesiumGeospatial::Cartographic& destination_)
```

kring: Alternatively, you could do this:

```cpp
GeocoderFeature(
      std::string displayName_,
      const CesiumGeospatial::Cartographic& destination_)
      : displayName(std::move(displayName_)), destination(destination_) {}
```

Which would be slightly more efficient because you can optionally move the string into the constructor (moving a GlobeRectangle isn't any different from copying it).

Slightly more efficient still would be two constructors, but that's almost certainly not worth the noise:

```cpp
GeocoderFeature(
      std::string&& displayName_,
      const CesiumGeospatial::Cartographic& destination_)
      : displayName(std::move(displayName_)), destination(destination_) {}

GeocoderFeature(
      const std::string& displayName_,
      const CesiumGeospatial::Cartographic& destination_)
      : displayName(displayName_), destination(destination_) {}
```

CesiumIonClient/include/CesiumIonClient/Geocoder.h
```diff
@@ -0,0 +1,131 @@
...
+   */
+  std::string html;
+
+  /**
+   * @brief If true, the credit should be visible in the main credit container.
+   * Otherwise, it can appear in a popover.
+   */
+  bool showOnScreen;
+
+  GeocoderAttribution(std::string& html_, bool showOnScreen_)
```
kring, line 112:
```suggestion
  GeocoderAttribution(const std::string& html_, bool showOnScreen_)
```

CesiumIonClient/include/CesiumIonClient/Geocoder.h
kring: Another possibility here is to just get rid of the constructors entirely for these simple value-holding types. Users could still create instances with almost the exact same syntax, except they'd have to use curly braces.

CesiumIonClient/include/CesiumIonClient/Connection.h
```diff
@@ -291,6 +292,21 @@ class CESIUMASYNC_API Connection {
...
+   * A geocoding service is used to make a plain text query (like an address,
+   * city name, or landmark) and obtain information about where it's located.
+   *
+   * @param provider The ion geocoding provider to use.
+   * @param type The type of request to make. See {@link GeocoderRequestType} for more information.
+   * @param query The query to make.
+   */
+  CesiumAsync::Future<Response<GeocoderResult>> geocode(
+      const GeocoderProviderType& provider,
+      const GeocoderRequestType& type,
```
kring, line 307: Generally we pass enums by value instead of const reference, cause they're just integers.

CesiumIonClient/src/Connection.cpp
```diff
@@ -1,10 +1,15 @@
 #include "CesiumIonClient/Connection.h"
 
+#include "CesiumGeospatial/BoundingRegion.h"
+#include "CesiumGeospatial/Cartographic.h"
+#include "CesiumGeospatial/GlobeRectangle.h"
+#include "CesiumIonClient/Geocoder.h"
```
kring, line 6: This file is inconsistent to begin with, but we should move toward `#include <CesiumHeader/Whatever.h>` instead of `#include "CesiumHeader/Whatever.h"`.

CesiumIonClient/src/Connection.cpp
```diff
@@ -582,6 +588,85 @@ Defaults defaultsFromJson(const rapidjson::Document& json) {
   return defaults;
 }
 
+GeocoderResult geocoderResultFromJson(const rapidjson::Document& json) {
+  GeocoderResult result;
+
+  auto featuresMemberIt = json.FindMember("features");
+  if (featuresMemberIt != json.MemberEnd() &&
+      featuresMemberIt->value.IsArray()) {
+    auto featuresIt = featuresMemberIt->value.GetArray();
```
kring, line 597: This isn't an iterator, is it?

CesiumIonClient/src/Connection.cpp
```diff
@@ -582,6 +588,85 @@ Defaults defaultsFromJson(const rapidjson::Document& json) {
...
+    for (auto& feature : featuresIt) {
+      const rapidjson::Value* pLabel =
+          rapidjson::Pointer("/properties/label").Get(feature);
```
kring, line 600: Not a huge deal, but it'd be nice to move the `rapidjson::Pointer` construction outside the loop. No reason to do it repeatedly.

CesiumIonClient/src/Connection.cpp
```diff
@@ -582,6 +588,85 @@ Defaults defaultsFromJson(const rapidjson::Document& json) {
...
+      if (!pLabel) {
+        SPDLOG_WARN("Missing label for geocoder feature");
+        continue;
+      }
+
+      std::string label(pLabel->GetString());
```
kring, line 606: What if `properties.label` isn't a string? It'd be safer to either check `IsString`, or use `JsonHelpers::getStringOrDefault` (which will do this for you).

CesiumIonClient/src/Connection.cpp
```diff
@@ -582,6 +588,85 @@ Defaults defaultsFromJson(const rapidjson::Document& json) {
...
+
+  auto featuresMemberIt = json.FindMember("features");
+  if (featuresMemberIt != json.MemberEnd() &&
+      featuresMemberIt->value.IsArray()) {
+    auto featuresIt = featuresMemberIt->value.GetArray();
+    for (auto& feature : featuresIt) {
+      const rapidjson::Value* pLabel =
+          rapidjson::Pointer("/properties/label").Get(feature);
+      if (!pLabel) {
+        SPDLOG_WARN("Missing label for geocoder feature");
```
kring, line 602: We should generally avoid logging this sort of thing directly, because it's not accessible programmatically and in some cases it might be viewed as spam. The pattern used elsewhere in this file (use the `OrDefault` methods to silently fill in a default if the JSON doesn't match what we expect) isn't perfect, but better to stick with that for consistency.

CesiumIonClient/src/Connection.cpp
```diff
@@ -582,6 +588,85 @@ Defaults defaultsFromJson(const rapidjson::Document& json) {
...
+
+  auto attributionMemberIt = json.FindMember("attributions");
+  if (attributionMemberIt != json.MemberEnd() &&
+      attributionMemberIt->value.IsArray()) {
+    const auto& valueJson = attributionMemberIt->value;
+
+    result.attributions.reserve(valueJson.Size());
+
+    for (rapidjson::SizeType i = 0; i < valueJson.Size(); ++i) {
+      const auto& element = valueJson[i];
```
kring, line 659: It's ok to use `auto` for iterators, but this is more clear without:
```suggestion
      const rapidjson::Value& element = valueJson[i];
```

CesiumIonClient/src/Connection.cpp
```diff
@@ -1179,3 +1264,64 @@ Connection::tokens(const std::string& url) const {
...
+    requestUrl = CesiumUtility::Uri::addQuery(requestUrl, "geocoder", "BING");
+    break;
+  case GeocoderProviderType::Google:
+    requestUrl = CesiumUtility::Uri::addQuery(requestUrl, "geocoder", "GOOGLE");
+    break;
+  case GeocoderProviderType::Default:
+    requestUrl =
+        CesiumUtility::Uri::addQuery(requestUrl, "geocoder", "DEFAULT");
+    break;
+  }
```
kring, line 1291: CesiumJS seems to use lowercase names, and leaves the parameter out entirely for `Default`:
https://github.com/CesiumGS/cesium/pull/12299/files#diff-1a5903d86aa48fe1b27e5f0d8e5da57e1afc40ee0e1ee6029fded82c23f5b1aaR28

kring (commented):
Looks good @azrogers! Just some nitpicky comments here.

azrogers:
Request to get this finished from the forums: https://community.cesium.com/t/geocoder-for-unity/37435

azrogers:
@kring Addressed review comments!

azrogers:
Removed the constructors for `GeocoderFeature` and `GeocoderAttribution`, as you're right, they can just be constructed by braces - but this causes an issue on clang with `emplace_back`, which surfaced a hazy memory that that's the reason I added the constructors in the first place. Do you know a way around this error? @kring

kring:
> but this causes an issue on clang with emplace_back

Oh right, I don't think you can use emplace_back to call a brace-initializer "constructor". But if you construct it directly and then let the emplace_back move it into the vector, that works fine. I pushed up a change to do that.

CesiumIonClient/test/TestConnection.cpp
```diff
@@ -1,3 +1,6 @@
+#include "CesiumGeospatial/Cartographic.h"
```
kring, line 1: Might as well fix the include style here.

kring (commented):
Thanks @azrogers, this looks great! Please just merge in main and then this is good to go. One super trivial comment below.

azrogers:
Made a CMake change to use `set(CESIUM_CLANG_TIDY_USE_THREADS ... CACHE ...)` instead of cmake_dependent_option, as apparently `option`/`cmake_dependent_option` only accept bool values, and the result was that clang-tidy was erroring when not specifying `CESIUM_CLANG_TIDY_USE_THREADS` on the command line as it was getting the flag `-jOFF`.

kring:
Thanks @azrogers!

PR #1021, opened 2024-12-03
Overlay tiles with a subset do not have more detail

kring:
With this PR, raster overlay tiles with a `subset` are always deemed to have no more detail available. This is correct because the presence of a subset indicates that the tile we want to load wasn't available, and so we're using the nearest ancestor tile that actually exists. That ancestor is, by construction, the most detailed tile available for this area.

Previously, we would think there was more detail available if the nearest ancestor tile happened to think so originally. For example, consider tile `B` which is a child of tile `A`. Tile `A` loads successfully, but because it's not at the tile provider's "max level", we assume that more detail is available. Then, for another geometry tile, we attempt to load tile `B`. This fails (because the tile doesn't actually exist), so we end up using a subset of `A` instead. But because `A` originally was marked "more detail available", the new geometry tile thinks its raster overlays have more detail available, too. Which can mean we waste time upsampling this geometry tile to hang this phantom extra detail on. In some contexts (such as when zoomed in really close, or when traversing exhaustively), this process can continue until we're working with some very tiny tiles, perhaps causing a crash when we run out of memory.


j9liu:
Thanks @kring ! Seems straightforward to me, so I'll go ahead and merge. But let me know if there's any specific cases I should have tested.

Issue #981, opened 2024-11-08
Document our tile selection algorithm


j9liu:
Our tile selection algorithm is fairly complex, and also different from CesiumJS's selection algorithm. It would be helpful to document its inner workings to improve developer understanding. Some ideas I can think of off the top of my head:

- How does the fundamental traversal algorithm work?
- How do the Cesium3DTileset settings (exposed to runtimes) affect loading?
- Other intricacies?
  - Implicit tileset loading
  - External tilesets

PR #1022, opened 2024-12-04
Add a guide to rendering 3D Tiles with Cesium Native.

### LINKED ISSUE #982: Create architecture diagrams for loading 3D Tiles through Cesium Native
This is copied from a post on our [community forum](https://community.cesium.com/t/feedback-requested-documentation-for-cesium-native-cesium-for-unreal/36317/) soliciting feedback from our users about documentation. 

> While exploring the codebases for these implementations, I spent some time identifying the entry/common point(s) where cesium-native integrates with Unreal and osgEarth. The method I was looking for was [updateView()](https://github.com/CesiumGS/cesium-native/blob/be7cfe27ba93c7f59bedb945a63857602cd0e302/Cesium3DTilesSelection/src/Tileset.cpp#L320). Once I understood how updateView() is called, understanding the responsibilities of cesium-native became clearer.

> More explicit and publicly available architecture design of the intended 3D tile loading system, i.e going from tileset to asset accessor and how individual tiles will be streamed, as this can vastly accelerate the learning curve of any future integration / implementation.

It would be great to develop an illustrated overview of how 3D Tiles are loaded through cesium-native. This should also include references to the interfaces that users must implement to actually do something with the 3D Tiles -- e.g., `IAssetAccessor`, `IPrepareRendererResources` -- at an abstract level, and what functions would need to be called by the client, e.g., `updateView`

kring:
Fixes #982 

Also added a topic for #981 but it's just a placeholder for now.


azrogers:
Looks good to me! Only one minor typo to fix.

azrogers (changes requested):

kring:
Thanks @azrogers, I fixed the typo.

azrogers:
Thanks for the guide @kring!

PR #1026, opened 2024-12-05
build as an installable package and use vcpkg manifest mode to build standalone

timoore:
This PR has several moving parts:

- Package files for cesium-native created on installation. Any hackery needed to create targets for external targets is moved to CMake Find modules that are installed with the package files.
- A single vcpkg overlay port for asycplusplus
- `vcpkg.json` and `vcpkg-configuration.json`
- A CMakePresets.json file demonstrating how to do a manifest mode bulid.

I'm using this branch to build vsgCs using cesium-native as a vcpkg overlay port. That usage needs to be documented here too. Also, while I've tried to preserve the old behavior using exvcpkg, I haven't tested this yet with Cesium for Unreal and other integrations in the classic style of using cesium-native as a CMake subdirectory.


azrogers:
Looks like this causes the following issue in cesium-unreal:
```
  CMake Error at cesium-native/cmake_install.cmake:940 (file):
    file cannot create directory:
    C:/Dev/prs/cesium-native-1026/Plugins/cesium-unreal/extern/../Source/ThirdParty/lib/Windows-AMD64-$<IF:$<CONFIG:Debug>,Debug,Release>/cmake/cesium-native.
    Maybe need administrative privileges.
  Call Stack (most recent call first):
    cmake_install.cmake:37 (include)
```

Not sure why this would have broken here - the cesium-unreal/extern CMakeLists.txt is unchanged here and I don't see why the `CONFIG` generator expression would stop evaluating correctly here.

timoore:
The problem is in the new cesium-native/CMakeLists.txt, which defines exports to install. the install(EXPORT...) command doesn't support generator functions in its destination argument... which makes a certain amount of sense, even if it's inconsistent. The export files won't be used in the Unreal build, but I'll change the expression to use a more suitable variable.

timoore:
I've started to look at how to use this in cesium-unreal and have hit a bit of a sticking point with specifying the MSVC toolset version to build those parts of cesium-unreal/extern that wouldn't be vcpkg port overlays. Maybe this isn't a real issue, because I think the GitHub build action would work the same. It's really only a problem for developers who have multiple toolsets installed. It is possible to set the toolset version through CMakePresets.json, and on the command line, but otherwise it is really obscure. Do you think this is worth worrying about?

Just to be clear, this isn't a new problem that comes with vcpkg manifest mode.

2xsaiko:
I've been trying to package this (Homebrew for now, later Gentoo) so I can use it as a dependency for vsgCs which I want to use in something I'm working on, however right now it seems pretty much impossible to do so without significantly changing the build files. This patchset looks like it could improve that, however with this it fails to even configure in the Homebrew build environment:

> error: In manifest mode, `vcpkg install` does not support individual package arguments.

https://gist.github.com/2xsaiko/9d2bc0691529d2599e2d43d8cf7e8ec5

Strangely enough, this error doesn't occur when running it outside of the Homebrew package. Any ideas?

timoore:
> I've been trying to package this (Homebrew for now, later Gentoo) so I can use it as a dependency for vsgCs which I want to use in something I'm working on, however right now it seems pretty much impossible to do so without significantly changing the build files. This patchset looks like it could improve that, however with this it fails to even configure in the Homebrew build environment:
> 
> > error: In manifest mode, `vcpkg install` does not support individual package arguments.
> 
> https://gist.github.com/2xsaiko/9d2bc0691529d2599e2d43d8cf7e8ec5
> 
> Strangely enough, this error doesn't occur when running it outside of the Homebrew package. Any ideas?

The `vcpkg-build` branch in vsgCs uses a commit to a personal copy of Cesium Native -- sucked in as a `vcpkg` port overlay -- that was the basis of this PR and is very close to it.  I encourage you to give that a try.

In your build that is crashing, it looks like the ezvcpkg code is getting confused by the vcpkg.json mainifest file and is then going into manifset mode. This is interesting as I would like to keep ezvcpkg working for a while, but for experimenting with this PR you probably want to specify CESIUM_USE_EZVCPKG=OFF.

azrogers:
Gave this a shot using vsgCs and I'm getting this build error when vcpkg tries to install cesium-native:
```
C:\PROGRA~1\MICROS~4\2022\PROFES~1\VC\Tools\MSVC\1442~1.344\bin\Hostx64\x64\cl.exe   /TP -DCESIUMGEOMETRY_BUILDING -DCESIUM_SHARED=ON -DCesiumGeometry_EXPORTS -DFMT_SHARED -DGLM_ENABLE_EXPERIMENTAL -DGLM_FORCE_EXPLICIT_CTOR -DGLM_FORCE_XYZW_ONLY -DSPDLOG_COMPILED_LIB -DSPDLOG_FMT_EXTERNAL -DSPDLOG_SHARED_LIB -IC:\Tools\vcpkg\buildtrees\cesium-native\src\a7d8bb629b-dfd76a50f3.clean\CesiumGeometry\include -IC:\Tools\vcpkg\buildtrees\cesium-native\src\a7d8bb629b-dfd76a50f3.clean\CesiumUtility\include -external:ID:\Dev\vsgCs\vcpkg_installed\x64-windows\include -external:W0 /nologo /DWIN32 /D_WINDOWS /utf-8 /GR /EHsc /MP  /MDd /Z7 /Ob0 /Od /RTC1  -std:c++20 -MDd /W4 /WX /wd4201 /bigobj /w45038 /w44254 /w44242 /w44191 /w45220 /wd4251 /wd4275 /utf-8 /showIncludes /FoCesiumGeometry\CMakeFiles\CesiumGeometry.dir\src\Availability.cpp.obj /FdCesiumGeometry\CMakeFiles\CesiumGeometry.dir\ /FS -c C:\Tools\vcpkg\buildtrees\cesium-native\src\a7d8bb629b-dfd76a50f3.clean\CesiumGeometry\src\Availability.cpp
C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.42.34433\include\vector(1416): error C2280: 'std::unique_ptr<CesiumGeometry::AvailabilityNode,std::default_delete<CesiumGeometry::AvailabilityNode>> &std::unique_ptr<CesiumGeometry::AvailabilityNode,std::default_delete<CesiumGeometry::AvailabilityNode>>::operator =(const std::unique_ptr<CesiumGeometry::AvailabilityNode,std::default_delete<CesiumGeometry::AvailabilityNode>> &)': attempting to reference a deleted function
C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.42.34433\include\memory(3452): note: see declaration of 'std::unique_ptr<CesiumGeometry::AvailabilityNode,std::default_delete<CesiumGeometry::AvailabilityNode>>::operator ='
C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.42.34433\include\memory(3452): note: 'std::unique_ptr<CesiumGeometry::AvailabilityNode,std::default_delete<CesiumGeometry::AvailabilityNode>> &std::unique_ptr<CesiumGeometry::AvailabilityNode,std::default_delete<CesiumGeometry::AvailabilityNode>>::operator =(const std::unique_ptr<CesiumGeometry::AvailabilityNode,std::default_delete<CesiumGeometry::AvailabilityNode>> &)': function was explicitly deleted
```

I've no idea what might be causing this to happen when vcpkg tries building it, since cesium-native builds fine on its own in the same environment. The only thing I can think is that this build is using `ninja`, whereas the normal CMake cesium-native build on Windows uses the `Visual Studio 17 2022` generator, but this only raises further questions since `cmake -B build -S . -G Ninja` fails on cesium-native in the configuration stage.

HellfireATGM:
For what its worth, I have the exact same issue as above. Is there a workaround or fix available?  Thanks.

timoore:
I'm taking a look at this, but I'm curious: do @azrogers or @HellfireATGM have more than one version of the Visual Studio toolset installed? It can be quite tricky to insure that vcpkg packages and the main project all use the same version.

timoore:
I can't reproduce this with the latest vsgCs sources.

HellfireATGM:
Sorry, my bad. Didn't realize there was a config for windows. I did the cmake with this preset:
cmake --preset Windows-Devel
then built in visual studio. 

All good now. Thanks.

kring:
Sorry @timoore I'm not really comfortable merging this before next week's release. I haven't reviewed it in depth, but one problem I see is that the CI build times are drastically increased. I think the caching isn't working anymore. Let's aim to get it in as soon after next week's release as we can, and we can even tag a mid-cycle native release if it's helpful.

timoore:
> Sorry @timoore I'm not really comfortable merging this before next week's release. I haven't reviewed it in depth, but one problem I see is that the CI build times are drastically increased. I think the caching isn't working anymore. Let's aim to get it in as soon after next week's release as we can, and we can even tag a mid-cycle native release if it's helpful.

The cache management page for cesium-native says:
`Approaching total cache storage limit (16.74 GB of 10 GB Used)`

This PR creates new caches for the VCPKG builds that are probably no smaller than the caches for ezvcpkg. I don't think that the caches for this PR are being kept around (or are even being created), and I wouldn't be surprised if caching for all the builds is being affected.

timoore:
I've tried to set up vcpkg caching with S3, but I'm getting a spew of errors like:
```
warning: error: aws failed with exit code: (254).

An error occurred (AccessDenied) when calling the ListObjectsV2 operation: Access Denied
```

so I clearly don't know what I'm doing.

kring:
@timoore I checked out a clean copy of this branch and ran this on my Windows machine:

```
cmake -B build -S .
```

And it seems to run ezvcpkg as normal and then also run some other vcpkg process with a different triplet.

```
PS C:\cesium\cesium-native> cmake -B build -S .
-- Building for: Visual Studio 17 2022
-- VCPKG_TRIPLET x64-windows-static-md
-- VCPKG_TARGET_TRIPLET
-- VCPKG_OVERLAY_PORTS C:/cesium/cesium-native/extern/vcpkg/ports
-- VCPKG_OVERLAY_TRIPLETS
-- EZVCPKG v0.1 starting up
        Website: https://github.com/jherico/ezvcpkg
-- EZVCPKG initializing
        commit:     2024.11.16
        repository: https://github.com/microsoft/vcpkg.git
        local dir:  f:/.ezvcpkg/2024.11.16
-- EZVCPKG Building/Verifying package asyncplusplus using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package expected-lite using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package fmt using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package glm using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package rapidjson using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package spdlog using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package stb using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package ada-url using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package abseil using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package draco using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package ktx using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package modp-base64 using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package meshoptimizer using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package openssl using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package s2geometry using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package libjpeg-turbo using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package sqlite3 using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package tinyxml2 using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package libwebp using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package zlib-ng using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package picosha2 using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package earcut-hpp using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package cpp-httplib[core] using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package libmorton using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package zstd using triplet x64-windows-static-md
-- EZVCPKG Building/Verifying package doctest using triplet x64-windows-static-md
-- EZVCPKG done
-- CMAKE_TOOLCHAIN_FILE f:/.ezvcpkg/2024.11.16/scripts/buildsystems/vcpkg.cmake
CMake Deprecation Warning at F:/.ezvcpkg/2024.11.16/scripts/buildsystems/vcpkg.cmake:40 (cmake_policy):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.31/Modules/CMakeDetermineSystem.cmake:146 (include)
  CMakeLists.txt:119 (project)


-- Running vcpkg install
Fetching registry information from https://github.com/microsoft/vcpkg (HEAD)...
Detecting compiler hash for triplet x64-windows...
Compiler found: C:/Program Files/Microsoft Visual Studio/2022/Professional/VC/Tools/MSVC/14.42.34433/bin/Hostx64/x64/cl.exe
The following packages will be built and installed:
  * abseil:x64-windows@20240722.0 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\95ed5fee53b8698ac2cc6687f470e6fa76474a3a
    ada-url:x64-windows@2.9.2 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\ca729da9e664e3a59d88115c9e8a8d59b775801c
    asyncplusplus:x64-windows@1.1#2 -- C:\cesium\cesium-native\./vcpkg-overlays\asyncplusplus
  * brotli:x64-windows@1.1.0#1 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\4e5b5ae1ad26c80535c893cc0307121f0398549e
    cpp-httplib[brotli,core]:x64-windows@0.18.0 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\22ef6ccb262c5a3a835b39278d4800035a967a94
    doctest:x64-windows@2.4.11 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\619b544477f70ef777fcc294e0b31650e2bd4c05
    draco:x64-windows@1.5.7 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\06799add74da6e9e4634e4c7f27b82b1149a8bb6
    earcut-hpp:x64-windows@2.2.4 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\f4f99272400624124150cb485899cdc3efb24313
    expected-lite:x64-windows@0.8.0 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\ffed518a0067b94d369307cc869b46263bbcd976
  * fmt:x64-windows@11.0.2#1 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\07a73a7565e5de9eb42e90c16c133bdfdfebbcda
    glm:x64-windows@1.0.1#3 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\f8b09027afe97c717e95bd81d1d4f37c057d1c78
    ktx:x64-windows@4.3.2 -- C:/cesium/cesium-native/extern/vcpkg/ports\ktx
    libjpeg-turbo:x64-windows@3.0.4 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\fbedc8ef954f9951c7d169c9bac3f9534b4b2c77
    libmorton:x64-windows@0.2.12 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\6f00ce462ab2d581aa3d0bb33c2ff37ff745377c
    libwebp[core,libwebpmux,nearlossless,simd,unicode]:x64-windows@1.4.0#1 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\9d0c00119dd6ec04488b4e0c926aa3ae73b95847
    meshoptimizer:x64-windows@0.21 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\9ff504432cfa14706fac154ca9c490a3df9922f8
    modp-base64:x64-windows@2020-09-26#2 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\85f2a557b0fdca31efb6e9e6c769309d583ea2a3
    ms-gsl:x64-windows@4.1.0 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\d71696a526519c3f680b1b77e87d39f8eb94e650
    openssl:x64-windows@3.4.0 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\ce504a83eb9627d54f1cffdb497a6bf5bd970d18
    picosha2:x64-windows@2018-07-30#2 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\c873a8748f3e35ae1be390b0de9cc3754191c987
    rapidjson:x64-windows@2023-07-17#1 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\a3dfe7dca3a1a27564b1fc5a9aea657cd9dae01c
    s2geometry:x64-windows@0.11.1 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\3fd94490409a17aef62732c674d5e12b4804fdd8
    spdlog:x64-windows@1.14.1 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\545d0306db761411e23623ae6f44fc196cb571f7
    sqlite3[core,json1]:x64-windows@3.47.0#2 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\543d6067a10a7ab97a4893997afaee4d34230a27
    stb:x64-windows@2024-07-29#1 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\c362152d1d02973e7b1ce777960f8a6876656560
    tinyxml2:x64-windows@10.0.0 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\e7763e8defdfa46b8d2e399756511f0c7ff00448
  * vcpkg-cmake:x64-windows@2024-04-23 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\e74aa1e8f93278a8e71372f1fa08c3df420eb840
  * vcpkg-cmake-config:x64-windows@2024-05-23 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\97a63e4bc1a17422ffe4eff71da53b4b561a7841
  * vcpkg-cmake-get-vars:x64-windows@2024-09-22 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\f23148add155147f3d95ae622d3b0031beb25acf
    zlib-ng:x64-windows@2.1.5#1 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\3f7c39c070cb4f55a281ebe3a178b4f7e896dea8
    zstd:x64-windows@1.5.6 -- C:\Users\kevin\AppData\Local\vcpkg\registries\git-trees\796171d4d359b6786b8e1380fb6da8e677a9087b
Additional packages (*) will be modified to complete this operation.
Restored 11 package(s) from C:\Users\kevin\AppData\Local\vcpkg\archives in 568 ms. Use --debug to see more details.
Installing 1/31 vcpkg-cmake-config:x64-windows@2024-05-23...
Elapsed time to handle vcpkg-cmake-config:x64-windows: 19.2 ms
vcpkg-cmake-config:x64-windows package ABI: eb83882cb50f9ebd78ad85880e437c0acbc0e7110d9ae1741b7663cf22f187af
Installing 2/31 vcpkg-cmake:x64-windows@2024-04-23...
Elapsed time to handle vcpkg-cmake:x64-windows: 9.81 ms
vcpkg-cmake:x64-windows package ABI: c9a09eb622adac56e4988734885488bd1ae3c275dca9df8c825038943ff9e0b2
Installing 3/31 ada-url:x64-windows@2.9.2...
```

(log continues after that, installing more packages)

I guess my mistake is that I'm not using the cmake preset that disables ezvcpkg. But ezvcpkg should still be a valid option at this point, right?

kring:
The same happens on macOS, but at least there the triplets are the same (`arm64-osx` in both cases). But it's still spending a bunch of time compiling packages for unclear reasons.

timoore:
After looking at this for "a while," I believe that it's normal that vcpkg.cmake is running again (several times in my experiments). vcpkg abuses the CMake toolchain file, which can be run multiple times during configuration as CMake configures different things. This PR includes new CMake targets etc., which I believe is the cause of more configuration "runs" than in the old code.

It is concerning that the triplet used for the subsequent runs of vcpkg is different from the first; I think I know what's up with that and will amend the PR.

timoore:
@kring The behavior you were seeing was caused by vcpkg getting invoked in manfest mode even when managed by ezvcpkg. Once vcpkg.cmake is established as the toolchain file, which ezvcpkg does, it will run in manifest mode if it finds vcpkg.json, which is now a part of the source tree... unless told otherwise, which I have now done, politely but firmly.

Cesium3DTilesSelection/CMakeLists.txt
```diff
@@ -25,16 +25,19 @@ target_sources(
...
 
 cesium_target_include_directories(
     TARGET 
         Cesium3DTilesSelection
     PUBLIC
-        ${CMAKE_BINARY_DIR}
-        ${CMAKE_CURRENT_LIST_DIR}/include
+        $<BUILD_INTERFACE:${CESIUM_NATIVE_RAPIDJSON_INCLUDE_DIR}>
+        $<BUILD_INTERFACE:${CESIUM_NATIVE_DRACO_INCLUDE_DIR}>
+        $<BUILD_INTERFACE:${CESIUM_NATIVE_LIBMORTON_INCUDE_DIR}>
```
kring, line 37: Are these needed? They don't seem to be defined anywhere.

Cesium3DTilesSelection/CMakeLists.txt
```diff
@@ -25,16 +25,19 @@ target_sources(
...
+        $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}>
```
kring, line 38: I think those one isn't new, but it's not obvious why it's needed, either. Do you happen to know?

CesiumAsync/CMakeLists.txt
```diff
@@ -24,15 +24,16 @@ target_sources(
...
         ${CESIUM_ASYNC_PUBLIC_HEADERS}
 )
 
 cesium_target_include_directories(
-    TARGET 
+    TARGET
         CesiumAsync
     PUBLIC
-        ${CMAKE_CURRENT_LIST_DIR}/include
+        $<BUILD_INTERFACE:${CESIUM_NATIVE_RAPIDJSON_INCLUDE_DIR}>
```
kring, line 34: Same here, not sure what this is for.

CesiumGeospatial/CMakeLists.txt
```diff
@@ -25,23 +25,23 @@ target_sources(
...
+        $<INSTALL_INTERFACE:include>
 )
 
 target_link_libraries(CesiumGeospatial
 PUBLIC
     CesiumGeometry
     CesiumUtility
-    earcut
 PRIVATE
+    $<BUILD_INTERFACE:earcut>
```
kring, line 44: I don't quite understand what this is doing, either.

CesiumGltfReader/CMakeLists.txt
```diff
@@ -37,16 +37,18 @@ target_sources(
...
 )
 
 cesium_target_include_directories(
     TARGET
         CesiumGltfReader
     PUBLIC
-        ${CMAKE_CURRENT_LIST_DIR}/include
-        ${CMAKE_CURRENT_LIST_DIR}/generated/include
+        # Necessary for `draco/draco_features.h`
+        $<BUILD_INTERFACE:${CESIUM_NATIVE_DRACO_INCLUDE_DIR}>
```
kring, line 48: Another occurrence of this. It has a comment this time, but it doesn't clear things up for me. We don't appear to be including `draco/draco_features.h` anywhere?

CesiumUtility/CMakeLists.txt
```diff
@@ -25,15 +25,16 @@ target_sources(
...
         ${CESIUM_UTILITY_PUBLIC_HEADERS}
 )
 
 cesium_target_include_directories(
     TARGET
         CesiumUtility
     PUBLIC
-        ${CMAKE_CURRENT_LIST_DIR}/include
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>
+        $<BUILD_INTERFACE:${CESIUM_NATIVE_RAPIDJSON_INCLUDE_DIR}>
```
kring, line 36: Here again.

CMakeLists.txt
```diff
@@ -244,6 +280,14 @@ list(APPEND CMAKE_PREFIX_PATH "${PACKAGE_BUILD_DIR}")
 # However, for some of the vcpkg built libraries where they don't provide a prope cmake config file, we have to declare
 # and imporeted library target ourselves. This is the case for modp_b64::modp_b64, picosha2::picosha2 and earcut. In
 # these cases, we *do* have the somewhat ugly and verbose details in the extern/CMakeLists.txt file.
+#
+# XXX Above comment should be obsoleted by these first calls to
```
kring, line 284: It'd be good to just update this entire comment block.

vcpkg-configuration.json
```diff
@@ -0,0 +1,17 @@
+{
+    "default-registry": {
+        "kind": "git",
+        "baseline": "4f746bc66438fce2b900c3ba6094a483b871b045",
```
kring, line 4: How was this chosen? It looks to be older than the 2024.11.16 tag we're currently using.

kring (commented):
Thanks @timoore, this is working well for me now and is getting pretty close! A few questions and comments below.

kring:
One other thing: It seems to be difficult to use the ezvcpkg mode from VS Code now, because it wants to use one of the presets, and there's no ezvcpkg preset. And using of the presets that do exist requires installing vcpkg, which was not previously required.

So, if we can make things work just as they did before (at least for non-manifest-mode users), then that's fine. But if not, then we at least need to update the developer setup instructions to reflect how things work now.

vcpkg-configuration.json
timoore: It was the vcpkg HEAD when I first started hacking on this. I'll update it to 2024.11.16; we can look at updating it again at a later date.

vcpkg-overlays/asyncplusplus/portfile.cmake
```diff
@@ -0,0 +1,21 @@
+vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
+
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO Amanieu/asyncplusplus
+    REF 4159da79e20ad6d0eb1f13baa0f10e989edd9fba
```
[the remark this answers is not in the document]
timoore: I totally agree about using vcpkg ports wherever possible. Do you think there's any risk in moving to asyncplusplus1.2 in the PR?

CesiumGeospatial/CMakeLists.txt
timoore: I think this correct. CesiumGeospatial depends on a header-only library which declares its target properties in the correct CMake fashion.

Cesium3DTilesSelection/CMakeLists.txt
timoore: The BUILD_INTERFACE / INSTALL_CHANGES were made before we switched to ezvcpkg, and I forgot that these redundant include directories were hanging around. I've removed them.

kring:
Thanks for adding the ezvcpkg preset @timoore, but unfortunately this is still pretty much unusable in VS Code. The problem is that there's no way to control which build configuration (Debug / Release / RelWithDebInfo) is built without defining extra configure and/or build presets (depending on whether or not it's a multi-config generator). We could add more presets, but we're talking about `platforms * (generators and dependency modes) * configurations`, which really adds up. From what I can tell, this is inherent in the (poor?) design of CMake presets, and compounded by the (poor?) design of the VS Code CMake integration.

It's possible to specify a configuration on the command-line, of course, but there's no easy way to do this within VS Code. As compared to before this PR, where you just choose the configuration from a dropdown at the bottom and it automatically handles the single-config versus multi-config generator difference for you. I change configurations a _lot_ while working with cesium-native, so IMO this is a pretty big step backwards.

Can we get rid of the presets file? I don't think it's actually adding that much, so that seems like the easiest solution. Won't almost everything in there be deduced correctly anyway? Even the `x64-linux` triplet should be deduced automatically in recent (but pre-4.0) versions of CMake. The only reason it doesn't happen automatically on some systems is because, amazingly, the `LINUX` CMake variable was only added in CMake v3.25:
https://cmake.org/cmake/help/latest/variable/LINUX.html

Users using vcpkg directly without ezvcpkg will need to manually specify the vcpkg toolchain file, but that's pretty basic stuff explained right in the vcpkg docs, so it doesn't seem too big a burden. And even in VS Code that isn't a problem, because it's just an extra argument added to every cmake configure command. It's not the same as the situation with configurations described above, because we need to frequently change configurations.

timoore:
> Thanks for adding the ezvcpkg preset @timoore, but unfortunately this is still pretty much unusable in VS Code. The problem is that there's no way to control which build configuration (Debug / Release / RelWithDebInfo) is built without defining extra configure and/or build presets (depending on whether or not it's a multi-config generator). We could add more presets, but we're talking about `platforms * (generators and dependency modes) * configurations`, which really adds up. From what I can tell, this is inherent in the (poor?) design of CMake presets, and compounded by the (poor?) design of the VS Code CMake integration.
> 

I downloaded VS Code and gave it a spin. I don't think it's going to see much use in the @timoore household. This problem is known to the CMake developers; see https://gitlab.kitware.com/cmake/cmake/-/issues/22538. The solutions presented there seem pretty out of control to me. In command-line CMake the configure preset specified in a build preset doesn't actually have to correspond to the configure preset in use; it will work as long as the build directories are the same, or at least works in my limited tests. Unfortunately IDEs like VS Code and CLion enforce a strict correspondence between the actual configure preset and the available build presets.

It seems easy enough to define a new build preset in VS Code, which would let the user set up whatever configure / build combos they want, but it insists in putting the new configuration in CMakePresets.json, not the more preferable CMakeUserPresets.json. :skull_and_crossbones: 

> It's possible to specify a configuration on the command-line, of course, but there's no easy way to do this within VS Code. As compared to before this PR, where you just choose the configuration from a dropdown at the bottom and it automatically handles the single-config versus multi-config generator difference for you. I change configurations a _lot_ while working with cesium-native, so IMO this is a pretty big step backwards.
> 

> Can we get rid of the presets file? I don't think it's actually adding that much, so that seems like the easiest solution. Won't almost everything in there be deduced correctly anyway? Even the `x64-linux` triplet should be deduced automatically in recent (but pre-4.0) versions of CMake. The only reason it doesn't happen automatically on some systems is because, amazingly, the `LINUX` CMake variable was only added in CMake v3.25: https://cmake.org/cmake/help/latest/variable/LINUX.html
> 

Do you mean deduced by `vcpkg` or by our own `detect-vcpkg-triplet.cmake`?

I had been trying to move vcpkg-related stuff out of Native's CMake machinery as much as possible; IMO, that makes it simpler to integrate Cesium Native into other projects that use vcpkg in their builds, which is where the real interest of all this work lies. So, no specifying the toolchain file in an included cmake file, etc. 


> Users using vcpkg directly without ezvcpkg will need to manually specify the vcpkg toolchain file, but that's pretty basic stuff explained right in the vcpkg docs, so it doesn't seem too big a burden. And even in VS Code that isn't a problem, because it's just an extra argument added to every cmake configure command. It's not the same as the situation with configurations described above, because we need to frequently change configurations.

This PR is a step in removing ezvcpkg, so I'd like to make the pure `vcpkg` workflow productive. I think that the `vcpkg` arguments are fairly verbose.  GitHub workflow files will be more verbose. I would be OK with putting a sample CMakePresets.json or CMakeUserPresets.json in the documentation.

VS Code can also be configured to ignore CMakePresets.json.

kring:
Sorry for the delay, and thanks for all your work on this @timoore. Merging!

PR #1031, opened 2024-12-06
Allow valid binary subtrees to load without requiring a binary chunk

j9liu:
`SubtreeFileReader` assumed the subtree would be invalid if its binary chunk was missing, which is not necessarily true. As described in the [spec](https://github.com/CesiumGS/3d-tiles/tree/main/specification/ImplicitTiling#subtree-binary-format), subtrees expressed in the binary format can exist without the binary chunk.

This PR fixes that. I also ended up refactoring how subtrees were procedurally created in `TestSubtreeAvailability`, so the diff is relatively large because of my changes there.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,9 @@
 ### ? - ?
 
 ##### Fixes :wrench:
+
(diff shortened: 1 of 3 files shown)


kring:
Thanks @j9liu!

PR #1035, opened 2024-12-09
Add more getters for `AccessorView` and `AccessorWriter`

lilleyse:
> - Added `offset` getter to `AccessorView`. 

This makes it possible to create an `AccessorView` from another `AccessorView` of a different type.

> - Added `stride`, `offset`, and `data` getters to `AccessorWriter`.
> - Added `value_type` typedef to `AccessorWriter`.

For convenience when working with `AccessorWriter`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,12 @@
 
 ### ? - ?
 
+##### Additions :tada:
+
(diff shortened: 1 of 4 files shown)


kring:
Thanks @lilleyse!

PR #1040, opened 2024-12-11
Fix all Doxygen warnings

azrogers:
This PR fixes *every* remaining Doxygen warning, and configures Doxygen to return a non-zero exit code when there's a warning, meaning the CI task will fail if code is added that causes new warnings. This not only helps improve our current documentation, but will allow us to keep a better eye on our documentation going forward!

Original description:
> This PR adds Doxygen documentation to a whole lot of things. How many things? I can't quite say, but the current CI Doxygen log [is 5052 lines long](https://github.com/CesiumGS/cesium-native/actions/runs/12266566422/job/34224945676) while the CI Doxygen log for this PR is [only 2991 lines long](https://github.com/CesiumGS/cesium-native/actions/runs/12284506412/job/34280419149) - so quite a few warnings have been taken care of. Plenty still remain (651, by my count of just the STDERR log), but this is still a pretty big step forward.


azrogers:
@kring pointed out that the ideal here is that we get Doxygen warnings to zero, so we can then start treating them as errors which will help make sure we keep the docs up to date going forward. I think that's an achievable goal, so I'm making this PR a draft for the moment while I work on that.

azrogers:
And with that, *every* remaining warning has been resolved!

PR #1041, opened 2024-12-12
Production doc script improvements.

kring:
- Use Doxygen 1.12.
- Use ccache and the vcpkg artifacts cache to improve performance.
- Run the doc deploy in both the `cesium.com` and `cesium.com-staging` branches. In the latter, deploy to `cesium-native/ref-doc-staging/` instead of `cesium-native/ref-doc/`. This allows us to test doc changes without affecting production.

### DIFF
--- a/.github/workflows/prod.yml
+++ b/.github/workflows/prod.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - 'cesium.com'
+      - 'cesium.com-staging'
 jobs:
   deploy-docs:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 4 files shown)


kring:
I've already merged this to `cesium.com-staging`, so it's live. I'm purposely avoiding linking directly to it because we don't want it to be indexed by search engines. I tried to prevent this by adding the `X-Robots-Tag: None` HTTP header, but unfortunately S3 doesn't allow adding arbitrary headers like this.

azrogers:
What about adding a line like `printf "User-agent: *\nDisallow: /" > build/doc/html/robots.txt` just before syncing with AWS? We could also add the [robots meta tag](https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag#robotsmeta) but this seems like an easier solution.

kring:
My understanding is that `robots.txt` has to be in the root directory of the web server, unfortunately. The robots meta tag would work, but I'm not sure if it's worth the trouble.

azrogers:
@kring How about this solution?

kring:
Looks great, thanks @azrogers! I guess we could consider using`none` instead of `noindex`, but it probably doesn't make much difference.

azrogers:
Great! Thanks for the change, @kring!

PR #1042, opened 2024-12-13
"Kick" only when the current tile is renderable

### LINKED ISSUE #1033: Consider "kicking" only when the current tile is renderable
After working on #1028 (and its fix in #1032), it occurred to me that the kicking mechanism might be more aggressive than it needs to be, and so it is sometimes preventing tiles from appearing as soon as they could.

The purpose of kicking is to allow arbitrary level skipping while also avoiding unnecessary holes while loading is in progress. Fundamentally, when we detect that a) some of a tile's selected children / descendants aren't done loading yet, and b) none of them were rendered last frame, we can throw them all out of the render list and render the current tile instead.

However, what if the current tile _also_ isn't renderable? We don't currently check that when choosing to kick. Which means one of the following will eventually happen:

1. This tile will get kicked, too, in favor of some other ancestor tile. Hopefully that one is renderable! Or,
2. This tile will be a hole. And it will continue to be a hole until all of its originally-selected descendants finish loading.

Case (1) is fine, or at least the best we can do. Case (2) is really not great. If we're going to have a hole, it would be better to have more smaller holes that visibly fill in, rather than one big hole that sticks around for a long time.

So I _think_ that we can make our loading better by only doing a "kick" when the current tile is actually renderable. This shouldn't affect case (1) because a renderable ancestor can do the kick even if the current (unrenderable) tile doesn't. And in case (2) the hole would be smaller and fill in more incrementally over time.

This is a pretty simple change, but we need to try it out and make sure it actually makes things better.

An alternate idea is to keep kicking as we currently do, but detect the case (2) hole and load the missing tile with highest priority, rather than waiting for all the more detailed tiles for that space to load. In other words, make it a feature of the algorithm that it eliminates holes as quickly as it can. This is likely to be trickier to implement than the above, but may give better results.

kring:
This is a PR into #1032 so merge that first.

Fixes #1033

This is a draft because they are some test failures. I think it's just tests asserting conditions that are purposely changed here, but I need to think through it more thoroughly.


kring:
I've fixed the test failures and this is ready for review now.

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -1430,7 +1430,16 @@ Tileset::TraversalDetails Tileset::_visitTile(
       lastFrameSelectionResult == TileSelectionState::Result::Rendered &&
       pRenderContent && pRenderContent->getLodTransitionFadePercentage() < 1.0f;
 
-  if (kickDueToNonReadyDescendant || kickDueToTileFadingIn) {
+  // Only kick if this tile is renderable, or if we've exceeding the
+  // loadingDescendantLimit. It's pointless to kick to a tile that is not
+  // loaded, because it means we will still have a hole, quite possibly a bigger
+  // one.
```
j9liu, line 1436:
```suggestion
  // Only kick the descendants of this tile if it is renderable, or if we've
  // exceeded the loadingDescendantLimit. It's pointless to kick the descendants
  // of a tile that is not yet loaded, because it means we will still have a
  // hole, and quite possibly a bigger one.
```

CHANGES.md
```diff
@@ -20,6 +20,7 @@
 - Fixed a raster overlay bug that could cause unnecessary upsampling with failed or missing overlay tiles.
 - Fixed a bug in  `SubtreeFileReader::loadBinary` that prevented valid subtrees from loading if they did not contain binary data.
 - Fixed a bug in the `Tileset` selection algorithm that could cause detail to disappear during load in some cases.
+- Improved the tileset selection's "kick" criteria to allow holes in a `Tileset`, when they do occur, to be filled with loaded tiles more incrementally.
```
j9liu, line 23: Just some slight tweaks for clarity.

```suggestion
- Improved the "kicking" mechanism in the tileset selection algorithm. The new criteria allows holes in a `Tileset`, when they do occur, to be filled with loaded tiles more incrementally.
```

Cesium3DTilesSelection/test/TestTilesetSelectionAlgorithm.cpp
```diff
@@ -600,20 +601,26 @@ TEST_CASE("Test additive refinement") {
...
+
+      // root is done loading and rendered.
       REQUIRE(root->getState() == TileLoadState::Done);
       REQUIRE(!doesTileMeetSSE(viewState, *root, tileset));
       REQUIRE(root->getChildren().size() == 1);
+      REQUIRE(std::find(ttr.begin(), ttr.end(), root) != ttr.end());
 
-      // root's children don't have content loading right now, so only root get
-      // rendered
+      // root's child is done loading and rendered, too.
```
j9liu, line 613:
```suggestion
      // root's children are done loading and rendered, too.
```

Cesium3DTilesSelection/test/TestTilesetSelectionAlgorithm.cpp
```diff
@@ -563,31 +563,32 @@ TEST_CASE("Test additive refinement") {
   SECTION("Load external tilesets") {
     ViewState viewState = zoomToTileset(tileset);
 
-    // 1st frame. Root will get rendered first and 5 of its children are loading
-    // since they meet sse
+    // 1st frame. Root, its child, and its four grandchildren will all be
+    // rendered because they meet SSE, even though they're not loaded yet.
```
j9liu, line 567: I see "render" getting used even when tiles aren't loaded in yet -- would it be more accurate to say tiles get "selected" vs. "rendered"?

Though, it seems like this verbiage is built into the API already (`tilesToRenderThisFrame`), so maybe it's better to stick with "render". But I hope our documentation clarifies that `render ~= selection`, and `render != content show up on screen` 🙂

Cesium3DTilesSelection/test/TestTilesetSelectionAlgorithm.cpp
```diff
@@ -1554,21 +1560,30 @@ void runUnconditionallyRefinedTestCase(const TilesetOptions& options) {
       pRawLoader->createRootTile(),
       options);
 
-  // On the first update, we should render the root tile, even though nothing is
-  // loaded yet.
+  // On the first update, we should refine right to the grandchild tile, even
```
j9liu, line 1563: The nitpickiest of nitpicks...

```suggestion
  // On the first update, we should refine down to the grandchild tile, even
```

j9liu (commented):
Thanks @kring ! I just have a few comments that just about clarity. Let me know if I should test this behavior in any way as well.

Cesium3DTilesSelection/test/TestTilesetSelectionAlgorithm.cpp
kring: There's only one in this case though, right?

Cesium3DTilesSelection/test/TestTilesetSelectionAlgorithm.cpp
kring: Yeah this is a good point. It's a bit of a quirk of this whole system. Arguably, tiles that are not renderable should never be selected/rendered, in general. That would make sense, and be least likely to surprise people. My thought process when I went the other way was that - just maybe! - it could be useful to know which tiles _would be_ rendered if they were loaded. For example, maybe you could render a grey cloudy thing in the space of their bounding box, and that would be better than a hole.

So whether this is a good idea or not, Cesium Native currently refers to selection as "rendering", and the rendered tiles aren't necessarily actually renderable. Changing this might be an excellent idea, but I think it's outside the scope of this PR.

kring: > But I hope our documentation clarifies that render ~= selection, and render != content show up on screen

I think it probably makes that apparent, but doesn't say so explicitly.

kring:
Thanks @j9liu! I think I've addressed your comments, or at least responded to them.

Cesium3DTilesSelection/test/TestTilesetSelectionAlgorithm.cpp
j9liu: Oh sorry, I got confused by the lines below this (`REQUIRE(parentB3DM.getChildren().size() == 4);`) and thought the comment was referring to this check. Nevermind 🙂

PR #1048, opened 2024-12-19
Fix clang-tidy warnings to make it actually useful

### LINKED ISSUE #203: Add clang-tidy and cppcheck to CI process

azrogers:
When we first added clang-tidy to Cesium Native, we left the warnings as merely informational, because there were too many of them to fix at the moment. Unfortunately, this leaves clang-tidy as an afterthought, because I don't imagine many of us are poking through the CI logs to see if our changes produced any new clang-tidy warnings when there's already hundreds of existing warnings. This change attempts to fix this: resolve the existing warnings, enable any useful checks that weren't already enabled, and treat the warnings as errors so we'll notice them.

Changes in this PR:
- clang-tidy on CI now uses clang-tidy version 19, because earlier versions don't support the `ExcludeHeaderFilterRegex` config option we're using to remove `ezvcpkg` headers from the build.
- We are now following Include-What-You-Use, thanks to clang-tidy's misc-include-cleaner check. This has resulted in pretty much every file in the project receiving a change, but it's worth it.
- Every `bugprone` clang-tidy check has been enabled, with the following exceptions:
    - `bugprone-exception-escape`, as fixing this is already a pending issue (#348).
    - `bugprone-misplaced-widening-cast`, as it results in extremely verbose lines (like `size_t pixels = static_cast<size_t>(image.width) * static_cast<size_t>(image.height) * static_cast<size_t>(image.channels)` instead of just `size_t pixels = static_cast<size_t>(image.width * image.height * image.channels)`.
    - `bugprone-unhandled-exception-at-new`, as this is both part of #348, and also extremely nitpicky - I don't think we need to wrap every call to `new` in a `try...catch`. 
- Most `modernize` and `misc` checks have been enabled, where they didn't conflict with the style guide or produce way too many changes for one PR. The following I decided were too much for just this PR, but they can be automatically applied at a later date using `clang-tidy-fix`:
    - `misc-const-correctness` detects every variable that could be declared `const`.
    - `modernize-loop-convert` uses range-based for loops instead of manually advancing the iterator.
    - `modernize-type-traits` changes `traits<...>::type` and `traits<..>::value` to `traits_t<...>` and `traits_v<...>`
    - `modernize-use-constraints` replaces `std::enable_if` with C++20 `requires` clauses.
    - `modernize-use-using` replaces `typedef` with `using`.
    - `modernize-return-braced-init-list` replaces statements like `return glm::dvec2(0.5, 0.5);` with `return {0.5, 0.5};`
    - `modernize-use-designated-initializers` would mean adding `.field=` to initializer statements, like `{.x = 0.5, .y = 0.5}`, which would prevent bugs when changing the order of fields.
- All `performance` checks have been enabled, with the exception of `performance-enum-size`, as I felt that saving three bytes by defining an enum with `int8_t` instead of the default `int32_t` didn't seem like it was worth the change.
- `readability` checks are mostly style guide-related, but might be worth looking into. `readability-else-after-return` and `readability-braces-around-statements` seem like they could be good to add.
- All warnings are now treated as errors.

Fixes #203


Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -347,6 +395,11 @@ std::vector<CesiumGeospatial::Projection> mapOverlaysToTile(
   return projections;
 }
 
+// Really, these two methods are risky and could easily result in bugs, but
+// they're private methods used only in this class, so it's easier to just tell
+// clang-tidy to ignore them.
```
kring, line 400: I don't think this is inherently risky, but it has particular semantics that are captured perfectly by the [clang::lifetimebound](https://clang.llvm.org/docs/AttributeReference.html#id11) attribute. Unfortunately that's clang-specific, and clang-tidy doesn't honor it anyway prior to version 20. https://clang.llvm.org/extra/clang-tidy/checks/bugprone/return-const-ref-from-parameter.html
So just disabling the check seems ok for now. It also wouldn't be terrible to change these functions to return a `BoundingVolume` by value.

Cesium3DTilesSelection/src/TilesetHeightQuery.cpp
```diff
@@ -112,12 +132,12 @@ void TilesetHeightQuery::intersectVisibleTile(
 
   // Set ray info to this hit if closer, or the first hit
   if (!this->intersection.has_value()) {
-    this->intersection = std::move(gltfIntersectResult.hit);
+    this->intersection = gltfIntersectResult.hit;
```
kring, line 135: Another move of a trivially-copyable type. This should be allowed IMO.

Cesium3DTilesSelection/src/TilesetHeightQuery.cpp
```diff
@@ -112,12 +132,12 @@ void TilesetHeightQuery::intersectVisibleTile(
...
   } else {
     double prevDistSq = this->intersection->rayToWorldPointDistanceSq;
     double thisDistSq = intersection->rayToWorldPointDistanceSq;
     if (thisDistSq < prevDistSq)
-      this->intersection = std::move(gltfIntersectResult.hit);
+      this->intersection = gltfIntersectResult.hit;
```
kring, line 140: Same here.

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -967,17 +1000,17 @@ TilesetJsonLoader::loadTileContent(const TileLoadInput& loadInput) {
                         logTileLoadResult(pLogger, tileUrl, result.errors);
                         if (result.errors) {
                           return TileLoadResult::createFailedResult(
-                              std::move(pAssetAccessor),
-                              std::move(pCompletedRequest));
+                              pAssetAccessor,
+                              pCompletedRequest);
```
kring, line 1004: These moves would be fine if the lambda were declared `mutable` and the `pCompletedRequest` were moved on capture.

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -967,17 +1000,17 @@ TilesetJsonLoader::loadTileContent(const TileLoadInput& loadInput) {
...
                         return TileLoadResult{
                             std::move(*result.model),
                             upAxis,
                             std::nullopt,
                             std::nullopt,
                             std::nullopt,
-                            std::move(pAssetAccessor),
-                            std::move(pCompletedRequest),
+                            pAssetAccessor,
+                            pCompletedRequest,
```
kring, line 1013: Same here.

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -990,7 +1023,7 @@ TilesetJsonLoader::loadTileContent(const TileLoadInput& loadInput) {
                       upAxis,
                       tileRefine,
                       pLogger,
-                      std::move(pAssetAccessor),
+                      pAssetAccessor,
```
kring, line 1026: And here.

Cesium3DTilesSelection/src/ViewState.cpp
```diff
@@ -1,14 +1,64 @@
...
 
+#include <optional>
+#include <variant>
+
 using namespace CesiumGeometry;
 using namespace CesiumGeospatial;
 
 namespace Cesium3DTilesSelection {
 
+namespace {
```
kring, line 28: Better to leave this code where it was in the file, and just change it from `static` to `namespace { ... }`.

kring (commented):
Thanks @azrogers! I've so far reviewed only the Cesium3DTiles, Cesium3DTilesContent, Cesium3DTilesReader, and Cesium3DTilesSelection namespaces. Lots of really good changes here! And a few things for me to nitpick (see below).

.github/workflows/build.yml
```diff
@@ -17,14 +17,18 @@ jobs:
     steps:
       - name: Check out repository code
         uses: actions/checkout@v4
-      - name: Install ninja
-        uses: seanmiddleditch/gha-setup-ninja@master
-      - name: Install nasm
-        uses: ilammy/setup-nasm@v1
+      - name: Install latest ninja and cmake
+        uses: lukka/get-cmake@latest
```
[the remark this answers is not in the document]
azrogers: The motivation here is to speed up clang-tidy a bit by not installing nasm, since it shouldn't affect anything clang-tidy is checking. I can add it back if we still need it for linting though.

Cesium3DTilesContent/src/BatchTableToGltfStructuralMetadata.cpp
```diff
@@ -791,6 +816,7 @@ void updateExtensionWithJsonScalarProperty(
 
   for (int64_t i = 0; i < propertyTable.count; ++i, ++p, ++it) {
     if (it->IsNull()) {
+      CESIUM_ASSERT(noDataValue.has_value());
```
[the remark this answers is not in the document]
j9liu: I think the intention is to guarantee that `*noDataValue` is a valid value before dereferencing the `optional`. And it's been a while since I've looked at this code, so I wasn't 100% sure that that will be the case. But after a lookover I'm fairly confident this won't happen.

TL;DR if `noDataValue` were `nullopt`, this would be a developer error, and I don't think it hurts to keep the assert.

Rationale below if you would like it....
<hr>

First we have to execute `findCompatibleTypes` to narrow down how the property could be represented. For scalar properties, we try to settle on a reasonable sentinel value (-1 or 0) if it contains `null` entries. At the end, we verify whether any numeric sentinel values are left. And if there are none, we actually default the type to `STRING` so that we can use the empty string sentinel.

https://github.com/CesiumGS/cesium-native/blob/80cb0d92c6b1e29a7cfa000318fb00f7762ccf41/Cesium3DTilesContent/src/BatchTableToGltfStructuralMetadata.cpp#L654-L659

Looking at where `noData` is set.... the scalar property _should_ have a sentinel value to fall back on. Because otherwise, it would have been turned into a string property. But this is definitely not obvious from a glance.

https://github.com/CesiumGS/cesium-native/blob/80cb0d92c6b1e29a7cfa000318fb00f7762ccf41/Cesium3DTilesContent/src/BatchTableToGltfStructuralMetadata.cpp#L1420-L1436

And in the way we handle JSON string property, we do check whether the sentinel value is present. We don't attempt to write the sentinel value if `noDataValue` is `std::nullopt`. I think we just give up and try to write `null` to the output. 😆 

https://github.com/CesiumGS/cesium-native/blob/80cb0d92c6b1e29a7cfa000318fb00f7762ccf41/Cesium3DTilesContent/src/BatchTableToGltfStructuralMetadata.cpp#L698-L716

Cesium3DTiles/src/MetadataQuery.cpp
```diff
@@ -1,5 +1,15 @@
+#include "Cesium3DTiles/Class.h"
+#include "Cesium3DTiles/ClassProperty.h"
+#include "Cesium3DTiles/MetadataEntity.h"
+#include "Cesium3DTiles/Schema.h"
+#include "CesiumUtility/JsonValue.h"
```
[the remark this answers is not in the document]
azrogers: Wrote a Node script (when all you have is a Node.js runtime...) to conform the includes, so local files (private headers included by cpp files in the same folder) use quotes and everything else uses `<>`.

Cesium3DTilesContent/src/I3dmToGltfConverter.cpp
```diff
@@ -405,8 +431,8 @@ CesiumAsync::Future<ConvertedI3dm> convertI3dmContent(
         decodedInstances.rotations.begin(),
         [](auto&& upOct, auto&& rightOct) {
           return rotationFromUpRight(
-              decodeOct32P(upOct),
-              decodeOct32P(rightOct));
+              decodeOct32P(reinterpret_cast<const uint16_t*>(upOct)),
+              decodeOct32P(reinterpret_cast<const uint16_t*>(rightOct)));
```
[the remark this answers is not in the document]
azrogers: These were fixes for [cppcoreguidelines-pro-bounds-array-to-pointer-decay](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/pro-bounds-array-to-pointer-decay.html) before I decided that it was too verbose for too little gain to leave enabled.

Cesium3DTilesSelection/src/ImplicitQuadtreeLoader.cpp
```diff
@@ -201,7 +221,7 @@ CesiumAsync::Future<TileLoadResult> requestTileContent(
                     logTileLoadResult(pLogger, tileUrl, result.errors);
                     if (result.errors || !result.model) {
                       return TileLoadResult::createFailedResult(
-                          std::move(pAssetAccessor),
+                          pAssetAccessor,
```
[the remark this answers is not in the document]
azrogers: Moving `pAssetAccessor` in the capture does nothing as `pAssetAccessor` is passed by constant reference.

azrogers:
Disabled `CheckTriviallyCopyableMove` and re-added what `std::move`s I could, but some of them are just passed to the functions they're in via const references, so attempting to move them does nothing and clang-tidy complains about this.

Cesium3DTilesSelection/src/ImplicitQuadtreeLoader.cpp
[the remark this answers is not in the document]
kring: Ah right, so the trick is to put `pAssetAccessor = pAssetAccessor` in the capture list. It doesn't matter that the parameter is passed by const reference, because the capture is by value. But by default (and for no good reason IMO), lambdas capture const things as const even when capturing by value. The silly looking `pAssetAccessor = pAssetAccessor` (which looks like self assignment but it isn't) will work around this problem.

.github/workflows/build.yml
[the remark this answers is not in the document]
kring: Oh right, of course, this is only for linting. That's fine.

azrogers:
@kring When you're back, I believe I've tackled the remaining review comments.

Cesium3DTilesContent/src/BatchTableToGltfStructuralMetadata.cpp
```diff
@@ -708,7 +732,7 @@ void updateExtensionWithJsonStringProperty(
       // Because serialized string json will add double quotations in the
       // buffer which is not needed by us, we will manually add the string to
       // the buffer
-      const auto& rapidjsonStr = it->IsNull() ? *noDataValue : it->GetString();
+      const auto& rapidjsonStr = it->GetString();
```
kring, line 735: The more I look at this code, the more I think it's wrong (both before and after this PR).

The intention of this if/else is, AFAICT, to make sure that property values are shown the same way they would be if serialize to JSON, except that strings should not be quoted. So we should go into the `if` when we have a non-string, and into the else when we have a string (so that we can remove the quotes).

Mostly this is straightforward, but there's on tricky bit. If the value is null, and the property has a default value, then we need to use that default value (which is always a string).

But that's not what this code is going to do. If the value is null, then `it->IsString()` will be false, and so we'll always go into the if block and never the else. That will end up writing the string `null` to the buffer instead of the default value.

I believe the correct code looks roughly like:

```cpp
if (it->IsString() || (it->IsNull() && noDataValue)) {
    // use string formatting that removes quotes
} else {
    // Use JSON formatting
}
```

I'll make this change and push it into this PR.

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -1306,10 +1330,9 @@ Tileset::TraversalDetails Tileset::_visitTile(
   // initially marked for RENDER here, it may later switch to REFINE as a
   // result of `mustContinueRefiningToDeeperTiles`.
   VisitTileAction action = VisitTileAction::Render;
-  if (unconditionallyRefine)
-    action = VisitTileAction::Refine;
-  else if (!meetsSse && !ancestorMeetsSse)
+  if (unconditionallyRefine || (!meetsSse && !ancestorMeetsSse)) {
```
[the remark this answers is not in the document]
kring: I'm going to revert this because I made a further change to it in #1059 that I think is better, and it will conflict if we change it here, too.

CesiumUtility/include/CesiumUtility/ReferenceCounted.h
```diff
@@ -99,6 +98,9 @@ class ReferenceCounted
   }
 
 private:
+  ReferenceCounted() noexcept = default;
+  friend T;
```
kring, line 102: This is clever! Did clang-tidy detect that this is a CRTP class and suggest this? If so, I'm impressed.

cmake/macros/configure_cesium_library.cmake
```diff
@@ -16,6 +16,12 @@ function(configure_cesium_library targetName)
         target_compile_options(${targetName} PRIVATE -Wno-dangling-reference)
     endif()
 
+    target_compile_definitions(
+        ${targetName}
+        PUBLIC
+            GLM_FORCE_CXX2A # Force GLM to use C++20 features, most importantly noexcept
```
kring, line 22: Are you sure this is needed? I could be wrong, but I don't think it's actually doing anything. We're using glm v1.0.1, and unlike older versions, that version appears to have no such define (it's C++20 now, not C++2a, right?). If we do need it, what is the impact on Cesium for Unreal? Because the other GLM_FORCE_* defines below we need to explicitly define in Unreal, too, or we get compiler errors.

cmake/macros/setup_clang_tidy.cmake
```diff
@@ -81,14 +81,17 @@ function(setup_clang_tidy)
             "Clang"
             RESULT_TOOL_PATH
             CLANG_TIDY_RUNNER_PATH)
+        list(APPEND CLANG_TIDY_RUNNER_PARAMS -extra-arg=-Wno-unknown-warning-option)
+        list(APPEND CLANG_TIDY_RUNNER_PARAMS -j14)
```
kring, line 85: This might lead to poor performance on CI, since it only has 2 cores.

tools/conform-includes.js
```diff
@@ -0,0 +1,89 @@
+const fs = require("fs");
```
kring, line 1: Nice! Would be good to have a comment at the top explaining what this is and how to use it.

kring:
I've reviewed everything now and it all looks great. Just a couple comments/questions above and this is ready to merge!

CesiumUtility/include/CesiumUtility/ReferenceCounted.h
azrogers: It did! The `bugprone-crtp-constructor-accessibility` check caught this.

cmake/macros/configure_cesium_library.cmake
azrogers: Nope, it was added to make sure GLM would have `noexcept` on its methods, but I disabled the check for that so it can be removed.

azrogers:
Addressed your comments @kring!

kring:
Thanks @azrogers! 🎉

PR #1049, opened 2024-12-23
Make SharedAssetDepot implementation not inline

kring:
This is just some really light refactoring of `SharedAssetDepot` before I start making some other changes to it. There are no functional changes here at all, it just moves the method definitions to be external to the class definition. This makes the class a little easier to lead, and avoids the compiler making these methods all inline for no real reason.


j9liu (approved):
Seems straightforward enough. Thanks @kring !

PR #1050, opened 2024-12-23
Ensure SharedAssetDepot stays alive while lock is held

kring:
This is a PR into #1049 so merge that first.

This fixes a problem I noticed while looking at CesiumgS/cesium-unity#540. The `SharedAssetDepot` keeps itself alive until the last asset it manages becomes unreferenced. It does this by maintaining an IntrusivePointer to itself in some situations. The problem is that it can release this reference while the current thread also holds the mutex. This will result in the SharedAssetDepot being destroyed (just) before the mutex is released, which causes undefined behavior (an assert in debug builds, to start).

This PR adds a test to demonstrate the problem. It fixes it by using a helper class to hold a reference to the depot while the mutex is locked, so that the depot always outlives the mutex lock.


j9liu:
Thanks @kring ! I think everything makes sense. Nothing jumps out at me so I'll go ahead and merge :smile:

PR #1053, opened 2025-01-06
Use Mermaid dark theme for dark mode

azrogers:
A simple fix to use the `dark` theme for Mermaid if the user has set a preference for dark mode. Currently, dark mode makes the diagrams unreadable as they remain in the light Mermaid `default` theme.


kring:
Thanks @azrogers! I've merged this to the `cesium.com` and `cesium.com-staging` branches so it is live on the web site. However, as I'm writing this, some pages are still in the 30 minute CloudFront cache so they'll still have the old style for a bit longer.

PR #1055, opened 2025-01-08
Add UrlTemplateRasterOverlay

### LINKED ISSUE #614: Add a UrlTemplateRasterOverlay
Similar to CesiumJS's UrlTemplateImageryProvider. This can be used for OpenStreetMap, WMTS, and a variety of other overlay tile sources.

azrogers:
Closes #614. We've had a lot of requests for an equivalent to CesiumJS's [UrlTemplateImageryProvider](https://cesium.com/learn/cesiumjs/ref-doc/UrlTemplateImageryProvider.html) to interface with services such as OpenStreetMap. This is an implementation of such a provider. It supports all of the template parameters of the UrlTemplateImageryProvider, with the exception of `{s}` - subdomain - as using multiple subdomains to work around browser limits isn't an issue we have to worry about.
- `{x}` - The tile X coordinate in the tiling scheme, where 0 is the westernmost tile.
- `{y}` - The tile Y coordinate in the tiling scheme, where 0 is the nothernmost tile.
- `{z}` - The level of the tile in the tiling scheme, where 0 is the root of the quadtree pyramid.
- `{reverseX}` - The tile X coordinate in the tiling scheme, where 0 is the easternmost tile.
- `{reverseY}` - The tile Y coordinate in the tiling scheme, where 0 is the southernmost tile.
- `{reverseZ}` - The tile Z coordinate in the tiling scheme, where 0 is equivalent to `urlTemplateOptions.maximumLevel`.
- `{westDegrees}` - The western edge of the tile in geodetic degrees.
- `{southDegrees}` - The southern edge of the tile in geodetic degrees.
- `{eastDegrees}` - The eastern edge of the tile in geodetic degrees.
- `{northDegrees}` - The northern edge of the tile in geodetic degrees.
- `{westProjected}` - The western edge of the tile in projected coordinates of the tiling scheme.
- `{southProjected}` - The southern edge of the tile in projected coordinates of the tiling scheme.
- `{eastProjected}` - The eastern edge of the tile in projected coordinates of the tiling scheme.
- `{northProjected}` - The northern edge of the tile in projected coordinates of the tiling scheme.
- `{width}` - The width of each tile in pixels.
- `{height}` - The height of each tile in pixels.


azrogers:
Added a utility header file to blit text to an `ImageAsset` using a simple bitmap font. We can revert it if it seems like clutter, but it helped me work out the issue I was struggling with so I at least wanted to get it into the git history if I need to find it later.

kring:
> Added a utility header file to blit text to an ImageAsset using a simple bitmap font.

This is cool, but the license makes me slightly nervous. Especially the "US Government Users Restricted Rights" part, which I don't think I've seen before and I'm not sure how to interpret. If we're going to leave this in, I think we need to run it by Larry. Given that it's test code, it might not be worth the trouble. 😬

CesiumRasterOverlays/include/CesiumRasterOverlays/UrlTemplateRasterOverlay.h
```diff
@@ -0,0 +1,137 @@
...
+      const std::string& name,
+      const std::string& url,
+      const std::vector<CesiumAsync::IAssetAccessor::THeader>& headers = {},
+      const UrlTemplateRasterOverlayOptions& urlTemplateOptions = {},
+      const RasterOverlayOptions& overlayOptions = {})
+      : RasterOverlay(name, overlayOptions),
+        _url(url),
+        _headers(headers),
+        _options(urlTemplateOptions) {}
+  virtual ~UrlTemplateRasterOverlay() override {}
```
kring, line 120: I think this can be removed. Or at least replaced with:

```
virtual ~UrlTemplateRasterOverlay() override = default;
```

CesiumRasterOverlays/src/DebugText.h
```diff
@@ -0,0 +1,242 @@
+#include <CesiumGltf/ImageAsset.h>
+
+#include <algorithm>
+#include <cstdint>
+
+namespace CesiumRasterOverlays {
+
+namespace {
```
kring, line 8: It's not great to put an anonymous namespace in a header file, because it means every cpp that includes this will get its own copy. The best approach is to add a `DebugText.cpp` and move as much as possible (pretty much all of it) into there.

CesiumRasterOverlays/src/UrlTemplateRasterOverlay.cpp
```diff
@@ -0,0 +1,189 @@
...
+using namespace CesiumGeospatial;
+using namespace CesiumUtility;
+
+namespace CesiumRasterOverlays {
+
+namespace {
+static bool caseInsensitiveCompare(std::string_view lhs, std::string_view rhs) {
+  return lhs.size() == rhs.size() &&
+         std::equal(lhs.begin(), lhs.end(), rhs.begin(), [](char a, char b) {
+           return std::tolower(a) == std::tolower(b);
```
kring, line 16: Dragons. To start with, std::tolower is locale-dependent, so it might not do what we want on all systems. For another, there's this note in the `std::tolower` docs:

> Similarly, they should not be directly used with standard algorithms when the iterator's value type is char or signed char. Instead, convert the value to unsigned char first:

Take a look at `HttpHeaders.h` / `.cpp`, which does case-insensitive comparisons in a way that might still be imperfect, but it should be at least a little more reliable.

CesiumRasterOverlays/src/UrlTemplateRasterOverlay.cpp
```diff
@@ -0,0 +1,189 @@
...
+    options.moreDetailAvailable = tileID.level < this->getMaximumLevel();
+
+    const GlobeRectangle unprojectedRect =
+        unprojectRectangleSimple(this->getProjection(), options.rectangle);
+
+    const std::string substitutedUrl = Uri::substituteTemplateParameters(
+        this->_url,
+        [this, tileID, &rectangle = options.rectangle, &unprojectedRect](
+            const std::string& placeholder) {
+          if (caseInsensitiveCompare(placeholder, "x")) {
```
kring, line 75: I think it would be worthwhile to build a std::map (with a case-insensitive key comparer) once and then use it for much faster lookups. This is what the WMTS tile provider does, minus the case-insensitivity.

kring (commented):
Looks great @azrogers! Just a few comments below.

azrogers:
> > Added a utility header file to blit text to an ImageAsset using a simple bitmap font.
> 
> This is cool, but the license makes me slightly nervous. Especially the "US Government Users Restricted Rights" part, which I don't think I've seen before and I'm not sure how to interpret. If we're going to leave this in, I think we need to run it by Larry. Given that it's test code, it might not be worth the trouble. 😬

Yeah, seems to me that it's the old SGI OpenGL license, which apparently wasn't considered a true "free software" license. Eventually OpenGL was relicensed under "SGI Free Software License B v2.0," which maybe applies to this code too, but I think it's worth cutting it just in case. I can come back to it in the future and see what I can do if it ends up being useful for other things (I could almost certainly just find another bitmap font to use).

azrogers:
@kring Removed radiansToDegrees and changed the naming to minimumX, maximumX, etc.

kring:
Thanks @azrogers!

PR #1056, opened 2025-01-08
Remove `GLM_FORCE_SIZE_T_LENGTH`

lilleyse:
This flag goes against our style guide recommendation to default to signed types: https://github.com/CesiumGS/cesium-native/blob/main/doc/topics/style-guide.md#-integers

Also, on a more practical note, this made it easier for me to integrate a library that uses glm that doesn't compile when this flag is set.


kring:
Thanks @lilleyse!

PR #1058, opened 2025-01-10
Support upsampling non-indexed triangles, triangle fans, and triangle strips

### LINKED ISSUE #918: Raster overlays don't work with non-indexed triangles, or with triangle fans or strips
The top of `upsamplePrimitiveForRasterOverlays` has this:

```cpp
  if (primitive.mode != MeshPrimitive::Mode::TRIANGLES ||
      primitive.indices < 0 ||
      primitive.indices >= static_cast<int>(parentModel.accessors.size())) {
    // Not indexed triangles, so we don't know how to divide this primitive
    // (yet). So remove it.
    return false;
  }
```

Fortunately, indexed triangle lists are the most common by far, so as far as I'm aware no one has ever actually complained about this. But still, we should round out this functionality.

azrogers:
Closes #918. Currently, raster overlay upsampling doesn't support primitives using the `TRIANGLE_STRIP` or `TRIANGLE_FAN` primitive modes, or primitives using the `TRIANGLES` mode that don't have indices. This change adds support for these less typical cases by more or less remapping them on the fly to indexed `TRIANGLES` primitives. This is done through the `IndicesViewRemapper` struct, which takes the place of the previous `AccessorView<TIndex>` type, accessing the internal `AccessorView` if one exists and remapping or generating new indices as needed. This allows the rest of the code to operate on the primitive as if it's the indexed `TRIANGLES` primitive that it expects.

### DIFF
--- a/CesiumRasterOverlays/src/RasterOverlayUtilities.cpp
+++ b/CesiumRasterOverlays/src/RasterOverlayUtilities.cpp
@@ -408,8 +408,8 @@ void copyMetadataTables(const Model& parentModel, Model& result);
  * @brief Helper struct for working with non-indexed triangles. Returns either
  * the indices from an index accessor view, or generates new indices.
  */
-template <typename TIndex> struct IndicesViewOrGenerator {
-  IndicesViewOrGenerator(
+template <typename TIndex> struct IndicesViewRemapper {
+  IndicesViewRemapper(
       const Model& model,
       const MeshPrimitive& primitive,
       int32_t primitiveIndices,
@@ -985,7 +985,7 @@ bool upsamplePrimitiveForRasterOverlays(
   const bool keepAboveV = !isSouthChild(childID);
 
   const AccessorView<glm::vec2> uvView(parentModel, uvAccessorIndex);
-  const IndicesViewOrGenerator<TIndex> indicesView(
+  const IndicesViewRemapper<TIndex> indicesView(
       parentModel,
       primitive,
       primitive.indices,
--- a/CesiumRasterOverlays/src/RasterOverlayUtilities.cpp
+++ b/CesiumRasterOverlays/src/RasterOverlayUtilities.cpp
@@ -403,6 +403,58 @@ bool isSouthChild(CesiumGeometry::UpsampledQuadtreeNode childID) noexcept {
 
 void copyImages(const Model& parentModel, Model& result);
 void copyMetadataTables(const Model& parentModel, Model& result);
+
+/**
+ * @brief Helper struct for working with non-indexed triangles. Returns either
+ * the indices from an index accessor view, or generates new indices.
+ */
+template <typename TIndex> struct IndicesViewOrGenerator {
+  IndicesViewOrGenerator(AccessorView<TIndex>&& accessorView_)
+      : accessorView(accessorView_), indicesCount(accessorView->size()) {}
+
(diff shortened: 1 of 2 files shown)


CesiumRasterOverlays/src/RasterOverlayUtilities.cpp
```diff
@@ -403,6 +403,89 @@ bool isSouthChild(CesiumGeometry::UpsampledQuadtreeNode childID) noexcept {
...
+  IndicesViewRemapper(
+      const Model& model,
+      const MeshPrimitive& primitive,
+      int32_t primitiveIndices,
+      int64_t numVertices)
+      : accessorView(std::nullopt),
+        indicesCount(0),
+        primitiveMode(primitive.mode) {
+    AccessorView<TIndex> view(model, primitiveIndices);
+    if (view.status() == AccessorViewStatus::Valid) {
```
kring, line 421: The logic here appears to be that if the indices accessor is invalid for any reason, the model is treated as if it's non-indexed. I think it would be better to only treat the model as non-indexed when `primitiveIndices` is less than zero, and pass through any other problems as actual errors.

kring (commented):
Thanks @azrogers! Just one comment below, and also please update CHANGES.md.

azrogers:
@kring Made some changes, now if the indices accessor is invalid and `primitiveIndices` isn't less than zero, `IndicesViewRemapper::status` will return the original (invalid) status of the accessor instead of treating it as non-indexed triangles.

CesiumRasterOverlays/src/RasterOverlayUtilities.cpp
```diff
@@ -1596,14 +1684,62 @@ bool upsamplePrimitiveForRasterOverlays(
...
+    const auto& positionIt = primitive.attributes.find("POSITION");
+    if (positionIt == primitive.attributes.end()) {
+      // No position buffer - nothing we can do here
+      return false;
+    }
+
+    // No indices buffer - pick the smallest indices type that will fit all
+    // vertices
+    const int64_t vertexCount =
+        parentModel.accessors[static_cast<size_t>(positionIt->second)].count;
```
kring, line 1706: I don't see anything handling the possibility of an out-of-range array access here, if the accessor referenced by the POSITION attribute is invalid. `parentModel.getSafe` may be useful.

CesiumRasterOverlays/src/RasterOverlayUtilities.cpp
```diff
@@ -1596,14 +1684,62 @@ bool upsamplePrimitiveForRasterOverlays(
...
+    if (vertexCount <= 0xff) {
```
kring, line 1707: There'a tricky glTF subtlety here. See https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#meshes-overview:

>indices accessor MUST NOT contain the maximum possible value for the component type used (i.e., 255 for unsigned bytes, 65535 for unsigned shorts, 4294967295 for unsigned ints).
> 
> Implementation Note
> The maximum values trigger primitive restart in some graphics APIs and would require client implementations to rebuild the index buffer.

So I believe this should be `vertexCount < 0xff` etc. instead of `<=`.

kring (commented):
Thanks for the changes @azrogers. A couple more small comments below.

azrogers:
@kring Fixed those two things!

kring:
Thanks @azrogers!

PR #1060, opened 2025-01-13
Add clang-tidy instructions to the developer setup guide

kring:
I also modified the script slightly to allow the location of `run-clang-tidy` to be specified explicitly.

@azrogers can you please add instructions for Linux and Windows?


azrogers:
@kring Added clang-tidy instructions for Linux and Windows. The Windows instructions are lackluster on account of not being able to build using the Ninja generator through anything but Visual Studio 2022 on Windows (at least, from what I've been able to achieve).

cmake/macros/setup_clang_tidy.cmake
```diff
@@ -52,36 +52,41 @@ function(setup_clang_tidy)
 
     # Generate compile_commands.json for clang-tidy to use.
     set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
+    # CMake has built-in support for running clang-tidy during the build
+    if(_ENABLE_CLANG_TIDY_ON_BUILD)
```
kring, line 57: Did you mean to move this outside the Ninja/Makefiles check below?

azrogers: I did, though looking it up, `CMAKE_CXX_CLANG_TIDY` only has an effect if on a Ninja or Makefile generator. I'll revert the change.

kring:
Thanks @azrogers!

PR #1059, opened 2025-01-13
Tile selection algorithm documentation, round 2

kring:
Fixes #981

PR #1061, opened 2025-01-15
Remove implicit-tiling.md, which is mostly incorrect at this point.

### LINKED ISSUE #986: Update / Remove `implicit-tiling.md`
`implicit-tiling.md` is outdated, so let's update it or consider removing it entirely. https://github.com/CesiumGS/cesium-native/blob/main/doc/implicit-tiling.md

kring:
Fixes #986 

This document rehashes a bit of the 3D Tiles spec, and makes some claims about the cesium-native implementation that are now completely wrong. It's not useful at all, so I've removed it.

There's a section about implicit tiling in #1059 which I think is sufficient.


azrogers:
Thanks @kring!

Issue #800, opened 2024-01-31
Use Doxygen's `\ref` syntax instead of JSDoc's `{@link}`


azrogers:
Throughout cesium-native, we currently use `{@link TypeName}` to represent a reference to another documented type or function. The problem with this syntax is that it's not Doxygen's syntax - it's JSDoc's, imported from CesiumJS. While Doxygen can interpret commands that start with `@`, [the `@link` command in Doxygen](https://www.doxygen.nl/manual/commands.html#cmdlink) is different than in JSDoc and expects a corresponding `@endlink` tag. This seems to currently work with Doxygen, the generated docs seem to handle the JSDoc tag just fine, it completely breaks Visual Studio's Intellisense, which will stop parsing the doc comment at the first `@link`.

We should instead use [`\ref`](https://www.doxygen.nl/manual/commands.html#cmdref), which is the equivalent to JSDoc's `{@link}`. This can likely be accomplished with a Regex find-and-replace, but should be done over the entire project at once.

PR #1062, opened 2025-01-15
Add Documentation section to style guide

j9liu:
I was writing this as part of https://github.com/CesiumGS/cesium-unreal/issues/1541, but since we're using Doxygen for both Native and Unreal, I figured I should move the section here.

This encapsulates what we do for public API documentation, just formally writing it down for the record.


doc/topics/style-guide.md
```diff
@@ -147,6 +147,56 @@ Not covered by C++ Core Guidelines.
...
 * Test files should be prefixed with `Test`, e.g. `TestModel` instead of `ModelTests`.
 
+## ✏️ Documentation
+
+Not covered by C++ Core Guidelines.
+
+Our documentation is generated by Doxygen before being published [online](https://cesium.com/learn/cesium-native/ref-doc). Therefore, any public API should be documented with Doxygen-compatible comments that use of the following tags (in the order listed):
+
+- `@brief`: Summarize the purpose of the class, function, etc.
+- `@details`: Use for lengthier explanations, after `@brief`.
```
lilleyse, line 157: I don't see `@details` being used much in native. Usually it's just any paragraphs after `@brief`. Is it different in Unreal?

doc/topics/style-guide.md
```diff
@@ -147,6 +147,56 @@ Not covered by C++ Core Guidelines.
...
+- `@tparam`: Describe template parameters.
+- `@param`: Denote and describe function parameters.
+- `@returns`: Describe the return value of a non-`void` function.
+- `@exception`: Describe any exceptions that a function throws.
+
+Additionally, make sure to:
+
+- Put comments **above** the thing being documented (instead of inline).
+- Use the `/** ... */` comment style (instead of `///`).
+- Use `\ref` when referencing other classes, functions, etc. in the Cesium Native API.
```
lilleyse, line 167: Would `@link` be a better choice? We seem to use both interchangeably and `@link` is more consistent with JSDoc.

doc/topics/style-guide.md
```diff
@@ -147,6 +147,56 @@ Not covered by C++ Core Guidelines.
...
+Not covered by C++ Core Guidelines.
+
+Our documentation is generated by Doxygen before being published [online](https://cesium.com/learn/cesium-native/ref-doc). Therefore, any public API should be documented with Doxygen-compatible comments that use of the following tags (in the order listed):
+
+- `@brief`: Summarize the purpose of the class, function, etc.
+- `@details`: Use for lengthier explanations, after `@brief`.
+- `@tparam`: Describe template parameters.
+- `@param`: Denote and describe function parameters.
+- `@returns`: Describe the return value of a non-`void` function.
+- `@exception`: Describe any exceptions that a function throws.
```
azrogers, line 161: We've also got a few examples in the codebase of `@remarks`, such as at the top of `CesiumUtility/SharedAsset.h`. It's useful to write a bunch of information that isn't necessary to get a picture of what the function does, but is useful to get the full scoop on its inner workings. Might be worth adding.

azrogers (commented):
Looks good! Just one small comment.

doc/topics/style-guide.md
```diff
@@ -147,6 +147,56 @@ Not covered by C++ Core Guidelines.
 
 * Test files should be prefixed with `Test`, e.g. `TestModel` instead of `ModelTests`.
 
+## ✏️ Documentation
+
+Not covered by C++ Core Guidelines.
+
+Our documentation is generated by Doxygen before being published [online](https://cesium.com/learn/cesium-native/ref-doc). Therefore, any public API should be documented with Doxygen-compatible comments that use of the following tags (in the order listed):
```
lilleyse, line 154: Worth including `@copydoc` and `@warning`?

doc/topics/style-guide.md
lilleyse, line 161: I sometimes see `@throws`. But `@exception` is more common in native and CesiumJS so I'm good with that.

doc/topics/style-guide.md
```diff
@@ -147,6 +147,56 @@ Not covered by C++ Core Guidelines.
...
+
+- `@brief`: Summarize the purpose of the class, function, etc.
+- `@details`: Use for lengthier explanations, after `@brief`.
+- `@tparam`: Describe template parameters.
+- `@param`: Denote and describe function parameters.
+- `@returns`: Describe the return value of a non-`void` function.
```
lilleyse, line 160: I sometimes see `@return` (singular). I like the plural version more.

lilleyse (commented):
Thanks for adding this @j9liu!

doc/topics/style-guide.md
j9liu: Hm it's not different in Unreal. I guess I included it because I saw `@details` used in CesiumGltfWriter.

https://github.com/CesiumGS/cesium-native/blob/1780e8769942c61cffd1d0fd6eb452703a7a3486/CesiumGltfWriter/include/CesiumGltfWriter/GltfWriter.h#L76

But that's literally the only instance where we use `@details` 😅  And based on @azrogers 's comment it looks like we prefer to use `@remarks` instead. I'll replace this.

doc/topics/style-guide.md
j9liu: Thanks for pointing that out @azrogers ! I'll let it replace `@details` since we don't actually use the latter too often.

doc/topics/style-guide.md
j9liu: We initially were using @link, but I think it wasn't actually working when we generated documentation? 😦 hence #800

azrogers: Technically `@ref` and `@link` aren't interchangeable. `\ref` can have an optional link text specified after the ref name, like `@ref CesiumUtility::Uri "URI parsing"`. Whereas `@link` is supposed to be paired with `@endlink`, like `@link CesiumUtility::Uri URI parsing @endlink`. Because we use the JSDoc inline link tag, `{@link name}`, Doxygen somehow seems to understand this and doesn't require us to use `@endlink`. But I'd prefer to use the suggested Doxygen syntax rather than this JSDoc compatibility feature.

doc/topics/style-guide.md
azrogers: We use both in the codebase, so I think they're worth it.

doc/topics/style-guide.md
azrogers: We should definitely standardize, we seem to use both.

doc/topics/style-guide.md
j9liu: @azrogers do you know if `@ref` works the same as `\ref`? If so, I think we should stick with `@ref` so it matches the rest of the tags we use

lilleyse: @azrogers @j9liu thanks and good to know, I'll stop using `@link` and use `@ref` instead

doc/topics/style-guide.md
j9liu: I'll go with `@throws`. Seems like there are more occurrences in Native, and it matches `@returns`

j9liu:
Thanks for the feedback @azrogers and @lilleyse ! I think I addressed all your comments, let me know if you catch anything else.

doc/topics/style-guide.md
azrogers: @j9liu `@ref` and `\ref` work the same. The `@` and `\` are interchangeable prefixes (I accidentally used `\ref` where I meant to use `@ref` in that comment). Personally, I like to use `@` in the directives at the start of the line and `\` for the directives in the middle of the line (like `@brief Some text with a \ref Item tag`) to distinguish between inline and non-inline tags. But that's just me and it's probably best if we keep it consistent instead.

doc/topics/style-guide.md
javagl: I think it's still singular, but simple present, as in "This function return**s** a this-and-that value". That nit being picked 😁 : It seems like tools generally cannot agree here. TypeScript uses `@returns`. Java uses `@return`. Doxygen supports both. No strong preference, as long as it's not `\return...`.

kring:
I'm merging this because it looks good to me and it seems all the previous review feedback has been addressed. But feel free to comment or write issues if you still have any feedback on it @azrogers and @lilleyse.

Issue #973, opened 2024-10-31
Can't load glTF URIs with unicode characters


lilleyse:
According to the glTF [URIs](https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/Specification.adoc#28-uris) section:

> Paths with non-ASCII characters MAY be written as-is, with JSON string escaping, or with percent-encoding; all these options are valid. For example, the following three paths point to the same resource:
>
> ```
> {
>    "images": [
>        {
>           "uri": "grande_sphère.png"
>        },
>        {
>            "uri": "grande_sph\u00E8re.png"
>        },
>        {
>            "uri": "grande_sph%C3%A8re.png"
>        }
>    ]
>}
>```

But when I try to use `GltfReader` to load a glTF with a URI to 🐶.bin I get a runtime error:

```
std::runtime_error caught: GET `🐶.bin` failed: Couldn't resolve host name
```

[🐶.gltf.zip](https://github.com/user-attachments/files/17591637/gltf.zip)

This may be a limitation in `uriparser` which only parses [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) URIs which are ASCII-based. It does not handle [RFC 3987](https://datatracker.ietf.org/doc/html/rfc3987) IRIs which allow Unicode characters verbatim.

kring:
Your explanation makes sense to me @lilleyse. I wonder if there's a better library for our purposes than uriparser. I couldn't find one when I originally selected it for cesium-native, but that was over four years ago now.

azrogers:
I've investigated a few alternatives to uriparser that each come with their own drawbacks:
- [skyr-url](https://github.com/cpp-netlib/url) is a more modern C++ URL parser as compared to uriparser's C API. It implements the [WhatWG URL specification](https://url.spec.whatwg.org/#url-class), which covers both RFC 3986 and RFC 3987 URLs. However, skyr-url seems to be unmaintained (last commit December 2023, compared to yesterday for uriparser). It also fails for URLs that uriparser can parse. For example, one of the test cases includes the URLs "test" and "buffer". skyr-url, perhaps correctly, believes these URLs to be invalid. We could probably work around this somehow, but that's a pain.
[...]

PR #1063, opened 2025-01-15
Allow URLs with non-ASCII characters to be parsed

azrogers:
As described in #973, glTFs that include URIs with non-ASCII characters in them don't currently work with Cesium Native. That's because the library we use for URI parsing, `uriparser`, follows the RFC 3986 standard. According to RFC 3986, only ASCII characters are allowed in URLs, and all other characters must be escaped. Unicode support came later, in RFC 3987 and now the WhatWG URL specification which seems to be the modern standard browsers aim to support. The glTF spec allows Unicode characters in URIs "as-is," meaning a strictly RFC 3986-compliant parser won't do the job. One option would be to substitute a different, WhatWG-compliant parser in its place - but as I described in [this comment](https://github.com/CesiumGS/cesium-native/issues/973#issuecomment-2593482114), this introduces more problems than it solves. 

Instead, the solution I went with was to encode all non-ASCII characters in the string before passing it to `uriparser`, then decoding them again before returning from the method. This seems to work flawlessly - `uriparser` gets the RFC 3986-compliant URLs it expects, and the user gets the WhatWG-compliant URLs they expect. Unfortunately, this solution does introduce an extra layer of string copies into URL parsing. This isn't ideal, but considering URLs are usually fairly short, I think it's a worthwhile tradeoff to gain this compliance with the glTF spec.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Fixed a crash in `GltfWriter` that would happen when the `EXT_structural_metadata` `schema` property was null.
+- `CesiumUtility::Uri` now fully supports non-ASCII characters.
 
 ### v0.43.0 - 2025-01-02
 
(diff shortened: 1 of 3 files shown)


azrogers:
Closed in favor of #1072.

PR #1064, opened 2025-01-15
Use doctest instead of Catch2 for tests

### LINKED ISSUE #69: Evaluate doctest instead of Catch2 for unit tests
Test build times are already noticeably long and we don't really have very many tests, so it's worth taking a look at a promising alternative.

azrogers:
Closes #69. We've been intending for quite a while to switch to doctest for tests instead of our current test framework, Catch2. Not only is doctest the test framework that other Cesium projects use, but the compile times are also quite a bit faster. It would be good to get this out of the way while we have the chance - as @kring mentioned, it will only get more difficult as time goes on.


azrogers:
I'll note that while the point of this is mostly to improve test build times, the test builds seem *slower* on CI. Subjectively, they're faster on my machine now, so maybe this is just an artifact of doctest not yet being cached? Not sure. A larger sample size is probably needed.

Cesium3DTilesReader/test/TestTilesetReader.cpp
```diff
@@ -15,6 +13,24 @@
 #include <string>
 #include <vector>
 
+using namespace doctest;
+
+bool compareVectors(
```
kring, line 18: This should be in the anonymous namespace. Or maybe moved to `CesiumNativeTests`?

CesiumGltf/test/TestPropertyTablePropertyView.cpp
```diff
@@ -32,16 +32,18 @@
...
 
-#include <catch2/catch.hpp>
-#include <catch2/catch_test_macros.hpp>
+#include <doctest/doctest.h>
 
 #include <bitset>
 #include <climits>
 #include <cstddef>
 #include <cstring>
+#include <ostream>
```
kring, line 41: Is this header needed?

CesiumGltf/test/TestPropertyView.cpp
```diff
@@ -21,13 +21,14 @@
 #include <glm/fwd.hpp>
 
 #include <cstdint>
+#include <ostream>
```
kring, line 24: Not sure why ostream is needed here, either.

kring (commented):
Thanks for doing this @azrogers! Minor comments...

CesiumGltf/test/TestPropertyView.cpp
azrogers: Don't think I could really tell you either, but it *is* apparently needed: https://github.com/CesiumGS/cesium-native/actions/runs/12797730812/job/35680280884

azrogers:
Ran though this script on this branch and the current `main`:
```sh
rm benchmark.log
for i in $(seq 1 10);
do
    cmake -B build_benchmark -S .
    /usr/bin/time -f "%e" -o benchmark.log -a cmake --build build_benchmark -j22
    rm -rf build_benchmark
done

awk '{s+=$1}END{print "average time:", s/NR}' RS="\n" benchmark.log
```

Average time per build on this branch (doctest): 210.464 seconds
Average time per build on `main` (Catch2): 230.288 seconds

So doctest *does* speed up the build, but only by around 20 seconds. That probably indicates our main slowdowns while building lie outside of the tests. I discovered `ClangBuildAnalyzer` while looking for how to benchmark this which could be worth trying out in the future to save build time elsewhere (#1066). But I think a 20 second speedup is good enough to make this change worthwhile.

kring:
That's a nice improvement from just changing the test framework! Thanks for doing the test. Merging!

PR #1065, opened 2025-01-16
Add raster overlay documentation

kring:
This is a PR into #1055 so merge that first.

This is meant to cover:

* What are raster overlays, and how can we use them?
* How do we develop a new type of raster overlay?
* How does the raster overlay system work under-the-hood?

A lot more detail is possible for all three, but I hope to have hit the most important points.


azrogers:
Looks good! Added the doc to the Developer Resources page (as well as ordering the raster overlays in alphabetical order). Thanks @kring!

PR #1068, opened 2025-01-16
Link to Geospatial Guide instead of including a glossary

azrogers:
I added the "Glossary" page as a temporary stand-in while we were working on getting the [Geospatial Guide](https://github.com/CesiumGS/community/blob/main/GeospatialGuide/README.md) together. Now that it's published, it's better that we link to the more detailed and informative explanations it contains instead of trying to reproduce them in our own glossary. To this end, I've created a tag file for the community repo. [Tag files](https://www.doxygen.nl/manual/external.html) are meant to be used to link between Doxygen docs (which we're currently using them for in Cesium for Unreal, to link back to the Native docs). But that's not a requirement, and we can just as easily write them ourselves to link to whatever pages we want. This change adds a tag file for the community repo, containing the information required to link to the headers on ECEF coordinates and ellipsoids within the geospatial guide. 

This is a proof of concept. If we like it, we could easily expand the tag file and reference it throughout our documentation. But at the very least I wanted to get rid of the single-item glossary I'd created.


kring:
This looks like a nice system to me @azrogers. Let's run with it!

PR #1074, opened 2025-01-20
Fix race condition in SharedAssetDepot

### LINKED ISSUE #1073: Assertion failures caused by SharedAssetDepot race condition
I see this pretty reliably in debug builds of Cesium for Unreal on an ARM MacBook by just flying around for a few seconds in PIE in Level 1 of the Samples, perhaps because ARM has a weaker memory model than x64 (see [here](https://developer.apple.com/documentation/apple-silicon/addressing-architectural-differences-in-your-macos-code#Synchronize-Access-to-Shared-Data-in-Memory)), or perhaps just because it's fast!

The cause is a sequence like this:

| *Time* | *Thread A*                    | *Thread B*                  |
|------|-------------------------------|-----------------------------|
|    0   | Decrements reference count, now 0.   |  getOrCreate - lock depot mutex.    |
|  1     | Attempts to lock depot mutex.              | Increments reference count, now 1. |
|  2     |                                                     | unmarkDeletionCandidateUnderLock |
|  3     |                                                      | 💥 Assertion failure because asset is not a deletion candidate!💥 |
|  4     |                                                      | Releases lock |
|  5     | Successfully acquires lock        |  |
|  6     | 💥 Added to deletion candidates even though it's now referenced. 💥 | |

I believe this can be solved with two parts:
1. Don't assert in `unmarkDeletionCandidateUnderLock` that the asset is in the deletion candidates list, cause it might not be yet.
2. In `markDeletionCandidateUnderLock`, don't add the asset to the deletion candidates if its reference count is greater than zero.

This should be reliable because mark/umark are serialized by the mutex, and because an addReference that takes the reference count from 0 to 1 can _only_ happen under the mutex in a call to `getOrCreate`, so it is serialized with both of these as well (unlike the releaseReference that takes the count from 1 to 0, which can happen whenever!).

kring:
Fixes #1073


j9liu:
@azrogers do you mind reviewing since you're most familiar with this code?

azrogers:
Thanks for the change @kring! I believe I've seen this in practice on my machine, but it was so intermittent that I chalked it up to something wrong with the code I was testing. It'll be good to have this fixed.

PR #1075, opened 2025-01-21
Fix macOS / ARM test failures

kring:
This is a PR into #1074 so merge that first.

Our test suite had several failures on an Apple Silicon Mac. This PR fixes them.

## Math::rotation

These tests were failing, but they were testing functionality that we don't really need anyway. As far as I can tell, `Math::rotation` does the same thing as `glm::rotation`. So rather than try to fix anything, I just removed the implementation and the tests. I also removed `Math::perpVector` for the same reason (`glm::perp`).

Both of these glm implementations are "experimental" though. We were already using some experimental glm stuff, which requires defining `GLM_ENABLE_EXPERIMENTAL`, but we were doing it in an ad hoc fashion. In this PR it's just defined in one place, alongside all the other GLM defines.

## losslessNarrow

This was a total mess. Our implementation came roughly from GSL, but it really didn't seem to be reliable. Especially in release builds, Clang in the latest Xcode seemed to be just totally optimizing out some of the checks for a lossless conversion. Some of the conversions it was doing, such as casting a double to an integer when the double is outside the range of the integer, are technically undefined behavior. The assumption was that it would at least do "something" and then we would check (by casting back) if it was correct. But technically a compiler can do anything in the face of undefined behavior, and Clang seemed to be doing something more like "pretend it never happened and optimize out the whole cast", and then our "check" would say "yep everything is fine!" because, if you skip the cast entirely, then sure the (non-) conversion is lossless, right? I saw cases where something like `static_cast<double>(static_cast<uint8_t>(257.0)) == 257.0` (not all on one line like that) was evaluated to true after the optimizer got done with it... which is just..... missing the point.

So anyway I rewrote losslessNarrow to explicitly check ranges. And moved all the implementation into a .cpp file to decrease the odds of it being inlined and optimized into oblivion. And wrote tests to demonstrate it's all working.

It's possible that other parts of cesium-native that used this, such as the metadata system, would have had bad behavior on some platforms in the previous implementation. For example, they could have ended up inadvertently doing some lossy type conversions, or breaking values entirely when attempting to convert them.


timoore:
What were the failures in Math::rotation and Math::perp? These were used until the code that used them turned out to be unsound, but I had thought that these functions were at least as good, if not better than, the glm ones.

kring:
@timoore on the ARM MacBook, the `Math::rotation` tests were failing that asserted that the quaternion `w` component was 0. It looked to me like the test was bad, rather than the implementation, but I didn't look too deeply.

kring:
Thanks @j9liu! I've addressed your comments and merged in main.

j9liu:
Thanks @kring ! Merging now :smile:

PR #1057, opened 2025-01-09
Allow specifying request headers in `TilesetOptions`

azrogers:
To support use cases like #738, this change allows specifying request headers in `TilesetOptions` that will be passed to the `TilesetContentManager`.

PR #1076, opened 2025-01-21
Allow specifying request headers in `RasterOverlayOptions`

### LINKED ISSUE #1067: Allow request headers to be specified in `RasterOverlayOptions`
We should implement an equivalent to #1057 for raster overlays to allow arbitrary HTTP headers to be specified for raster overlay loading. This functionality already exists for many of the raster overlay implementations, but there's no reason it can't be implemented on the base `RasterOverlayTileProvider` level instead.

azrogers:
Fixes #1067. This is a companion to #1057, applying a similar change to raster overlays. One wrinkle in the implementation is that some raster overlays already accept headers in their constructor. Not only would it be best to avoid changing the public API, but `IonRasterOverlay` in particular relies on this functionality being present so it can pass the `Authorization` header to the underlying `TileMapServiceRasterOverlay`. To support this, I've added a `mergeHeaders` method that allows the headers specified in the constructor to override any headers with the same name specified in the `RasterOverlayOptions`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Additions :tada:
 
 - Added `requestHeaders` field to `TilesetOptions` to allow per-tileset request headers to be specified.
+- Added `requestHeaders` field to `RasterOverlayOptions` to allow per-raster-overlay request headers to be specified.
 
 ### v0.44.0 - 2025-02-03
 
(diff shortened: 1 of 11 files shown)


azrogers:
Closing in favor of using the existing constructor parameters for headers, as used in CesiumGS/cesium-unreal#1597.

PR #1077, opened 2025-01-22
Use pre-release version of ada-url

kring:
This is a PR into #1072 that uses the current head of main of ada via an overlay port, allowing us to use the new URLPattern functionality.


kring:
Well, this seems to be broken on Linux. Good enough to play around with though hopefully, and if it looks promising we can figure out the next step.

azrogers:
The Linux build failure appears to be a compiler / standard library bug. [Testing it out on godbolt](https://godbolt.org/z/adWrnoE95), this error only comes up with GCC versions lower than 13.1.

anonrig:
Hello folks. Ada author here. URLPattern implementation requires a regex provider. I strongly recommend not using std_regex_provider since it's a security nightmare. I'm planning on only enabling it on ADA_TESTING flag. 

Happy to see that you're adopting Ada!

anonrig:
It seems you're compiling Ada with C++17. Main branch currently supports C++20

kring:
Hey @anonrig, thanks for chiming in. And thanks for Ada as well, it looks really nice!

It's looking right now like we'll go back to the current released version of Ada on vcpkg and avoid the URLPattern functionality.

We have a use-case where we have "URLs" with placeholders (they're not actually valid URLs) that look like `https://example.com/{z}/{x}/{y}.png`. We need to manipulate these as if they're valid, such as adding query parameters, and only later substitute real values for the placeholders in curly brackets. So URLPattern isn't quite the same use-case (it's almost the opposite), but we were exploring whether it might be made to work. Would love to have your opinion on this if you have ideas, but otherwise we'll likely solve it by encoding the invalid bits when manipulating the URLs and then decoding them later.

anonrig:
Yes. URLPattern doesn't give you the ability to manipulate the original input. I suggest making url valid and later parsing it. By default Ada will encode it, and you can use ada::idna (which is included with Ada) to decode it.

PR #1078, opened 2025-01-22
Return the highest sampled height, not the first one.

kring:
When `TilesetHeightQuery` sees multiple tiles that overlap a given location, it is supposed to return the highest height. But it was doing this incorrectly, so it would always end up returning the first one instead.

Reported here:
https://community.cesium.com/t/how-to-handle-overlapping-tiles-with-sampleheightmostdetailed/38109


j9liu:
Thanks @kring ! I didn't test this out personally but I trust the test that you added.

PR #1079, opened 2025-01-22
Add helper functions to Cesium3DTiles::Tileset

lilleyse:
Added some helper functions to `Cesium3DTiles::Tileset`

* `forEachTile` (similar to `CesiumGltf::Model::forEachNodeInScene`)
* `forEachContent` (similar to `CesiumGltf::Model::forEachPrimitiveInScene`)
* `addExtensionUsed`
* `addExtensionRequired`
* `removeExtensionUsed`
* `removeExtensionRequired`
* `isExtensionUsed`
* `isExtensionRequired`

### DIFF
--- a/Cesium3DTiles/include/Cesium3DTiles/Tileset.h
+++ b/Cesium3DTiles/include/Cesium3DTiles/Tileset.h
@@ -24,6 +24,9 @@ struct CESIUM3DTILES_API Tileset : public TilesetSpec {
   /**
    * @brief Apply the given callback to all tiles.
    *
+   * This will only invoke the callback for explicit tiles. It won't follow
+   * external tilesets or implicit tiles.
+   *
    * @param callback The callback to apply
    */
   void forEachTile(std::function<ForEachTileCallback>&& callback);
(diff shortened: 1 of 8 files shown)


Cesium3DTiles/include/Cesium3DTiles/Tileset.h
```diff
@@ -0,0 +1,129 @@
...
+  /**
+   * @brief A callback function for {@link forEachTile}.
+   */
+  typedef void ForEachTileCallback(
+      Tileset& tileset,
+      Tile& tile,
+      const glm::dmat4& transform);
+
+  /**
+   * @brief Apply the given callback to all tiles.
```
kring, line 25: I think it's worth mentioning in the doc that this will only invoke the callback for explicit tiles. It won't follow external tilesets, and it won't reify implicit tiles.

kring (commented):
Thanks @lilleyse, looks great! Just one doc suggestion.

Cesium3DTiles/include/Cesium3DTiles/Tileset.h
lilleyse: Thanks for the suggestion, added a note.

kring:
Thanks @lilleyse!

PR #1080, opened 2025-01-23
Document current deprecation policy

j9liu:
This puts a blurb in the style guide about our current deprecation policy (in that we don't really have one right now). Not sure if the style guide is the right place, but it seemed pretty short for a standalone document.


kring:
The content looks great @j9liu, thanks for writing it! I think I might like to move it to a separate topic, though, rather than putting it in the style guide. Because the style guide is kind of applicable across other projects at Cesium, but this deprecation policy is specific to Native. What do you think? I'm happy to make that change, if you don't have any objections.

lilleyse:
On a related note, should the style guide move to https://github.com/CesiumGs/community/?

kring:
I moved the deprecation info to README.md. Thanks @j9liu!

> On a related note, should the style guide move to https://github.com/CesiumGs/community/?

That's ok by me @lilleyse!

PR #1081, opened 2025-01-23
Fix error with `catch2` header

azrogers:
Foolishly I merged the raster overlay doc changes without checking the CI results, thinking it was just documentation so it wouldn't be an issue. Unfortunately, one of the example bits of code was using the catch2 header, which has been removed in favor of doctest! This fixes the error.

### DIFF
--- a/CesiumRasterOverlays/test/ExamplesRasterOverlays.cpp
+++ b/CesiumRasterOverlays/test/ExamplesRasterOverlays.cpp
@@ -1,7 +1,7 @@
 #include <CesiumRasterOverlays/RasterOverlay.h>
 #include <CesiumRasterOverlays/UrlTemplateRasterOverlay.h>
 
-#include <catch2/catch_test_macros.hpp>
+#include <doctest/doctest.h>
 
 using namespace CesiumRasterOverlays;
 


j9liu:
Well the change is straightforward @azrogers , and I would have merged already, but CI is weirdly failing for some of the builds because it can't download the KTX library?

[This](https://github.com/CesiumGS/cesium-native/actions/runs/12935699159/job/36080509811#step:10:73) is from `ubuntu-22.04/gcc/Debug`... 
```
Installing 2/2 ktx:x64-linux@4.3.2...
Building ktx:x64-linux@4.3.2...
Downloading KhronosGroup-KTX-Software-v4.3.2.tar.gz
Failed to download KhronosGroup-KTX-Software-v4.3.2.tar.gz.
```

But this is weirdly not a problem for some of the actions, like `ubuntu-22.04/gcc/RelWithDebInfo` 🤔 Not sure what's going on, but it doesn't seem like it should be related to the content of this PR?

azrogers:
Re-ran the failed jobs, and *some* of them succeeded, but others failed with the same kind of errors. It looks like curl is just failing to download all of the KTX repository at the given tag for one reason or another. GitHub Status doesn't show any issues. Maybe it's some sort of rate limiting on downloading the archive of a repository? I would think that would show up for the rest of the packages though...

azrogers:
Worth noting this issue seems to be showing up on other branches on CI now, including ones that don't have the raster overlay doc changes. I think it's something external to the project that's messing us up here, but I can't think of what that might be.

azrogers:
Third time's the charm. [Re-ran it again](https://github.com/CesiumGS/cesium-native/actions/runs/12935718529) and everything's green. Must've been the wind.

PR #1082, opened 2025-01-23
Report error to the user when layer.json .terrain file fails to load

### LINKED ISSUE #975: No error reported to the user when quantized-mesh / layer.json tiles fail to load
I saw this in Cesium for Unity, but it's most likely a cesium-native problem, not specific to Unity.

If a particular tile from a tileset.json is missing, users will see a message in the log:

> [2024-11-04 18:13:29.274] [error] [TilesetJsonLoader.cpp:924] Received status code 404 for tile content http://localhost:8080/batchedColors.b3dm

However, if one of the `.terrain` files in quantized-mesh / layer.json is missing, the error is silently ignored, leaving users wondering why there's a hole in their terrain.

Originally reported by a user here:
https://community.cesium.com/t/tile-not-rendering-which-closer-to-camera/36033

azrogers:
Fixes #975. Currently, errors and warnings are added to the `QuantizedMeshLoadResult` when a .terrain file fails to load, but nothing every logs those errors and warnings for the user to see. This PR makes sure that they get logged, along with adding tests for `LayerJsonTerrainLoader` and `TilesetJsonLoader` to ensure that errors are logged correctly when loading a tile encounters a 404.


kring:
Thanks @azrogers, looks good! Can you please just add a note to CHANGES.md? It's worth mentioning there as a bug fix, I think.

azrogers:
@kring Done!

kring:
Thanks @azrogers!

PR #1083, opened 2025-01-24
Fix crash while creating culling volume when the camera is very far from the globe

### LINKED ISSUE #633: Exception creating culling volume when camera is very far from the globe
When the camera is very far away from the globe, cesium-native throws an exception while attempting to construct a `Plane` because the normal is NaN.

I ran into this while testing globe scaling in Unity. If we zoom out so that we can see the whole globe, and then scale the globe down until it's very tiny (e.g. scale factor of 1e-12), that's equivalent to the camera being extraordinarily far away from the globe. In my testing, I was seeing the camera-to-globe distance be about 1e18 meters at the time of the crash. At that distance, the globe is much too small to see, but we shouldn't crash.

The problem is in the `createCullingVolume` function in `CullingVolume.cpp`. For tile selection, cesium-native doesn't use the near and far planes for culling. In fact, they're not even provided to the selection algorithm. But this `createCullingVolume` function, which was ported from CesiumJS, uses the near plane distance as part of the computation of the top, bottom, left, and right planes of the frustum. Since the overall goal here is just to form a direction, I thought the actual near plane value wouldn't matter and just hard-coded it to 1.0.

This is mostly a fine assumption, _except_ if the position value is so huge that adding 1.0 to it results in the same value. When that happens, our normal vector has zero magnitude and we throw an exception while trying to create a Plane for it.

I'm not sure if there's a smarter way to compute these planes, but a simple way to fix this bug is to use the same algorithm but simply to choose the "near plane" distance more carefully. Something like "half the distance to the globe" would probably work fine. Maybe put a minimum on it so that a camera at the center of the Earth won't cause a different sort of problem.

azrogers:
Fixes #633. As Kevin describes in the linked issue, the calculation to create a culling volume currently uses `1.0` as the near plane distance. This is fine for most operations we need to do, but causes problems when the value of `position` is *really* large. In this case, adding `1.0` to the position results in the same position value, so the normal is calculated with a magnitude of 0 and an exception is thrown when trying to create a plane with this normal. 

This change uses `std::nextafter` to obtain the next representable value based on the position's magnitude, subtracting the position's magnitude to obtain the minimum value we need to add to the position to not result in the same value. We only use this value if it's greater than 1.0, so the previous near plane behavior is preserved in most cases (otherwise the near plane distance would be `std::numeric_limits<double>::epsilon()` for values close to zero!). 

This should work to prevent a crash for any positions with a magnitude less than `std::numeric_limits<double>::max()`. If the position is at the max double value, we will still crash, but I'm not sure there's any tricks we can do to make the math work out in that case.

In the process, I noticed `CullingVolume` was in the `Cesium3DTilesSelection` namespace even though the files are in `CesiumGeometry`. I assume they were moved here at some point, but the namespace wasn't updated (was this on purpose or accidental?). I've updated the namespace to `CesiumGeometry`.


kring:
Thanks @azrogers!

PR #1084, opened 2025-01-28
ada-url: Substitute layer.json placeholders before treating as URLs

kring:
This is a PR into #1072.

It bothered me that the URL placeholders (like `{z}`) in the tile URL templates in layer.json were escaped, and then `substituteTemplateParameters` had to deal with this potential escaping. So this is an attempt to avoid that in a fairly simple way: substitute the values into the template URL before treating it as a URL. 

I think this is fairly reasonable. A URL with placeholders is not a valid URL, so don't try to pretend it is. If this looks reasonable to you @azrogers, then I think the changes to `substituteTemplateParameters` to allow escaped placeholders can be reverted.

The other places we were using placeholders in URLs (`UrlTemplateRasterOverlay` and `WebMapTileServiceRasterOverlay`) were already doing this, perhaps accidentally.


azrogers:
Thanks @kring!

PR #1088, opened 2025-01-30
Add parent property to `CesiumGltf::Class` and `Cesium3DTiles::Class`

lilleyse:
Updating the generated code, which now includes a `parent` property on `CesiumGltf::Class` and `Cesium3DTiles::Class`. See https://github.com/CesiumGS/glTF/pull/74 for more details.

Since this is experimental I don't think it's worth including in `CHANGES.md`, but let me know otherwise.

Also, `npm run generate-3d-tiles` now targets `3d-tiles/cesium-native` instead of `3d-tiles/main`.


kring:
Thanks @lilleyse!

PR #1089, opened 2025-01-31
Download KTX via vcpkg_from_git instead of vcpkg_from_github

kring:
`vcpkg_from_github` downloads a source ZIP file, which has been extremely unreliable lately. The leading theory is that it has something to do with GitHub packing the ZIP on demand and including files from LFS, which is either generally unreliable or is bumping into some kind of quota. Using git instead, and avoiding LFS, will hopefully be more reliable.

If CI is happy with this, I'm going to merge it right away and update cesium-unreal and cesium-unity to use it, too.


GeorgeS2019:
@kring 
We need ktx@4.4.0 instead of ktx@4.3.2

https://github.com/KhronosGroup/KTX-Software/releases/tag/v4.4.0

azrogers:
@GeorgeS2019 This is an old pull request that seems unrelated to your version issue. Can you open a new GitHub issue with information about your problem and why upgrading KTX is necessary?

PR #1091, opened 2025-01-31
Add <algorithm> to Cesium3DTiles/test/TestTileset.cpp

timoore:
The new use of std::find in TestTileset.cpp  was failing on Fedora 41.


lilleyse:
Thanks @timoore. Those were my changes so I'll go ahead and merge.

PR #1097, opened 2025-02-04
Upgrade to Doxygen 1.13.2

azrogers:
Apparently, some point in the Doxygen releases from 1.12.0 to 1.13.2, our documentation build started breaking. I've filed a bug report for the main issue here - a bug in predefined macro expansion, doxygen/doxygen#11393 - but it turns out we can get away with just disabling `EXPAND_ONLY_PREDEF`, which means the issue doesn't appear. It seems whatever reason there was to turn this on in the first place is no longer present, as the documentation builds fine without it enabled. 

Besides fixing the documentation build for 1.13.2 and updating the Doxygen version on CI, this PR also changes the way the `EXAMPLE_PATH` variable is set. Previously, we gave it the entire source tree as the example path, which led to every file in the directory (including `build`, `node_modules`, and any other cruft) being parsed before the documentation could be generated. Apparently, `EXCLUDE_PATTERNS` kicks in *after* this step in the process, so it had no effect. Instead, we glob every `Cesium*` folder and add each of those to the path to avoid any extra build files from slowing down our doc build.


azrogers:
...and just realized I hadn't fixed Mermaid diagrams. Simple fix, the issue seems to be that `^^` was getting replaced with `\ilinebr` at a later stage than `\ilinebr` was getting replaced with `\n`. Replacing each `^^` with `\ilinebr` did the trick.

kring:
Thanks @azrogers!

PR #1098, opened 2025-02-05
`useBaseQuery: true` for `LayerJsonTerrainLoader`

azrogers:
[As a user on our forums ran into](https://community.cesium.com/t/how-to-pass-token-in-terrain-tile-requests/38382), `LayerJsonTerrainLoader` doesn't currently specify the `useBaseQuery` argument to be able to transfer query parameters from the base URL to the URLs of each terrain tile. While this is technically "correct" behavior as far as URL resolution goes, `TilesetJsonTerrainLoader` and other parts of our codebase do inherit query parameters from the base URL, and we don't currently have a better way of specifying query parameters for each loaded URL other than this. With that in mind, I think this change is worth it.


asafMasa:
@azrogers Hi, when will this merge will come in?

kring:
Thanks @azrogers!

PR #1099, opened 2025-02-05
Make CesiumGltf headers public and fix warnings

### LINKED ISSUE #460: Fix warnings in public headers

azrogers:
Closes #460. As part of #1004 we stopped marking most of the public headers as `SYSTEM` so that they would be analyzed by clang-tidy, but we skipped over `CesiumGltf` as the number of warnings didn't seem worth resolving at the time. I ended up fixing some of those warnings while working on #1085, so I decided to finish the job here.

### DIFF
--- a/CesiumUtility/include/CesiumUtility/SharedAsset.h
+++ b/CesiumUtility/include/CesiumUtility/SharedAsset.h
@@ -78,7 +78,7 @@ public:
    */
   IDepotOwningAsset<T>* getDepot() { return this->_pDepot; }
 
-protected:
+private:
   SharedAsset() = default;
   ~SharedAsset() { CESIUM_ASSERT(this->_referenceCount == 0); }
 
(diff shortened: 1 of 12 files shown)


azrogers:
Great, looks like we'd need to upgrade to GCC 13.3 or 14 to fix the "unknown pragma" bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89038

CesiumUtility/include/CesiumUtility/SharedAsset.h
```diff
@@ -103,20 +103,23 @@ class CESIUMUTILITY_API SharedAsset : public CesiumUtility::ExtensibleObject {
    * useful to assign the data in derived classes.
    */
   SharedAsset& operator=(const SharedAsset& rhs) {
-    CesiumUtility::ExtensibleObject::operator=(rhs);
+    if (rhs != this) {
```
kring, line 106: This doesn't look right, comparing an instance to a pointer. Shouldn't it be:
```suggestion
    if (&rhs != this) {
```

Same with the move assignment operator below.

kring (commented):
Looks good, thanks @azrogers. Just one small comment. Please merge in main as well, since there's a conflict right now.

azrogers:
Unfortunately I think I'm going to have to scrap this PR and make a new one since we're no longer going with the `enum-metadata-support` branch :sweat_smile:

azrogers:
Closing in favor of #1133.

PR #1100, opened 2025-02-06
Fix scheme detection in CesiumUtility::Uri

lilleyse:
`urlHasScheme` checks if a scheme is followed by `://` but just `:` should be allowed too.

URI_syntax_diagram svg

For example, `"geopackage:/home/courtyard_imagery.gpkg"` wasn't being parsed correctly. This seems to be a regression from https://github.com/CesiumGS/cesium-native/pull/1072.

### DIFF
--- a/CesiumUtility/src/Uri.cpp
+++ b/CesiumUtility/src/Uri.cpp
@@ -28,19 +28,24 @@ using UrlResult = ada::result<ada::url_aggregator>;
 bool isAsciiAlpha(unsigned char c) {
   return c >= 0x41 && c <= 0x7a && (c <= 0x5a || c >= 0x61);
 }
-bool isAscii(unsigned char c) { return c <= 0x7f; }
+bool isAsciiAlphanumeric(unsigned char c) {
+  return isAsciiAlpha(c) || (c >= '0' && c <= '9');
+}
 
 /**
  * A URI has a valid scheme if it starts with an ASCII alpha character and has a
- * sequence of ASCII characters followed by a ":"
+ * sequence of ASCII characters followed by a "://"
  */
 bool urlHasScheme(const std::string& uri) {
   for (size_t i = 0; i < uri.length(); i++) {
     unsigned char c = static_cast<unsigned char>(uri[i]);
     if (c == ':') {
       return true;
-    } else if ((i == 0 && !isAsciiAlpha(c)) || !isAscii(c)) {
-      // Scheme must start with an ASCII alpha character and be an ASCII string
+    } else if (
+        (i == 0 && !isAsciiAlpha(c)) ||
+        (!isAsciiAlphanumeric(c) && c != '+' && c != '-' && c != '.')) {
+      // Scheme must start with an ASCII alpha character and be a string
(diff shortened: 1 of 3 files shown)


CesiumUtility/src/Uri.cpp
```diff
@@ -32,13 +32,13 @@ bool isAscii(unsigned char c) { return c <= 0x7f; }
...
  * A URI has a valid scheme if it starts with an ASCII alpha character and has a
- * sequence of ASCII characters followed by a "://"
+ * sequence of ASCII characters followed by a ":"
  */
 bool urlHasScheme(const std::string& uri) {
   for (size_t i = 0; i < uri.length(); i++) {
     unsigned char c = static_cast<unsigned char>(uri[i]);
     if (c == ':') {
-      return uri.length() > i + 2 && uri[i + 1] == '/' && uri[i + 2] == '/';
+      return true;
```
lilleyse, line 41: My only concern is that this would lead to false positives for URIs that don't have a scheme. It looks like `:` is allowed in the path portion of a URI.

Does ada have a more robust way to check if a URI has a scheme rather than rolling our own check?

azrogers: Not without parsing the URL and checking to see if it fails, which won't necessarily tell us if the URL has a scheme or not, just if it is valid at all. However, I do note that in [the WhatWG URL spec](https://url.spec.whatwg.org/#scheme-state), scheme parsing should stop if it encounters a character that's not ASCII alphanumeric, a `+`, a `-`, or a `.`. With that in mind, I believe this would identify all valid schemes:
```cpp
// C++ locale settings might change which values std::isalpha checks for. We
// only want ASCII.
bool isAsciiAlpha(unsigned char c) {
  return c >= 0x41 && c <= 0x7a && (c <= 0x5a || c >= 0x61);
}
bool isAsciiAlphanumeric(unsigned char c) {
  return isAsciiAlpha(c) || (c >= '0' && c <= '9');
}

/**
 * A URI has a valid scheme if it starts with an ASCII alpha character and has a
 * sequence of ASCII characters followed by a "://"
 */
bool urlHasScheme(const std::string& uri) {
  for (size_t i = 0; i < uri.length(); i++) {
    unsigned char c = static_cast<unsigned char>(uri[i]);
    if (c == ':') {
      return true;
    } else if (
        (i == 0 && !isAsciiAlpha(c)) ||
        (!isAsciiAlphanumeric(c) && c != '+' && c != '-' && c != '.')) {
      // Scheme must start with an ASCII alpha character and be a string
      // containing only alphanumeric ASCII or the characters '+', '-', and '.'.
      return false;
    }
  }

  return false;
}
```

lilleyse: That does seem better. If we got a URI like `"my:file.txt"` then it would correctly identify "my" as the scheme. And if we got `"path/to/my:file.txt"` it would correctly identify it as not having a scheme.

lilleyse: @azrogers I pushed your change.

PR #1105, opened 2025-02-07
Fix exception when parsing copyright with trailing semicolon

azrogers:
By coincidence, I happened to run into the same error while working on unloading external tilesets that a user reported on our forums yesterday: https://community.cesium.com/t/crash-with-google-tiles/38444

The crash is pretty simple. Some copyright strings end with a trailing semicolon - `Google;` being the one I saw. This causes an exception in `parseGltfCopyright`. It goes through the loop correctly once, but continues to loop again, because `end` isn't `npos` - after all, there is a semicolon, implying there's a next entry to read! But when we start the second run through the loop, `start` is pointing at the end of the string, so every value becomes `npos`, or `-1`. Trying to perform a substring between `-1` and `-1 - -1 + 1` gives us an `out_of_range` exception, and we crash.

The fix is also pretty simple - we exit the loop if we're at the end of the string, even if we saw another semicolon.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@
 ##### Fixes :wrench:
 
 - Fixed parsing URIs that have a scheme followed by `:` instead of `://`.
+- Fixed crash in `GltfUtilities::parseGltfCopyright` when the copyright string ends with a trailing semicolon.
 
 ### v0.44.1 - 2025-02-03
 
--- a/CesiumGltfContent/include/CesiumGltfContent/GltfUtilities.h
+++ b/CesiumGltfContent/include/CesiumGltfContent/GltfUtilities.h
@@ -130,6 +130,16 @@ struct CESIUMGLTFCONTENT_API GltfUtilities {
   parseGltfCopyright(const CesiumGltf::Model& gltf);
 
   /**
+   * @brief Parse a semicolon-separated string, such as the copyright field of a
+   * glTF model, and return the individual parts (credits).
+   *
+   * @param s The string to parse.
+   * @return The semicolon-delimited parts.
+   */
+  static std::vector<std::string_view>
+  parseGltfCopyright(const std::string_view& s);
+
+  /**
    * @brief Merges all of the glTF's buffers into a single buffer (the first
    * one).
    *
--- a/CesiumGltfContent/src/GltfUtilities.cpp
+++ b/CesiumGltfContent/src/GltfUtilities.cpp
@@ -283,26 +283,45 @@ GltfUtilities::computeBoundingRegion(
 
 std::vector<std::string_view>
 GltfUtilities::parseGltfCopyright(const CesiumGltf::Model& gltf) {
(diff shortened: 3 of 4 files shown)


azrogers:
Actually, turns out it needs one additional check for a trailing semicolon with only whitespace after it.

kring:
Thanks @azrogers! I added a couple more tests, and then ended up reworking the algorithm because some of my new tests weren't passing and the overall approach seemed unnecessarily complicated / slow (not your fault, this was pre-existing). Sorry to just run with it myself like this, but I think this is a pretty serious bug that impacts almost all our users, and I want to do a patch release with it ASAP.

martindlabaja:
I get this eror and I am unable to start my project atm.

martindlabaja:
This morning we woke up and are apps are crashing with FATAL ERROR when online and using Cesium. Crash log seems fine and unrelated to Cesium but when I unplug the internet, the app doesnt crash. 

Not sure if it is Cesium 100% but is the culprit so I am just asking if maybe you know if this is happening on other places, too?

Thanks, Martin. :)

Not sure if this is system wide issue or only our problem

martindlabaja:
I can verify the FATAL CRASHES are caused by Cesium. I have removed it from project and it works and do not crash.

This seems to be issue on multiple PCs.

azrogers:
@martindlabaja Have you tried using the new version 2.13.2 (Unreal) or 1.15.2 (Unity)? These should include the fix for this issue.

martindlabaja:
Ah ok! Thanks, will test it out. I figured out that removing the cache files in APP data and packaged project temporary fixed the issue.

ovigbinobaro:
experiencing this crash as well! @azrogers any one could the new version 2.13.2 fix for U.E 5.2.1? I understand it could not be updated on fab, i tried packaging using the developer setup guide but after getting through the CMake step, I could not get the plugin to package for 5.2 using both Visual Studio 2017 and 2022

kring:
@ovigbinobaro we can't provide versions for old UE releases, but if you post your compilation problem to the [community forum](https://community.cesium.com/c/cesium-for-unreal/10) we may be able to help. I'm curious, too, what keeps you from upgrading to 5.3+?

javagl:
This caused _so_ much noise in the forum that it actually caught my attention. I'm a bit surprised to see that revolving around a function called `parseGltfCopyright` and involving multiple appearances of `" \t"` and `';'`, and not some generic
`tokens = tokenize(input, delimiters);`
(maybe with some `boolean includeEmptyTokens` thrown in).

mschult2:
Fyi the parse errors are fixed for me in Cesium for Unity 1.15.3.

kring:
Yeah, I hear you @javagl. In any language but C++, there's a standard library function that could do this. When we added this three years ago, I think we were doing it rather hastily in support of Google's efforts, which still weren't public at the time, and I think we may have perceived it as largely temporary. An actual extension would be a lot nicer, and not require string parsing.

PR #1108, opened 2025-02-11
Fix conversion functions for KHR_mesh_quantization decompression

timoore:
This looks like a cut-and-paste error, but the result was that geometry that used unsigned byte positions would be approximately doubled in size, and geometry that used signed 16 bit integers would be half size.

Before:
Screenshot From 2025-02-11 15-31-14

After:
Screenshot From 2025-02-11 15-33-43

I tried to come up with a test based on the compressed Duck.glb files, but the compressed normals in those files seem kind of random. I had thought that normals in glb files should be already normalized, but that is not the case for the duck files that use mesh quantization.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@
 ##### Fixes :wrench:
 
 - Fixed parsing URIs that have a scheme followed by `:` instead of `://`.
+- Fixed decoding of KHR_mesh_quantization normalized values.
 
 ### v0.44.2 - 2025-02-10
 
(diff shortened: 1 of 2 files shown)


azrogers:
Thanks @timoore! This change looks good to me - can you just make a note of this fix in CHANGES.md? If you're able to figure out a test for this, definitely add it to the PR, but I think the change is straightforward enough to commit as is once the changelog is updated.

azrogers:
Thanks @timoore!

PR #1109, opened 2025-02-12
Fix another problem in parseGltfCopyright

kring:
Sadly, my fix for `parseGltfCopyright` in was subtly broken.

As reported here:
https://github.com/CesiumGS/cesium-unity/issues/551#issuecomment-2652714772

Explanation copied from the comment there:

> There's a line of code like this:
> 
> ```
> s.find_first_not_of(" \t", 0, end + 1)
> ```
> 
> Which I _thought_ would find the first character that isn't space or tab starting at position 0 in `s` and continuing for `end+1` characters. But no! That last parameter is the length of the string in the first parameter, not the number of characters to search. That is... _really_ unexpected. And amazingly, our rather extensive tests of this function passed on all platforms, in both debug and release even with this mistake. But it seems that if the contents of memory after that `" \t"` string is juuuust right, which it happens to be in your model and in Unity (but only in a Release build), then this function will do the wrong thing.

This has already gone out in a hotfix for both Unreal (v2.13.3) and Unity (v1.15.3).


j9liu:
Thanks @kring for dealing with this string parsing madness 🥲

PR #1104, opened 2025-02-07
Add generated classes for `3DTILES_content_voxels`

j9liu:
Depends on #1103, so merge that first.

PR #1111, opened 2025-02-12
Add support for `3DTILES_bounding_volume_cylinder`

j9liu:
Depends on #1104 so merge that first. But I'm going to mark this as a draft PR because I think there's a spec question to be had.

This PR adds support for the `3DTILES_bounding_volume_cylinder` extension found [here](https://github.com/CesiumGS/3d-tiles/tree/voxels/extensions/3DTILES_bounding_volume_cylinder). The extension is mainly used to support cylinder voxels in the [`3DTILES_content_voxels extension`](https://github.com/CesiumGS/3d-tiles/tree/voxels/extensions/3DTILES_bounding_volume_cylinder).

The cylinder is meant to be combined with implicit octree tiling, but that results in some weirdness for the `BoundingCylinder` implementation. When other volumes are subdivided (e.g. box, region) they are still the same shape. But the sub-regions of a cylinder are only parts of a cylinder, not fully cylinders itself. The `3DTILES_bounding_volume_cylinder` extension does not currently indicate a way to divide that initial cylinder.

The thing to do here, regardless of spec, is probably to rename `BoundingCylinder` -> `BoundingCylinderRegion`, to more accurately indicate that the volume may be a subregion of the fully cylinder. And that way, the division of a `BoundingCylinderRegion` into sub-regions is logically correct. But I'm wondering if the `3DTILES_bounding_volume_cylinder` should also allow angular / radial portions itself? Or, will we restrict it to being a full cylinder, but note that implementations should be wary of handling subregions with angular / radial bounds.


j9liu:
I also should explicitly note that for efficiency / time reasons, the underlying implementation of `BoundingCylinder` relies on `OrientedBoundingBox` as an approximation.

I had actually added cylinder-specific math to the compute distance function (derived from this [paper](https://liris.cnrs.fr/Documents/Liris-1297.pdf)) in this (https://github.com/CesiumGS/cesium-native/pull/1111/commits/e5e323f4116c74551b7c5c555ca4d57c07f6e7a2). Unfortunately I had to remove it when I remembered that we're often dealing with cylinder subregions instead of the full cylinder. The math _might_ carry over still but I haven't worked it out.

j9liu:
@lilleyse and I are reworking the spec here: https://github.com/CesiumGS/3d-tiles/pull/791

Once that's merged in, I'll make changes here accordingly

j9liu:
I'm marking this as ready for review. The `BoundingCylinder` class is now `BoundingCylinderRegion`, with fields that mirror the ones in https://github.com/CesiumGS/3d-tiles/pull/791.

Cesium3DTilesSelection/src/ImplicitOctreeLoader.h
```diff
@@ -15,7 +16,8 @@
 namespace Cesium3DTilesSelection {
 using ImplicitOctreeBoundingVolume = std::variant<
     CesiumGeospatial::BoundingRegion,
-    CesiumGeometry::OrientedBoundingBox>;
+    CesiumGeometry::OrientedBoundingBox,
+    CesiumGeometry::BoundingCylinderRegion>;
```
kring, line 20: Does `ImplicitQuadtreeLoader` need similar changes?

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -34,12 +36,17 @@
 #include <CesiumUtility/Assert.h>
 #include <CesiumUtility/ErrorList.h>
 #include <CesiumUtility/JsonHelpers.h>
+#include <CesiumUtility/Math.h>
 #include <CesiumUtility/Uri.h>
 
 #include <fmt/format.h>
 #include <glm/common.hpp>
+#include <glm/detail/setup.hpp>
```
kring, line 44: It's weird to need a detail header. Is this something clang-tidy wanted you to add?

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -180,6 +187,72 @@ std::optional<BoundingVolume> getBoundingVolumeProperty(
           maximumHeight,
           ellipsoid);
     }
+
+    const auto cylinderIt = extensionsIt->value.FindMember(
```
kring, line 191: The code above sets a poor example (probably because it was written before this was possible), but you can use the generated readers here. Something like:

```
Extension3dTilesBoundingVolumeCylinderReader reader;
ReadJsonResult<Cesium3DTiles::Extension3dTilesBoundingVolumeCylinder> cylinderExtensionResult = reader.readFromJson(etensionsIt->value);
```

In fact you can read the entire `BoundingVolume` using `BoundingVolumeReader`.

Cesium3DTilesSelection/test/TestBoundingVolume.cpp
```diff
@@ -1,12 +1,16 @@
-#include <Cesium3DTilesSelection/BoundingVolume.h>
+#include "Cesium3DTilesSelection/BoundingVolume.h"
```
kring, line 1: Angle brackets.

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -351,6 +426,16 @@ void createImplicitOctreeLoader(
         *pBox);
     pImplicitLoader = pLoader.get();
     currentLoader.addChildLoader(std::move(pLoader));
+  } else if (pCylinder) {
+    auto pLoader = std::make_unique<ImplicitOctreeLoader>(
```
kring, line 430: Is similar code needed for quadtrees above?

Cesium3DTilesContent/src/ImplicitTilingUtilities.cpp
```diff
@@ -1,9 +1,13 @@
+#include "Cesium3DTilesContent/ImplicitTilingUtilities.h"
+
+#include "Cesium3DTilesContent/TileBoundingVolumes.h"
```
kring, line 3: Angle brackets.

Cesium3DTilesContent/src/TileBoundingVolumes.cpp
```diff
@@ -1,13 +1,18 @@
+#include "Cesium3DTilesContent/TileBoundingVolumes.h"
```
kring, line 1: Angle brackets.

Cesium3DTilesContent/test/TestTileBoundingVolumes.cpp
```diff
@@ -1,11 +1,15 @@
+#include "Cesium3DTilesContent/TileBoundingVolumes.h"
```
kring, line 1: Angle brackets.

CesiumGeometry/test/TestBoundingCylinderRegion.cpp
```diff
@@ -0,0 +1,266 @@
+#include "CesiumGeometry/BoundingCylinderRegion.h"
```
kring, line 1: Angle brackets.

kring (commented):
A few comments (mostly nitpicking include style), but overall looks great!

CesiumGeometry/include/CesiumGeometry/OrientedBoundingBox.h
```diff
@@ -1,16 +1,17 @@
...
+#include "CesiumGeometry/CullingResult.h"
+#include "CesiumGeometry/Library.h"
 
 #include <glm/mat3x3.hpp>
 #include <glm/vec3.hpp>
 
 namespace CesiumGeometry {
 
 class Plane;
+class BoundingCylinder;
```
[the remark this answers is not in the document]
j9liu: Yep, leftover from the previous implementation. Thanks for the catch!

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
j9liu: I _think_ clang-tidy wanted me to add this, but I could be wrong. I'll see if I can get away with removing it, or if there's a better file to include.

j9liu:
Thanks for the thorough review @kring ! I think I addressed all your comments, let me know if anything else comes up.

kring:
Thanks @j9liu!

PR #1112, opened 2025-02-13
ENUM metadata support, treating them as scalars rather than using a container type

azrogers:
This is an alternative version of #1085, implementing enums by leveraging our existing scalar metadata support rather than implementing them as an entirely new container type. This requires some changes to the assumptions we're currently making in the metadata system, chief among them the breaking of the connection between C++ type and `PropertyType`. However, it does simplify the implementation and reduce the amount of copy-and-paste necessary to support enums. Please review and discuss which approach you prefer!


CesiumGltf/include/CesiumGltf/PropertyTableView.h
```diff
@@ -1095,19 +1127,43 @@ class PropertyTableView {
...
-    const PropertyComponentType componentType =
-        convertStringToPropertyComponentType(
-            classProperty.componentType.value_or(""));
-    if (TypeToPropertyType<T>::component != componentType) {
+
+    PropertyComponentType componentType = convertStringToPropertyComponentType(
+        classProperty.componentType.value_or(""));
+    const CesiumGltf::Enum* pEnumDefinition = nullptr;
+    if (type == PropertyType::Enum && classProperty.enumType &&
+        !IsMetadataBoolean<T>::value) {
```
kring, line 1139: Why is this check for boolean needed here?

CesiumGltf/test/TestPropertyAttributePropertyView.cpp
```diff
@@ -1,3 +1,5 @@
+#include "CesiumNativeTests/TypeToPropertyTypeString.h"
```
kring, line 1: Angle brackets here and in the other files.

kring (commented):
I have a few comments below, but overall this looks reasonable to me. I'm going to look at the Unreal implementation next, though, so that could change my opinion!

CesiumGltf/include/CesiumGltf/PropertyView.h
```diff
@@ -153,25 +161,91 @@ validatePropertyType(const ClassProperty& classProperty) {
...
  */
 template <typename T>
-PropertyViewStatusType
-validateArrayPropertyType(const ClassProperty& classProperty) {
-  using ElementType = typename MetadataArrayType<T>::type;
-  if (TypeToPropertyType<ElementType>::value !=
-      convertStringToPropertyType(classProperty.type)) {
+PropertyViewStatusType validatePropertyType(
+    const ClassProperty& classProperty,
+    const CesiumGltf::Enum* pEnumDefinition) {
```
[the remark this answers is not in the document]
azrogers: Went with adding a null check instead, since a nullptr enum definition should be an invalid enum status.

CesiumGltf/include/CesiumGltf/PropertyView.h
```diff
@@ -153,25 +161,91 @@ validatePropertyType(const ClassProperty& classProperty) {
...
+ *
+ * @returns A \ref PropertyViewStatus value representing the error found while
+ * validating, or \ref PropertyViewStatus::Valid if no errors were found.
+ */
+template <typename T>
+PropertyViewStatusType
+validateArrayPropertyType(const ClassProperty& classProperty) {
+  using ElementType = typename MetadataArrayType<T>::type;
+
+  if (convertStringToPropertyType(classProperty.type) != PropertyType::Enum) {
```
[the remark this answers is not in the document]
azrogers: Originally I was just skipping the parts of validation that wouldn't work for enums, then I changed my mind and made a separate validation function for enums instead. Just forgot to remove this part.

CesiumGltf/include/CesiumGltf/PropertyTableView.h
azrogers: Because we can't have a boolean enum. So if the type is `bool` and the property type is `Enum`, we will go to the else if branch where we return an `ErrorInvalidEnum` status.

CesiumGltf/include/CesiumGltf/PropertyView.h
```diff
@@ -1435,6 +1670,66 @@ class PropertyView<PropertyArrayView<ElementType>, false> {
     }
   }
 
+  /**
+   * @brief Constructs a property instance from a class definition only.
```
j9liu, line 1674:
```suggestion
   * @brief Constructs a property instance from a class and enum definition.
```

CesiumGltf/include/CesiumGltf/PropertyTexturePropertyView.h
```diff
@@ -386,6 +385,85 @@ class PropertyTexturePropertyView<ElementType, false>
...
+   * @brief Construct a view of the data specified by a {@link PropertyTextureProperty}.
+   *
+   * @param property The {@link PropertyTextureProperty}
+   * @param classProperty The {@link ClassProperty} this property conforms to.
+   * @param pEnumDefinition The {@link CesiumGltf::Enum} definition this property uses.
+   * @param sampler The {@link Sampler} used by the property.
+   * @param image The {@link ImageAsset} used by the property.
+   * @param options The options for constructing the view.
+   */
+  PropertyTexturePropertyView(
```
j9liu, line 398: If this is largely copy and paste, can the other constructor reference this one with a `nullptr` pEnumDefinition?

CesiumGltf/include/CesiumGltf/PropertyTablePropertyView.h
```diff
@@ -258,6 +259,34 @@ class PropertyTablePropertyView<ElementType, false>
...
+   * @param values The raw buffer specified by {@link PropertyTableProperty::values}
+   * @param pEnumDefinition A pointer to the enum definition used for this
+   * value.
+   */
+  PropertyTablePropertyView(
+      const PropertyTableProperty& property,
+      const ClassProperty& classProperty,
+      int64_t size,
+      std::span<const std::byte> values,
+      const CesiumGltf::Enum* pEnumDefinition) noexcept
```
j9liu, line 278: Sorry, this is my OCD talking -- it feels inconsistent that the `pEnumDefinition` is at the bottom for `PropertyTablePropertyView` , but comes right after `ClassProperty` in `PropertyTexturePropertyView` . Can we pick one? :')

CesiumGltf/include/CesiumGltf/PropertyTableView.h
j9liu: I'm also a bit confused, would that not get caught by `!canRepresentPropertyType<T>(type)`? (And if not, maybe the function should be changed to reflect this?)

j9liu (changes requested):
@azrogers I haven't done a full review, but I at least wanted to check the areas that Kevin commented on. I'll do another pass early next week

CesiumGltf/include/CesiumGltf/PropertyTexturePropertyView.h
azrogers: Unfortunately, no - this constructor calls out to the `PropertyView` constructor which calls out to `validatePropertyType`. `validatePropertyType` with the enum overload will return `ErrorInvalidEnum` with a nullptr enum definition (as it should).

CesiumGltf/include/CesiumGltf/PropertyTypeTraits.h
```diff
@@ -201,9 +201,12 @@ struct MetadataArrayType<CesiumGltf::PropertyArrayCopy<T>> {
   using type = T;
 };
 
-/**
- * @brief Convert a C++ type to PropertyType and PropertyComponentType
- */
+/** * @brief Infer the best-fitting PropertyType and PropertyComponentType for a
```
j9liu, line 204: Something went wrong with the asterisks in this comment...

CesiumGltf/test/TestPropertyTypeTraits.cpp
```diff
@@ -193,365 +193,65 @@ TEST_CASE("Test MetadataArrayType") {
   REQUIRE(std::is_same_v<type, uint32_t>);
 }
 
-TEST_CASE("TypeToPropertyType") {
```
j9liu, line 196: Now that we're preserving `TypeToPropertyType`, we should preserve the unit tests for it too.

CesiumGltf/test/TestPropertyView.cpp
```diff
@@ -3058,3 +3187,215 @@ TEST_CASE("String Array PropertyView") {
     REQUIRE(view.status() == PropertyViewStatus::ErrorInvalidNoDataValue);
   }
 }
+
+TEST_CASE("Enum PropertyView") {
```
j9liu, line 3191: Sorry for the nitpick, but can this be put under the `String PropertyView` tests? Just so the non-array tests are grouped together.

CesiumGltf/test/TestPropertyView.cpp
```diff
@@ -3058,3 +3187,215 @@ TEST_CASE("String Array PropertyView") {
...
+    REQUIRE(!view.noData());
+    REQUIRE(!view.defaultValue());
+  }
+
+  SUBCASE("Correct property type") {
+    ClassProperty classProperty;
+    classProperty.type = ClassProperty::Type::ENUM;
+    classProperty.array = true;
+    PropertyView<PropertyArrayView<uint8_t>> view(classProperty);
+    REQUIRE(view.propertyType() == PropertyType::Enum);
```
j9liu, line 3220:
```suggestion
    PropertyView<uint8_t> view(classProperty);
    REQUIRE(view.propertyType() == PropertyType::Enum);
```

CesiumGltf/test/TestPropertyView.cpp
```diff
@@ -3058,3 +3187,215 @@ TEST_CASE("String Array PropertyView") {
...
+    REQUIRE(view.status() == PropertyViewStatus::ErrorInvalidDefaultValue);
+
+    classProperty.noData = "null";
+    view = PropertyView<uint8_t>(classProperty, &enumDef);
+    REQUIRE(view.status() == PropertyViewStatus::ErrorInvalidNoDataValue);
+
+    classProperty.noData = "Qux";
+    view = PropertyView<uint8_t>(classProperty, &enumDef);
+    REQUIRE(view.status() == PropertyViewStatus::ErrorInvalidNoDataValue);
+  }
```
j9liu, line 3271: Pick one? These seem like they're testing the same thing, just with different random strings

CesiumGltf/include/CesiumGltf/PropertyView.h
```diff
@@ -182,6 +225,39 @@ validateArrayPropertyType(const ClassProperty& classProperty) {
...
 
+/**
+ * @brief Validates a \ref ClassProperty representing an array of enum values,
+ * checking for any type mismatches.
+ *
+ * @returns A \ref PropertyViewStatus value representing the error found while
+ * validating, or \ref PropertyViewStatus::Valid if no errors were found.
+ */
+template <typename T>
+PropertyViewStatusType validateArrayPropertyType(
```
j9liu, line 236: See comment on `PropertyTextureView`. I think the same consolidation could be applied here, too.

CesiumGltf/include/CesiumGltf/PropertyView.h
```diff
@@ -381,6 +459,68 @@ template <typename ElementType> class PropertyView<ElementType, false> {
     }
   }
 
+  /**
+   * @brief Constructs a property instance from a class definition and enum
+   * definition.
+   */
+  PropertyView(
```
j9liu, line 466: I kinda feel the same way about the duplication here. I think allowing `pEnumDefinition = nullptr` and doing an extra branch for enums would reduce duplication. Assuming everything else can maintain correctness.

CesiumGltf/include/CesiumGltf/PropertyView.h
```diff
@@ -145,6 +153,40 @@ validatePropertyType(const ClassProperty& classProperty) {
   return PropertyViewStatus::Valid;
 }
 
+/**
+ * @brief Validates a \ref ClassProperty representing an array, checking for any
```
j9liu, line 157: Think the doc is wrong here, this isn't an array type

CesiumGltf/include/CesiumGltf/PropertyView.h
```diff
@@ -417,6 +558,22 @@ template <typename ElementType> class PropertyView<ElementType, false> {
     getNumericPropertyValues(property);
   }
 
+  /**
+   * @brief Constructs a property instance from a property table property, class
+   * definition, and an associated enum definition.
+   */
+  PropertyView(
```
j9liu, line 565: Same with the rest of the duplicate constructors in this file. It would be great to consolidate as much as possible.

CesiumGltf/include/CesiumGltf/PropertyView.h
```diff
@@ -565,6 +755,12 @@ template <typename ElementType> class PropertyView<ElementType, false> {
     return _defaultValue;
   }
 
+  /**
+   * @brief Returns the \ref PropertyType of the property this view is
```
j9liu, line 759:
```suggestion
   * @brief Returns the \ref PropertyType of the property that this view is
```

CesiumGltf/include/CesiumGltf/PropertyTypeTraits.h
```diff
@@ -392,6 +395,48 @@ template <> struct TypeToPropertyType<std::string_view> {
   static constexpr PropertyType value = PropertyType::String;
 };
 
+/**
+ * @brief The number of dimensions that this type contains.
+ */
+template <typename T> struct TypeToDimensions;
```
j9liu, line 401: Just wanted to say I like this change 😄 thanks for making it!

CesiumGltf/include/CesiumGltf/PropertyView.h
```diff
@@ -1435,6 +1678,67 @@ class PropertyView<PropertyArrayView<ElementType>, false> {
...
+    if (_status != PropertyViewStatus::Valid) {
+      return;
+    }
+
+    if (classProperty.normalized) {
+      _status = PropertyViewStatus::ErrorNormalizationMismatch;
+      return;
+    }
+
+    getNumericPropertyValues(classProperty);
```
j9liu, line 1712: `getNumericPropertyValues` handles `offset`, `scale`, `max`, `min`, which don't apply to enum properties. When you consolidate the constructors, just check the property type here.

CesiumGltf/include/CesiumGltf/PropertyTablePropertyView.h
```diff
@@ -258,6 +259,34 @@ class PropertyTablePropertyView<ElementType, false>
...
 
+  /**
+   * @brief Construct an instance pointing to data specified by a {@link PropertyTableProperty}, with an enum definition attached.
+   * Used for non-array or fixed-length array data.
+   *
+   * @param property The {@link PropertyTableProperty}
+   * @param classProperty The {@link ClassProperty} this property conforms to.
+   * @param size The number of elements in the property table specified by {@link PropertyTable::count}
+   * @param values The raw buffer specified by {@link PropertyTableProperty::values}
+   * @param pEnumDefinition A pointer to the enum definition used for this
```
j9liu, line 270: Can `@param pEnumDefinition` be moved to reflect the new parameter order?

CesiumGltf/include/CesiumGltf/PropertyTablePropertyView.h
```diff
@@ -258,6 +259,34 @@ class PropertyTablePropertyView<ElementType, false>
...
+   * value.
+   */
+  PropertyTablePropertyView(
+      const PropertyTableProperty& property,
+      const ClassProperty& classProperty,
+      const CesiumGltf::Enum* pEnumDefinition,
+      int64_t size,
+      std::span<const std::byte> values) noexcept
+      : PropertyView<ElementType>(classProperty, property, pEnumDefinition),
```
j9liu, line 279: Could you consolidate this with the other constructor?

CesiumGltf/include/CesiumGltf/PropertyTablePropertyView.h
```diff
@@ -290,6 +319,41 @@ class PropertyTablePropertyView<ElementType, false>
...
+   * @param size The number of elements in the property table specified by {@link PropertyTable::count}
+   * @param values The raw buffer specified by {@link PropertyTableProperty::values}
+   * @param arrayOffsets The raw buffer specified by {@link PropertyTableProperty::arrayOffsets}
+   * @param stringOffsets The raw buffer specified by {@link PropertyTableProperty::stringOffsets}
+   * @param arrayOffsetType The offset type of arrayOffsets specified by {@link PropertyTableProperty::arrayOffsetType}
+   * @param stringOffsetType The offset type of stringOffsets specified by {@link PropertyTableProperty::stringOffsetType}
+   * @param pEnumDefinition A pointer to the enum definition used for this
+   * value.
+   */
+  PropertyTablePropertyView(
```
j9liu, line 336: Same here.

CesiumGltf/include/CesiumGltf/PropertyTableView.h
```diff
@@ -194,13 +195,44 @@ class PropertyTableView {
...
+
+        if (componentType == PropertyComponentType::Float32 ||
+            componentType == PropertyComponentType::Float64) {
+          callback(
+              propertyId,
+              PropertyTablePropertyView<uint8_t>(
+                  PropertyTablePropertyViewStatus::ErrorInvalidEnum));
+          return;
+        }
+      }
```
j9liu, line 221: Might be nitpicky, but can be flatten the `if else` here so there's less indentation? It helps with readability at the end of the day

CesiumGltf/include/CesiumGltf/PropertyTableView.h
```diff
@@ -1138,11 +1193,28 @@ class PropertyTableView {
...
+            _pPropertyTable->count,
+            values);
+      }
+    } else {
+      return PropertyTablePropertyView<T, Normalized>(
+          propertyTableProperty,
+          classProperty,
+          _pPropertyTable->count,
+          values);
+    }
```
j9liu, line 1217: Will these different branches be needed after all the constructor changes?

CesiumGltf/include/CesiumGltf/PropertyTextureView.h
```diff
@@ -595,14 +596,37 @@ class PropertyTextureView {
...
+        componentType = convertStringToPropertyComponentType(
+            enumDefinitionIt->second.valueType);
+        pEnumDefinition = &enumDefinitionIt->second;
+
+        if (componentType == PropertyComponentType::Float32 ||
+            componentType == PropertyComponentType::Float64) {
+          return PropertyTexturePropertyView<T, Normalized>(
+              PropertyTexturePropertyViewStatus::ErrorInvalidEnum);
+        }
+      }
```
j9liu, line 624: Could we flatten this `if`/`else`?

j9liu (changes requested):
@azrogers Thank you for responding quickly to feedback and putting effort into these changes!

My one remaining reservation is about the way that we handle the constructors with `pEnumDefinition`. There is a silent assumption that if we call the constructor that includes the enum definition parameter -- even if `nullptr` -- that the property must have been an enum. I feel like this isn't right; we could be handling the inputs more generally, starting with `PropertyView` then working its way up into `PropertyTablePropertyView`, etc.

Let me know what you think -- overall I think it's going to reduce duplication.

CesiumGltf/test/TestPropertyView.cpp
azrogers: They're not! The "null" string is interpreted as a null JSON value, which tests that the `noData` handling correctly ignores non-string values. The "Qux" string, on the other hand, is a JSON string that isn't in the enum definition, so it's testing that those types of invalid `noData` values will also be handled correctly.

CesiumGltf/include/CesiumGltf/PropertyTableView.h
azrogers: I don't see how it can be flattened?

CesiumGltf/include/CesiumGltf/PropertyTextureView.h
azrogers: Similarly, I don't see how this can be flattened. We can't do the lookup unless it's an enum value.

CesiumGltf/include/CesiumGltf/PropertyTextureView.h
```diff
@@ -680,14 +706,37 @@ class PropertyTextureView {
...
+        componentType = convertStringToPropertyComponentType(
+            enumDefinitionIt->second.valueType);
+        pEnumDefinition = &enumDefinitionIt->second;
+
+        if (componentType == PropertyComponentType::Float32 ||
+            componentType == PropertyComponentType::Float64) {
+          return PropertyTexturePropertyView<PropertyArrayView<T>, Normalized>(
+              PropertyTexturePropertyViewStatus::ErrorInvalidEnum);
+        }
+      }
```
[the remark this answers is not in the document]
azrogers: Same here, how would you flatten it?

azrogers:
@j9liu Addressed your review comments. Was able to consolidate most of the constructors - I did have to add a `pEnumDefinition` parameter to some of the protected constructors that would never be enums, just so the templating worked out, but since the parameter is ignored I don't think it's a big deal.

CesiumGltf/include/CesiumGltf/PropertyTableView.h
j9liu: Sorry, I just meant the outer if/else statement. Like this:

```suggestion
      if (enumDefinitionIt == this->_pEnumDefinitions->end()) {
        callback(
            propertyId,
            PropertyTablePropertyView<uint8_t>(
                PropertyTablePropertyViewStatus::ErrorInvalidEnum));
        return;
      }
        componentType = convertStringToPropertyComponentType(
            enumDefinitionIt->second.valueType);

        if (componentType == PropertyComponentType::Float32 ||
            componentType == PropertyComponentType::Float64) {
          callback(
              propertyId,
              PropertyTablePropertyView<uint8_t>(
                  PropertyTablePropertyViewStatus::ErrorInvalidEnum));
          return;
        }
      
```

j9liu: Same with all the others.

CesiumGltf/include/CesiumGltf/PropertyTextureView.h
j9liu:
```suggestion
      if (enumDefinitionIt == this->_pEnumDefinitions->end()) {
        return PropertyTexturePropertyView<T, Normalized>(
            PropertyTexturePropertyViewStatus::ErrorInvalidEnum);
      }
        componentType = convertStringToPropertyComponentType(
            enumDefinitionIt->second.valueType);
        pEnumDefinition = &enumDefinitionIt->second;

        if (componentType == PropertyComponentType::Float32 ||
            componentType == PropertyComponentType::Float64) {
          return PropertyTexturePropertyView<T, Normalized>(
              PropertyTexturePropertyViewStatus::ErrorInvalidEnum);
        }
      
```

CesiumGltf/include/CesiumGltf/PropertyTextureView.h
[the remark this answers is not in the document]
j9liu:
```suggestion
      if (enumDefinitionIt == this->_pEnumDefinitions->end()) {
        return PropertyTexturePropertyView<PropertyArrayView<T>, Normalized>(
            PropertyTexturePropertyViewStatus::ErrorInvalidEnum);
      } 
        componentType = convertStringToPropertyComponentType(
            enumDefinitionIt->second.valueType);
        pEnumDefinition = &enumDefinitionIt->second;

        if (componentType == PropertyComponentType::Float32 ||
            componentType == PropertyComponentType::Float64) {
          return PropertyTexturePropertyView<PropertyArrayView<T>, Normalized>(
              PropertyTexturePropertyViewStatus::ErrorInvalidEnum);
        }
      
```

j9liu:
Thanks so much @azrogers! I'll take another look soon.

> Was able to consolidate most of the constructors - I did have to add a pEnumDefinition parameter to some of the protected constructors that would never be enums, just so the templating worked out, but since the parameter is ignored I don't think it's a big deal.

Yeah, this would usually irk me on a conceptual level, but we have a pattern of ignoring parameters anyway (like how `PropertyTableProperty` is ignored for strings and booleans). And I think the reduced duplication is worth it.

CesiumGltf/test/TestPropertyView.cpp
j9liu: I don't think that's right, because we have this test for string property views:

```
    ClassProperty classProperty;
    classProperty.type = ClassProperty::Type::STRING;
    classProperty.required = false;
    classProperty.noData = "null";
    classProperty.defaultProperty = "default";

    PropertyView<std::string_view> view(classProperty);
    REQUIRE(view.status() == PropertyViewStatus::Valid);
    REQUIRE(!view.required());

    REQUIRE(view.noData() == "null");
    REQUIRE(view.defaultValue() == "default");
```

If `noData` didn't actually exist -- i.e., was a null JSON value--it wouldn't pass that `REQUIRE(view.noData() == "null");` check.

The reason that this test is failing is actually because `required = true` in the class property. Required properties are not allowed to have a `noData` value. Hence why the unit tests end up being duplicate.

j9liu: It sounds like we need a separate test to check whether the input `noData` / `default` values actually exist in the enum -- and yes, we should test that.

CesiumGltf/include/CesiumGltf/PropertyTableView.h
```diff
@@ -1220,11 +1306,28 @@ class PropertyTableView {
                 ErrorBufferViewSizeDoesNotMatchPropertyTableCount);
       }
 
-      return PropertyTablePropertyView<PropertyArrayView<T>, Normalized>(
-          propertyTableProperty,
-          classProperty,
-          _pPropertyTable->count,
-          values);
+      if constexpr (!IsMetadataBoolean<T>::value && !Normalized) {
```
j9liu, line 1309: We shouldn't be calling `getNumericArrayPropertyValues` for `T = bool` so I don't think this check is necessary

CesiumGltf/include/CesiumGltf/PropertyTableView.h
```diff
@@ -1259,6 +1362,20 @@ class PropertyTableView {
           arrayOffsets,
           arrayOffsetType);
     } else {
+      if constexpr (!IsMetadataBoolean<T>::value) {
```
j9liu, line 1365: Same here.

CesiumGltf/include/CesiumGltf/PropertyView.h
```diff
@@ -108,34 +111,59 @@ class PropertyViewStatus {
...
+   */
+  static const PropertyViewStatusType ErrorInvalidEnum = 14;
 };
 
 /**
- * @brief Validates a \ref ClassProperty, checking for any type mismatches.
+ * @brief Validates a \ref ClassProperty representing a property, checking for
+ * any type mismatches.
+ *
+ * @tparam T The value type of the PropertyView we are validating.
```
j9liu, line 125: I tried not to nitpick this since I doubt this function will be used by anyone else, but... I think it's better to use a neutral voice in the `/** */` documentation

```suggestion
 * @tparam T The value type of the PropertyView that was constructed for this ClassProperty.
```

CesiumGltf/include/CesiumGltf/PropertyView.h
```diff
@@ -146,33 +174,52 @@ validatePropertyType(const ClassProperty& classProperty) {
 }
 
 /**
- * @brief Validates a \ref ClassProperty representing an array, checking for any
- * type mismatches.
+ * @brief Validates a \ref ClassProperty representing an array of values,
+ * checking for any type mismatches.
+ *
+ * @tparam T The array type of the PropertyView we are validating.
```
j9liu, line 180:
```suggestion
 * @tparam T The array element type of the PropertyView that was constructed for this ClassProperty.
```

CesiumGltf/include/CesiumGltf/PropertyView.h
```diff
@@ -349,14 +408,20 @@ template <typename ElementType> class PropertyView<ElementType, false> {
...
+      getNumericPropertyValues(classProperty);
+    }
+
     if (_status != PropertyViewStatus::Valid) {
       return;
     }
 
     if (classProperty.noData) {
-      if (!_required) {
+      if (!_required && _propertyType == PropertyType::Enum &&
```
j9liu, line 420: Could we rationalize that if we get to this point and `propertyType == Enum` that `pEnumDefinition` is also non-null, so we no longer have to do a `nullptr` check? (Maybe an `assert` would help, or prove me wrong? 😝 )

CesiumGltf/test/TestPropertyTypeTraits.cpp
```diff
@@ -193,275 +193,643 @@ TEST_CASE("Test MetadataArrayType") {
...
-  SUBCASE("Works for scalar types") {
-    REQUIRE(TypeToPropertyType<uint8_t>::value == PropertyType::Scalar);
-    REQUIRE(
-        TypeToPropertyType<uint8_t>::component == PropertyComponentType::Uint8);
+TEST_CASE("Test canRepresentPropertyType") {
```
j9liu, line 196: Sorry to trouble but do you mind ordering the tests starting from `TypeToPropertyType` -> `CanBeNormalized` -> `canRepresentPropertyType`? I'm sure there's nothing wrong with the content, but the visual diff is large on Github so it's a bit hard to compare

j9liu (commented):
Looking much better @azrogers ! :smile: I have a few remaining points but nothing major about the design. The constructors make everything much cleaner :smile:

Could you also make a note of this in CHANGES.md when you get the chance?

CesiumGltf/include/CesiumGltf/PropertyView.h
azrogers: What do you mean by "neutral voice?" If you mean the phrasing using "we," we're using that phrasing, by my count, at over 500 places in our documentation currently.

CesiumGltf/include/CesiumGltf/PropertyView.h
azrogers: `T` is the array type, not the element type.

CesiumGltf/include/CesiumGltf/PropertyView.h
j9liu: Yes I'm referring to the phrasing using "we", and personally I'd prefer avoiding it. I'm surprised by how much we are apparently using it in our in-code documentation.

It's different if we're talking through logic in a code body or an algorithm, but these are top-level comments 😕

CesiumGltf/include/CesiumGltf/PropertyView.h
j9liu: Sorry I read too fast. Could you keep the rest of the wording without `element` added?

CesiumGltf/include/CesiumGltf/PropertyView.h
azrogers: I guess I just don't see the issue with using "we"? Not only do we use it in top-level comments in Cesium Native and Cesium for Unreal, but in CesiumJS as well. And my accidental ctrl-shift-f through the whole Cesium for Unreal VS solution says that "we" is used throughout the Unreal source code and even in the MSVC standard library. And I think it makes sense - if you were explaining an algorithm to someone, you say "First we do this step, then we do this second step, and then we take the result and do this to it."

azrogers: I've changed the phrasing for the sake of getting the review done with, but I think if we want to decide on particular dos and don'ts for doc comments, that's something that should be in the style guide.

CesiumGltf/test/TestPropertyView.cpp
j9liu: @azrogers I think you may have missed this comment

CesiumGltf/include/CesiumGltf/PropertyTableView.h
j9liu: @azrogers and this one.

azrogers:
Sorry I missed those comments @j9liu, I believe I've addressed them all now.

azrogers:
@j9liu Ah, there it is, it was hidden under a "show more" that I hadn't clicked. I've removed the redundant part of the test and made a separate test for correct enum `noData` and `defaultProperty` values.

j9liu:
All good @azrogers , I know there's a lot 😁

Looks great now! Thanks for all the hard work on this PR.

Issue #991, opened 2024-11-13
Decide on #include style


lilleyse:
We're not totally consistent with how we include headers. What about something like this?

* Use quotation marks for header files in the same directory e.g. `#include "Library.h"`
* Use angle braces for everything else e.g. `#include <CesiumUtility/JsonValue.h>`

It would be great if clang-format could enforce this. In any case, once we agree on something it should go in the [C++ Style Guide](https://github.com/CesiumGS/cesium-native/blob/main/doc/style-guide.md).

kring:
Previous issue about this:
https://github.com/CesiumGS/cesium-native/issues/80
Which pretty much came to the same conclusion that you wrote above. I think that's been the decision all along, we just haven't enforced it, or retroactively converted all code to that style.

In theory we don't need to put this in the Style Guide, because it's already in the C++ Core Guidelines, which are included by reference. 🤷 
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf12-prefer-the-quoted-form-of-include-for-files-relative-to-the-including-file-and-the-angle-bracket-form-everywhere-else

kring:
Just to add a little bit to this... We used to do this thing where public header files would include other public header files _from the same library_ with quotes and a relative path. There's nothing wrong with that, except it was a special rule that we constantly forgot to follow, and it wasn't really buying us anything, anyway. At one point, Ashley wrote a tool to go through and fix all of our includes, and that tool used angle brackets and the qualified path instead, which I thought was just fine. For example, this:

```
#include "Library.h"
```

became:

```
#include <Cesium3DTilesSelection/Library.h>
```

I don't mind going back to the old pattern, but it seemed much more trouble than it was worth.

However, there is another place where we previously used quotes, and that shouldn't change. In a .cpp file, when including a _private_ header in the same directory, quotes are absolutely the right thing to use. That shouldn't - and perhaps _can't_ - change.

azrogers:
Whether or not we actually ever sat down to decide on an include style, we've been using `#include <LibraryName/HeaderName.h>` for public headers and `#include "HeaderName.h"` for private headers, and have been applying this style in PR reviews. We should update the style guide to reflect this choice, which I have changed the issue name to reflect.

PR #1114, opened 2025-02-14
Add `AxisAlignedBox::fromPoints`

j9liu:
This is just a helper function that I plan to use in #1111. That PR is getting a bit heavy, so I wanted to break it up into more digestible PRs.

### DIFF
--- a/CesiumGeometry/include/CesiumGeometry/AxisAlignedBox.h
+++ b/CesiumGeometry/include/CesiumGeometry/AxisAlignedBox.h
@@ -128,8 +128,8 @@ struct CESIUMGEOMETRY_API AxisAlignedBox final {
    * @brief Creates a tight-fitting, axis-aligned bounding box that contains all
    * of the input positions.
    *
-   * @param position The positions.
-   * @returns An axis-aligned bounding box containing the positions.
+   * @param positions The positions.
+   * @returns An axis-aligned bounding box derived from the input positions.
    */
   static AxisAlignedBox
   fromPositions(const std::vector<glm::dvec3>& positions) {
--- a/CesiumGeometry/include/CesiumGeometry/AxisAlignedBox.h
+++ b/CesiumGeometry/include/CesiumGeometry/AxisAlignedBox.h
@@ -2,6 +2,8 @@
 
 #include <CesiumGeometry/Library.h>
 
+#include <glm/ext/vector_double3.hpp>
+
 #include <vector>
 
 namespace CesiumGeometry {
--- a/CesiumGeometry/include/CesiumGeometry/AxisAlignedBox.h
+++ b/CesiumGeometry/include/CesiumGeometry/AxisAlignedBox.h
@@ -2,9 +2,6 @@
 
 #include <CesiumGeometry/Library.h>
 
-#include <glm/common.hpp>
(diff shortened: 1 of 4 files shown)


CesiumGeometry/include/CesiumGeometry/AxisAlignedBox.h
```diff
@@ -1,9 +1,12 @@
 #pragma once
 
-#include <CesiumGeometry/Library.h>
+#include "Library.h"
```
kring, line 3: We used to do this for headers in the same directory, but lately we've just been using the qualified name in angle brackets for consistency.

CesiumGeometry/include/CesiumGeometry/AxisAlignedBox.h
```diff
@@ -120,6 +123,30 @@ struct CESIUMGEOMETRY_API AxisAlignedBox final {
...
+
+  /**
+   * @brief Creates a tight-fitting, axis-aligned bounding box that contains all
+   * of the input positions.
+   *
+   * @param positions The positions.
+   * @returns An axis-aligned bounding box derived from the input positions.
+   */
+  static AxisAlignedBox
+  fromPositions(const std::vector<glm::dvec3>& positions) {
```
kring, line 135: Please create an AxisAlignedBox.cpp and put this implementation there. Less code in headers means faster compile times! This class hasn't needed a .cpp so far because all of its methods are constexpr.

kring (commented):
Thanks @j9liu, just a couple of minor comments here.

CesiumGeometry/include/CesiumGeometry/AxisAlignedBox.h
j9liu: Okay! I thought that went against our style guide, but I'll revert it.

CesiumGeometry/include/CesiumGeometry/AxisAlignedBox.h
j9liu: Makes sense. I noticed the lack of .cpp but didn't realize the reason 😅

CesiumGeometry/include/CesiumGeometry/AxisAlignedBox.h
lilleyse: It might be worth adding a section about #include style to the style guide? Especially since this is a change from [SF.12](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf12-prefer-the-quoted-form-of-include-for-files-relative-to-the-including-file-and-the-angle-bracket-form-everywhere-else).

lilleyse: I reopened https://github.com/CesiumGS/cesium-native/issues/991 noting this change.

lilleyse: That might be fixable with `.clang-format`

lilleyse: Hm maybe not: https://github.com/llvm/llvm-project/issues/39735

kring:
Thanks @j9liu!

PR #1115, opened 2025-02-14
Fix cmake debug

constant5:
I could not build the debug config of cesium-native when included as a submodule with cesium-omniverse.  I was receiving a build  error that spdlogd.lib could not be found.  I discovered that that the debug libraries were not getting copied from the right location so I added some logic to append "debug/lib/" to the PACKAGE_DIR path.


constant5:
I see this was already patched.

PR #1117, opened 2025-02-28
Fix variable length string array handling in `EXT_structural_metadata`

azrogers:
Closes CesiumGS/cesium-unreal#1619. When testing with the [ComplexTypes](https://github.com/CesiumGS/3d-tiles-samples/tree/a8d4e1f47fe8c7ffc44e78c5f9b970532f8c8bc2/glTF/EXT_structural_metadata/ComplexTypes) sample tileset in Cesium for Unreal, @j9liu noticed that the handling of variable length string arrays doesn't currently work. It turns out this is actually a problem in Cesium Native, not Unreal, and just somehow went unnoticed until now. The root of the problem is actually fairly simple: the `arrayOffsets` buffer specifies the _indices_ into the `stringOffsets` buffer, but we were treating it as if it was specifying the _byte offsets_ into that buffer. Some tests needed changing to make this work, but it wasn't much in the end. I also added the ComplexTypes sample as a test.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -25,6 +25,7 @@
 - `Tile` children of external tilesets will now be cleared when the external tileset is unloaded, fixing a memory leak that happened as a result of these `Tile` skeletons accumulating over time.
 - Requests headers specified in `TilesetOptions` are now included in tile content requests. Previously they were only included in the root tileset.json / layer.json request.
 - Fixed a crash when loading a `tileset.json` without a valid root tile.
+- Fixed a bug that could cause variable length string arrays in `EXT_structural_metadata` to be interpreted incorrectly.
 
 ### v0.44.3 - 2025-02-12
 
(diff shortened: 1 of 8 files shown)


j9liu:
Thanks @azrogers ! Just need an entry in `CHANGES.md` and this should be set to go.

kring:
Thanks @azrogers! (and thanks for the review, @j9liu!)

javagl:
Stumbling over this, some musings: The duplication between things like [`TestPropertyTableView`](https://github.com/CesiumGS/cesium-native/blob/61f2fb31ae128150fa41612cf6fc4d238f549d40/CesiumGltf/test/TestPropertyTableView.cpp#L3716) and [`BinaryPropertyTableValidationSpec`](https://github.com/CesiumGS/3d-tiles-validator/blob/31e37f85658b7e79345b3128dd8d9385740bf808/specs/metadata/BinaryPropertyTableValidationSpec.ts#L607) makes we wonder whether we could, at some point, just dump out a bunch of glTF files with all these cases, as some sort of CTS (Conformance Test Suite) for implementors to test their stuff against. 

(Technically, this could be simple (e.g. based on the validator tests - just adding some `fs.writeSync(thatTest...)` at the right place), but with some higher-level engineering questions coming in...)

PR #1118, opened 2025-02-28
Assorted fixes for `BoundingCylinderRegion`

j9liu:
- Updates for the axis change in https://github.com/CesiumGS/3d-tiles/tree/voxels/extensions/3DTILES_bounding_volume_cylinder, where `angle = 0` now happens at the `+x` axis.
- Fixed incorrect transform math. (For some reason, I did the opposite of what I intended -- transforming at the region's center instead of the reference cylinder's center. Not sure how that got flipped in my head!)
- Fixed the math for computing the new cylinder angles in implicit tiling.


kring:
Thanks @j9liu!

PR #1119, opened 2025-03-03
Fix angle math for implicit tiling of `BoundingCylinderRegion`

j9liu:
This PR fixes a bug that I unfortunately didn't realize until just now. I assumed that my unit test for `[pi, -pi`] would suffice, but turns out it's a special case that hides the problem!

Previously, the angle range of the root volume was calculated to be `glm::abs(maxAngle - minAngle)`. That works when `maxAngle >= minAngle`. But it doesn't work for `minAngle < maxAngle` cases, **unless** `minAngle = pi` and `maxAngle = -pi`.

For instance, in the below case, the actual range is `3pi / 2`, but the computed range was `pi/2`. The difference between the angles is actually a slice that is taken out of the full circle:

image

Also, I didn't put this change in the changelog... and that was intentional at first because the original implementation wasn't robust, so I thought to treat it as "experimental". Now that the implementation is more finalized, should I add this back in?

### DIFF
--- a/Cesium3DTilesContent/src/ImplicitTilingUtilities.cpp
+++ b/Cesium3DTilesContent/src/ImplicitTilingUtilities.cpp
@@ -368,8 +368,18 @@ ImplicitTilingUtilities::computeBoundingVolume(
   const glm::dvec2& rootRadialBounds = rootBoundingVolume.getRadialBounds();
   const glm::dvec2& rootAngularBounds = rootBoundingVolume.getAngularBounds();
 
+  // The angle range is "reversed" when the region sweeps over the -pi / pi
+  // discontinuity line.
+  bool angleReversed = rootAngularBounds.x > rootAngularBounds.y;
+
   double rootRadiusRange = rootRadialBounds.y - rootRadialBounds.x;
   double rootAngularRange = glm::abs(rootAngularBounds.y - rootAngularBounds.x);
+
+  if (angleReversed) {
+    // When the angle range is "reversed", the difference must be subtracted
+    // from the full circle.
+    rootAngularRange = CesiumUtility::Math::TwoPi - rootAngularRange;
+  }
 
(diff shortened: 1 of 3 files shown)


Cesium3DTilesContent/src/ImplicitTilingUtilities.cpp
```diff
@@ -368,9 +368,19 @@ ImplicitTilingUtilities::computeBoundingVolume(
...
 
+  // The angle range is "reversed" when the region sweeps over the -pi / pi
+  // discontinuity line.
+  bool angleReversed = rootAngularBounds.x > rootAngularBounds.y;
+
   double rootRadiusRange = rootRadialBounds.y - rootRadialBounds.x;
   double rootAngularRange = glm::abs(rootAngularBounds.y - rootAngularBounds.x);
 
+  if (angleReversed) {
+    // When the angle range is "reversed", the difference must be subtracted
```
timoore, line 379: I think the logic on 371-373 and here is equivalent to:
```
rootAngularRange = std::fmod(rootAngularBounds.y - rootAngularBounds.x, CesiumUtility::Math::TwoPi);
```
I don't know if there's any advantage in expressing that way. Are rootAngularBounds.{x,y} constrained to be in [-pi,pi]?

kring:
This looks good to me @j9liu. And yes, let's include it in CHANGES.md. Up to you if you want to go with Tim's alternate formulation mentioned above.

Cesium3DTilesContent/src/ImplicitTilingUtilities.cpp
j9liu: Sorry that I didn't see this earlier Tim. Yes, the bounds are meant to be constrained in `[-pi, pi]`

kring:
Thanks @j9liu!

PR #1121, opened 2025-03-05
Add `TilesetContentLoader` for content from the iTwin Cesium Curated Content service

azrogers:
The [iTwin Cesium Curated Content](https://developer.bentley.com/apis/cesium-curated-content/overview/) service allows a set of Cesium ion assets to be obtained using an iTwin token with the `itwin-platform` scope. This change adds an `ITwinCesiumCuratedContentLoader` that allows accessing this content, mostly using the existing `CesiumIonTilesetLoader` but with a different API URL and passing the token through the `Authorization` header rather than through a query parameter. It also adds the concept of a `TilesetLoaderFactory` that can be passed to the `Tileset` constructor, saving us from having to come up with a new way to overload the `Tileset` constructor for each new tileset loader we add.


azrogers:
You can test out this change with [the equivalent branch](https://github.com/CesiumGS/cesium-unreal/tree/itwin-ccc-loader) in Cesium for Unreal. I'm not sure that that one is worth making a PR for at the moment though.

azrogers:
Closing in favor of #1132.

PR #1124, opened 2025-03-06
Disable `-Wstringop-overflow`

lilleyse:
I'm getting false positives when building with `aarch64-linux-gnu-g++-13`. It looks like we disabled this warning in a different, internal project because of false positives so I think it's ok to do the same here.


kring:
@lilleyse I have no problem disabling the warning if it's giving false positives, but that's pretty unfortunate. The [docs](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wstringop-overflow) make it sound like it will flag things that _will_ overflow (not just "might"). So just to make sure, are you certain whatever it's flagging isn't a legitimate problem?

lilleyse:
I was going to get the exact error but I can't seem to reproduce this anymore. I'll reopen if it comes up again.

kring:
Thanks @lilleyse!

PR #1128, opened 2025-03-15
Credit Text: add a way to prioritize user-defined credit text

Julot94:
Only available for the user-defined credit text for now (the one applied to the whole tileset independently from the tiles actually viewed). It can be used to display a logo on the left of all credit texts, typically.

### DIFF
--- a/CesiumUtility/include/CesiumUtility/CreditSystem.h
+++ b/CesiumUtility/include/CesiumUtility/CreditSystem.h
@@ -23,6 +23,9 @@ public:
     return this->id == rhs.id;
   }
 
+  /**
+   * @brief Gets the priority of this credit.
+   */
   int32_t getPriority() const noexcept { return priority; }
 
 private:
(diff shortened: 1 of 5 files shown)


CesiumUtility/include/CesiumUtility/CreditSystem.h
```diff
@@ -23,10 +23,13 @@ struct CESIUMUTILITY_API Credit {
     return this->id == rhs.id;
   }
 
+  int32_t getPriority() const noexcept { return priority; }
```
azrogers, line 26: This should have a doc comment.

azrogers:
@Julot94 This looks like a great addition! Besides my one review comment, can you update `CHANGES.md` with a line about this under the Additions heading? It also looks like you'll need to run `clang-format` on the PR (very easy - just run `npm install` and `npm run format` in the project root directory).

CesiumUtility/include/CesiumUtility/CreditSystem.h
Julot94: I just added some doc

Julot94:
Hi @azrogers I just made the modifications.

j9liu:
One thing I'm a little worried about is whether it could cause a user to override the priority of the "Cesium ion" logo, or the "Google" logo. I don't know how much of a big deal that would be re: attribution and terms of service?

PR #1130, opened 2025-03-17
GltfReader.cpp: call Uri::resolve() with useBaseQuery = true.

AlexandreWalensa:
This is needed to fetch the textures used by tilesets coming from the Bentley mesh export service,
because the base URL contains authorization info in its parameters, which must be kept when requesting the textures.

For example, the base url looks like this:
https://xxx/tileset.json?sv=2024-05-04&spr=https&se=yyy...
The query part (everything after the question mark) contains the authorization info.
This part must be kept when building the URL for downloading textures (or buffers).

I think this is a bug, since useBaseQuery = true is already set when downloading the gltf meshes themselves.


j9liu:
Thanks for the PR @AlexandreWalensa ! Could you add a note for this to `CHANGES.md`? Other than that this seems straightforward to merge! @kring do you want to take a look as well?

AlexandreWalensa:
@j9liu I have updated CHANGES.md.

kring:
This looks good to me @j9liu!

PR #1133, opened 2025-03-20
Make CesiumGltf headers public and fix warnings

### LINKED ISSUE #460: Fix warnings in public headers

azrogers:
Closes #460, supersedes #1099. As part of #1004 we stopped marking most of the public headers as SYSTEM so that they would be analyzed by clang-tidy, but we skipped over `CesiumGltf` as the number of warnings didn't seem worth resolving at the time. This PR exposes `CesiumGltf` headers to clang-tidy analysis and compiler warnings, and fixes all the current issues.


CesiumGltf/test/TestPropertyView.cpp
```diff
@@ -23,7 +23,12 @@
 #include <glm/fwd.hpp>
 
 #include <cstdint>
-#include <ostream>
+// We're not using anything from <iostream> directly in this file, but MSVC
+// removed <iostream> from the <string> header, meaning comparisons between
+// string_view and const char* will fail unless we include it manually. Only on
+// MSVC.
+// https://developercommunity.visualstudio.com/t/visual-studio-2019-compilation-issue/525532#T-N549764.
+#include <iostream>
```
kring, line 31: Comparing `std::string_view` to `const char*` requires streams for some reason? I can't really imagine why that would be. In any case, the link above mentions you can include `<sstream>`, which I think would be better than `<iostream>`.

kring (commented):
Looks good! Just one comment/question.

CesiumGltf/test/TestPropertyView.cpp
azrogers: Now that I look at it, I believe it's doctest that's making us include the header: https://github.com/CesiumGS/cesium-native/actions/runs/13976740318/job/39132293455#step:8:631. I've changed it to `<sstream>`.

kring:
Thanks @azrogers!

PR #1134, opened 2025-03-21
Add -ftime-trace option for clang

### LINKED ISSUE #1066: Investigate `ClangBuildAnalyzer` to identify areas where we can improve build performance
While benchmarking #1064, I found out about [ClangBuildAnalyzer](https://github.com/aras-p/ClangBuildAnalyzer). Using this tool in conjunction with clang's `-ftime-trace` flag, we can generate a report of which parts of our codebase take the longest to build. The granularity is pretty impressive, allowing us to break down which files take longest to parse, which templates take the longest to generate, which functions take the longest to compile, and more. It would be nice to have this as an additional compile target we can use, like clang-tidy.

azrogers:
Closes #1066. This is a very small change that adds a `CESIUM_CLANG_TIME_TRACE` option to the CMake build process which will add the `-ftime-trace` option to the Clang command line. This option [generates JSON files compatible with the Chrome profiler](https://aras-p.info/blog/2019/01/16/time-trace-timeline-flame-chart-profiler-for-Clang/) for each object compiled. This alone is kinda neat, but it gets particularly useful when we use it in conjunction with [ClangBuildAnalyzer](https://github.com/aras-p/ClangBuildAnalyzer). This will collate all of the individual JSON files and produce a report on the whole build process.

I've [attached a report](https://github.com/user-attachments/files/19391764/report.txt) generated using the following commands:
```
cmake -B build -S . -DCESIUM_CLANG_TIME_TRACE=1
ClangBuildAnalyzer --start build
cmake --build build --config debug --target cesium-native-tests -j22
ClangBuildAnalyzer --stop build session.dat
ClangBuildAnalyzer --analyze session.dat
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -186,6 +186,7 @@ option(CESIUM_GLM_STRICT_ENABLED "Whether to force strict GLM compile definition
 option(CESIUM_DISABLE_DEFAULT_ELLIPSOID "Whether to disable the WGS84 default value for ellipsoid parameters across cesium-native." OFF)
 option(CESIUM_MSVC_STATIC_RUNTIME_ENABLED "Whether to enable static linking for MSVC runtimes" OFF)
 option(CESIUM_DEBUG_TILE_UNLOADING "Whether to enable tracking of tile _doNotUnloadSubtreeCount modifications for tile unloading debugging." OFF)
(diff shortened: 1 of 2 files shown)


kring:
Thanks @azrogers!

PR #1136, opened 2025-03-24
Fix memory leak in GltfReader::postprocessGltf

timoore:
A vsgCs user provided an ASAN log that showed that GltfReader::postprocessGltf() was leaking, and the problem was that release() was called on the unique_ptr pointing to the GltfReaderResult before it was moved. Unless I'm missing some subtlety, it should be fine to do the move and then let the unique_ptr destructor deallocate / destroy that object; that happpens after the lambda body is exited.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Fixed missing URI query part when downloading glTF textures or buffers.
+- Fixed memory leak in `CesiumGltfReader`.
 
 ### v0.45.0 - 2025-03-03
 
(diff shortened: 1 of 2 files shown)


timoore:
The last commit comment should read "make lambda immutable"...

azrogers:
@timoore I'm going to test this out but your reasoning seems to be correct. Can you update CHANGES.md?

azrogers:
Seems to work fine in Cesium for Unreal. I'll merge this once you update CHANGES.md, unless @kring notices anything wrong here I've overlooked.

kring:
Nope, it looks right to me!

kring:
Do update CHANGES.md though please!

PR #1141, opened 2025-03-28
Force CI clang-tidy job to use CMake 3

azrogers:
Currently, the clang-tidy CI task is set up to use the latest version of CMake available, for reasons that probably made sense at the time. Unfortunately, CMake just released version 4.0.0, which breaks compatibility with older versions of CMake. This is causing our CI to fail. Explicitly specifying that we want CMake 3.31.6 solves the issue.


timoore:
Thanks for diagnosing this more coherently then I!

PR #1150, opened 2025-04-01
Upsample point clouds for raster overlays

### LINKED ISSUE #1145: Raster overlays on point clouds develop holes when you get too close
Raster overlays work nicely on point clouds in Cesium for Unreal. Here's Mount St. Helens with Bing draped over it:

Image

However, when you get too close, tiles start disappearing:

Image

I believe this is because zooming in close to a tileset with a raster overlay requires upsampling, and we currently only upsample triangles. There's no good reason for this, though. Implementing [upsampleGltfForRasterOverlays](https://cesium.com/learn/cesium-native/ref-doc/structCesiumRasterOverlays_1_1RasterOverlayUtilities.html#a1fd281b14966000de37d080c10ebdad1) for points should be very straightforward. We just need to discard points that are outside the new tile bounds.

azrogers:
Fixes #1145. Currently, point clouds are not upsampled when raster overlays are applied to the tileset. This can cause holes to appear in the tileset. Fortunately, upsampling point cloud tiles is pretty simple - we just discard points outside of the region we're upsampling to fill.


CesiumRasterOverlays/src/RasterOverlayUtilities.cpp
```diff
@@ -875,8 +875,250 @@ T getVertexValue(
...
+    return false;
+  }
+
+  std::vector<uint32_t> clipVertexToIndices;
+  std::vector<CesiumGeometry::TriangleClipVertex> clippedA;
+  std::vector<CesiumGeometry::TriangleClipVertex> clippedB;
+
+  std::vector<uint32_t> vertexMap(
+      size_t(uvView.size()),
+      std::numeric_limits<uint32_t>::max());
```
kring, line 1020: These are not used.

CesiumRasterOverlays/src/RasterOverlayUtilities.cpp
```diff
@@ -875,8 +875,250 @@ T getVertexValue(
...
+  // Populate the buffers
+  Buffer& vertexBuffer = model.buffers[vertexBufferIndex];
+  vertexBuffer.cesium.data.resize(newVertexFloats.size() * sizeof(float));
+  float* pAsFloats = reinterpret_cast<float*>(vertexBuffer.cesium.data.data());
+  std::copy(newVertexFloats.begin(), newVertexFloats.end(), pAsFloats);
+  vertexBuffer.byteLength = vertexBufferView.byteLength =
+      int64_t(vertexBuffer.cesium.data.size());
+  vertexBufferView.byteStride = vertexSizeFloats * int64_t(sizeof(float));
+
+  bool onlyWater = false;
```
kring, line 1062: All this water mask stuff looks duplicated with the triangle version of the function. Can it be factored out? We don't have, and may never have, a POINTS dataset with a water mask, but it's reasonable enough to support it as long as it's easy.

kring (changes requested):
Thanks @azrogers! This looks good, but I think it can be cleaned up and simplified a little. A test or two might be good as well. If I'm reading this correctly, I don't think that it will currently copy any attributes other than POSITION to the output.

azrogers:
@kring Should be ready for another look!

kring:
This all looks good to me, but I'd still like to see some tests. Tests for the pre-existing functionality are in `TestUpsampleGltfForRasterOverlay.cpp`. Hopefully it won't be much work to create versions of a couple of these that exercise the point cloud path.

azrogers:
@kring Added a couple of tests!

kring:
Thanks @azrogers!

PR #1157, opened 2025-04-15
Fix broken ktx dependency

kring:
CC CesiumGS/cesium-unreal#1657.

This will break again the next time msys2 removes the particular verisons listed here, but a proper solution to this problem is elusive.


kring:
I've also opened an upstream PR with this change:
https://github.com/microsoft/vcpkg/pull/45040

kring:
I'm merging this in order to unbreak main.

PR #1163, opened 2025-04-24
Rasterize GeoJSON data for a RasterOverlay

azrogers:
Another step on the long road to CesiumGS/cesium-unreal#628 - a raster overlay for `VectorDocument` objects. This allows vector data loaded from GeoJSON (or, in theory, any other format once support is added) to be rasterized and overlaid onto a tileset. While not the most optimal choice in terms of visual quality, this does have the benefit of being a relatively quick-and-easy to add support for rendering vector data. 

While I originally tried to write my own triangle rasterizer, I quickly realized that it would be a better decision to find something else that does the job better than I could. [Skia](https://skia.org/) is the obvious choice - widely used, actively developed, well tested - but the fact that even the smallest binary with as many optional features turned off is still 14MB spooked me. For now, I've gone with Blend2D. This is a very fast vector rasterization library (possibly even faster than Skia), but has the disadvantage of fewer features. For one, it only really supports ARGB32 as a pixel format, which is inconvenient for this raster overlay (we have to manually swizzle it into RGBA32), but just plain wasteful for the RasterizedPolygonsOverlay. Also, pretty important for full support for vector data - it doesn't support polygons with holes in them, from what I can tell. Maybe there's an approach I'm missing, but this is necessary to actually render some polygons from a GeoJSON file or from other formats.

Also, we should do something about the line width so it looks fairly consistent across LODs - it would make the change between LODs less obvious.


azrogers:
After asking the blend2d developers, it turns out that polygons with holes are supported - I just needed to get the winding order right!

j9liu (changes requested):
Thanks @azrogers ! I took a first pass, though I need to do some homework on the raster overlays portion. @kring might have more to say if he gets a chance to look

CesiumGltf/include/CesiumGltf/ImageAsset.h
```diff
@@ -118,6 +119,33 @@ struct CESIUMGLTF_API ImageAsset final
...
   ImageAsset() = default;
 
+  /**
+   * @brief Writes this image to a TGA file. If the image has any mip levels,
+   * they will be written as `filename-mip{level}.tga`, where `{level}` is the mip level.
+   *
+   * This method is only meant to be used for debugging. It does not support any
+   * compressed pixel formats.
+   */
+  void writeTga(const std::string& outputPath) const;
```
kring: We have `ImageManipulation::savePng` already. This new function should follow that pattern.

CesiumGltf/src/ImageAsset.cpp
```diff
@@ -0,0 +1,167 @@
...
+    writeTgaImpl(
+        outputPath,
+        this->pixelData.data(),
+        this->channels,
+        this->width,
+        this->height);
+  } else {
+    std::filesystem::path outputPathParsed(outputPath);
+    for (size_t i = 0; i < this->mipPositions.size(); i++) {
+      std::filesystem::path thisPath(fmt::format(
```
kring: I'm not sure, but I think we might have problems with this code on Android. That's one reason that cesium-native avoids doing any file I/O itself, and `savePng` writes to a buffer instead of a file. The code in tests can write to files.

CesiumGltf/src/ImageAsset.cpp
```diff
@@ -0,0 +1,167 @@
...
+    // We're using fewer channels than previous, so we can perform the
+    // conversion in-place.
+    size_t writePos = 0;
+    for (size_t i = 0; i < this->pixelData.size();
+         i += (size_t)this->channels) {
+      for (size_t j = 0; j < (size_t)newChannels; j++) {
+        this->pixelData[writePos + j] = this->pixelData[i + j];
+      }
+
+      writePos += (size_t)newChannels;
```
kring:
```suggestion
      for (size_t j = 0; j < (size_t)newChannels; ++j) {
        this->pixelData[writePos] = this->pixelData[i + j];
        ++writePos;
      }
```

CesiumGltf/src/ImageAsset.cpp
```diff
@@ -0,0 +1,167 @@
...
+         i += (size_t)this->channels) {
+      for (size_t j = 0; j < (size_t)this->channels; j++) {
+        newPixelData[writePos + j] = this->pixelData[i + j];
+      }
+
+      for (size_t j = 0; j < (size_t)(newChannels - this->channels); j++) {
+        newPixelData[writePos + (size_t)this->channels + j] = defaultValue;
+      }
+
+      writePos += (size_t)newChannels;
```
kring:
```suggestion
      for (size_t j = 0; j < (size_t)this->channels; j++) {
        newPixelData[writePos] = this->pixelData[i + j];
        ++writePos;
      }

      for (size_t j = 0; j < (size_t)(newChannels - this->channels); j++) {
        newPixelData[writePos] = defaultValue;
        ++writePos;
      }
```

vcpkg.json
```diff
@@ -24,7 +24,9 @@
     "earcut-hpp",
     "cpp-httplib",
     "libmorton",
-    "zstd"
+    "zstd",
+    "blend2d",
+    "asmjit"
```
kring: It would be good to create an Unreal and Unity branch that uses this branch of cesium-native, in order to get early information about whether these new depenencies have any problems on some of our weirder platforms.

CesiumVectorData/CMakeLists.txt
```diff
@@ -48,5 +48,10 @@ target_link_libraries(CesiumVectorData
     PUBLIC
         CesiumUtility
         CesiumGeometry
+        CesiumGeospatial
         CesiumAsync
+        CesiumGltf
```
kring, line 57: Looks like this is only needed for `ImageAsset`. That's a bit unfortunate.

Probably the right solution is to create a new `CesiumImage` library for ImageAsset and a few other things. I guess. Probably not worth the trouble at the moment.

CesiumRasterOverlays/src/VectorDocumentRasterOverlay.cpp
```diff
@@ -0,0 +1,679 @@
...
+};
+
+struct Quadtree {
+  GlobeRectangle rectangle = GlobeRectangle::EMPTY;
+  uint32_t rootId = 0;
+  std::vector<QuadtreeNode> nodes;
+  std::vector<QuadtreeGeometryData> data;
+  std::vector<uint32_t> dataIndices;
+  std::vector<uint32_t> dataNodeIndicesBegin;
+};
```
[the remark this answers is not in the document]
azrogers: With `RasterizedPolygonsOverlay`, we're rendering a small handful of polygons. It's simple enough to just check each polygon to see if it's renderable for each tile, and then render it. With `VectorDocumentRasterOverlay`, we're rendering thousands, or tens of thousands, or more lines and polygons. The sample data we got from Jeroen is over 6,700 lines. The Philadelphia streets dataset is over 40,000 lines. [Here's a 1.8GB GeoJSON file](https://catalog.data.gov/dataset/blm-natl-mlrs-oil-and-gas-leases-c73da) that contains 456,542 polygons. It would be incredibly expensive if every time we need to render a tile we need to check its bounding box against 6,700+, or 40,000+, or 456,542 candidates. Using a quadtree lets us efficiently query which of those candidates we actually need to consider for this rendering, allowing us to render tiles performantly even with large datasets.

I don't think we can reuse `QuadtreeRectangleAvailability` here as 1) it doesn't look like we can attach any other data to what we put in the quadtree like we do here, 2) it seems designed around the idea that one rectangle will correspond with a single quadtree node, which isn't the case here (geometry can appear in a quadtree node as well as in the children of that node), and 3) it would require many more heap allocations than our implementation here does.

CesiumVectorData/src/VectorRasterizer.cpp
```diff
@@ -0,0 +1,282 @@
...
+VectorRasterizer::finalize() {
+  if (_finalized) {
+    return this->_imageAsset;
+  }
+
+  this->_context.end();
+
+  // Blend2D writes in BGRA whereas ImageAsset is RGBA.
+  // We need to swap the channels to fix the values.
+  // TODO: use BLPixelConverter which has SIMD support
```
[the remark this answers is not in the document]
azrogers: It doesn't seem to impact performance that much as-is. The compiler optimization vectorization seems to be doing fine already.

CesiumGltf/include/CesiumGltf/ImageAsset.h
```diff
@@ -118,6 +119,33 @@ struct CESIUMGLTF_API ImageAsset final
...
+   * is allocated and the missing channels are filled with `defaultValue`.
+   *
+   * @param newChannels The new number of channels for this image.
+   * @param defaultValue The default value for missing channels when increasing
+   * the channel count.
+   *
+   * @warning Currently only works with one-byte-per-channel images.
+   */
+  void
+  convertToChannels(int32_t newChannels, std::byte defaultValue = std::byte{0});
```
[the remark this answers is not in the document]
azrogers: I'm not sure this is a better name.

CesiumGltf/src/ImageAsset.cpp
```diff
@@ -0,0 +1,167 @@
...
+  }
+}
+
+void ImageAsset::convertToChannels(
+    int32_t newChannels,
+    std::byte defaultValue) {
+  if (newChannels == this->channels) {
+    // Nothing to do.
+    return;
+  } else if (newChannels < this->channels) {
```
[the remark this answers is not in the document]
azrogers: 1. There's a comment here for consistency with the other branches.
```
if(conditionA) { 
    // We do nothing.
}
else if(conditionB) {
    // We do A, B, C
}
else {
    // We do D, E, F
}
```
While we could remove it, I don't think we have a character count limit for this file.
2. For similar reasons to the above, I prefer the `else` being here. It shows that these three branches are connected, it's a fork in the road with three possible paths. We don't really gain anything by dropping the `else` keyword but I think it would lose a little bit of readability.

CesiumGltf/include/CesiumGltf/ImageAsset.h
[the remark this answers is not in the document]
azrogers: "Resize" as an action has a very common meaning for images that I think this method name would get confused with. Meanwhile, Photoshop calls the operation of changing an image's color profile (the definition of how many channels and what each channel means) "Convert to Profile." I think "convert" is the more correct name for this operation.

CesiumVectorData/src/VectorRasterizer.cpp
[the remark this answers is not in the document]
j9liu: Alright, then perhaps the remove the TODO since it's no longer a plan?

CesiumRasterOverlays/src/VectorDocumentRasterOverlay.cpp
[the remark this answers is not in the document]
j9liu: Thanks for fielding my questions! It'd be great to summarize a couple lines of your explanation and document this approach as comments, then.

CesiumVectorData/src/VectorRasterizer.cpp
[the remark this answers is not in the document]
azrogers: I wouldn't say it's no longer a plan, just that it doesn't seem essential to getting this PR in. It's a not insignificant amount of work that would give some minor performance benefit, which would be good in the long term but doesn't make a lot of sense right now.

CesiumGltf/include/CesiumGltf/ImageAsset.h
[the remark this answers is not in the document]
j9liu: Hm I'm coming from the context of the code itself, where we'd literally be resizing the underlying vector of data. I'd also think that the inclusion of `Channels` would help with this understanding. But I'm not going to fight hard for this change.

My brain would still like to autocorrect `convertToChannels` to `convertChannels`, but that is another nitpick as well.

CesiumVectorData/src/VectorRasterizer.cpp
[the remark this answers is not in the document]
j9liu: I see, I misunderstood your comment. Thanks!

CesiumGltf/include/CesiumGltf/ImageAsset.h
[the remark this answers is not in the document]
kring: When in doubt, be verbose! `changeNumberOfChannels` maybe? 😀

CesiumUtility/include/CesiumUtility/Color.h
```diff
@@ -0,0 +1,36 @@
...
+  /**
+   * @brief Creates a new Color from the given components.
+   *
+   * @param r_ The red component.
+   * @param g_ The green component.
+   * @param b_ The blue component.
+   * @param a_ The alpha component.
+   */
+  Color(uint8_t r_, uint8_t g_, uint8_t b_, uint8_t a_ = 0xff)
+      : r(r_), g(g_), b(b_), a(a_) {}
```
kring: Move the implementation to the .cpp file, and remove the underscores from the header (only).

CesiumVectorData/include/CesiumVectorData/GeoJsonObject.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "CesiumVectorData/VectorStyle.h"
 #include "GeoJsonObjectTypes.h"
```
kring:
```suggestion
#include <CesiumVectorData/VectorStyle.h>
#include <CesiumVectorData/GeoJsonObjectTypes.h>
```

CesiumVectorData/include/CesiumVectorData/VectorStyle.h
```diff
@@ -0,0 +1,114 @@
...
+   *
+   * For `ColorMode::Normal`, this just returns the value of `color`. For
+   * `ColorMode::Random`, this returns a randomized value obtained based on
+   * the rules described in \ref ColorMode.
+   */
+  CesiumUtility::Color getColor() const;
+};
+
+/** @brief The style used to draw a Polygon. */
+struct PolygonStyle : public ColorStyle {
```
kring: I think I'd prefer aggregation rather than inheritance here. In other words, give `PolygonStyle` two `ColorStyle` properties, one named `fillColor` and the other named `outlineColor`. With inheritance, the `fillColor` property will be named simply `color`, which is less clear. And also it prevents using a different colors for lines versus polygon outlines. Unless there's some compelling reason to say that a `PolygonStyle` IS A `ColorStyle` that isn't obvious to me?

kring: Using `optional` here might be a nice design, too, instead of the two booleans. Something like:

```cpp
std::optional<ColorStyle> fill;
std::optional<LineStyle> outline;
```

CesiumVectorData/include/CesiumVectorData/VectorStyle.h
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * @brief The line width will cover this number of meters of the ellipsoid
+   * it's rendered on. This may cause the line to disappear as the user zooms
+   * out.
+   */
+  Meters = 1
+};
+
+/** @brief The style used to draw polylines and strokes. */
+struct LineStyle : public ColorStyle {
```
kring, line 66: Inheritance makes more sense here, though.

CesiumVectorData/include/CesiumVectorData/VectorRasterizer.h
```diff
@@ -0,0 +1,120 @@
+#pragma once
+
+#include "VectorStyle.h"
```
kring, line 3: I'm largely ok with this, since it's a header in the same directory. But we seem to have moved toward using the explicit `<CesiumVectorData/VectorStyle.h>` everywhere now, so I think we should be consistent about that. Just curious, do you have an IDE that is automatically adding headers with the incorrect style on you? If so, there might be a setting you can tweak so we don't keep hassling you about it in reviews. 😀

CesiumVectorData/include/CesiumVectorData/VectorRasterizer.h
```diff
@@ -0,0 +1,120 @@
...
+
+  /**
+   * @brief Draws a \ref CesiumGeospatial::CartographicPolygon to the canvas.
+   *
+   * @param polygon The polygon to draw.
+   * @param style The \ref VectorStyle to use when drawing the polygon.
+   */
+  void drawPolygon(
+      const CesiumGeospatial::CartographicPolygon& polygon,
+      const VectorStyle& style);
```
kring: If you take my previous suggestion for changes to `PolygonStyle`, then I think this method could simply take a `PolygonStyle` instead of an entire `VectorStyle`. Similar with the other methods; they can just take the style type appropriate for what they're drawing.

CesiumVectorData/include/CesiumVectorData/VectorRasterizer.h
```diff
@@ -0,0 +1,120 @@
...
+   * @brief Rasterizes the provided `GeoJsonObject` to the canvas.
+   *
+   * Polygons are equivalent to calls to `drawPolygon`. Polylines are equivalent
+   * to calls to `drawPolyline`. Points are currently not drawn.
+   *
+   * @param geoJsonObject The GeoJSON object to draw.
+   * @param style The \ref VectorStyle to use when drawing the object.
+   */
+  void drawGeoJsonObject(
+      const GeoJsonObject* geoJsonObject,
```
kring: Why a pointer here, when the other methods take references?

CesiumVectorData/include/CesiumVectorData/VectorRasterizer.h
```diff
@@ -0,0 +1,120 @@
...
+   */
+  void drawPolyline(
+      const std::span<const glm::dvec3>& points,
+      const VectorStyle& style);
+
+  /**
+   * @brief Rasterizes the provided `GeoJsonObject` to the canvas.
+   *
+   * Polygons are equivalent to calls to `drawPolygon`. Polylines are equivalent
+   * to calls to `drawPolyline`. Points are currently not drawn.
```
kring: I guess I'd expect this method to draw everything in the GeoJsonObject. For example, if it were a `FeatureCollection`, it would draw every line and polygon in any feature in the collection. Is that a crazy expectation? If so, we should at least update the docs to make it very clear that this is not the case. It mentions that points are not drawn, but it doesn't mention other things.

CesiumRasterOverlays/include/CesiumRasterOverlays/GeoJsonDocumentRasterOverlay.h
```diff
@@ -0,0 +1,135 @@
+#pragma once
+
+#include "CesiumAsync/IAssetAccessor.h"
```
kring:
```suggestion
#include <CesiumAsync/IAssetAccessor.h>
```

CesiumRasterOverlays/include/CesiumRasterOverlays/GeoJsonDocumentRasterOverlay.h
```diff
@@ -0,0 +1,135 @@
...
+using GeoJsonDocumentRasterOverlaySource = std::variant<
+    std::shared_ptr<CesiumVectorData::GeoJsonDocument>,
+    IonGeoJsonDocumentRasterOverlaySource,
+    UrlGeoJsonDocumentRasterOverlaySource>;
+
+/**
+ * @brief A raster overlay made from rasterizing a \ref
+ * CesiumVectorData::GeoJsonDocument.
+ */
+class CESIUMRASTEROVERLAYS_API GeoJsonDocumentRasterOverlay final
```
kring, line 48: I think a simpler design might be possible here.

At a high level, this raster overlay just need a `GeoJsonDocument`, which already includes styling information, and it can render it. The complexity comes from two places:
- Because the raster overlay loads the GeoJsonDocument itself, a callback is necessary to allow the user to style it before it's rendered.
- Because the raster overlay loads the GeoJsonDocument itself, we need to provide users with multiple sources for the GeoJsonDocument.

So really that's just one source of complexity. What if the raster overlay doesn't load the GeoJsonDocument itself? Things get much simpler, right?

```cpp
GeoJsonDocumentRasterOverlay(
      const std::string& name,
      const std::shared_ptr<GeoJsonDocument>& geoJson,
      const GeoJsonDocumentRasterOverlayOptions& vectorOverlayOptions,
      const RasterOverlayOptions& overlayOptions = {});
```

But one downside to this is that it means there's no way to create and add the raster overlay until the GeoJsonDocument is loaded. That might be annoying. But we can solve that with another constructor overload that takes a Future:

```cpp
GeoJsonDocumentRasterOverlay(
      const std::string& name,
      const CesiumAsync::Future<std::shared_ptr<GeoJsonDocument>>&& futureGeoJson,
      const GeoJsonDocumentRasterOverlayOptions& vectorOverlayOptions,
      const RasterOverlayOptions& overlayOptions = {});
```

The nice part about this is that there's no need for a callback. Users just walk the document after it's loaded and apply styling, if they want. And it will work reasonably well with GeoJSON from URLs, Cesium ion, iTwin, or whatever else, without needing to teach this class about all the things.

Obviously this is different from how `Tileset` works. But I think that's justified because constructing a `Tileset` really represents the start of a conversation where we need to make lots of different requests, and even refresh tokens in some cases. Whereas with GeoJSON it's just download one file and we're done.

kring (commented):
So far I've really only reviewed the header files, to get a sense of the design. I think it looks awesome so far, though I do have a few comments and suggestions.

CesiumVectorData/include/CesiumVectorData/VectorRasterizer.h
azrogers: Yep, VS Code keeps doing it to me. I've been unable to find a setting to change that behavior.

vcpkg.json
azrogers: Already created an Unreal branch: https://github.com/CesiumGS/cesium-unreal/pull/1677

Non-iOS platforms seem to be failing for mundane reasons. iOS is the only one that has issues, because `sys_icache_invalidate` doesn't exist on iOS. More broadly, JIT in general isn't an option for iOS - it requires the `com.apple.security.cs.allow-jit` entitlement in the app bundle to allow memory to be mapped with the `MAP_JIT` flag, and [this entitlement only exists on macOS](https://developer.apple.com/documentation/BundleResources/Entitlements/com.apple.security.cs.allow-jit).

CesiumGltf/include/CesiumGltf/ImageAsset.h
azrogers: I'm going to move the `writeTga` stuff to `CesiumNativeTests`, since it's only there for testing purposes anyways.

CesiumVectorData/include/CesiumVectorData/VectorStyle.h
azrogers: The reason for the way it's designed currently is that I cribbed it from KML, which isn't really a good excuse.

CesiumRasterOverlays/include/CesiumRasterOverlays/GeoJsonDocumentRasterOverlay.h
azrogers: I think that is a lot simpler! I don't like how the constructor now needs `asyncSystem` so that it can create the resolved future for `createTileProvider` to use, but besides that I think it's a good change.

CesiumVectorData/include/CesiumVectorData/VectorRasterizer.h
```diff
@@ -0,0 +1,128 @@
...
+   * CesiumGltf::ImageAsset must be four channels, with
+   * only one byte per channel (RGBA32).
+   * @param mipLevel The mip level that the rasterizer should rasterize for the
+   * image.
+   * @param ellipsoid The ellipsoid to use.
+   */
+  VectorRasterizer(
+      const CesiumGeospatial::GlobeRectangle& bounds,
+      CesiumUtility::IntrusivePointer<CesiumGltf::ImageAsset>& imageAsset,
+      uint32_t mipLevel = 0,
```
kring, line 44: Would it be worthwhile to rasterize all mip levels simultaneously? I think there would be a bit of efficiency to be gained that way. It would save creating the `std::vector<BLPoint>` multiple times, at least.

CesiumVectorData/src/VectorRasterizer.cpp
```diff
@@ -0,0 +1,280 @@
...
+void setStrokeWidth(
+    BLContext& context,
+    const LineStyle& style,
+    const Ellipsoid& ellipsoid,
+    const GlobeRectangle& bounds) {
+  if (style.widthMode == LineWidthMode::Pixels) {
+    context.setStrokeWidth(style.width);
+  } else if (style.widthMode == LineWidthMode::Meters) {
+    const double radians = style.width / ellipsoid.getRadii().x;
+    context.setStrokeWidth(radians / bounds.computeWidth());
```
kring: Is this right? Unless I'm missing something, it needs to incorporate the width of the canvas in pixels. A line that is a certain number of meters wide will be some number of radians wide, and that is a function of the ellipsoid equitorial radius, as you have here. But the number of _pixels_ wide that line should be will also depend on the total resolution of the canvas.

kring: My hasty (and quite possibly wrong) derivation:

```
// In the geographic projection, angles and pixels are proportional.
(lineWidthPixels / lineWidthRadians) = (canvasWidthPixels / canvasWidthRadians)
lineWidthPixels = (canvasWidthPixels * lineWidthRadians) / canvasWidthRadians

// This is just the arc length formula
equatorialDistanceInMeters = angleInRadians * ellipsoid.getRadii().x

// Same equation as above, applied to line widths.
lineWidthMeters = lineWidthRadians * ellipsoid.getRadii().x
lineWidthRadians = lineWidthMeters / ellipsoid.getRadii().x

// Substitute and reorganize
lineWidthPixels = (canvasWidthPixels * (lineWidthMeters / ellipsoid.getRadii().x)) / canvasWidthRadians
lineWidthPixels = (canvasWidthPixels * lineWidthMeters) / (canvasWidthRadians * ellipsoid.getRadii().x)
```

So I think it'd be this:

```
context.setStrokeWidth((context.targetWidth() * style.width) / (bounds.computeWidth() * ellipsoid.getRadii().x));
```

CesiumVectorData/include/CesiumVectorData/VectorStyle.h
```diff
@@ -0,0 +1,113 @@
...
+   * @brief The line will always be this number of pixels in width, no matter
+   * how close the user gets to the line.
+   */
+  Pixels = 0,
+  /**
+   * @brief The line width will cover this number of meters of the ellipsoid
+   * it's rendered on. This may cause the line to disappear as the user zooms
+   * out.
+   */
+  Meters = 1
```
kring, line 62: Clarify that this is meters _at the equator_. These "meters" get smaller as we move toward the equator.

Alternatively, the line width could be expressed directly in radians, and wouldn't need as much explanation. That might be less confusing. Maybe.

Maintaining an actual fixed width in meters would require adjusting the line width by latitude even within a single tile. That'd be cool if we could do it, but it's probably unrealistic. Setting the line width based on the average latitude or something would cause a discontinuity between adjacent tiles, so I don't think that would be good.

CesiumRasterOverlays/src/RasterizedPolygonsOverlay.cpp
```diff
@@ -83,88 +86,38 @@ void rasterizePolygons(
...
+      0,
+      ellipsoid);
+  rasterizer.clear(CesiumUtility::Color{outsideColor, 0x00, 0x00, 0xff});
+
+  CesiumVectorData::PolygonStyle insideStyle{
+      CesiumVectorData::ColorStyle{
+          CesiumUtility::Color{insideColor, 0x00, 0x00, 0xff}},
+      std::nullopt};
+  for (const CartographicPolygon& selection : cartographicPolygons) {
+    rasterizer.drawPolygon(selection, insideStyle);
```
kring, line 115: This new implementation is probably much faster than the old one too, right? Worth mentioning in the changelog, if so.

CesiumRasterOverlays/src/GeoJsonDocumentRasterOverlay.cpp
```diff
@@ -0,0 +1,586 @@
...
+    rasterizer.finalize();
+  }
+}
+} // namespace
+
+class CESIUMRASTEROVERLAYS_API GeoJsonDocumentRasterOverlayTileProvider final
+    : public RasterOverlayTileProvider {
+
+private:
+  std::shared_ptr<GeoJsonDocument> _document;
```
kring:
```suggestion
  std::shared_ptr<GeoJsonDocument> _pDocument;
```

kring (commented):
I haven't finished reviewing the rasterizer implementation just yet, but here are a few more comments in the meantime.

CesiumGltf/src/ImageAsset.cpp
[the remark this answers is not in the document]
j9liu: 1. My intention isn't to reduce character count. Comments serve best when they clarify non-obvious things in code, and this first branch is a bit obvious. I would also argue for removing redundant comments in other parts of the code, but I'm not gonna get up in arms about it.

2. Readability is actually the reason I made that suggestion, but it seems like we have different preferences for that, which is okay. 😌 For me the lack of `else` helps with readability, given that things can `return` early. But if it genuinely helps you with readability, then keep it in.

CesiumVectorData/include/CesiumVectorData/VectorRasterizer.h
azrogers: I took a look into implementing it, and it's definitely doable, but one catch is that we would still have to *fill* the `std::vector<BLPoint>` each time, as the radians need to be converted to pixel positions based on the width and height of the image each time - so we'd only be saving some repeated allocations. Feels like a lot of added complexity for something that might not have a significant impact on performance. If we could use the transformation matrix of the canvas to convert between radians and pixel coordinates, it'd be a lot more efficient, but that caused issues with stroke width last time I gave that a shot.

CesiumRasterOverlays/src/GeoJsonDocumentRasterOverlay.cpp
```diff
@@ -0,0 +1,665 @@
...
+       i < std::max(result.pImage->mipPositions.size(), (size_t)1);
+       i++) {
+    VectorRasterizer rasterizer(
+        rectangle,
+        result.pImage,
+        (uint32_t)i,
+        ellipsoid);
+    // Keeps track of primitives that have already been rendered to avoid
+    // re-drawing the same primitives that appear in multiple quadtree nodes.
+    std::vector<bool> primitivesRendered(tree.data.size(), false);
```
kring: Small and easy optimization opporunity: move this outside the loop, then just `resize` and `assign` inside the loop.

CesiumVectorData/src/VectorRasterizer.cpp
```diff
@@ -0,0 +1,281 @@
...
+  const int32_t imageWidth =
+      std::max(this->_imageAsset->width >> this->_mipLevel, 1);
+  const int32_t imageHeight =
+      std::max(this->_imageAsset->height >> this->_mipLevel, 1);
+  std::byte* pData =
+      this->_mipLevel == 0
+          ? this->_imageAsset->pixelData.data()
+          : this->_imageAsset->pixelData.data() +
+                this->_imageAsset->mipPositions[this->_mipLevel].byteOffset;
+  _image.createFromData(
```
kring:
```suggestion
  this->_image.createFromData(
```

CesiumVectorData/src/VectorRasterizer.cpp
```diff
@@ -0,0 +1,281 @@
...
+      imageWidth,
+      imageHeight,
+      BL_FORMAT_PRGB32,
+      reinterpret_cast<void*>(pData),
+      (int64_t)imageWidth * (int64_t)this->_imageAsset->channels);
+
+  _context.begin(this->_image);
```
kring:
```suggestion
  this->_context.begin(this->_image);
```

CesiumVectorData/src/VectorRasterizer.cpp
```diff
@@ -0,0 +1,281 @@
...
+  // Initialize the image as all transparent.
+  _context.clearAll();
```
kring:
```suggestion
  this->_context.clearAll();
```

CesiumVectorData/src/VectorRasterizer.cpp
```diff
@@ -0,0 +1,281 @@
...
+}
+
+void VectorRasterizer::drawPolygon(
+    const CesiumGeospatial::CartographicPolygon& polygon,
+    const PolygonStyle& style) {
+  if (_finalized || (!style.fill && !style.outline)) {
```
kring:
```suggestion
  if (this->_finalized || (!style.fill && !style.outline)) {
```

CesiumVectorData/src/VectorRasterizer.cpp
```diff
@@ -0,0 +1,281 @@
...
+        vertices.data(),
+        vertices.size(),
+        BLRgba32(style.outline->getColor().toRgba32()));
+  }
+}
+
+void VectorRasterizer::drawPolygon(
+    const std::vector<std::vector<glm::dvec3>>& polygon,
+    const PolygonStyle& style) {
+  if (_finalized || (!style.fill && !style.outline)) {
```
kring:
```suggestion
  if (this->_finalized || (!style.fill && !style.outline)) {
```

CesiumVectorData/src/VectorRasterizer.cpp
```diff
@@ -0,0 +1,281 @@
...
+        vertices.data(),
+        vertices.size(),
+        BLRgba32(style.outline->getColor().toRgba32()));
+  }
+}
+
+void VectorRasterizer::drawPolyline(
+    const std::span<const glm::dvec3>& points,
+    const LineStyle& style) {
+  if (_finalized) {
```
kring:
```suggestion
  if (this->_finalized) {
```

CesiumVectorData/src/VectorRasterizer.cpp
```diff
@@ -0,0 +1,281 @@
...
+    void operator()(const GeoJsonGeometryCollection& /*catchAll*/) {}
+  };
+
+  for (const GeoJsonObject& object : geoJsonObject) {
+    object.visit(PrimitiveDrawVisitor{*this, style});
+  }
+}
+
+void VectorRasterizer::clear(const CesiumUtility::Color& clearColor) {
+  if (_finalized) {
```
kring:
```suggestion
  if (this->_finalized) {
```

CesiumVectorData/src/VectorRasterizer.cpp
```diff
@@ -0,0 +1,281 @@
...
+    return;
+  }
+
+  this->_context.fillAll(BLRgba32(clearColor.toRgba32()));
+}
+
+CesiumUtility::IntrusivePointer<CesiumGltf::ImageAsset>
+VectorRasterizer::finalize() {
+  if (_finalized) {
```
kring, line 248:
```suggestion
  if (this->_finalized) {
```

CesiumVectorData/src/VectorRasterizer.cpp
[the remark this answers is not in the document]
kring: We should generally avoid introducing new TODOs into main. One problem is it lacks the context in these comments. Someone coming along later may wonder if this TODO is a major issue.

So it would be better to write an issue for it and remove the TODO. If it doesn't seem worth writing an issue for, then simply changing it to a non-TODO comment saying essentially what you wrote above would be better. Something like: "We could use BLPixelConverter here, which should be faster because it has SIMD support. This doesn't seem to be much of a performance impact as-is, though, perhaps because the compiler's vectorization is doing a fine job."

CesiumVectorData/src/VectorRasterizer.cpp
```diff
@@ -0,0 +1,281 @@
...
+    // ARGB to ABGR.
+    uint32_t* pPixel = reinterpret_cast<uint32_t*>(pData + i);
+    const uint32_t pixel = *pPixel;
+    const uint32_t newPixel =
+        (pixel & 0xff000000) | ((pixel & 0x00ff0000) >> 16) |
+        (pixel & 0x0000ff00) | ((pixel & 0x000000ff) << 16);
+    *pPixel = newPixel;
+  }
+
+  _finalized = true;
```
kring:
```suggestion
  this->_finalized = true;
```

CesiumRasterOverlays/src/GeoJsonDocumentRasterOverlay.cpp
```diff
@@ -0,0 +1,665 @@
...
+          documentRegionBuilder,
+          useStyle ? *useStyle : style);
+    }
+  }
+
+  void operator()(const auto& /*catchAll*/) {}
+};
+
+void addPrimitivesToData(
+    const GeoJsonObject* geoJsonObject,
```
kring, line 260:
```suggestion
    const GeoJsonObject* pGeoJsonObject,
```

CesiumRasterOverlays/src/GeoJsonDocumentRasterOverlay.cpp
```diff
@@ -0,0 +1,665 @@
...
+    std::vector<QuadtreeGeometryData>& data,
+    BoundingRegionBuilder& documentRegionBuilder,
+    const VectorStyle& style) {
+  BoundingRegionBuilder thisBuilder;
+  std::visit(
+      GlobeRectangleFromObjectVisitor{thisBuilder},
+      geoJsonObject->value);
+  GlobeRectangle rect = thisBuilder.toGlobeRectangle();
+  documentRegionBuilder.expandToIncludePosition(rect.getSouthwest());
+  documentRegionBuilder.expandToIncludePosition(rect.getNortheast());
```
kring: I don't think this will be entirely reliable, because of anti-meridian handling. Imagine that `thisBuilder` is a giant region that stretches across the entire globe, except that it stops 5 degrees short of the anti-meridian on both sides. So -175 to 175 degrees longitude, let's say. If you add the corners of that region to a new builder, you'll end up with a much smaller region that crosses the anti-meridian and only covers 175 to -175 degrees longitude.

I think a reasonable solution is to add a new method to `BoundingRegionBuilder`: `expandToIncludeRectangle`.

CesiumRasterOverlays/src/GeoJsonDocumentRasterOverlay.cpp
```diff
@@ -0,0 +1,665 @@
...
+
+  uint32_t resultId = (uint32_t)tree.nodes.size();
+  tree.nodes.emplace_back(rectangle);
+
+  uint32_t indicesBegin = (uint32_t)tree.dataIndices.size();
+  tree.dataIndices.insert(tree.dataIndices.end(), begin, end);
+
+  tree.dataNodeIndicesBegin.emplace_back(indicesBegin);
+
+  if (begin + 1 == end || depth >= DEPTH_LIMIT ||
```
kring: I'm a little worried about this depth limit. For a dataset with a lot of lines in a small space, it could mean that all the lines end up in the same quadtree node, right? Allowing unlimited depth obviously isn't the solution, though. Any ideas? Off the top of my head, I'm thinking maybe some kind of heuristic related to the number of objects in the node? So the first object just goes in the root node, but after `n` objects are added to the root node, the algorithm declares that too many and pushes them down to the children. And pushing down process may repeat if they all push into the same child of the root.

CesiumRasterOverlays/src/GeoJsonDocumentRasterOverlay.cpp
```diff
@@ -0,0 +1,665 @@
...
+      std::equal(begin + 1, end, begin)) {
+    return resultId;
+  }
+
+  Cartographic southWest = rectangle.getSouthwest();
+  Cartographic northEast = rectangle.getNortheast();
+  Cartographic center = rectangle.computeCenter();
```
kring: It might not matter too much here, but it's a good idea to form quadtree bounds by multiplication instead of repeated division at the center point. In other words, compute level/x/y for the node and then use `QuadtreeTilingScheme` to compute its rectangle. The problem is that repeated division becomes a bit numerically unstable as you go deeper in the tree.

kring (commented):
Looks great @azrogers! Mostly nitpick comments below, but also a couple of larger issues for your consideration.

CesiumRasterOverlays/src/GeoJsonDocumentRasterOverlay.cpp
kring: If you implement something like I described above, watch out for crazy cases. Like a thousand different points all at the same location.

CesiumGeospatial/include/CesiumGeospatial/GlobeRectangle.h
```diff
@@ -61,6 +61,21 @@ class CESIUMGEOSPATIAL_API GlobeRectangle final {
...
+  /**
+   * @brief Constructs a new `GlobeRectangle` from the provided `Rectangle`.
+   *
+   * The rectangle's minimum coordinates will be interpreted as the southwest
+   * position and the maximum coordinates as the northeast.
+   *
+   * @param rectangle The rectangle defining the bounds of the new
+   * `GlobeRectangle`.
+   */
+  constexpr GlobeRectangle(const CesiumGeometry::Rectangle& rectangle) noexcept
```
kring: I think this implicit conversion is dangerous. It means that a `Rectangle` will be silently turned into a `GlobeRectangle`, but that only works when the original `Rectangle` is expressed in radians, which is not guaranteed. In other cases, it will silently produce wrong results. Better to leave this as an explicit conversion, even if it's more verbose.

CesiumGeospatial/src/BoundingRegionBuilder.cpp
```diff
@@ -123,4 +125,31 @@ bool BoundingRegionBuilder::expandToIncludePosition(
...
+
+  if (rectangle.getNorth() > this->_rectangle.getNorth()) {
+    this->_rectangle.setNorth(rectangle.getNorth());
+    modified = true;
+  }
+
+  if (rectangle.getEast() < this->_rectangle.getEast()) {
+    this->_rectangle.setEast(rectangle.getEast());
+    modified = true;
+  }
```
kring: I don't think this will handle the anti-meridian correctly. It'd be good to see some tests for this, particular where either the builder, the rectangle, or both cross the anti-meridian.

CesiumGeospatial/include/CesiumGeospatial/BoundingRegionBuilder.h
```diff
@@ -71,6 +71,17 @@ class CESIUMGEOSPATIAL_API BoundingRegionBuilder {
...
   bool expandToIncludePosition(const Cartographic& position);
 
+  /**
+   * @brief Expands the bounding region to include the given globe rectangle.
+   *
+   * The region will be kept as small as possible.
+   *
+   * @param rectangle The rectangle to be included in the region.
+   * @returns True if the region was modified, or false if the region already
+   * contained the position.
```
kring:
```suggestion
   * contained the rectangle.
```

kring (commented):
Just a couple more minor comments. I think my previous comment about the depth limit still needs to be addressed, too. Not necessarily with a code change, but at least some thought and a response. Otherwise, this looks ready to go to me!

azrogers:
Was going to push a fix to the `BoundingRegionBuilder` antimeridian handling but looks like you already solved it 😅 Your solution is more elegant than mine was going to be anyways!

kring:
Thanks @azrogers, great work on this!

PR #1164, opened 2025-04-25
Enable SIMD code paths for GLM

azrogers:
This PR enables the `GLM_FORCE_INTRINSICS` define that forces GLM to use SIMD code paths on platforms that have SIMD support. It also moves the `GLM_ENABLE_EXPERIMENTAL` define out of the `CESIUM_GLM_STRICT_ENABLED` CMake option, since I don't believe this define has anything to do with strict mode.


shehzan10:
@azrogers Can we run brief tests to check if there are notable performance gains by enabling this?

azrogers:
@shehzan10 We don't really have any benchmarks set up in Native, nor do we really have any in our integrations. We could write some benchmarks, but it wouldn't be a particularly quick job, and I wouldn't expect it to show any notable improvements; glm isn't really our bottleneck.

kring:
@azrogers we have some automated performance tests in Cesium for Unreal that we could run pretty easily. I also would be surprised to see any measurable difference from this change, though, so agree it's not worth spending any significant time on.

kring:
I created `glm-simd` branches of both cesium-unreal and cesium-unity. If CI is happy for both, I'll merge this.
* https://github.com/CesiumGS/cesium-unreal/tree/glm-simd
* https://github.com/CesiumGS/cesium-unity/tree/glm-simd

kring:
Thanks @azrogers!

PR #1167, opened 2025-04-29
Include `Tile.h` in `ViewUpdateResult.h`

j9liu:
This was needed to compile the `multiple-views` changes in Cesium for Unity. The build step could not resolve the `std::vector<Tile::Pointer>`s in `ViewUpdateResult` unless we included `Tile.h`.


kring:
Thanks @j9liu!

PR #1170, opened 2025-04-30
Don't reference Microsoft.GSL in Config.cmake.in

timoore:
It's not used anymore and, because it's not mentioned in the vcpkg.json manifest file, will cause an error if no other package uses it.


kring:
Thanks @timoore!

PR #1171, opened 2025-05-02
Add `SubtreeWriter::writeSubtreeBinary`

lilleyse:
Adds `SubtreeWriter::writeSubtreeBinary` for writing the [subtree binary format](https://github.com/CesiumGS/3d-tiles/tree/main/specification/ImplicitTiling#subtree-binary-format). Previously we only supported writing the JSON format.


kring:
Thanks @lilleyse!

PR #1173, opened 2025-05-09
Fix use-after-free bug in `SubtreeAvailability`

lilleyse:
Steps to reproduce

1. Create unavailable subtree
2. Set a tile's availability to true
3. Set a content's availability to true
4. Set a tile's availability to true
5. `undefined behavior`

Step 3 causes `subtree.buffers[0].cesium.data` to grow which invalidates the span held onto by tile availability.

Fixed by updating all the spans when `convertConstantAvailabilityToBitstream` is called so that they point to valid memory.


j9liu (approved):
Thanks @lilleyse, this looks good to me! I confirmed that the unit test fails on main but passes on this PR.

PR #1174, opened 2025-05-09
Add boolean availability parameter in `SubtreeAvailability::createEmpty`

j9liu:
This PR adds `setTilesAvailable` to `SubtreeAvailability::createEmpty`. The default behavior was sufficiently documented (initializing all tiles as available), but a bit unexpected because of the implications of "empty". Either way, it helps to have the option to explicitly set how to initialize the tile availability.


lilleyse:
Looks good!

PR #1177, opened 2025-05-13
Add a "render-readiness" flag on tiles

GhisBntly:
Render engine or user-developer customizations can toggle it off to delay the point at which the tile can be shown, while waiting for some asynchronous post-processing to finish


GhisBntly:
Related to https://github.com/CesiumGS/cesium-unreal/pull/1669

PR #1178, opened 2025-05-14
Several maintenance items

kring:
## vcpkg upgrade

The vcpkg release we were using, 2024.11.16, doesn't work with CMake 4. On most platforms, this isn't too hard to work around by installing CMake 3.31 or by setting the `CMAKE_POLICY_VERSION_MINIMUM` environment variable to `3.5`. But on macOS, homebrew makes it hard to install an old cmake, and setting that environment variable isn't sufficient due to a [separate problem](https://github.com/microsoft/vcpkg/pull/44712) on macOS. We need to upgrade to at least 2025.04.09 to get the fix for both of these problems.

However, 2025.04.09 has a problem in the `ktx` port. I opened [a PR](https://github.com/microsoft/vcpkg/pull/45040) to fix this, and the vcpkg folks merged it, but it hasn't gone out in an official release yet. I've been waiting for a new vcpkg release, but it has been longer than I expected, and in the meantime using CMake 4, especially on a Mac, is really annoying.

So this PR updates our vcpkg version to the current head. We'll switch to an official release in a separate PR as soon as one comes out.

## Use Ubuntu 24.04 for documentation builds

Previously we were using 22.04. ada-url in the latest vcpkg release doesn't seem to like 22.04's version of GCC. Moving off the old Ubuntu release is an easy fix.

## Doxygen version

Previously, we were using Doxygen 1.13.2 to build doc as part of the normal process, but then using 1.12.0 to build for production. Not what we want. Now it's all 1.13.2.

## Overlay ports and triplets

I removed the now-unnecessary `ktx` vcpkg overlay port. And had to add README files in the ports and triplets directories to prevent vcpkg from complaining that they don't exist.


j9liu:
Thanks @kring ! I don't have a Mac to this, but at least I don't see any unexpected changes here on Windows.

PR #1179, opened 2025-05-16
Fix Tile reference counting assertion with implicit tiling

kring:
@j9liu found a tileset that was causing reference count assertion errors when clicking "Refresh Tileset" in Cesium for Unreal. The problem essentially boiled down to the fact that the tileset used implicit tiling and some tiles that exist but do not have content, and as a result `ImplicitOctreeLoader` (also quadtree) creates tiles that initially have `TileEmptyContent`. Loaded content (including empty content) "sometimes" counts as a reference against the Tile. In this case, `addReference` was not being called when the tile was initially created with empty content, but `releaseReference` _was_ being called when the tile's content was unloaded. Oops!

This is more complicated than I'd like it to be. Loaded content counts as a `Tile` reference so that we don't destroy the tile before its content is unloaded. But we only want to do that for tile that _can_ be unloaded, right? And our rule is that we can only unload content if the tile has a valid ID, because without a valid ID, there's no possible way we can _reload_ the content if we unload it. Unloading content that we can't reload will break the tileset.

So, the overly-complicated rule is that content only counts as a reference against the `Tile` when the `Tile` also has an ID.

The `Tile` constructor has this code:

```cpp
if (this->hasReferencingContent()) {
  // Add a reference for the loaded content.
  this->addReference("Constructor with content");
}
```

Which looks like it should do the job.

Unfortunately, `hasReferencingContent` will _always_ return false, because at the time this code is invoke, the `Tile` can't possibly have an ID yet. So we're never adding a content reference.

In the vast majority of cases, this is fine. Most tiles with initial content will never get an ID anyway.

But in `ImplicitQuadtreeLoader` and `ImplicitOctreeLoader`, a `Tile` was constructed like this:

```cpp
if (subtreeAvailability.isContentAvailable(
                relativeChildLevel,
                relativeChildMortonID,
                0)) {
  children.emplace_back(&loader);
} else {
  children.emplace_back(&loader, TileEmptyContent{});
}

Tile& child = children.back();
// ...
child.setTileID(childID);
```

So, when an implicit tile doesn't have content available, we created it with `TileEmptyContent`, which will _not_ add a reference, as described above. And _then_ we give it an ID, which makes this `TileEmptyContent` unloadable, and when it's unloaded the reference count will be decremented. That's a `releaseReference` without an `addReference`, which eventually causes an assertion failure.

There are a number of possible solutions to this. The one I went with in this PR is not necessarily the best, but it's simple and low impact, and at least makes this sort of problem less likely to happen in the future.

I changed the `Tile` constructors that take content to _also_ require a `TileID`. With the `TileID` supplied at construction time, the constructor code will properly increment the reference count when it needs to do so.

It's still possible to subvert this system and run into reference counting issues by initially supplying a default `TileID` (empty string) and changing it later, but... don't do that. Really the `Tile` class could use some refactoring to better control its own internal state and make this sort of problem impossible. That's a much bigger change, though.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@
 
 - Renamed `SubtreeWriter::writeSubtree` to `SubtreeWriter::writeSubtreeJson`.
 - `SubtreeAvailability::createEmpty` now requires a boolean parameter to set initial tile availability.
-- `Cesium3DTilesSelection::Tile` constructors taking initial empty or external content now also require that a `TileID` be supplied.
+- `Cesium3DTilesSelection::Tile` constructors that take initially empty or external content now also require a `TileID` to be supplied.
 
 ##### Additions :tada:
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 
 - Renamed `SubtreeWriter::writeSubtree` to `SubtreeWriter::writeSubtreeJson`.
 - `SubtreeAvailability::createEmpty` now requires a boolean parameter to set initial tile availability.
+- `Cesium3DTilesSelection::Tile` constructors taking initial empty or external content now also require that a `TileID` be supplied.
 
 ##### Additions :tada:
 
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tile.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tile.h
@@ -136,19 +136,33 @@ public:
    * with this constructor.
    *
    * @param pLoader The {@link TilesetContentLoader} that is used to load the tile.
+   * @param tileID The ID of this tile. If not specified, the ID will initially
+   * be an empty string.
    */
-  explicit Tile(TilesetContentLoader* pLoader) noexcept;
+  explicit Tile(
+      TilesetContentLoader* pLoader,
+      const TileID& tileID = {}) noexcept;
 
   /**
    * @brief Construct a tile with an external content and a loader that is
    * associated with this tile. Tile has ContentLoaded status when initializing
    * with this constructor.
    *
+   * If the supplied `TileID` is not an empty string, the tile's reference count
+   * will be incremented on account of the the loaded, referencable content.
+   * Otherwise, if it is an empty string, the reference count will not be
+   * incremented. It is essential that this be accounted for later if the tile
+   * ID is changed in order to avoid reference count assertion failures at
+   * tileset destruction.
+   *
    * @param pLoader The {@link TilesetContentLoader} that is assiocated with this tile.
+   * @param tileID The ID of this tile. If it is an empty string, then the
+   * external content will not be unloable.
    * @param externalContent External content that is associated with this tile.
    */
   Tile(
       TilesetContentLoader* pLoader,
+      const TileID& tileID,
       std::unique_ptr<TileExternalContent>&& externalContent) noexcept;
 
   /**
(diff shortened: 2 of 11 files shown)


CHANGES.md
```diff
@@ -6,6 +6,7 @@
 
 - Renamed `SubtreeWriter::writeSubtree` to `SubtreeWriter::writeSubtreeJson`.
 - `SubtreeAvailability::createEmpty` now requires a boolean parameter to set initial tile availability.
+- `Cesium3DTilesSelection::Tile` constructors taking initial empty or external content now also require that a `TileID` be supplied.
```
j9liu, line 9: Nitpick:

```suggestion
- `Cesium3DTilesSelection::Tile` constructors that take initially empty or external content now also require a `TileID` to be supplied.
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tile.h
```diff
@@ -136,30 +136,56 @@ class CESIUM3DTILESSELECTION_API Tile final {
...
+      TilesetContentLoader* pLoader,
+      const TileID& tileID = {}) noexcept;
 
   /**
    * @brief Construct a tile with an external content and a loader that is
    * associated with this tile. Tile has ContentLoaded status when initializing
    * with this constructor.
    *
+   * If the supplied `TileID` is not an empty string, the tile's reference count
+   * will be incremented on account of the the loaded, referencable content.
```
j9liu, line 152:
```suggestion
   * will be incremented on account of the loaded, referencable content.
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tile.h
```diff
@@ -136,30 +136,56 @@ class CESIUM3DTILESSELECTION_API Tile final {
...
+      const TileID& tileID,
       std::unique_ptr<TileExternalContent>&& externalContent) noexcept;
 
   /**
    * @brief Construct a tile with an empty content and a loader that is
    * associated with this tile. Tile has ContentLoaded status when initializing
    * with this constructor.
    *
+   * If the supplied `TileID` is not an empty string, the tile's reference count
+   * will be incremented on account of the the loaded, referencable content.
```
j9liu, line 174:
```suggestion
   * will be incremented on account of the loaded, referencable content.
```

Cesium3DTilesSelection/test/TestTilesetContentManager.cpp
```diff
@@ -476,6 +476,7 @@ TEST_CASE("Test tile state machine") {
     pMockedLoader->mockCreateTileChildren = {{}, TileLoadResultState::Success};
```
j9liu, line 476: Is it possible to design a unit test to catch this case? (Maybe in the implicit loader tests?)

j9liu (commented):
Thanks so much @kring ! I confirmed that this fixes the error I was seeing for that tileset. I just had a few comments, mostly just pointing out some typos 😋

kring:
Thanks for the review @j9liu! I fixed the typos (apparently I was aiming for a typo in every comment! 😆) and added some tests, so this should ready for another look.

Cesium3DTilesSelection/test/TestImplicitOctreeLoader.cpp
```diff
@@ -708,4 +708,68 @@ TEST_CASE("Test tile subdivision for implicit octree loader") {
...
+        // of scope.
+        REQUIRE(child.getReferenceCount() == 1);
+        child.releaseReference();
+      } else {
+        REQUIRE(child.getReferenceCount() == 0);
+      }
+    }
+
+    // The seven tiles without content should have referencing EmptyContent.
+    // The fourth should not yet have any referencing content because it has not
```
j9liu, line 771:
```suggestion
    // The eighth should not yet have any referencing content because it has not
```

PR #1186, opened 2025-05-28
Add an optional glTF model modification stage after tile loading

GhisBntly:
* Rationale:
  - Because a glTF primitive translates to a `UStaticMesh` instance, shaded by a single material instance derived from the base material, customizing the shading is limited by the tileset provider's decisions regarding the grouping/merging of geometries into glTF primitives. When using per-vertex metadata to customize the shading, some faces of a single primitive, thus sharing the same material instance, may require incompatible material settings.
  - A processing stage at the end of the tile loading process allows to generate new primitives representing the same geometry that can be shaded as intended.
  - Another use case could be to optimize rendering of tilesets which geometry has been split into too many primitives, by merging them into more efficient larger primitives.
* Implementation:
  - The tuner is a simple interface which implementation is left to users of this extension, because supplying a generic implementation supporting the full glTF specification is a complex task and could be a performance bottleneck.
  - The tuner has a version to tag each processed model, to allow the processing strategy to evolve over time depending on application logic. Outdated models are thus automatically reprocessed.
  - The code supplied in this PR applies the tuning process in two situations:
    - On a newly loaded tile, before it can be considered for display, to avoid any visual glitch,
	- On already loaded tiles, after the tuner version has been incremented.
* Limitation:
  - For simplicity, the consistency of the tileset after the tuner version has been incremented is not enforced: models corresponding to different tuner versions may coexist for as long as it will take to reprocess all outdated models.


Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetExternals.h
```diff
@@ -20,6 +24,66 @@ class CreditSystem;
 namespace Cesium3DTilesSelection {
 class IPrepareRendererResources;
 
+/** Abstract class that allows tuning a glTF model.
+ * "Tuning" means reorganizing the primitives, eg. merging or splitting them.
```
j9liu: May I propose using "modify" instead of "tune" throughout this PR? The meaning of "modify" is a bit more obvious and aligned with the intent of this code (at least to me). So this would become something like `GltfModifier` instead.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetExternals.h
```diff
@@ -20,6 +24,66 @@ class CreditSystem;
...
+ * Splitting primitives allows to assign different materials to parts that were
+ * initially in the same primitive. Tuning is done in 2 phases: first phase in
+ * worker thread, then second phase in main thread. Tuning can occur several
+ * times during the lifetime of the model, depending on current needs. Hence
+ * the use of a "tune version" which allows to know if the mesh is up-to-date,
+ * or must be re-processed.
+ * A just constructed tuner is considered nilpotent, ie. tuning
+ * will not happen until retune() has been called at least once
+ */
+class GltfTuner {
```
j9liu: I would suggest putting this in a separate file, since it's a substantial enough class to warrant one.

CesiumGltf/include/CesiumGltf/Model.h
```diff
@@ -12,6 +12,10 @@ namespace CesiumGltf {
 
 /** @copydoc ModelSpec */
 struct CESIUMGLTF_API Model : public ModelSpec {
+
+  //! When produced by the glTF tuner, identifies the model version
+  int _tuningVersion = -1;
```
j9liu: I'm wary of coupling a detail in `CesiumGltf` to something that doesn't exist in the same scope... What about rewording this like so?

Also, it would help to make this optional to indicate a lack of "versioning" for clients that don't need it.

```suggestion
  // An optional version number for the model. This can be used by clients to indicate state about the model.
  std::optional<int> version = std::nullopt;
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tile.h
```diff
@@ -498,8 +498,11 @@ class CESIUM3DTILESSELECTION_API Tile final {
 
   /**
    * @brief Determines if this tile is currently renderable.
+   * @param minTuningVersionNeeded Minimum tuning version needed to allow this
+   * tile to render, or -1 to ignore the Gltf tuning version of the tile. See
+   * CesiumGltf::Model::_tuningVersion.
    */
-  bool isRenderable() const noexcept;
+  bool isRenderable(int minTuningVersionNeeded) const noexcept;
```
j9liu: Can we make this `optional` instead of using -1 as a special value? (same for the other modified functions in this file)

```suggestion
  bool isRenderable(std::optional<int> minTuningVersionNeeded = std::nullopt) const noexcept;
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileContent.h
```diff
@@ -193,9 +193,61 @@ class CESIUM3DTILESSELECTION_API TileRenderContent {
...
+  /** The state of the tuning process of the glTF model of a given tile content.
+   */
+  enum class TunerState {
+    /** No tuning is in progress. */
+    Idle,
+    /** Worker-thread phase is in progress. */
+    WorkerRunning,
+    /** Worker-thread phase is complete, main-thread phase is not done yet. */
+    WorkerDone,
+  };
```
j9liu: Move this to a separate file? (the same one that `GltfTuner` would be moved to)

Cesium3DTilesSelection/src/Tile.cpp
```diff
@@ -271,16 +275,37 @@ bool anyRasterOverlaysNeedLoading(const Tile& tile) noexcept {
...
+        renderContent->getTunerState() == TileRenderContent::TunerState::Idle) {
+      // Need to account for tuneModel's version too in case finishLoading
+      // hasn't yet been called
+      int latestVersion = renderContent->getTunedModel()._tuningVersion;
+      if (-1 == latestVersion)
+        latestVersion = renderContent->getModel()._tuningVersion;
+      if (latestVersion < tunerVersion)
+        return true;
+    }
+  }
```
j9liu: Having the `Tile` keep track of its assigned version number makes sense, but I think these checks are a little awkward -- or at least extraneous for implementations that won't ever use a `GltfTuner`.

What about an alternative approach where `TileState` has a new state called `NeedsReload` or `Dirty`, or something like that? Then `Tile` could have a method like `setDirty`, and the `needs____ThreadLoading` or `isRenderable` functions could account for that?

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileContent.h
```diff
@@ -193,9 +193,61 @@ class CESIUM3DTILESSELECTION_API TileRenderContent {
...
+
 private:
   CesiumGltf::Model _model;
   void* _pRenderResources;
+
+  /** Current state of the glTF tuning for this tile content */
+  TunerState _tunerState = TunerState::Idle;
+  /** Temporary model used during tuning when the tile content already has a
+   * model. */
+  CesiumGltf::Model _tunedModel;
```
j9liu: Make this optional, so that it's not required to store two models per tile with content.

```suggestion
  /** A modified copy of the tile's original model (if it had one).*/
  std::optional<CesiumGltf::Model> _modifiedModel;
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
```diff
@@ -635,7 +637,8 @@ class CESIUM3DTILESSELECTION_API Tileset final {
 
   static TraversalDetails createTraversalDetailsForSingleTile(
       const TilesetFrameState& frameState,
-      const Tile& tile);
+      const Tile& tile,
+      int minGltfTuningVersion);
```
j9liu: **Question** (and sorry if I missed this in the code): The `min` version implies that it's acceptable for tiles to be of mixed tuning versions. Is that intended and/or desired? Would there ever be a need for more granular control over this behavior, e.g., restricting tile versions to a single number instead of a range?

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetExternals.h
GhisBntly: Makes sense indeed, clearer semantics!

Cesium3DTilesSelection/src/Tile.cpp
GhisBntly: I had considered using a `TileLoadState::Dirty` state indeed, but it means that when incrementing the tuner/modifier's version, we also have to change the state of _all_ tiles at once, possibly synchronizing with other threads to avoid race conditions.... Other code testing the tile state have to consider this new state, too.
Thus the current solution looked safer and less intrusive.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
GhisBntly: Good point, it is indeed misleading, and I don't see any use case, especially since the glTF modifier version can only grow. I will just use `modelVersion` instead.

PR #1187, opened 2025-05-29
Enable access to `3DTILES_content_voxels` on a tileset's content

j9liu:


j9liu:
I'm putting this under a new changelog section so that we don't feel pressured to get it into the upcoming release. 🙂 I also recommend holding off on this until I open the PR that uses it in cesium-unreal.

j9liu:
Thanks for the suggestion @kring ! I like that a lot more, and I just pushed some commits that implement it.

> Aside: why? If the extension is only allowed on the root tile, it seems strange that it's part of the Tile.

Yeah that's a fair point. It feels like it should semantically belong in the `Content` level, since it's meant to describe the content. But our voxel implementation in CesiumJS (and soon Unreal) operates on a lot of assumptions at a higher level. It would be interesting if we moved it higher up, like on the `tileset` level itself.

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -548,17 +553,14 @@ std::optional<Tile> parseTileJsonRecursively(
...
-    if (uriIt != contentIt->value.MemberEnd() && uriIt->value.IsString()) {
-      contentUri = uriIt->value.GetString();
-    }
+  std::optional<Content> maybeContent;
+  if (hasContentMember) {
+    ContentReader contentReader;
+    auto contentResult = contentReader.readFromJson(contentIt->value);
+    maybeContent = std::move(contentResult.value);
   }
+  const char* contentUri = maybeContent ? maybeContent->uri.c_str() : nullptr;
```
kring, line 575: This code previously also allowed a `url` property instead of `uri`. It's non-standard, but I believe a pre-release version of 3D Tiles used that name and there's still content in the wild that uses it. We can support it here by looking for it in `unknownProperties`.

j9liu:
Thanks @kring ! How does my latest commit look?

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -548,16 +553,21 @@ std::optional<Tile> parseTileJsonRecursively(
...
-    if (uriIt != contentIt->value.MemberEnd() && uriIt->value.IsString()) {
-      contentUri = uriIt->value.GetString();
-    }
+  const char* contentUri = nullptr;
+  if (maybeContent && !maybeContent->uri.empty()) {
+    contentUri = maybeContent->uri.c_str();
+  } else if (maybeContent && maybeContent->unknownProperties.contains("url")) {
+    std::string url =
+        maybeContent->unknownProperties["url"].getStringOrDefault({});
+    contentUri = !url.empty() ? url.c_str() : nullptr;
```
kring: This will be a use-after-free error. The memory pointed to by `contentUri` will be freed when `url` is destroyed.

I'd suggest something like this:

```cpp
if (maybeContent && maybeContent->uri.empty()) {
  auto it = maybeContent->unknownProperties.find("url");
  if (it != maybeContent->unknownProperties.end()) {
    maybeContent->uri = it->getStringOrDefault({});
  }
}
```

j9liu: whoops 🤦 thanks for the catch!

PR #1188, opened 2025-05-29
Pass Tile SSEs to ViewUpdateResult

j9liu:
In this PR, `updateViewGroup` will return the SSE values for selected tiles in a parallel vector in `ViewUpdateResult`.

This was used to help prototype voxel rendering, where SSE is incorporated into a selection metric to influence what tiles are retained as the view is changed. However, I'm making this a draft because I'm not entirely happy with that approach.

### DIFF
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
@@ -532,10 +532,12 @@ private:
       const TilesetFrameState& frameState,
       Tile& tile,
       double tilePriority,
+      double tileSse,
       ViewUpdateResult& result);
   TraversalDetails _renderInnerTile(
       const TilesetFrameState& frameState,
       Tile& tile,
+      double tileSse,
       ViewUpdateResult& result);
   bool _kickDescendantsAndRenderTile(
       const TilesetFrameState& frameState,
@@ -545,7 +547,8 @@ private:
       size_t firstRenderedDescendantIndex,
       const TilesetViewGroup::LoadQueueCheckpoint& loadQueueBeforeChildren,
       bool queuedForLoad,
-      double tilePriority);
+      double tilePriority,
+      double tileSse);
   TileOcclusionState _checkOcclusion(const Tile& tile);
 
   TraversalDetails _visitTile(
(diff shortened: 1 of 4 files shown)


Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
timoore, line 601: Document `tileSse`.

Cesium3DTilesSelection/src/Tileset.cpp
timoore: Not your code, but if I understand correctly `frustums` and `distances` are always the same size. Seems like it's redundant to to check the loop against the size of distances too.

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -1076,11 +1083,13 @@ bool Tileset::_loadAndRenderAdditiveRefinedTile(
     Tile& tile,
     ViewUpdateResult& result,
     double tilePriority,
+    double tileSse,
     bool queuedForLoad) {
   // If this tile uses additive refinement, we need to render this tile in
   // addition to its children.
   if (tile.getRefine() == TileRefine::Add) {
     result.tilesToRenderThisFrame.emplace_back(&tile);
+    result.tileScreenSpaceErrorThisFrame.emplace_back(tileSse);
```
timoore: Maybe a `ViewUpdateResult::addTileToRender(tile, tileSse)` function is in called for? It would be unfortunate to forget to emplace the tileSse in the future...

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -1130,14 +1140,21 @@ bool Tileset::_kickDescendantsAndRenderTile(
...
       renderList.end());
+  sseList.erase(
+      sseList.begin() +
+          static_cast<std::vector<double>::iterator::difference_type>(
+              firstRenderedDescendantIndex),
+      sseList.end());
 
   if (tile.getRefine() != Cesium3DTilesSelection::TileRefine::Add) {
     renderList.emplace_back(&tile);
+    sseList.push_back(tileSse);
```
timoore: Any reason for `push_back` vs `emplace_back`?

j9liu: No, not sure what happened there :/ Fixing.

Cesium3DTilesSelection/src/Tileset.cpp
j9liu: Yeah that's not my code, but I can change this to use one `size()` and have an assert for them being equal.

PR #1191, opened 2025-06-03
Add UrlAssetAccessor based on libcurl

kring:
This code was originally written by @timoore for [vsgCs](https://github.com/timoore/vsgCs), and is arriving here by way of an internal project.

I debated where to put it. The obvious choice is CesiumAsync, but that adds a large new dependency (libcurl) to a very widely used library. I decided in the end of put it in its own library (CesiumCurl).


timoore:
Good stuff. 

Current vsgCs adds the X-Cesium-Client headers to each request, in the same way as Cesium Unreal's `UnrealAssetAccessor`. I don't know if that's appropriate for a general asset accessor like this PR, but this functionality is a bit awkward for a client to add otherwise. One could subclass `CurlAssetAccessor`and wrap `get()` and `request()` to add the client headers to each request, but with the current function signatures that means a lot of copying of headers.

Perhaps these extra header fields could be passed to the constructor?

CesiumCurl/include/CesiumCurl/CurlAssetAccessor.h
```diff
@@ -0,0 +1,93 @@
...
+#include <vector>
+
+namespace CesiumCurl {
+
+/**
+ * @brief An implementation of `IAssetAccessor` that can make network and local
+ * requests to a variety of servers using libcurl.
+ */
+class CESIUMCURL_API CurlAssetAccessor final
+    : public std::enable_shared_from_this<CurlAssetAccessor>,
```
timoore, line 88: Why is `enable_shared_from_this` needed?

CesiumCurl/src/CurlAssetAccessor.cpp
```diff
@@ -0,0 +1,528 @@
...
+      CurlAssetResponse::headerCallback);
+  curl_easy_setopt(curl, CURLOPT_HEADERDATA, this);
+}
+
+curl_slist* setCommonOptions(
+    CURL* curl,
+    const std::string& url,
+    const HttpHeaders& headers,
+    const std::string& userAgent,
+    const std::string& certificatePath,
```
timoore, line 319: I came to the conclusion that not finding the certificate file is a special case of broken OpenSSL configuration and take a different approach in vsgCs. I don't think the Curl options are general enough to fix the problems I've encountered. Here's my understanding of the situation:

* On Windows, Curl from vcpkg uses Secure Channel, not OpenSSL. I haven't personally seen any certificate issues with vcpkg builds on windows.
* On Linux, the key to smooth sailing with OpenSSL is to use the configuration file that goes with the library used. On Fedora, for example, there are options in the system-wide OpenSSL configuration files that will break vcpkg OpenSSL. vsgCs has options to either use the system OpenSSL (via an empty ports file) or to install the config files from the vcpkg port and use them. The code that does that is in https://github.com/timoore/vsgCs/blob/a5d30688fb4d5d1d686b1f6b2be05e60c95a227c/src/vsgCs/RuntimeEnvironment.cpp#L71. It's done more globally in vsgCs because there are other users of Curl (PROJ).

This is all to say that I think the path issues are OpenSSL issues and should be attacked that way. @weegeekps , any comments?

kring:
Good idea @timoore. I've added an option to add certain headers to all requests. I also moved all of the constructor parameters into an "options" class.

timoore:
> Good idea @timoore. I've added an option to add certain headers to all requests. I also moved all of the constructor parameters into an "options" class.

Cool. I think initializing OpenSSL can be kept totally separate in an application.

CesiumCurl/include/CesiumCurl/CurlAssetAccessor.h
```diff
@@ -0,0 +1,130 @@
...
+   * false if the initialization and cleanup are done elsewhere.
+   */
+  bool doGlobalInit{true};
+};
+
+/**
+ * @brief An implementation of `IAssetAccessor` that can make network and local
+ * requests to a variety of servers using libcurl.
+ */
+class CESIUMCURL_API CurlAssetAccessor final
```
timoore: Why `final`? I would say that we are putting this out there for users to customize.

CesiumCurl/include/CesiumCurl/CurlAssetAccessor.h
timoore: > Why is `enable_shared_from_this` needed?

I suppose the fact that the externals interface takes a shared pointer to `IAssetAccessor` is reason enough.

kring: It's used in `get` and `request` to ensure that the asset accessor stays alive while requests are in flight. The next best alternative I can think of would add a lot of complexity. We'd have to move all of the asset accessor's state into another nested object so its lifetime can extend past the asset accessor's lifetime.

CesiumCurl/src/CurlAssetAccessor.cpp
kring: > On Linux, the key to smooth sailing with OpenSSL is to use the configuration file that goes with the library used.

This makes me nervous. If I understand correctly, it means that the set of certificate authorities will be baked into the application at compile time, and can't (easily?) be updated.

Using the operating system's certificate configuration seems like a much better idea. I'd be interested to learn what about Fedora's OpenSSL configuration breaks the vcpkg OpenSSL. Using the system OpenSSL makes sense where it's an option.

In any case, I don't think we're making any strong commitments here in this PR, just giving users options.

CesiumCurl/include/CesiumCurl/CurlAssetAccessor.h
kring: You'll see this on almost all cesium-native classes. I agree it's not great here, though. Users could customize by aggregation as a workaround, but there's no particular reason to force them to do that. I'll remove it.

CesiumCurl/include/CesiumCurl/CurlAssetAccessor.h
timoore: OK, I see now. The original `UrlAssetAccessor` doesn't do this, relying unsafely on the acessor being alive for the duration of the application.

CesiumCurl/src/CurlAssetAccessor.cpp
timoore: > Using the operating system's certificate configuration seems like a much better idea. I'd be interested to learn what about Fedora's OpenSSL configuration breaks the vcpkg OpenSSL. Using the system OpenSSL makes sense where it's an option.

It's the option `rh-allow-sha1-signatures`. Not anything to do with certificates, but a random option that causes vcpkg to give up when it doesn't recognize it in opensslcnf.config... or at least what I could tell from debugging.

> In any case, I don't think we're making any strong commitments here in this PR, just giving users options.

PR #1192, opened 2025-06-04
Fix crash when unloading ellipsoid tileset with raster overlays

azrogers:
Closes CesiumGS/cesium-unreal#1682. As documented in that issue, when creating a tileset using the `EllipsoidTilesetLoader`, if the tileset has raster overlays, it will crash when unloading. This turns out to be a result of simply forgetting to pass the `LoadedTileEnumerator` to the `RasterOverlayCollection` when using the `TilesetContentManager` constructor that is currently only used by the `EllipsoidTilesetLoader`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@
 
 - Added `CesiumVectorData` library for loading data from vector formats. Currently only GeoJSON is supported.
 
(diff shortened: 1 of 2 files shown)


kring:
Thanks @azrogers!

PR #1194, opened 2025-06-11
Add `GeoJsonDocument::fromUrl` method

azrogers:
As pointed out in CesiumGS/cesium-unreal#1651, it would be useful to have a method to load GeoJSON data from an arbitrary URL, to go along with our methods to load data from Cesium ion. This PR adds that method.

### DIFF
--- a/CesiumVectorData/src/GeoJsonDocument.cpp
+++ b/CesiumVectorData/src/GeoJsonDocument.cpp
@@ -725,8 +725,11 @@ Future<Result<GeoJsonDocument>> GeoJsonDocument::fromUrl(
           -> Result<GeoJsonDocument> {
             const IAssetResponse* pAssetResponse = pAssetRequest->response();
 
-            if (pAssetResponse->statusCode() < 200 ||
-                pAssetResponse->statusCode() >= 300) {
+            // Curl returns a status code of zero with a file URL, so we want to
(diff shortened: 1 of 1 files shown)


CesiumVectorData/src/GeoJsonDocument.cpp
```diff
@@ -714,6 +714,28 @@ Future<Result<GeoJsonDocument>> GeoJsonDocument::fromCesiumIonAsset(
...
+    const std::shared_ptr<IAssetAccessor>& pAssetAccessor,
+    const std::string& url,
+    const std::vector<CesiumAsync::IAssetAccessor::THeader>& headers) {
+  return pAssetAccessor->get(asyncSystem, url, headers)
+      .thenImmediately(
+          [](std::shared_ptr<IAssetRequest>&& pAssetRequest) mutable
+          -> Result<GeoJsonDocument> {
+            const IAssetResponse* pAssetResponse = pAssetRequest->response();
+
+            if (pAssetResponse->statusCode() < 200 ||
```
kring: We usually allow a status code of zero, too, because libcurl uses code zero on successful load from a file URL, for some reason. You'll see that scattered all over the place.

azrogers:
@kring Should be addressed now!

PR #1195, opened 2025-06-11
Update projects list in the README

azrogers:
With the release of 3D Tiles for Godot - as well as [a user on the forums](https://community.cesium.com/t/cesium-native-dart-package/41196) sharing a Dart package they've created wrapping Cesium Native - I think it's about time we update the list of community projects in the README!


kring:
Thanks @azrogers!

Issue #1196, opened 2025-06-12
Dodgy `Authorization: Bearer ` header included in Google P3DT request when loaded from Cesium ion


kring:
When loading Google Photorealistic 3D Tiles from Cesium ion, the request for the root tile (`root.json`) includes an HTTP header that looks like this:

```
Authorization: Bearer
```

It's added here:
https://github.com/CesiumGS/cesium-native/blob/v0.48.0/Cesium3DTilesSelection/src/CesiumIonTilesetLoader.cpp#L215

Most Cesium ion assets have this `accessToken` field, but Google Photorealistic 3D Tiles does not, so we should simply skip adding this header in this case.

It doesn't seem to do any harm, but it's dodgy to add a blank `Authorization` header like this, and we shouldn't do it. Other servers might not be so forgiving.

PR #1198, opened 2025-06-12
Fixed sending empty authorization header `Authorization: Bearer` when no access token is provided.

animal89042:
Fixes: #1196
Fixed sending empty authorization header `Authorization: Bearer` when no access token is provided while using `CesiumIonTilesetLoader`. Prevents potential future issues with some servers including GP3D Tiles.


azrogers:
Thank you for the fix @animal89042!

PR #1199, opened 2025-06-13
Fix reusing the If-None-Match header for future requests

### LINKED ISSUE #1190: Caching issue with Google 3D Photorealistic and status code 304
When using Cesium Native to load Google 3D Photorealistic tiles directly, we observe repeated HTTP 304 (Not Modified) responses even though we are using SQLiteCache. In theory, the cache should prevent redundant requests, but it appears that our tile requests continue to miss the cache.

Upon inspection of the SQLiteCache database, we discovered that the full request URL, including both the API key and the session parameter, is stored as the cache key. While the API key remains constant, the session value changes over time. As soon as the session parameter differs from what was originally cached, the cache lookup fails and Cesium Native re-issues the request, resulting in multiple 304 responses.

Would it be possible to adjust the caching logic so that URLs are normalized by removing or ignoring the session query parameter? This way, the cache would match on the remaining stable components of the URL (e.g. dataset, file path) and avoid unnecessary round trips when only the session token has changed.

We are using Cesium Native v0.48.0

germanmarques:
Fixes #1190 

It avoids storing the If-None-Match header in the cache and in the TilesetContentManager. This avoids Cesium Native to reuse this header to download tiles that are not in the cache.

### DIFF
--- a/Cesium3DTilesSelection/src/TilesetContentManager.cpp
+++ b/Cesium3DTilesSelection/src/TilesetContentManager.cpp
@@ -2034,6 +2034,15 @@ void TilesetContentManager::propagateTilesetContentLoaderResult(
           creditResult.showOnScreen));
     }
 
+    {
+      // Remove the If-None-Match header if exists to avoid adding it to other requests
+      CesiumAsync::CaseInsensitiveCompare cmp;
+      std::erase_if(result.requestHeaders,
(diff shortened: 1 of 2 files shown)


kring:
Thanks for diagnosing this issue and opening a PR with a fix @germanmarques! We'll take a look soon.

kring:
Closing in favor of #1206, which builds on this. Thanks @germanmarques!

PR #1203, opened 2025-06-19
Fixes ezvcpkg silently skipping outdated ports

GhisBntly:
`vcpkg install` does not upgrade already installed ports, it silently skips them.

Normally, the command to run is `vcpkg upgrade --no-dry-run`, but for some reason it removes the port's folder in the
packages/ directory, and cesium-native's CMakeLists.txt rely on it to install headers and libs!

So in this PR I am suggesting `vcpkg remove --outdated --recurse` instead, which does the job.

Note that relying on the content of the `packages/` folder is not such a good idea, because when cleaning the (ez)vcpkg folder and letting vcpkg re-install using "binarysource" (archives cache), only the `installed/` folder is populated, not the `packages/`, thus breaking the cesium-native build.
The header and library file list to install should rather be extracted from the `installed/vcpkg/info/<port>_<version>_<triplet>.list` files instead.


kring:
Thanks for the PR @GhisBntly!

I love what you've done here! It's annoying to have to manually remove outdated packages, which is what I've mostly done in the past. This PR nicely avoids the need to do that sort of thing.

But unfortunately, it also currently breaks an important development workflow that we use every day, and I don't know how to fix it.

The problem is that `vcpkg remove --outdated --triplet x64-windows-unreal` does not (at least on my system?!) make vcpkg limit itself to remove outdated packages associated with the `x64-windows-unreal` triplet. It removes _all_ outdated packages across all triplets.

This is a problem because we commonly use one vcpkg "classic mode" directory (e.g. `.ezvcpkg/dbe35ceb30c688bf72e952ab23778e009a578f18`) across multiple products (e.g. cesium-native, cesium-unreal, cesium-unity) where each product has a different triplet and a different set of overlay ports. Because of the different overlay ports, `vcpkg remove --outdated` considers cesium-unreal's packages outdated when running cesium-native's build, and vice-versa.

So if I run the cmake configure for cesium-native, it invokes ezvcpkg, which builds all the third-party packages for the `x64-windows` triplet. So far so good. Then, if I run the cmake configure in the cesium-unreal `extern` directory, I get this:

```
[cmake] -- EZVCPKG Removing outdated packages for triplet x64-windows-unreal
[cmake] The following packages will be removed:
[cmake]     abseil:x64-windows
[cmake]     ada-url:x64-windows
[cmake]     openssl:x64-windows
[cmake]   * s2geometry:x64-windows
[cmake]     sqlite3:x64-windows
[cmake] warning: Additional packages (*) need to be removed to complete this operation.
[cmake] Removing 1/5 s2geometry:x64-windows
[cmake] Removing 2/5 abseil:x64-windows
[cmake] Removing 3/5 ada-url:x64-windows
[cmake] Removing 4/5 openssl:x64-windows
[cmake] Removing 5/5 sqlite3:x64-windows
```

Yikes, it just deleted packages cesium-native needs. So now when I configure cesium-native again, it will have to rebuild those packages that were just removed. And worse, running the cesium-native build will delete the new ones that the cesium-unreal build just built for the `x64-windows-unreal` triplet.

I (and presumably others on the Native Runtimes team, at least) frequently switch between cesium-native and cesium-unreal builds many times per day. So rebuilding a bunch of packages each time we do is going to add a lot of friction.

We're actively moving away from vcpkg classic mode and toward manifest mode, so a lot of this is going to change and hopefully improve. But in the meantime, I don't think it makes sense to merge this PR in its current state.

kring:
I'm going to close this. We'll revisit the issue once we make the jump to manifest mode.

PR #1204, opened 2025-06-21
Include query parameters when loading external subtree buffers

kring:
Even though it's a somewhat unusual thing to do, we always include query parameters from the base URL when resolving a relative URL. This is essential when those query parameters include authorization information, but is pretty common.

This PR fixes a bug where we weren't including those query parameters when loading external subtree buffers in implicit tiling.


j9liu:
Thanks @kring !

Issue #1201, opened 2025-06-17
i3dm BATCH_ID parsing does not use default component type


timoore:
If the BATCH_ID in an i3dm feature table does not provide a component type, then the BATCH_ID is assumed to be invalid, despite uint16 being the specified default. This was reported on the [Community Forums](https://community.cesium.com/t/problem-with-instanced-tilesets/41369). The tileset mentioned there, https://www.3drotterdam.nl/datasource-data/b9294d5b-79e0-4744-ac88-b61e3e90d5bc/tileset.json illustrates the problem: a spew of "Invalid Batch ID Semantic" messages are produced when displaying the tileset with a native client.

PR #1205, opened 2025-06-25
Fix problem parsing i3dm BATCH_ID

timoore:
The code for initializing BatchIdSemantic would fail when no component type was specified for the BATCH_ID semantic, but that is incorrect; uint16 is the default.

Addresses #1201


j9liu:
@timoore Could you update CHANGES and add a unit test before I take a deeper look?

Cesium3DTilesContent/src/BatchTableToGltfStructuralMetadata.cpp
```diff
@@ -2073,38 +2073,45 @@ struct BatchIdSemantic {
       return;
     }
     uint32_t byteOffset = byteOffsetIt->value.GetUint();
+    std::optional<MetadataProperty::ComponentType> componentType;
```
j9liu: It doesn't look like `componentType` would ever be nullopt. if `componentTypeIt` doesn't exist, then it defaults. If it exists, but is invalid, then the function returns. It'd be cleaner if this wasn't an optional, and instead started off with the default value. Then, it can be overridden as necessary.

```suggestion
    MetadataProperty::ComponentType componentType = MetadataProperty::ComponentType::UNSIGNED_SHORT;
```

timoore: @j9liu I went back and forth on this myself; while `UNSIGNED_SHORT` is the default, there can still be an error if e.g. an invalid component type is specified. It's true that the code doesn't do anything with an empty optional value, so I think I agree that it would be simpler to either not use `std::optional` or else not initialize the value at all.

timoore: I ended up changing `componentType` to not be optional.

timoore:
> @timoore Could you update CHANGES and add a unit test before I take a deeper look?

I've done that now. Good thing you insisted on the unit test; it revealed a more serious bug!

Cesium3DTilesContent/src/BatchTableToGltfStructuralMetadata.cpp
```diff
@@ -2073,38 +2071,45 @@ struct BatchIdSemantic {
       return;
     }
     uint32_t byteOffset = byteOffsetIt->value.GetUint();
+    MetadataProperty::ComponentType componentType;
     const auto componentTypeIt = batchIdIt->value.FindMember("componentType");
-    if (componentTypeIt != batchIdIt->value.MemberEnd() &&
-        componentTypeIt->value.IsString()) {
+    if (componentTypeIt == batchIdIt->value.MemberEnd()) {
+      componentType = MetadataProperty::ComponentType::UNSIGNED_SHORT;
+    } else {
```
j9liu: **NITPICK**: I think this could be condensed further. You would still return early if the `componentType` property is incorrectly formed:

```c++
    MetadataProperty::ComponentType componentType = MetadataProperty::ComponentType::UNSIGNED_SHORT;

    const auto componentTypeIt = batchIdIt->value.FindMember("componentType");
    if (componentTypeIt != batchIdIt->value.MemberEnd()) {

File: CHANGES.md [hunk 3]
```diff
@@ -24,6 +24,8 @@
 - Fixed sending empty authorization header `Authorization: Bearer` when no access token is provided while using `CesiumIonTilesetLoader`. Prevents potential future issues with some servers including GP3D Tiles.
 - Fixed a bug where `CachingAssetAccessor` would include "revalidation" headers like `If-None-Match` in the returned `IAssetRequest` when the remote server returned new content rather than a 304 response. This could cause the header to be incorrectly included in later requests for different content.
 - Fixed a bug in `SubtreeFileReader` where it did not include query parameters from the base URL when requesting an external subtree buffer.
+- Fixed a bug in the parsing of the i3dm BATCH_ID semantic.
+- Fixed a bug in the conversion of i3dm batch ids to EXT_instance_features feature ids.
```
j9liu: Minor nitpicks:

```suggestion
- Fixed a bug in the parsing of the i3dm `BATCH_ID` semantic.
- Fixed a bug in the conversion of i3dm batch ids to `EXT_instance_features` feature ids.
```

j9liu (commented):
Thanks for the changes @timoore ! I have two minor suggestions but it's nothing show-stopping. I'll also sanity-check this in Unreal just to verify that instanced data still loads fine

PR #1206, opened 2025-06-27
Don't include cache revalidation request headers in CachingAssetAccessor's IAssetRequest.

kring:
This builds on #1199, but I'm not allowed to commit to that PR's branch.

@germanmarques I've made some further changes based on your PR. Mostly I wanted to avoid `TilesetContentManager` needing to know about cache headers. I also modified some tests to demonstrate that this is working. Can you please take a look too and confirm that it still solves the problem for you? (I expect that it should)


azrogers (approved):
Looks good to me!

germanmarques:
Thanks @kring! Looks good to me. It's the way I wanted to approach it, but as an outsider it seemed excessive to add just another class for this edge case. I didn't know if you already had wrappers to handle this. But I get your point of not wanting the `TilesetContentManager` to know about specific headers. 

Again, thanks a lot for taking the time to review it.

PR #1207, opened 2025-06-30
Fix build on iOS prior to v13.

kring:
iOS prior to version 13 is missing `std::filesystem` capabilities used by `CurlAssetAccessor`. It's only used to create directories in PUTs to file URLs, and off by default anyway. So this PR ifdefs out the feature out on older iOS versions in order to make it compile.


j9liu:
Thanks @kring ! I don't really have a way to test this, but I trust you've tried this yourself 😛

PR #1210, opened 2025-07-01
Exclude unused functions which triggered a build error on XCode 16, clang 17

david-lively:
When building on MacOS 15.3.1, XCode 16.4 clang 17.0, two warnings are generated (and treated as errors) due to unused functions in `CesiumCurl/src/CurlAssetAccessor.cpp`.  
This PR adds an `#if` block around those functions to exclude them on iOS devices with version < 13.


timoore:
Looks good to me, though I don't have a local way to test it.

PR #1218, opened 2025-07-28
Add Cesium ion token refresh for raster overlays

kring:
Fixes CesiumGS/cesium-unreal#1698

This turned out to be a whole lot more work than I expected! It's working, but it's draft because there are some things left to do:

* [x] Support for refreshing TMS overlays. Currently only Bing overlays are refreshed.
* [x] Check lifetime of the CesiumIonAssetAccessor created by IonRasterOverlay. CesiumIonTilesetLoader calls `notifyOwnerIsBeingDestroyed`, but `IonRasterOverlay` does not, which is (probably) a problem.
* [x] clang-tidy warnings. In particular, there's one suggesting use of dynamic_cast instead of static_cast, which would be reasonable if it weren't against our coding guidelines (for good reason): https://cesium.com/learn/cesium-native/ref-doc/style-guide.html#%EF%B8%8F-run-time-type-information


kring:
@azrogers I'm running into a problem in this branch where clang-tidy is complaining that `<CesiumUtility/Log.h>` is unnecessary, but it's wrong. It contains a formatter for `rapidjson::ParseErrorCode`, and the code won't compile without it. I can just disable the warning, but do you have any other thoughts here?

CesiumRasterOverlays/include/CesiumRasterOverlays/BingMapsRasterOverlay.h
```diff
@@ -116,6 +116,20 @@ class CESIUMRASTEROVERLAYS_API BingMapsRasterOverlay final
...
+  /**
+   * @brief Refresh a previously-created tile provider using a new key.
+   *
+   * Calling this method on a tile provider that was not created by this \ref
+   * BingMapsRasterOverlay will lead to undefined behavior.
+   *
+   * @param pProvider The previously-created tile provider.
+   * @param newKey The new key to use.
+   */
+  CesiumAsync::Future<void> refreshTileProviderWithNewKey(
```
j9liu, line 128: General question I guess, is there a reason why a raster overlay doesn't "own" or even keep track of its provider, which would remove the need for this assumption/condition?

CesiumRasterOverlays/src/IonRasterOverlay.cpp
```diff
@@ -100,13 +105,86 @@ IonRasterOverlay::createTileProvider(
...
-      pAssetAccessor,
-      pCreditSystem,
-      pPrepareRendererResources,
-      pLogger,
-      std::move(pOwner));
+  struct ProviderHolder {
+    RasterOverlayTileProvider* pProvider = nullptr;
+  };
+
+  std::shared_ptr<ProviderHolder> pHolder = std::make_shared<ProviderHolder>();
```
j9liu, line 117: I'm probably missing something obvious, but why is this necessary?

CesiumAsync/src/CesiumIonAssetAccessor.cpp
```diff
@@ -0,0 +1,302 @@
...
+                  this->_pLogger,
+                  "Request failed while attempting to refresh the Cesium "
+                  "ion token.");
+              return asyncSystem.createResolvedFuture(UpdatedToken());
+            }
+
+            uint16_t statusCode = pIonResponse->statusCode();
+            if (statusCode >= 200 && statusCode < 300) {
+              auto accessToken =
+                  getNewAccessToken(pIonResponse, this->_pLogger);
```
j9liu: **NITPICK**: In accordance with style guide, avoid `auto`? (Or at least rename to `maybeAccessToken`?)

```suggestion
              std::optional<std::string> accessToken =
                  getNewAccessToken(pIonResponse, this->_pLogger);
```

CesiumAsync/src/CesiumIonAssetAccessor.cpp
```diff
@@ -0,0 +1,302 @@
...
+              if (accessToken) {
+                std::string authorizationHeader = "Bearer " + *accessToken;
+                UpdatedToken update{*accessToken, authorizationHeader};
+                return this->_updatedTokenCallback.value()(update)
```
j9liu: **NITPICK**: Our style guide says to prefer `*` over `value`, but it's not the end of the world...

```suggestion
                return (*this->_updatedTokenCallback)(update)
```

CesiumAsync/src/CesiumIonAssetAccessor.cpp
```diff
@@ -0,0 +1,302 @@
...
+                                      pLogger = this->_pLogger,
+                                      url = this->_assetEndpointUrl]() {
+                      SPDLOG_LOGGER_INFO(
+                          pLogger,
+                          "Successfuly refreshed Cesium ion token for url {}.",
+                          url);
+
+                      return update;
+                    });
+
```
j9liu: **NITPICK:** remove extra space?

j9liu (commented):
Thanks @kring ! I have a couple of questions / comments but nothing major. I'm also testing this in Unreal on the side just to sanity check it works :)

CesiumRasterOverlays/include/CesiumRasterOverlays/BingMapsRasterOverlay.h
kring: I guess the idea is that a `RasterOverlay` is kind of like a factory for a tile provider, but doesn't own it. A single overlay could be used on multiple tilesets, for example.

The real reason for the "undefined behavior" is that we've decided that cesium-native must work without RTTI, so `dynamic_cast` is not available. Which means we can't downcast safely.

CesiumRasterOverlays/src/IonRasterOverlay.cpp
kring: To solve a chicken-and-egg problem. The `CesiumIonAssetAccessor` must be given to `createTileProvider`, so of course the `CesiumIonAssetAccessor` needs to be created first. But the "new token" callback needs to modify the created tile provider, which can't be known when that callback needs to be created. So we give the callback a "holder" created ahead of time, and fill it with the provider once it's created.

CesiumAsync/src/CesiumIonAssetAccessor.cpp
kring: This code isn't new, it just moved here. But I'll bring it up to the standard.

CesiumRasterOverlays/include/CesiumRasterOverlays/BingMapsRasterOverlay.h
j9liu: I see -- not a matter of ownership, but literally because each raster overlay has its own private tile provider and we `static_cast` in the function.

I'm curious why the tile providers are privated then, since it seems like it sort of matters, though maybe it didn't matter up until now? But it's not worth holding up this PR over it.

kring: Making the tile providers public wouldn't help, because the larger system still holds the instance as a `RasterOverlayTileProvider*`. There has to be a cast somewhere.

kring: Well.... the alternative to casting is a new virtual function that classes derived from `RasterOverlayTileProvider` are required to implement. But I don't think that would work well here.

j9liu: Meh, sounds like a refactor wouldn't even help then. Don't mind me, I'm just throwing ideas at a system I don't fully understand 😁

PR #1221, opened 2025-08-08
Only provide `const` access to the root tile

kring:
The `getRootTile` method on `Tileset` returns a pointer to the root tile, once there is one. Previously, the returned pointer would be to const if called on a const Tileset, and to non-const otherwise. After this PR, the returned pointer is _always_ to const.

This is an important improvement because some modifications to `Tile` and the data it holds are unsafe, even when done from the main thread. For example, upsampling for raster overlays happens in a worker thread, which is given a pointer to the parent tile's glTF. The `Tileset` makes sure this is safe, but if users are allowed to make arbitrary modifications to `Tile` while this is happening, then there is potential for undefined behavior. And #1186 introduces another place where a tile's model is read in a background thread.

Fortunately, almost nothing that I'm aware of is using the non-const version of `getRootTile`. There's only one place in cesium-unreal, and it's in a test, and is solved by simply adding a missing `const`. So this is just tightening things up hopefully without a large impact on users.

Other changes in this PR:

* `slowlyGetCurrentStates` and `slowlyGetPreviousStates` on `TreeTraversalState` now use a const raw pointer as the key in the map. Because if you only have access to `const Tile` instances, it's a hassle to provide an `IntrusivePointer<Tile>` just for map lookup.
* `IntrusivePointer<T>` now has an `element_type` typedef, making it play nicely with `std::pointer_traits`. This was motivated by the change above.

### DIFF
--- a/Cesium3DTilesSelection/src/Tile.cpp
+++ b/Cesium3DTilesSelection/src/Tile.cpp
@@ -362,7 +362,7 @@ bool isContentReferenced(const Tile& tile) {
 
 } // namespace
 
-void Tile::addReference([[maybe_unused]] const char* reason) noexcept {
+void Tile::addReference([[maybe_unused]] const char* reason) const noexcept {
   ++this->_referenceCount;
 
 #ifdef CESIUM_DEBUG_TILE_UNLOADING
@@ -396,11 +396,13 @@ void Tile::addReference([[maybe_unused]] const char* reason) noexcept {
   // "ineligible for content unloading" with the TilesetContentManager.
   if (isContentReferenced(*this) && this->_pLoader &&
       this->_pLoader->getOwner()) {
-    this->_pLoader->getOwner()->markTileIneligibleForContentUnloading(*this);
+    this->_pLoader->getOwner()->markTileIneligibleForContentUnloading(
+        const_cast<Tile&>(*this));
   }
 }
 
-void Tile::releaseReference([[maybe_unused]] const char* reason) noexcept {
+void Tile::releaseReference(
+    [[maybe_unused]] const char* reason) const noexcept {
   CESIUM_ASSERT(this->_referenceCount > 0);
   --this->_referenceCount;
 
@@ -424,7 +426,8 @@ void Tile::releaseReference([[maybe_unused]] const char* reason) noexcept {
   // content will be unloaded immediately.
   if (!isContentReferenced(*this) && this->_pLoader &&
       this->_pLoader->getOwner()) {
-    this->_pLoader->getOwner()->markTileEligibleForContentUnloading(*this);
+    this->_pLoader->getOwner()->markTileEligibleForContentUnloading(
+        const_cast<Tile&>(*this));
   }
 
   // When the reference count goes from 1 to 0, this Tile is once again
--- a/Cesium3DTilesSelection/src/Tileset.cpp
+++ b/Cesium3DTilesSelection/src/Tileset.cpp
@@ -170,10 +170,6 @@ void Tileset::setShowCreditsOnScreen(bool showCreditsOnScreen) noexcept {
   }
 }
 
-Tile* Tileset::getRootTile() noexcept {
-  return this->_pTilesetContentManager->getRootTile();
-}
-
 const Tile* Tileset::getRootTile() const noexcept {
   return this->_pTilesetContentManager->getRootTile();
 }
@@ -392,7 +388,7 @@ const ViewUpdateResult& Tileset::updateViewGroup(
 
   ViewUpdateResult& result = viewGroup.getViewUpdateResult();
 
-  Tile* pRootTile = this->getRootTile();
+  Tile* pRootTile = this->_pTilesetContentManager->getRootTile();
   if (!pRootTile) {
     return result;
   }
@@ -438,7 +434,7 @@ void Tileset::loadTiles() {
 
   this->_asyncSystem.dispatchMainThreadTasks();
 
-  Tile* pRootTile = this->getRootTile();
+  Tile* pRootTile = this->_pTilesetContentManager->getRootTile();
   if (!pRootTile) {
     // If the root tile is marked as ready, but doesn't actually exist, then
     // the tileset couldn't load. Fail any outstanding height requests.
--- a/Cesium3DTilesSelection/src/Tileset.cpp
+++ b/Cesium3DTilesSelection/src/Tileset.cpp
@@ -491,20 +491,6 @@ LoadedConstTileEnumerator Tileset::loadedTiles() const {
           : nullptr);
 }
 
-LoadedTileEnumerator Tileset::loadedTiles() {
-  return LoadedTileEnumerator(
-      this->_pTilesetContentManager
-          ? this->_pTilesetContentManager->getRootTile()
-          : nullptr);
-}
-
-void Tileset::forEachLoadedTile(
-    const std::function<void(Tile& tile)>& callback) {
-  for (Tile& tile : this->loadedTiles()) {
-    callback(tile);
-  }
-}
-
 void Tileset::forEachLoadedTile(
     const std::function<void(const Tile& tile)>& callback) const {
   for (const Tile& tile : this->loadedTiles()) {
--- a/Cesium3DTilesSelection/src/Tileset.cpp
+++ b/Cesium3DTilesSelection/src/Tileset.cpp
@@ -259,8 +259,8 @@ void Tileset::_updateLodTransitions(
     // Update fade out
     for (auto tileIt = result.tilesFadingOut.begin();
          tileIt != result.tilesFadingOut.end();) {
-      TileRenderContent* pRenderContent =
-          (*tileIt)->getContent().getRenderContent();
+      TileRenderContent* pRenderContent = const_cast<TileRenderContent*>(
+          (*tileIt)->getContent().getRenderContent());
 
       if (!pRenderContent) {
         // This tile is done fading out and was immediately kicked from the
@@ -287,9 +287,9 @@ void Tileset::_updateLodTransitions(
     }
 
     // Update fade in
-    for (const Tile::Pointer& pTile : result.tilesToRenderThisFrame) {
-      TileRenderContent* pRenderContent =
-          pTile->getContent().getRenderContent();
+    for (const Tile::ConstPointer& pTile : result.tilesToRenderThisFrame) {
+      TileRenderContent* pRenderContent = const_cast<TileRenderContent*>(
+          pTile->getContent().getRenderContent());
       if (pRenderContent) {
         float transitionPercentage =
             pRenderContent->getLodTransitionFadePercentage();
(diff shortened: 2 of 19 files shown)


Cesium3DTilesSelection/test/TestTilesetSelectionAlgorithm.cpp
```diff
@@ -765,10 +767,10 @@ TEST_CASE("Render any tiles even when one of children can't be rendered for "
   initializeTileset(tileset);
   ViewState viewState = zoomToTileset(tileset);
 
-  Tile* pTilesetJson = tileset.getRootTile();
+  const Tile* pTilesetJson = tileset.getRootTile();
   REQUIRE(pTilesetJson);
   REQUIRE(pTilesetJson->getChildren().size() == 1);
-  Tile* root = &pTilesetJson->getChildren()[0];
+  const Tile* root = &pTilesetJson->getChildren()[0];
```
j9liu, line 780:
```suggestion
  const Tile* pRoot = &pTilesetJson->getChildren()[0];
```

Cesium3DTilesSelection/src/Tile.cpp
```diff
@@ -396,11 +396,13 @@ void Tile::addReference([[maybe_unused]] const char* reason) noexcept {
   // "ineligible for content unloading" with the TilesetContentManager.
   if (isContentReferenced(*this) && this->_pLoader &&
       this->_pLoader->getOwner()) {
-    this->_pLoader->getOwner()->markTileIneligibleForContentUnloading(*this);
+    this->_pLoader->getOwner()->markTileIneligibleForContentUnloading(
+        const_cast<Tile&>(*this));
```
j9liu, line 400: Maybe instead of `const_cast`ing, `markTileIneligibleForContentUnloading` should take a `const Tile&`?

j9liu: Though, now that I trace it to the `DoublyLinkedListAdvanced::remove` call I see that there's a `node.*Pointers` property that gets modified from the `Tile`. Could/should that be made `mutable`? Especially since it seems like a more "external" property of the `Tile` itself, like the reference counting, unless I'm misunderstanding the intentions of the list.

This is probably nitpicking at this point but IMO I'd want to feel secure about justifying the `const_cast`s -- we're otherwise violating the intentions of the API.

Cesium3DTilesSelection/src/Tile.cpp
```diff
@@ -424,7 +426,8 @@ void Tile::releaseReference([[maybe_unused]] const char* reason) noexcept {
   // content will be unloaded immediately.
   if (!isContentReferenced(*this) && this->_pLoader &&
       this->_pLoader->getOwner()) {
-    this->_pLoader->getOwner()->markTileEligibleForContentUnloading(*this);
+    this->_pLoader->getOwner()->markTileEligibleForContentUnloading(
```
j9liu, line 429: Looking at `markTileEligible...`, indeed the `Tile` can be unloaded by this function so we need that `const_cast`. But then the `const` in `releaseReference` is basically a lie.

Is it hard to work with this if we remove the `const`? Maybe there's another place where it's more reasonable to do a `const_cast`?

Cesium3DTilesSelection/test/TestTileLoadRequester.cpp
```diff
@@ -23,45 +23,51 @@ class TestTileLoadRequester : public TileLoadRequester {
...
   }
 
-  Tile* getNextTileToLoadInMainThread() override {
+  const Tile* getNextTileToLoadInMainThread() override {
     CESIUM_ASSERT(!this->_mainThread.empty());
-    Tile* pResult = this->_mainThread.back();
+    const Tile* pResult = this->_mainThread.back();
     this->_mainThread.pop_back();
-    return pResult;
+    return const_cast<Tile*>(pResult);
```
j9liu: Doesn't this return a `const Tile*`? And `pResult` is already `const Tile*` ? 
```suggestion
    return pResult;
```

Cesium3DTilesSelection/test/TestTileLoadRequester.cpp
```diff
@@ -23,45 +23,51 @@ class TestTileLoadRequester : public TileLoadRequester {
...
   }
 
-  Tile* getNextTileToLoadInWorkerThread() override {
+  const Tile* getNextTileToLoadInWorkerThread() override {
     CESIUM_ASSERT(!this->_workerThread.empty());
-    Tile* pResult = this->_workerThread.back();
+    const Tile* pResult = this->_workerThread.back();
     this->_workerThread.pop_back();
-    return pResult;
+    return const_cast<Tile*>(pResult);
```
j9liu: Same here...

```suggestion
    return pResult;
```

Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -1593,7 +1593,9 @@ class WeightedRoundRobin {
       this->recomputeRequesterFractions();
     }
 
-    return pToLoad;
+    // The Tile is const from the perspective of the TileLoadRequester. But the
+    // TilesetContentManager is going to load it, so cast away the const.
+    return const_cast<Tile*>(pToLoad);
```
j9liu, line 1598: I'm not sure I understand why the `const` in `TileLoadRequester` is tacked on to the `getNextTile___` methods?

The way I'm interpreting this, a `const Tile*` output doesn't necessarily guarantee that the `TileLoadRequester` won't modify the `Tile`. It just means that the return of the function is not meant to be modified. But in this case, it's supposed to be modified. And in the body of both `getNextTile___` methods, the `Tile*` is initially retrieved as non-`const`. So unless I'm missing something (and I probably am!), this `const` seems like it's enforcing a limitation that we don't even respect here

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -263,8 +259,8 @@ void Tileset::_updateLodTransitions(
     // Update fade out
     for (auto tileIt = result.tilesFadingOut.begin();
          tileIt != result.tilesFadingOut.end();) {
-      TileRenderContent* pRenderContent =
-          (*tileIt)->getContent().getRenderContent();
+      TileRenderContent* pRenderContent = const_cast<TileRenderContent*>(
```
j9liu, line 262: This `const_cast` I can more easily understand; we want the `Tile`s in `ViewUpdateResult` to be non-modifiable by public API users, but the `Tileset` must modify them under-the-hood to apply these transitions.

Cesium3DTilesSelection/src/Tile.cpp
kring: Yeah this gets awkward. In kinda predictable ways, though.

Ok, so it's sometimes useful to hold const pointers to objects whose lifetime is controlled by reference counting, right? That's what we're doing here. It just means whoever holds that pointer isn't supposed to modify it. It's true that releasing a reference counted pointer to a const object can cause that object to be destroyed, and - oh no! - that's a modification of a const object! But that is best viewed as the object _itself_ destroying itself (which it is free to do), not as the holder of the const pointer destroying the object.

By the way, did you know that it's perfectly valid in C++ to delete a const pointer? Lifetime is separate from mutability.

```cpp
const int* p = new int(42);
delete p; // no problem!
```

So if we accept that all that is reasonable, then the next problem is that the C++ const rules work against us. `addReference` / `releaseReference` need to modify the reference count. If they're const, they can't. If they're not const, then they can't be called on a const instance.

The usual solution is to make those methods const, and make the reference count mutable, and that's what I've done here. It's also possible to keep `addReference` / `releaseReference` non-const, and then `IntrusivePointer` can const-cast. That's not really an improvement, though, because `IntrusivePointer` should just be a helper. We should be able to explicitly call `addReference` / `releaseReference` on e.g., `const Tile*`, too.

Ok so that's just the normal intrusive pointer design so far. You can find an example like this with Boost's intrusive pointer at the top of this page:
https://www.boost.org/doc/libs/1_56_0/doc/html/atomic/usage_examples.html

But our case is a little more complicated. We're not directly deleting heap-allocated objects when their reference count goes to zero. We're controlling whether their content can be unloaded, and maybe, eventually, in some cases, allowing them to be destroyed (but certainly not by the `delete` operator).

In practice, that means `addReference` and `releaseReference` can indirectly modify the tile's `_unusedTilesLinks` field. And in some cases they can also release the tile's content, which is stored as a field on the Tile.

Conceptually, making `_unusedTilesLinks` mutable is just fine, for the same reason making the reference count mutable is fine. That wouldn't be sufficient, though. We'd also have to make some slightly tricky changes to `CesiumUtility::DoublyLinkedList`. It's probably doable, but after thinking about it for a few minutes it wasn't obvious how it should work, so I convinced myself the const_cast was simpler and not obviously worse in terms of semantics.

Especially in light of the fact that the content can also be modified. Making tile content mutable would be... really weird. I think at the point where we do that, we've given up all pretense of const-correctness. Casting away const-ness in this one scenario where we know it's safe is a lot less problematic.

So that's what I did. I considered making `markTileEligibleForContentUnloading` const and putting the const cast inside there instead, but I think that's worse. For one thing, it would mean more const_casts, but that's not too important. More important, I think, is that it gets the semantics right. Users are allowed to reference count `Tile` instances they're not allowed to modify. Sometimes, as a result of the reference counting, the `Tile` will choose to modify itself, which it will do by casting away const.

It's definitely messier than, "A const method means the method won't change the thing. At all. For sure." And arguably this is a case where we should do in C++ what we would be forced to do in languages without a concept of const: separate the "external" interface from the "internal" one, and restrict what clients can do to our instance by only handling them the external interface. But that's a lot more work, and at least in C++ it comes with some performance caveats as well.

So overall, I think what's here is a reasonable tradeoff.

Cesium3DTilesSelection/src/Tileset.cpp
kring: Yeah, although this is actually pointing to a bigger problem, well outside the scope of this PR. Now that we have multiple view groups, it makes no sense for the fade percentage to be stored on the Tile.

Cesium3DTilesSelection/src/TilesetContentManager.cpp
kring: The way I think about it is that only the TilesetContentManager (and things it explicitly delegates the responsibility to, like `IPrepareRendererResources` and eventually `GltfModifier`) are allowed to modify Tile instances. A `TileLoadRequester` - or whoever implements that interface - isn't allowed to modify Tiles. So how could they have a non-const pointer to a Tile? Unless they casted away the const?

So the reason `getNextTile___` returns a const Tile pointer is because that's all the implementors of that method are going to have. By returning it, they're requesting modification of a `Tile` that they themselves are not allowed to modify.

Crystal clear, right?

You're right that returning something by const pointer usually means "and you better not modify this thing I'm giving you." But not here. In this case, the thing (Tile) is actually owned by the entity it's being returned to. And the implementor of the method is identifying the particular thing that it was previously given restricted access to.

Cesium3DTilesSelection/test/TestTileLoadRequester.cpp
kring: Yep, that's left over from a previous iteration. Good catch!

Cesium3DTilesSelection/test/TestTilesetSelectionAlgorithm.cpp
kring: Best to make a change like this in a separate PR, since this was an existing case of the name not matching our naming conventions (and there are heaps more in this file). Otherwise we end up with tons of changes in this file and it's hard to separate what actually changed from all the noise.

Cesium3DTilesSelection/src/Tile.cpp
j9liu: Ahh okay, I think I got it now... it'll take me a bit to reprogram my brain's interpretation of `const`, but it helps to have the justification. Thanks and sorry I made you write a novel! 😓

PR #1222, opened 2025-08-11
Make more properties from tileset.json available on `TilesetMetadata`

kring:
This is  PR into #1221 so merge that first.

We continue to have an unfortunate distinction between `Cesium3DTiles::Tileset` and `Cesium3DTilesSelection::Tileset`. This PR doesn't fix that, but it does make a more complete set of information from tileset.json available on the `Selection` version of `Tileset`. It expands the `TilesetMetadata` class added in #709 to include every tileset property in the 3D Tiles schema, including extensions.

This should enable us to avoid the custom JSON parsing method, `parseTilesetJson`, on the `GltfModifier` class in #1186. In addition to being easier for users (no need to deal with RapidJSON), it avoids a likely backward compatibility issue in the future if and when we consolidate our two `Tileset` classes.

### DIFF
--- a/Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
+++ b/Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
@@ -761,6 +761,7 @@ void removeRootPropertyAndParseTilesetMetadata(
           CesiumUtility::Uri::resolve(baseUrl, *tileset.schemaUri);
     }
     metadata.statistics = std::move(tileset.statistics);
+    metadata.unknownProperties = std::move(tileset.unknownProperties);
   }
 }
 
--- a/Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
+++ b/Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
@@ -18,6 +18,7 @@
 #include <Cesium3DTilesReader/PropertiesReader.h>
 #include <Cesium3DTilesReader/SchemaReader.h>
 #include <Cesium3DTilesReader/StatisticsReader.h>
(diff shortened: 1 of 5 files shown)


Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -9,12 +9,11 @@
 #include <Cesium3DTiles/ExtensionContent3dTilesContentVoxels.h>
 #include <Cesium3DTilesContent/GltfConverterResult.h>
 #include <Cesium3DTilesContent/GltfConverters.h>
+#include <Cesium3DTilesReader/AssetReader.h>
```
j9liu: I'm not seeing this used, remove?

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -710,41 +711,55 @@ TilesetContentLoaderResult<TilesetJsonLoader> parseTilesetJson(
...
-    }
+  // Remove the root tile from the RapidJSON document. Parsing the complete tile
+  // tree will take too long, and we don't need it.
+  tilesetJson.RemoveMember("root");
+
+  Cesium3DTilesReader::TilesetReader tilesetReader;
+  auto tilesetResult = tilesetReader.readFromJson(tilesetJson);
+
+  if (!tilesetResult.errors.empty() || !tilesetResult.warnings.empty()) {
+    ErrorList el;
```
j9liu: **NITPICK**

```suggestion
    ErrorList errorList;
```

j9liu (approved):
Thanks @kring ! Nothing major, just two minor suggestions that I'll commit myself.

PR #1224, opened 2025-08-15
Fix an endless loop in `updateViewGroupOffline` when it's given an empty list of frustums

kring:
Reported here:
https://community.cesium.com/t/playmoviesequencer-blueprint-node-freezes-mrq-render-after-v2-15-0/42499

It seems that the first time `ACesium3DTileset::Tick` is called under the control of a Level Sequence / Movie Render Queue, the player camera has a width and height of 0 pixels. As a result, Cesium for Unreal discards that camera and calls `updateViewGroupOffline` with an empty list of frustums. This causes `updateViewGroupOffline` to get stuck in an endless loop waiting for the load progress to reach 100% (it will never go above 0%).

The fix here is to end the loop after one iteration when there are no frustums.

This probably worked before #1125 because we used to initialize the load progress to 100%. This was misleading and could cause applications to think loading was complete before it really was, but it also avoided this endless loop.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@
 
 ##### Fixes :wrench:
 
-- Fixed a bug in `Tileset::updateViewGroupOffline` that would cause it to get stuck in an endless loop if invoked with no frustums.
+- Fixed a bug in `Tileset::updateViewGroupOffline` that would cause it to get stuck in an endless loop when invoked with no frustums.
 
 ### v0.50.0 - 2025-08-01
 


CHANGES.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
+- Fixed a bug in `Tileset::updateViewGroupOffline` that would cause it to get stuck in an endless loop if invoked with no frustums.
```
j9liu: Super nitpick:

```suggestion
- Fixed a bug in `Tileset::updateViewGroupOffline` that would cause it to get stuck in an endless loop when invoked with no frustums.
```

PR #1226, opened 2025-08-21
add extra info for converted i3dm gltf

tomlee-dev:
for distingushing other type of converted gltf (e.g. b3dm, vctr, pnts, cmpt)


TomLeeLive:
@timoore thank you the comment!

my purpose is that checking 3dtiles type of converted gltf in UnityPrepareRendererResources.cpp
what is the best way to check it with considering runtime performance? checking glTF extension of onverted gltf in UnityPrepareRendererResources.cpp will be better way?

TomLeeLive:
@timoore oh I get it. for my comment above never mind!

I just add a method for checking it in in UnityPrepareRendererResources.cpp

bool CheckForExtMeshGpuInstancing(
    const CesiumGltf::Model& gltf,
    const CesiumGltf::Node& node) {
  return node.getExtension<CesiumGltf::ExtensionExtMeshGpuInstancing>() !=
         nullptr;
}

PR #1231, opened 2025-08-26
Further improvements to the GltfModifier in #1186

kring:
This PR builds on the work in #1186. Some of the key changes from that original PR are:

* `GltfModifier` previously managed its version with an `atomic_int`. Now, like most things in cesium-native, it is intended to only be used from the main thread. Except for `apply`, which is still invoked in a worker.
* `GltfModifier::apply`:
  * now returns a `Future`, so it can work asynchronously.
  * takes `GltfModifierInput` and returns `GltfModifierOutput`, making it easier to expand the inputs and outputs in the future.
  * is given the version in the `GltfModifierInput`, so it doesn't need to call `getCurrentVersion`. Which is good because doing so from a thread is now discouraged.
  * no longer needs to manually add/update the version of the modified model. The system does this automatically.
  * no longer receives the root tile's transform. Instead, a new `onRegister` method can be overridden to receive the root tile and the `TilesetMetadata` instance. It's better to use the bounding volume of the root tile to determine approximately where in the world the model is located. This will be much more reliable than looking at the root tile translation.
* Removed `GltfModifier::parseTilesetJson`. The root tile and `TilesetMetadata` mentioned above should cover the possible use-cases of this.
* `isRenderable` is no longer dependent on the GltfModifier version being up-to-date. So new versions will appear incrementally and without causing the LOD to shift.
* `GltfModifier` now provides a `TileLoadRequester`, which is used to cause all already-loaded tiles to (eventually) update to the new version. This avoids a situation where visible tiles update to a new version, but if you spin the camera around to previously-hidden tiles, they'll all still be the old version momentarily. The `TileLoadRequester` uses a "weight" of 0.5, which means that about two visible tiles will load for each background-updating tile. This could easily be made configurable, but I just hard-coded it in this PR.
* Instead of adding a `version` property directly to `CesiumGltf::Model`, the version is now attached by adding a `GltfModifierVersionExtension` to it.
* I've tried to simplify the state management. I'm not completely sure I've succeeded.
* Various documentation improvements and small reorganizations.

### DIFF
--- a/Cesium3DTilesSelection/src/TilesetContentManager.cpp
+++ b/Cesium3DTilesSelection/src/TilesetContentManager.cpp
@@ -1131,9 +1131,7 @@ void TilesetContentManager::reapplyGltfModifier(
           const CesiumGltf::Model& model =
               modified ? modified->modifiedModel : previousModel;
           const auto it = model.extras.find("gltfUpAxis");
-          if (it == model.extras.end()) {
-            tileLoadResult.glTFUpAxis = CesiumGeometry::Axis::Y;
-          } else {
+          if (it != model.extras.end()) {
             tileLoadResult.glTFUpAxis = static_cast<CesiumGeometry::Axis>(
                 it->second.getSafeNumberOrDefault(1));
           }
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileLoadResult.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileLoadResult.h
@@ -78,7 +78,7 @@ struct CESIUM3DTILESSELECTION_API TileLoadResult {
   /**
    * @brief The up axis of glTF content.
    */
-  CesiumGeometry::Axis glTFUpAxis;
+  CesiumGeometry::Axis glTFUpAxis = CesiumGeometry::Axis::Y;
 
   /**
    * @brief A tile can potentially store a more fit bounding volume along with


kring:
This should be ready to go. Merging this will merge #1186, too. @GhisBntly I'd really appreciate if you or someone from your team could take a look at this. Mostly to ensure that it still works well for your use-cases, but I'm interested in any other feedback as well.

kring:
Here's a modification to `Cesium3DTileset.cpp` in cesium-unreal that I used to test this.

```
diff --git a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
index 07554d88..e5245bd9 100644
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -5,6 +5,7 @@
 #include "Camera/CameraTypes.h"
 #include "Camera/PlayerCameraManager.h"
 #include "Cesium3DTilesSelection/EllipsoidTilesetLoader.h"
+#include "Cesium3DTilesSelection/GltfModifier.h"
 #include "Cesium3DTilesSelection/Tile.h"
 #include "Cesium3DTilesSelection/TilesetLoadFailureDetails.h"
 #include "Cesium3DTilesSelection/TilesetOptions.h"
@@ -872,6 +873,7 @@ void ACesium3DTileset::UpdateLoadStatus() {
 }

 namespace {
+
 const TSharedRef<CesiumViewExtension, ESPMode::ThreadSafe>&
 getCesiumViewExtension() {
   static TSharedRef<CesiumViewExtension, ESPMode::ThreadSafe>
@@ -879,6 +881,60 @@ getCesiumViewExtension() {
           GEngine->ViewExtensions->NewExtension<CesiumViewExtension>();
   return cesiumViewExtension;
 }
+
+class TestGltfModifier : public Cesium3DTilesSelection::GltfModifier {
+public:
+  virtual CesiumAsync::Future<void> onRegister(
+      const CesiumAsync::AsyncSystem& asyncSystem,
+      const std::shared_ptr<CesiumAsync::IAssetAccessor>& pAssetAccessor,
+      const std::shared_ptr<spdlog::logger>& pLogger,
+      const Cesium3DTilesSelection::TilesetMetadata& tilesetMetadata,
+      const Cesium3DTilesSelection::Tile& rootTile) override {
+    return asyncSystem.createResolvedFuture();
+  }
+
+  virtual CesiumAsync::Future<
+      std::optional<Cesium3DTilesSelection::GltfModifierOutput>>
+  apply(Cesium3DTilesSelection::GltfModifierInput&& input) override {
+    std::vector<double> colors[] = {
+        {1.0, 0.0, 0.0, 1.0},
+        {0.0, 1.0, 0.0, 1.0},
+        {0.0, 0.0, 1.0, 1.0},
+    };
+    Cesium3DTilesSelection::GltfModifierOutput output{
+        .modifiedModel = input.previousModel};
+
+    size_t index =
+        input.version % (sizeof(colors) / sizeof(std::vector<double>));
+
+    for (auto& material : output.modifiedModel.materials) {
+      if (!material.pbrMetallicRoughness) {
+        material.pbrMetallicRoughness.emplace();
+      }
+      material.pbrMetallicRoughness->baseColorFactor = colors[index];
+    }
+
+    using namespace std::chrono_literals;
+    std::chrono::milliseconds multiplier = 250ms;
+
+    double r = double(std::rand()) / double(RAND_MAX);
+    std::this_thread::sleep_for(multiplier * r);
+
+    return input.asyncSystem.createResolvedFuture(
+        std::make_optional(std::move(output)));
+  }
+
+  void tick(float deltaTime) {
+    float nextTime = this->_time + deltaTime;
+    if (std::floor(nextTime / 10.0f) > std::floor(this->_time / 10.f))
+      this->trigger();
+    this->_time = nextTime;
+  }
+
+private:
+  float _time = 0.0f;
+};
+
 } // namespace

 void ACesium3DTileset::LoadTileset() {
@@ -992,7 +1048,11 @@ void ACesium3DTileset::LoadTileset() {
            ->EnableExperimentalOcclusionCullingFeature &&
        this->EnableOcclusionCulling && this->BoundingVolumePoolComponent)
           ? this->BoundingVolumePoolComponent->getPool()
+    double r = double(std::rand()) / double(RAND_MAX);
+    std::this_thread::sleep_for(multiplier * r);
+
+    return input.asyncSystem.createResolvedFuture(
+        std::make_optional(std::move(output)));
+  }
+
+  void tick(float deltaTime) {
+    float nextTime = this->_time + deltaTime;
+    if (std::floor(nextTime / 10.0f) > std::floor(this->_time / 10.f))
+      this->trigger();
+    this->_time = nextTime;
+  }
+
+private:
+  float _time = 0.0f;
+};
+
 } // namespace

 void ACesium3DTileset::LoadTileset() {
@@ -992,7 +1048,11 @@ void ACesium3DTileset::LoadTileset() {
            ->EnableExperimentalOcclusionCullingFeature &&
        this->EnableOcclusionCulling && this->BoundingVolumePoolComponent)
           ? this->BoundingVolumePoolComponent->getPool()
-          : nullptr};
+          : nullptr,
+      Cesium3DTilesSelection::TilesetSharedAssetSystem::getDefault(),
+      std::make_shared<TestGltfModifier>()};
+
+  externals.pGltfModifier->trigger();

   this->_startTime = std::chrono::high_resolution_clock::now();

@@ -2007,6 +2067,10 @@ void ACesium3DTileset::Tick(float DeltaTime) {
     }
   }

+  static_cast<TestGltfModifier*>(
+      this->_pTileset->getExternals().pGltfModifier.get())
+      ->tick(DeltaTime);
+
   if (this->BoundingVolumePoolComponent && this->_cesiumViewExtension) {
     TRACE_CPUPROFILER_EVENT_SCOPE(Cesium::UpdateOcclusion)
     const TArray<USceneComponent*>& children =
```

kring:
Oh I just noticed your code is parsing the `extras` property of the `asset`. So perhaps if you expand the `asset` property on the `TilesetMetadata` you'll find what you're looking for there?

GhisBntly:
Oops, you're right, thanks! Thought I had unfolded them all, sorry about that... ,^^

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetExternals.h
```diff
@@ -76,6 +78,14 @@ class CESIUM3DTILESSELECTION_API TilesetExternals final {
    */
   CesiumUtility::IntrusivePointer<TilesetSharedAssetSystem> pSharedAssetSystem =
       TilesetSharedAssetSystem::getDefault();
+
+  /**
+   * Optional user-controlled tile loading post-processing stage that can modify
+   * the glTF meshes (eg. split or merge them).
```
j9liu: **NIT:**

```suggestion
   * the glTF meshes (e.g., split or merge them).
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetViewGroup.h
```diff
@@ -96,8 +98,13 @@ class CESIUM3DTILESSELECTION_API TilesetViewGroup final
...
    *
    * @param task The tile load task to add to the queue.
+   * @param pModifier The optional glTF modifier. If not `nullptr`, this method
+   * will also add the tile to load queue if it needs glTF modification. See
+   * {@link TilesetExternals::pGltfModifier}.
    */
-  void addToLoadQueue(const TileLoadTask& task);
+  void addToLoadQueue(
+      const TileLoadTask& task,
+      const std::shared_ptr<GltfModifier>& pModifier);
```
j9liu: Maybe `pModifier` could have a default value of `nullptr` since it's optional?

Cesium3DTilesSelection/include/Cesium3DTilesSelection/GltfModifierState.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+namespace Cesium3DTilesSelection {
+
+/** The state of the glTF modifier process for a given tile content's model.
```
j9liu: Worth linking to the actual `GltfModifier` class?

```suggestion
/** The state of the @ref GltfModifier process for a given tile content's model.
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/GltfModifierState.h
```diff
@@ -0,0 +1,16 @@
...
+ */
+enum class GltfModifierState {
+  /** Modifier is not currently processing this tile. */
+  Idle,
+  /** Worker-thread phase is in progress. */
+  WorkerRunning,
+  /** Worker-thread phase is complete, main-thread phase is not done yet. */
+  WorkerDone,
```
j9liu: Would someone ever need a state to detect if the main thread phase is done? Or nah?

Cesium3DTilesSelection/include/Cesium3DTilesSelection/GltfModifier.h
```diff
@@ -0,0 +1,278 @@
...
+#include <Cesium3DTilesSelection/GltfModifierState.h>
+#include <Cesium3DTilesSelection/Tile.h>
+#include <Cesium3DTilesSelection/TileLoadRequester.h>
+#include <CesiumAsync/AsyncSystem.h>
+#include <CesiumAsync/Future.h>
+#include <CesiumGltf/Model.h>
+
+#include <spdlog/fwd.h>
+
+#include <atomic>
```
j9liu: Remove? (I'm surprised clang-tidy isn't complaining here)

Cesium3DTilesSelection/include/Cesium3DTilesSelection/GltfModifier.h
```diff
@@ -0,0 +1,278 @@
...
+  CesiumGltf::Model modifiedModel;
+};
+
+/**
+ * @brief An abstract class that allows modifying a tile's glTF model after it
+ * has been loaded.
+ *
+ * An example modification is merging or splitting the primitives in the glTF.
+ * Merging primitives can lead to improved rendering performance. Splitting
+ * primitives allows to assign different materials to parts that were initially
```
j9liu:
```suggestion
 * primitives allows different materials to be assigned to parts that were initially
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/GltfModifier.h
```diff
@@ -0,0 +1,278 @@
...
+public:
+  /**
+   * @brief Gets the current version number, or `std::nullopt` if the
+   * `GltfModifier` is currently inactive.
+   *
+   * Returns `std::nullopt` when in the default nilpotent state where glTFs will
+   * not be modified at all. Call {@link trigger} once to set the version
+   * number to 0 and activate the `GltfModifier`. Call it successive times to
+   * increment the version number and re-apply modification to all
+   * previously-modified models.
```
j9liu:
```suggestion
   * not be modified at all. Calling @ref trigger once will set the version
   * number to 0 and activate the `GltfModifier`. Calling it successive times will
   * increment the version number and re-apply modification to all
   * previously-modified models.
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/GltfModifier.h
```diff
@@ -0,0 +1,278 @@
...
+   */
+  static bool
+  needsMainThreadModification(const GltfModifier* pModifier, const Tile& tile);
+
+protected:
+  GltfModifier();
+  virtual ~GltfModifier();
+
+  /**
+   * @brief Notifies this instance that is has been registered with a
```
j9liu:
```suggestion
   * @brief Notifies this instance that it has been registered with a
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/GltfModifier.h
```diff
@@ -0,0 +1,278 @@
...
+   * {@link Tileset}.
+   *
+   * This method is called after the tileset's root tile is known but
+   * before {@link Tileset::getRootTileAvailableEvent} has been raised.
+   *
+   * This method is called from the main thread. Override this method respond to
```
j9liu:
```suggestion
   * This method is called from the main thread. Override this method to respond to
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileContent.h
```diff
@@ -195,9 +198,67 @@ class CESIUM3DTILESSELECTION_API TileRenderContent {
...
+
+  /**
+   * @brief Sets the state of the {@link GltfModifier} processing of this
+   * tile's content.
+   */
+  void setGltfModifierState(GltfModifierState modifierState) noexcept;
+
+  /**
+   * @brief Gets the modified model produced by the {@link GltfModifier} but
+   * not yet available for rendering.
```
j9liu: Is this a fair change? I was confused by this at first but after some reads, I think I get it.

```suggestion
   * @brief Gets the modified model produced by the {@link GltfModifier} that
   * is not yet available for rendering.
```

Cesium3DTilesSelection/src/GltfModifier.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+bool GltfModifier::isActive() const {
+  return this->getCurrentVersion().has_value();
+}
+
+void GltfModifier::trigger() {
+  if (!this->_currentVersion) {
+    this->_currentVersion = 0;
+  } else {
+    ++*this->_currentVersion;
```
j9liu: **NIT** for readability:

```suggestion
    ++(*this->_currentVersion);
```

Cesium3DTilesSelection/src/GltfModifier.cpp
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  if (!this->isRegistered()) {
+    return;
+  }
+
+  // Add all already-loaded tiles to this requester worker thread load queue.
```
j9liu:
```suggestion
  // Add all already-loaded tiles to this requester's worker thread load queue.
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/GltfModifier.h
```diff
@@ -0,0 +1,278 @@
...
+  /**
+   * @brief Checks if the given tile needs to be processed by the given
+   * `GltfModifier` in a worker thread.
+   * @param pModifier The `GltfModifier` to check.
+   * @param tile The tile to check.
+   * @returns `true` if the tile needs to be processed by the `GltfModifier` in
+   * a worker thread, or `false` otherwise.
+   */
+  static bool needsWorkerThreadModification(
+      const GltfModifier* pModifier,
```
j9liu: This is a `static` function with a pointer to `GltfModifier`, but you need a non-null pointer `GltfModifier` to even do anything here. Is there a reason it isn't a member function?

j9liu: I see that there's a call in `Tile.cpp` with a potentially null `GltfModifier`, but we could check `pModifier && pModifier->needsMainThreadModification(*this)` ?

Cesium3DTilesSelection/include/Cesium3DTilesSelection/GltfModifier.h
```diff
@@ -0,0 +1,278 @@
...
+  /**
+   * @brief Checks if the given tile needs to be processed by the given
+   * `GltfModifier` in the main thread.
+   * @param pModifier The `GltfModifier` to check.
+   * @param tile The tile to check.
+   * @returns `true` if the tile needs to be processed by the `GltfModifier` in
+   * the main thread, or `false` otherwise.
+   */
+  static bool
+  needsMainThreadModification(const GltfModifier* pModifier, const Tile& tile);
```
j9liu: Same concern as `needsWorkerThreadModification`

Cesium3DTilesSelection/include/Cesium3DTilesSelection/GltfModifierState.h
j9liu: Hm okay, after going through more of the code, I see that the main thread phase is done once the `modifiedModel` is swapped with the old model version. So no reason to put an enum here, but I do feel like a comment would be helpful!

Cesium3DTilesSelection/src/TileContent.cpp
```diff
@@ -14,22 +17,68 @@ namespace Cesium3DTilesSelection {
...
+void TileRenderContent::resetModifiedModelAndRenderResources() noexcept {
+  this->_pModifiedRenderResources = nullptr;
+  this->_modifiedModel.reset();
+}
+
+void TileRenderContent::replaceWithModifiedModel() noexcept {
+  CESIUM_ASSERT(this->_modifiedModel);
+  if (this->_modifiedModel) {
+    this->_model = std::move(*this->_modifiedModel);
+    // reset after move because tested in Tile::needsWorkerThreadLoading:
```
j9liu: **NIT**:

```suggestion
    // reset after move because this is tested for nullopt in Tile::needsWorkerThreadLoading:
```

Cesium3DTilesSelection/test/TestTilesetContentManager.cpp
```diff
@@ -496,6 +498,8 @@ TEST_CASE("Test tile state machine") {
             std::move(pMockedLoader),
             std::move(pRootTile)};
 
+    pManager->waitUntilIdle();
```
j9liu, line 501: Was this added in _because_ of the glTF modifier changes, or should this have been here from the beginning?

Just trying to understand the impact of the former; this implies that tiles will take more ticks to resolve than before, but maybe I'm misunderstanding here.

Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -645,6 +653,51 @@ postProcessContentInWorkerThread(
...
+
+                    CesiumGltf::Model* pModel =
+                        std::get_if<CesiumGltf::Model>(&result.contentKind);
+                    if (pModel) {
+                      GltfModifierVersionExtension::setVersion(
+                          *pModel,
+                          *version);
+                    }
+
+                    return result;
```
j9liu, line 687: Trying to check my understanding: this is done before any `pRenderContent` is created, otherwise we would just call `pRenderContent->replaceWithModifiedModel`. Is that right?

Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -981,11 +1056,206 @@ TilesetContentManager::~TilesetContentManager() noexcept {
...
+            pRenderContent->getRasterOverlayDetails();
+        tileLoadResult.initialBoundingVolume = tileBoundingVolume;
+        tileLoadResult.initialContentBoundingVolume = tileContentBoundingVolume;
+
+        {
+          const CesiumGltf::Model& model =
+              modified ? modified->modifiedModel : previousModel;
+          const auto it = model.extras.find("gltfUpAxis");
+          if (it == model.extras.end()) {
+            tileLoadResult.glTFUpAxis = CesiumGeometry::Axis::Y;
```
j9liu: Silly question, but isn't this redundant with the glTF spec?

j9liu (commented):
Thanks @kring ! I appreciate the docs / comments you left throughout.

I still have to digest some of what's going on, especially in `TilesetContentManager.cpp`, but I gathered some initial comments with questions + small docfixes. I'll avoid troubling you for those small fixes and push a commit myself!

Cesium3DTilesSelection/include/Cesium3DTilesSelection/GltfModifier.h
kring: We could, but it would make that code a lot less readable:

Currently:
```cpp
  return this->getState() == TileLoadState::Unloaded ||
         this->getState() == TileLoadState::FailedTemporarily ||
         GltfModifier::needsWorkerThreadModification(pModifier, *this) ||
         anyRasterOverlaysNeedLoading(*this);
```

If it were an instance method:

```cpp
  return this->getState() == TileLoadState::Unloaded ||
         this->getState() == TileLoadState::FailedTemporarily ||
         (pModifier != nullptr && pModifier->needsWorkerThreadModification(*this)) ||
         anyRasterOverlaysNeedLoading(*this);
```

Or we break it up into separate return statements. I guess that would be better, so I'll do that.

Cesium3DTilesSelection/test/TestTilesetContentManager.cpp
kring: It's required now, and it wasn't before. Previously, constructing a `TilesetContentManager` with an already-constructed loader (which was pretty much only done in tests) would yield an immediately-ready-to-use content manager complete with a root tile. Now, with the `registerGltfModifier` and `thenInMainThread` continuation, that is no longer case. The root tile won't become available until the first call to `dispatchMainThreadTasks`. It only delays the root tile in this specific case, though, not any later tiles.

If we wanted to avoid this, we could use the new `AsyncSystem::enterMainThread` to tell the async system that we're already in the main thread, and then the `thenInMainThread` continuation would be invoked immediately in the case that there is no GltfModifier or its Future resolves immediately.

I think this is probably fine and not worth the trouble, but I can be convinced. One thing to consider is that adding that `enterMainThread` call might also require changes to tests, because it is likely to make other things happen _sooner_. Not that that's a bad thing.

Cesium3DTilesSelection/src/TilesetContentManager.cpp
kring: Yes that's right. We can tell this method is operating on content that isn't available on the Tile yet because we're in a worker thread and `TileLoadResult` is passed in as an r-value reference. That wouldn't make sense if the content were "owned" by the Tile (i.e., if it were already renderable).

Cesium3DTilesSelection/src/TilesetContentManager.cpp
kring: I don't think so. glTF doesn't have an up-axis property, it's always Y-up.

kring:
Thanks for the review @j9liu! I think I've addressed all your comments.

Cesium3DTilesSelection/src/TilesetContentManager.cpp
j9liu: Right, that's what I mean... glTF is always y-up, and this property says the glTF up axis is Y. 👀

kring: Oh, I see. Yeah, certainly that's the default, but here is where we're setting that default. I guess there are other possible ways to structure this:

1. Make `TileLoadResult::glTFUpAxis` a `std::optional` so it's only has a value if it's not the default.
2. Give `TileLoadResult::glTFUpAxis` a default value consistent with the glTF default in the class definition.

I don't think I like (1). It would just require any place the up axis is _consumed_ to use a `.value_or(Y)`. While there are probably fewer occurrences of that compared to loaders that are currently setting this property to Y, I still think the current setup is better. In the current setup, consumers don't need to be aware of the default, while producers only need to think about what _their_ data's up axis is (even if it happens to be the default), which doesn't seem like a bad thing.

(2) could be ok. It would let us simplify this conditional and remove any knowledge of the glTF default from this particular code. I don't think it would change any other code in cesium-native, though, because every other place we create a `TileLoadResult` we do it using brace initialization.

Note that this particular case you're reviewing here is unusual in that it's just trying to make sure that the up-axis of the new, modified model matches the original. And actually it may very well be the case that we'll _always_ take the `else` block, because there will always be an `extras["gltfUpAxis"]` property.

Cesium3DTilesSelection/test/TestTilesetContentManager.cpp
j9liu: Gotcha, thanks! Although that `enterMainThread` change might be nice, I guess I'm not sure how it would affect other tests. And if this scenario only realistically happens for our tests, I agree it's not worth the trouble to change this.

Cesium3DTilesSelection/src/TilesetContentManager.cpp
j9liu: I like how (2) sounds, and it sounds simple enough to do! But I recognize your statement that this is a fairly small scope for the change. I wouldn't hold up this PR over it.

kring: Done

Cesium3DTilesSelection/test/TestTilesetContentManager.cpp
kring: I think I'll leave this alone for now. Longer term, I want to look at making `updateViewGroup` call `enterMainThread`. And maybe rig up the tests to know they're in the main thread, too. But those are slightly bigger changes outside the scope of this PR and I don't see much reason to do something hackier here and now.

kring:
I think everything is addressed here @j9liu. Thanks again!

PR #1232, opened 2025-08-28
Fix raster overlay crash with external tilesets

kring:
This should fix the instability we've recently been seeing with clipping polygons and Google Photorealistic 3D Tiles. Here's the sequence that I think leads to the crash:

1. A new raster overlay is added to Google Photorealistic 3D Tiles.
2. A placeholder for the new overlay is added to all tiles that are already loaded.
3. The code wasn't checking to ensure the tiles were renderable, so it was added to external tileset tiles, too.
4. The unloadTileContent has a newish separate path for external tilesets, which returns before raster overlays are detached.
5. So unloaded external tilesets keep their (useless) raster overlays.
6. Later, the raster overlay is removed.
7. Unloaded tiles aren't enumerated by the LoadedTileEnumerator, so the mapped raster overlay tiles are not removed from the external tileset tiles.
8. When we destroy the placeholder tile provider, we detect that there are still placeholder tiles (attached to the unloaded external tileset tiles), triggering an assertion failure.
9. In release builds, we eventually crash or have sketchy behavior due to violation of the assertion condition (A RasterOverlayTile outliving its RasterOverlayTileProvider, mostly).

But even with that understanding of what's happening, it's remarkably hard to reproduce consistently. Which makes it hard to be 100% sure it's fixed.

In any case, this PR fixes this in two ways that I think should be pretty reliable:

- Raster overlay placeholders are no longer added to tiles that are known to not be renderable (e.g., loaded external tilesets).
- When a tile's content is unloaded, its raster overlays are now detached. Even if the content being unloaded is an external tileset. This should cover us in case one sneaks through the above (though I don't know of a way that can happen).

See also #1161, which previously fixed a very similar bug.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@
 ##### Fixes :wrench:
 
 - Fixed a bug in `Tileset::updateViewGroupOffline` that would cause it to get stuck in an endless loop when invoked with no frustums.
+- Fixed a bug that could lead to a crash when using raster overlays with tilesets that use "external tilesets", such as Google Photorealistic 3D Tiles.
 
 ### v0.50.0 - 2025-08-01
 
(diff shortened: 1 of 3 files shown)


azrogers:
The series of steps seem plausible to me, but unfortunately I was still able to reproduce the crash with this branch active. I think it's worth getting these fixes in, as they at worst *can't hurt*, but this doesn't seem to solve the issue.

PR #1233, opened 2025-08-28
Tileset geojson

timoore:
Support tilesets with GeoJson content via the MAXAR_content_geojson extension.


kring:
Thanks @timoore! While this looks reasonable to me, I'm a little hesitant to merge it at the moment. Mostly because I suspect we're eventually going to go in a direction where vector data is encoded in glTF, rather than create a separate Tile content type for GeoJSON specifically. I think we should wait to merge this until either we have some more clarity on that, or we have a more pressing need to deliver this functionality. Does that seem ok to you?

j9liu:
Closing in favor of #1332.

PR #1238, opened 2025-09-08
Add `SharedAssetDepot::invalidate`

kring:
This PR adds an `invalidate` method to `SharedAssetDepot`, allowing a previously-cached asset to effectively be removed from the depot. The next time the same key is requested, a new asset will be created.

This sounds pretty simple, but things are a bit more complicated than they appear because of thread safety requirements. In particular, `SharedAsset` has a pointer to the `SharedAssetDepot`. When an asset is invalidated, we'd ideally like to set this pointer to nullptr. The `SharedAsset` is still valid, it's just not associated with the depot anymore.

We can't do that, though, because releasing that pointer could allow the depot to be destroyed. Meanwhile, another thread might be calling through that pointer. That would result in undefined behavior (probably a crash).

I don't see any way around that short of giving each `SharedAsset` its own mutex, which I'd really rather not do.

So instead of trying to detach the asset from the depot entirely, this PR takes a different approach. When an asset is invalidated, a new property on the `SharedAsset` marks it as such and the corresponding `AssetEntry` is removed. These two operations are done under the depot lock. Invalidated assets keep the depot alive (they need to, because they still have a pointer to it), but their lifetime becomes controlled exclusively by reference counting (rather than the depot's deletion candidate list). Invalidated assets don't count against the depot's `inactiveAssetSizeLimitBytes`.

When an invalidated asset's reference count goes to zero, the depot deletes it immediately instead of adding it to the unused list.

This PR also changes the public property, `SharedAssetDepot::inactiveAssetSizeLimitBytes`, to be a `std::atomic<int64_t>` (previously it was just `int64_t`). This should _mostly_ not matter as an int64_t is likely to be atomic enough anyway on our supported platforms, but it is more correct.


j9liu:
@azrogers mind giving this a pass when you get the chance?

azrogers (approved):

kring:
Your explanation is perfect, @azrogers!

azrogers:
Great! Thanks @kring!

PR #1239, opened 2025-09-09
Add `RasterOverlayExternals`, make `RasterOverlayTileProvider::loadTileImage` receive a const `RasterOverlayTile`

kring:
This is some light but slightly noisy raster overlay refactoring.

- Added `RasterOverlayExternals` to wrap up all the little interfaces that raster overlays use. I've so far not threaded this throughout the whole system. Mostly because it's pending some other WIP changes.
- `loadTileImage` now receives a const reference to the `RasterOverlayTile`. There's no good reason this method should modify the tile.
- Moved implementations of some `RasterOverlayTileProvider` methods to the .cpp file.

### DIFF
--- a/CesiumRasterOverlays/include/CesiumRasterOverlays/RasterOverlayExternals.h
+++ b/CesiumRasterOverlays/include/CesiumRasterOverlays/RasterOverlayExternals.h
@@ -1,0 +1,62 @@
+#pragma once
+
+#include <CesiumAsync/AsyncSystem.h>
+#include <CesiumAsync/IAssetAccessor.h>
+#include <CesiumRasterOverlays/IPrepareRasterOverlayRendererResources.h>
+#include <CesiumRasterOverlays/Library.h>
+#include <CesiumUtility/CreditSystem.h>
+
+#include <spdlog/spdlog.h>
+
+#include <memory>
+
+namespace CesiumRasterOverlays {
+
+/**
+ * @brief External interfaces used by a {@link RasterOverlay}.
+ */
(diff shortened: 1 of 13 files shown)


CesiumRasterOverlays/include/CesiumRasterOverlays/RasterOverlayExternals.h
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * @brief The async system to use to do work in threads.
+   */
+  CesiumAsync::AsyncSystem asyncSystem;
+
+  /**
+   * @brief The {@link CesiumUtility::CreditSystem} that can be used to manage
+   * credit strings and periodically query which credits to show and and which
+   * to remove from the screen.
```
j9liu:
```suggestion
   * credit strings and periodically query which credits to show and which to
   * remove from the screen.
```

CesiumRasterOverlays/include/CesiumRasterOverlays/RasterOverlayExternals.h
```diff
@@ -0,0 +1,62 @@
...
+#include <CesiumUtility/CreditSystem.h>
+
+#include <spdlog/spdlog.h>
+
+#include <memory>
+
+namespace CesiumRasterOverlays {
+
+/**
+ * @brief External interfaces used by a {@link RasterOverlay}.
```
j9liu: Prefer `@ref` in accordance with our style guide:

```suggestion
 * @brief External interfaces used by a @ref RasterOverlay.
```

PR #1240, opened 2025-09-11
Make sure vcpkg dependencies with [features] are installed.

kring:
Fixes CesiumGS/cesium-unity#598

`libblend2d.a` was not being installed for iOS because it (alone) uses a vcpkg `[feature]` specification (specifically: `blend2d[core]`) to turn off asmjit, which is enabled by default. This triggered a bug in the part of our CMakeLists.txt that is meant to find the .lib associated with the vcpkg dependencies and install them.

The fix implemented here is to remove the `[feature]` from the package name using a regular expression.

We were previously _trying_ to do this, but it didn't work because:

- It was only implemented for public dependencies, and blend2d is a private dependency.
- The backslashes were not properly escaped, so the regex didn't find the feature spec it was looking for.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@
 ##### Fixes :wrench:
 
 - Fixed a build system bug that prevented `libblend2d.a` from being installed for iOS.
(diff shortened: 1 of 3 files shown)


j9liu:
Thanks @kring ! I saw feedback from one of our users that the Unity iOS build now succeeds, so I'll merge once CI passes.

PR #1243, opened 2025-09-16
Add move constructor / assignment operator for TileProviderAndTile.

kring:
`TileProviderAndTile` has an explicit destructor. Therefore, the compiler does not generate a move constructor:

> If no user-defined move constructors are provided for a class type, and all of the following is true:
> 
> - ...
> - there is no user-declared destructor.
> - ...
>
> Then the compiler will declare a move constructor as a non-[explicit](https://en.cppreference.com/w/cpp/language/explicit.html) inline public member of its class with the signature T::T(T&&).

https://en.cppreference.com/w/cpp/language/move_constructor.html

This could result in the `AsyncSystem` copying an instance of this class in a non-main thread, which is disastrous because the IntrusivePointers it holds are to non-thread-safe classes. Moving is safe.

### DIFF
--- a/CesiumRasterOverlays/include/CesiumRasterOverlays/RasterOverlayTileProvider.h
+++ b/CesiumRasterOverlays/include/CesiumRasterOverlays/RasterOverlayTileProvider.h
@@ -133,7 +133,16 @@ struct TileProviderAndTile {
    * RasterOverlayTile used for this tile. */
   CesiumUtility::IntrusivePointer<RasterOverlayTile> pTile;
 
+  TileProviderAndTile(
+      const CesiumUtility::IntrusivePointer<RasterOverlayTileProvider>&
+          pTileProvider_,
+      const CesiumUtility::IntrusivePointer<RasterOverlayTile>&
+          pTile_) noexcept;
   ~TileProviderAndTile() noexcept;
+  TileProviderAndTile(const TileProviderAndTile&) = delete;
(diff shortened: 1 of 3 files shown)


CesiumRasterOverlays/include/CesiumRasterOverlays/RasterOverlayTileProvider.h
```diff
@@ -133,7 +133,31 @@ struct TileProviderAndTile {
...
+  TileProviderAndTile(
+      const CesiumUtility::IntrusivePointer<RasterOverlayTileProvider>&
+          pTileProvider_,
+      const CesiumUtility::IntrusivePointer<RasterOverlayTile>&
+          pTile_) noexcept;
+
   ~TileProviderAndTile() noexcept;
+
+  TileProviderAndTile(const TileProviderAndTile&) = delete;
+  TileProviderAndTile& operator=(const TileProviderAndTile&) = delete;
```
j9liu: Is it worth mentioning that these were deleted in the changelog?

kring: Yeah, good point. Maybe I'll just make them `default` instead of `delete`, actually. There's no reason copying needs to be forbidden. The main reason I deleted them was to be certain they were being moved instead of copied.

PR #1244, opened 2025-09-16
Refactor `RasterOverlayTileProvider` to make it easier to "decorate" it

kring:
This builds on #1239 so merge that first.

`RasterOverlayTileProvider` has two responsibilities.

The first is that, via its derived classes, it knows how to obtain an image that covers a rectangle. This is the `loadTileImage` abstract method. It also has related `getProjection` and `getCoverageRectangle` that tell the raster overlay system what projection its images will be in, and the maximum total rectangle they might cover.

The second responsibility is that it manages the lifecycle of `RasterOverlayTile` instances. Its `loadTile` and `loadTileThrottled` methods are used to kick off loading of a particular overlay tile. It keeps track of the number of tiles currently loading and handles throttling. It tracks the total number of bytes used by the overlay's loaded data.

While working on automatic token refresh for raster overlays (e.g., Bing Maps key expires after ~1 hour, re-request the Cesium ion `/asset/2/endpoint` service to get a new one), I found that it would be very convenient to be able to use the good old GoF [Decorator pattern](https://wiki.c2.com/?DecoratorPattern) with `RasterOverlayTileProvider`. This would allow us to wrap the tile provider created by the `BingMapsRasterOverlay` with one that knows how to interact with Cesium ion. The decorating tile provider would call the underlying `loadTileImage` and add some extra handling for 401 errors when the token expires.

Unfortunately, the poor, dual-responsibility design of `RasterOverlayTileProvider` made this difficult. It's easy enough to wrap `loadTileImage`, but that second responsibility gets in the way. Which `RasterOverlayTileProvider` instance should track the throttled loads in progress? The Bing one, or the wrapping Cesium ion one? There's no good answer.

So this PR refactors the raster overlay system to separate the two responsibilities. In this PR, `RasterOverlayTileProvider` _exclusively_ has the first responsibility. This means that our existing tile provider implementations don't need to change, nor will any that may have been developed by our users.

The second responsibility is now found in a new class called `ActivatedRasterOverlay`. Call `activate` on a `RasterOverlay` and it will synchronously create an `ActivatedRasterOverlay` which internally and asynchronously creates the `RasterOverlayTileProvider` that will be used to get the image data. `RasterOverlayTile` now points back to the `ActivatedRasterOverlay` that created it (instead of to the `RasterOverlayTileProvider`), and the `ActivatedRasterOverlay` manages all the tile lifecycle and accounting.

Conceptually it's pretty straightforward, but it involved a lot of small changes throughout the raster overlay system, including some breaking changes. It's unlikely that users will run into any of the breakage, though.

This PR also adds a `CesiumUtility::TransformIterator` which has a similar purpose to Boost's [transform_iterator](https://www.boost.org/doc/libs/1_89_0/libs/iterator/doc/transform_iterator.html) (but the implementations aren't all that similar).

### DIFF
--- a/CesiumRasterOverlays/include/CesiumRasterOverlays/RasterOverlay.h
+++ b/CesiumRasterOverlays/include/CesiumRasterOverlays/RasterOverlay.h
@@ -191,9 +191,10 @@ public:
   /**
    * @brief Activates this overlay.
    *
-   * This method is called by a @ref Cesium3DTilesSelection::RasterOverlayCollection
-   * when an overlay is added to it. The returned @ref ActivatedRasterOverlay
-   * is used by the collection to manage the overlay.
+   * This method is called by a @ref
+   * Cesium3DTilesSelection::RasterOverlayCollection when an overlay is added to
+   * it. The returned @ref ActivatedRasterOverlay is used by the collection to
+   * manage the overlay.
    *
    * @param externals The external interfaces for use by the raster overlay.
    * @param ellipsoid The @ref CesiumGeospatial::Ellipsoid.


CesiumRasterOverlays/include/CesiumRasterOverlays/RasterOverlay.h
```diff
@@ -186,19 +188,32 @@ class RasterOverlay
     return this->_credits;
   }
 
+  /**
+   * @brief Activates this overlay.
+   *
+   * This method is called by a {@link Cesium3DTilesSelection::RasterOverlayCollection}
```
j9liu: I know a pass for `@link -> @ref` might be noisy, but we should at least do it for anything new we're adding so that we're consistent with our style guide. (I can do this)

CesiumRasterOverlays/include/CesiumRasterOverlays/RasterOverlayTile.h
```diff
@@ -245,17 +244,24 @@ class RasterOverlayTile final
...
-private:
-  friend class RasterOverlayTileProvider;
-
+  /**
+   * @brief Sets the load state of this tile.
+   *
+   * This function is not supposed to be called by clients.
+   *
+   * @private
+   */
```
j9liu, line 253: What's the reason for this being brought out of `private:`?

CesiumRasterOverlays/include/CesiumRasterOverlays/ActivatedRasterOverlay.h
```diff
@@ -0,0 +1,300 @@
...
+   * @brief Returns the number of tiles that are currently loading.
+   */
+  uint32_t getNumberOfTilesLoading() const noexcept;
+
+  /**
+   * @brief Removes a no-longer-referenced tile from this provider's cache and
+   * deletes it.
+   *
+   * This function is not supposed to be called by client. Calling this method
+   * in a tile with a reference count greater than 0 will result in undefined
```
j9liu: **NIT:**

```suggestion
   * on a tile with a reference count greater than 0 will result in undefined
```

CesiumRasterOverlays/include/CesiumRasterOverlays/ActivatedRasterOverlay.h
```diff
@@ -0,0 +1,300 @@
...
+
+  /**
+   * @brief Gets the tile provider created for this activated overlay. This will
+   * be `nullptr` before `getReadyEvent` resolves.
+   */
+  const CesiumRasterOverlays::RasterOverlayTileProvider*
+  getTileProvider() const noexcept;
+
+  /** @copydoc getTileProvider */
+  CesiumRasterOverlays::RasterOverlayTileProvider* getTileProvider() noexcept;
```
j9liu, line 129: Should be `const` since the function itself does not modify `ActivatedRasterOverlay`, even if the provider might later be modified?

```suggestion
  CesiumRasterOverlays::RasterOverlayTileProvider* getTileProvider() const noexcept;
```

CesiumRasterOverlays/include/CesiumRasterOverlays/ActivatedRasterOverlay.h
```diff
@@ -0,0 +1,300 @@
...
+   *
+   * The placeholder may be used prior to {@link getReadyEvent} resolving,
+   * but it will create placeholder tiles only.
+   */
+  const CesiumRasterOverlays::RasterOverlayTileProvider*
+  getPlaceholderTileProvider() const noexcept;
+
+  /** @copydoc getPlaceholderTileProvider */
+  CesiumRasterOverlays::RasterOverlayTileProvider*
+  getPlaceholderTileProvider() noexcept;
```
j9liu, line 157: Same comment about `const`?

```suggestion
  CesiumRasterOverlays::RasterOverlayTileProvider*
  getPlaceholderTileProvider() const noexcept;
```

CesiumRasterOverlays/include/CesiumRasterOverlays/ActivatedRasterOverlay.h
```diff
@@ -0,0 +1,300 @@
...
+
+  /**
+   * @brief Gets the placeholder tile created by the
+   * {@link getPlaceholderTileProvider}.
+   */
+  const CesiumRasterOverlays::RasterOverlayTile*
+  getPlaceholderTile() const noexcept;
+
+  /** @copydoc getPlaceholderTile */
+  CesiumRasterOverlays::RasterOverlayTile* getPlaceholderTile() noexcept;
```
j9liu, line 167:
```suggestion
  CesiumRasterOverlays::RasterOverlayTile* getPlaceholderTile() const noexcept;
```

CesiumUtility/include/CesiumUtility/TransformIterator.h
```diff
@@ -0,0 +1,124 @@
...
+#include <type_traits>
+
+namespace CesiumUtility {
+
+/**
+ * @brief An iterator that wraps another iterator and applies a transformation
+ * function to each element as it is accessed.
+ */
+template <typename TTransformFunction, typename TIterator>
+class TransformIterator {
```
j9liu, line 13: Is it worth adding unit tests for this or nah?

CesiumRasterOverlays/src/RasterOverlay.cpp
```diff
@@ -72,15 +87,70 @@ RasterOverlay::getAsyncDestructionCompleteEvent(
...
+          externals.asyncSystem,
+          externals.pAssetAccessor,
+          externals.pCreditSystem,
+          externals.pPrepareRendererResources,
+          externals.pLogger,
+          this);
+
+  // This continuation, by capturing pResult, keeps the instance from being
+  // destroyed. But it does not keep the RasterOverlayCollection itself alive.
+  std::move(future)
```
j9liu, line 107: What's the purpose of the `std::move(future)` here, since it's not being passed somewhere else explicitly?

Cesium3DTilesSelection/include/Cesium3DTilesSelection/RasterOverlayCollection.h
```diff
@@ -89,100 +126,93 @@ class CESIUM3DTILESSELECTION_API RasterOverlayCollection final {
...
+   * @param tilesetOptions The {@link TilesetOptions} for the tileset to which
+   * the tile belongs.
+   * @param tile The tile for which to add the overlays.
+   * @returns The list of projections required by the overlays that were added
+   * to the tile.
    */
-  CesiumRasterOverlays::RasterOverlayTileProvider* findTileProviderForOverlay(
-      CesiumRasterOverlays::RasterOverlay& overlay) noexcept;
+  std::vector<CesiumGeospatial::Projection>
+  addTileOverlays(const TilesetOptions& tilesetOptions, Tile& tile) noexcept;
```
j9liu: Nitpick, but I thought we generally put "options" at the end of the method parameters?

```suggestion
  addTileOverlays(Tile& tile, const TilesetOptions& tilesetOptions) noexcept;
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/RasterOverlayCollection.h
```diff
@@ -89,100 +126,93 @@ class CESIUM3DTILESSELECTION_API RasterOverlayCollection final {
...
+   * @param tilesetOptions The {@link TilesetOptions} for the tileset to which
+   * the tile belongs.
+   * @param tile The tile for which to update the overlays.
+   * @returns Details of the raster overlays attached to this tile.
    */
-  CesiumRasterOverlays::RasterOverlayTileProvider*
-  findPlaceholderTileProviderForOverlay(
-      CesiumRasterOverlays::RasterOverlay& overlay) noexcept;
+  TileRasterOverlayStatus
+  updateTileOverlays(const TilesetOptions& tilesetOptions, Tile& tile) noexcept;
```
j9liu:
```suggestion
  updateTileOverlays(Tile& tile, const TilesetOptions& tilesetOptions) noexcept;
```

Cesium3DTilesSelection/src/RasterOverlayCollection.cpp
```diff
@@ -32,34 +24,20 @@ using namespace CesiumUtility;
...
 RasterOverlayCollection::~RasterOverlayCollection() noexcept {
-  if (this->_pOverlays) {
-    OverlayList& list = *this->_pOverlays;
-    if (!list.overlays.empty()) {
-      for (int64_t i = static_cast<int64_t>(list.overlays.size() - 1); i >= 0;
-           --i) {
-        this->remove(list.overlays[static_cast<size_t>(i)].get());
-      }
-    }
+  for (int64_t i = static_cast<int64_t>(this->_activatedOverlays.size() - 1);
```
j9liu: This may be unnecessary at the end of the day, but I think it's safer to cast `size()` before subtracting to avoid arithmetic overflow?

```suggestion
  for (int64_t i = static_cast<int64_t>(this->_activatedOverlays.size()) - 1;
```

Cesium3DTilesSelection/src/RasterOverlayCollection.cpp
```diff
@@ -205,133 +142,121 @@ void RasterOverlayCollection::remove(
...
-
-  for (size_t i = 0; i < overlays.size() && i < tileProviders.size(); ++i) {
-    if (overlays[i].get() == &overlay)
-      return tileProviders[i].get();
+std::vector<CesiumGeospatial::Projection>
+RasterOverlayCollection::addTileOverlays(
+    const TilesetOptions& tilesetOptions,
+    Tile& tile) noexcept {
+  // when tile fails temporarily, it may still have mapped raster tiles, so
+  // clear it here
```
j9liu: **NIT**:

```suggestion
  // When a tile temporarily fails to load, it may still
  // have mapped raster tiles, so clear them here
```

j9liu (commented):
Thanks @kring ! I had a couple questions / suggestions but hopefully nothing too major.

CesiumRasterOverlays/include/CesiumRasterOverlays/RasterOverlay.h
kring: I'm trying to remember why we added this to the style guide. Originally, we used `{@link Whatever}` because this is jsdoc style, which is what CesiumJS uses and it works in Doxygen, too. My thought process at the time was that it would make it easier to move between the code bases, and minimize the changes necessary when copying/pasting doc comments between them (that was more common in the early days of cesium-native).

But if we're not going to do jsdoc style anymore, why `@ref` versus the normal Doxygen `\ref` syntax?

kring: To answer my own question, here's the discussion on it: https://github.com/CesiumGS/cesium-native/pull/1062#discussion_r1917208089

kring: It wouldn't be possible to call it otherwise. Previously, this method wasn't intended to be called from outside the class. It was only called from the `loadTile` / `loadTileThrottled` public methods also found on `RasterOverlayTileProvider`. But now, those methods don't exist. So if this method weren't public, only friends would be able to use a `RasterOverlayTileProvider` at all.

Also, it makes perfect conceptual sense for this method to be part of the public API, given the purpose of this class.

CesiumRasterOverlays/include/CesiumRasterOverlays/RasterOverlayTile.h
kring: It wouldn't be possible to call it otherwise. Previously, this method wasn't intended to be called from outside the class. It was only called from the `loadTile` / `loadTileThrottled` public methods also found on `RasterOverlayTileProvider`. But now, those methods don't exist. So if this method weren't public, only friends would be able to use a `RasterOverlayTileProvider` at all.

Also, it makes perfect conceptual sense for this method to be part of the public API, given the purpose of this class.

CesiumRasterOverlays/include/CesiumRasterOverlays/ActivatedRasterOverlay.h
kring: There's a const overload right above this. Unless you're suggesting we should _only_ have the const version, and it should return a non-const pointer? There are exceptions sometimes, but by default, a const method should return its field as const. I can't think of any particular reason this should be an exception.

CesiumRasterOverlays/include/CesiumRasterOverlays/ActivatedRasterOverlay.h
kring: Same situation here.

CesiumUtility/include/CesiumUtility/TransformIterator.h
kring: Good idea, I'll add some.

CesiumRasterOverlays/src/RasterOverlay.cpp
kring: It's required. `catchInMainThread` is declared like this:

```
template <typename Func> Future<T> catchInMainThread(Func&& f) && { ... }
```

That `&&` at the end means it may only be called on an r-value reference. This is intentional, because `catchInMainThread` "consumes" the future you call it on.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/RasterOverlayCollection.h
kring: Hmm this seems "right" to me, though I'm struggling to explain why.  In the other cases where it's last, that's usually necessary because we want to give it a default value. Here there is flexibility, and my instinct is to put the "most varying" thing last. I think maybe this comes from functional programming where it's a good habit in order to [support currying](https://stackoverflow.com/questions/5863128/ordering-of-parameters-to-make-use-of-currying). But that's not really a relevant argument here, so I guess putting `TilesetOptions` last is not meaningfully worse and perhaps better for the consistency with constructors. I'll change it.

Cesium3DTilesSelection/src/RasterOverlayCollection.cpp
kring: Definitely! Good call.

kring:
Thanks for the review @j9liu! I think I've addressed everything.

CesiumRasterOverlays/include/CesiumRasterOverlays/ActivatedRasterOverlay.h
j9liu: Hm okay, maybe I'm being too pedantic with things. But a function that is const means that calling the function won't modify the object it's being called on, right? So even though the function is returning a non-const pointer, the function itself does not modify ActivatedRasterOverlay. That's why I'd think `Thing* getThing() const;` is technically correct.

I could see the interpretation, though, that after the non-const pointer is returned, it's possible for the RasterOverlayTileProvider to be modified, which would result in the modification of the ActivatedRasterOverlay. So not using const here can more semantically reflect that... is that what you're getting at?

kring: Right, if I have a const reference to an object, that _mostly_ (again, there are exceptions) means that I shouldn't be allowed to make any sort of modification to that object. If a method on that object returned some other piece of itself by non-const reference, then that would allow the const reference to be used to modify the object (or at least that piece of it). Now, is the RasterOverlayTileProvider "part of" the ActivatedRasterOverlay? Or is it fine for the `RasterOverlayTileProvider` to be modified even when the `ActivatedRasterOverlay` shouldn't be? Assuming they are completely independent objects, not a part/whole, is it a good idea for someone with a const reference to one to be able to obtain a non-const reference to the other?

I don't know! I'm not even sure when this would come up. But "const methods return their fields by const" is a very normal pattern that isn't going to surprise anyone. It's the safer choice. It's the default we should start with until we have a reason to do differently.

It's a bit snarky, but there are some useful insights in here:
https://isocpp.org/wiki/faq/const-correctness#const-member-fns

PR #1247, opened 2025-09-17
Optimize culling volume setting

jiangheng90:
fix calculation for potential precision lose


jiangheng90:
close in favor of this method has been deprecated

PR #1252, opened 2025-09-22
Improve raster overlay token refresh by treating `ExternalAssetEndpoint` as a `SharedAsset`

### LINKED ISSUE #1230: Caching implementation can often result in use of extra Bing Maps sessions
`IonRasterOverlay` has a caching mechanism to avoid making extra, unnecessary requests to the `/assets/[assetID]/endpoint` service during a single run of an application. This is important for assets like Bing Maps because each request to this service counts as a new session for Cesium ion account limit purposes. It works well for things like switching levels or repeatedly entering Play-in-Editor mode. A Bing Maps session will be counted when the level is first loaded, but then users can switch levels and start Play-in-Editor as much as they want without using any additional sessions.

However, the caching mechanism, as currently implemented, works poorly when the same Cesium ion asset is accessed from multiple places nearly simultaneously. For example, if the level has both Cesium World Terrain and Cesium OSM Buildings, and both have a Bing Maps raster overlay, then the initial loading of this level will initiate two requests to `/assets/2/endpoint` almost simultaneously. Both requests will see a cache miss, and so both will do a real network request and two sessions will be used.

Fixing this will require detecting that the first request is already in flight when initiating the second one, and waiting for its response.

kring:
This PR reworks how raster overlays manage Cesium ion token refreshes. Previously, we detected 401 errors and initiated a refresh whenever we saw one. Now, we simply refresh the token every 55 minutes (the token expires after an hour).

This works by creating a "wrapping" `RasterOverlayTileProvider` with some extra logic to get the underlying Bing or TMS `RasterOverlayTileProvider` as a shared asset. The `SharedAssetDepot` and the new `DerivedValue` class guarantee that only one fresh happens at a time and everyone who needs it waits for it.

This builds on both #1238 and #1244, but I can only choose one as the base branch. So this diff will get smaller once #1238 is merged to main and main is merged to `raster-overlay-wrappability`.

Fixes #1230 

Fixes a problem that can theoretically happen when a raster overlay, like Bing Maps, is accessed via Cesium ion and returns a 401 error that can not be resolved with a Cesium ion token refresh. This situation could previously cause cesium-native to spam token refresh attempts, potentially running up the user's session count until it exceeds the quota.

Bonus improvements:

* Added `CesiumUtility::DerivedValue`, which is just a really simple cache of a single output of a function for as long as the input remains the same.
* `SharedAssetDepot` now allows a context object to be specified via template parameter. Previously, the "context" passed to the factory function consisted of an AsyncSystem and an IAssetAccessor and this wasn't changeable.
* Fixed a typo in the token refresh message, and changed `url` to `URL`.


azrogers (approved):

azrogers:
Code looks good, and I tested it out and it seems to correctly handle refreshing the token after 55 minutes. Thanks @kring!

PR #1255, opened 2025-09-26
Don't hardcode WGS84 ellipsoid in `TilesetHeightQuery`

### LINKED ISSUE #1236: sampleHeightMostDetailed only works on Earth
The problem is here:
https://github.com/CesiumGS/cesium-native/blob/v0.51.0/Cesium3DTilesSelection/src/TilesetHeightQuery.cpp#L104

Despite an `Ellipsoid` passed in, the code incorrectly uses `WGS84`.

j9liu:
Fixes #1236.


kring:
Thanks @j9liu!

Issue #1249, opened 2025-09-18
Cartographic polygon artifacts with self-intersection polygons


kring:
This might be firmly in the "well, don't do that" category. But cartographic polygons with polygons with self intersections do work remarkably well:

<img width="881" height="870" alt="Image" src="https://github.com/user-attachments/assets/56d7d53d-4802-4a3c-af3f-71f7dd191f7a" />

Except when you zoom in close, some tiles - perhaps upsampled ones - seem to mysteriously be missing completely. This is even though "Exclude Tiles" is off on the raster overlay,

<img width="1912" height="1176" alt="Image" src="https://github.com/user-attachments/assets/8c0c0975-caba-4863-8fc5-03575f358f46" />

This is on Google Photorealistic 3D Tiles, in case it matters.

PR #1256, opened 2025-09-30
Fix GeoJSON rasterization issues

azrogers:
Issues have been reported about incorrect rasterization of certain documents:

- The antimeridian is not handled correctly, [as reported here](https://community.cesium.com/t/errors-about-geojsondocumentrasteroverlay/43478). 
- Complex polygons with holes do not render correctly - reported in the same thread as above, in CesiumGS/cesium-unreal#1721, and possibly related to #1249.

Antimeridian issues were a simple fix - one step in the rasterization involves transforming the cartographic coordinates into pixel coordinates in the tile being rendered. First, the coordinates were normalized to the bounds of the tile, then the normalized coordinates were multiplied by the pixel dimensions. The problem here is that normalizing longitude cannot just be done with a simple (x - min) / (max - min), and needs to be antimeridian-aware. This has been fixed.

The issue with polygons with holes is a bit more puzzling. I've tried it every which way with different winding orders, but there's always an artifact when using polygons with multiple rings. It always looks like this:
polygon-holes-outline-wrong
When it's supposed to look like this:
polygon-holes-outline

This causes some significant issues with more complex polygons, as seen in the posts linked above. Ultimately, it was @yuxueliyuxl who provided a fix in [this post](https://github.com/CesiumGS/cesium-unreal/issues/1721#issuecomment-3216207871). The fix uses a path to replicate the same behavior as a polygon, which seems to avoid whatever strange issues there are with the polygon. I would be more confident in this solution if I knew *why* the polygon rendering was messing up - but since the fact you can use polygons with multiple holes *at all* is something not at all mentioned in the documentation of Blend2D (I had to ask them directly), I'm very willing to believe it's an oversight.

Anyways, this means that we'll want to get a CLA signed by @yuxueliyuxl before merging it, as it's his code that provided the fix.


azrogers:
@kring The issue with the zero-height rectangle I mentioned - here's the GeoJSON data I was testing with:
```json
{
  "type": "LineString",
  "coordinates": [
    [
      -175,
      30
    ],
    [
      -170,
      30
    ]
  ]
}
```

This data causes [this branch](https://github.com/CesiumGS/cesium-native/blob/aa5646dbf58d6cf8164b2e7a5e09c120184beba1/CesiumRasterOverlays/src/GeoJsonDocumentRasterOverlay.cpp#L5710) to fail because `computeIntersection` has [this check](https://github.com/CesiumGS/cesium-native/blob/aa5646dbf58d6cf8164b2e7a5e09c120184beba1/CesiumGeospatial/src/GlobeRectangle.cpp#L106) that returns `nullopt` if the rectangle is zero height.

kring:
@azrogers I think where we're running into trouble is that we're treating lines as if they have zero area when computing their bounding rectangle, but, when rendered, they very clearly don't have zero area. If they truly had no width, there would be no need to draw them at all, because they'd be invisible.

Of course, the actual width of a line on the globe is a big hard to pin down when we're drawing fixed-pixel-width lines that vary in real-world width based on the eventual view. I think that kind of needs to be figured out - at least approximately - for correct rendering, though. Imagine we have a line that is near the northern border of a tile, drawn with a wide pixel width such that some of its pixels lie in the tile to the north. If we didn't draw the line in the northern tile, too, then the line would appear too narrow.

Perhaps a simple buffering (offsetting) operation would work? We have the (degenerate in this case) rectangle for the dataset. And we know we're drawing the lines, say, 10 pixels wide. To determine if the dataset overlaps a given tile, we take the dataset's (degenerate?) rectangle, and increase its size in all directions by the spatial size of 10 pixels in this particular tile's raster overlay. Then test if that (no longer degenerate) rectangle overlaps the tile's rectangle.

azrogers:
@IndigoCloud6 Thank you for noticing this! That's definitely an oversight on my part, I've re-added the `seedForObject` bits.

@kring I've done my best to implement some sort of solution like what you've recommended, but it's complicated by geometry rectangles being pre-computed, and only the combination of all geometry rectangles being used to test against the tile rectangle. This means:
- For line width that needs to update based on zoom level, the best we can do is stretch the full tree's rectangle by the max line width of all of its children. This works but results in (possibly significant) wasted rasterization calls in all but the worst cases.
- We can't recompute which quadtree nodes each piece of geometry is in every time the zoom level changes, so the quadtree will be generated with rectangles based only on the vertices of the geometry, not the line width. Besides the empty rectangle issue, this results in geometry not being rendered for tiles where its line width overlaps, because it's not in the quadtree node.

I've pushed the fixes I have so far regardless.

CesiumGeospatial/src/GlobeRectangle.cpp
```diff
@@ -142,6 +143,22 @@ GlobeRectangle::computeUnion(const GlobeRectangle& other) const noexcept {
       glm::max(this->_north, other._north));
 }
 
+glm::dvec2 GlobeRectangle::computeNormalizedCoordinates(
+    const Cartographic& cartographic) const noexcept {
+  double east = this->_east;
+  double cartoLong = cartographic.longitude;
+  if (east < this->_west) {
+    east += CesiumUtility::Math::TwoPi;
+    if (cartographic.longitude < 0) {
```
kring: This logic isn't quite right. If the west=179, east=10, and longitude=5 (all degrees), it's necessary to add 180 degrees to longitude in order to get the right answer, even though it's greater than 0.

I think you want to add 180 degrees to longitude whenever it is less than west.

CesiumGeospatial/include/CesiumGeospatial/GlobeRectangle.h
```diff
@@ -259,6 +261,17 @@ class CESIUMGEOSPATIAL_API GlobeRectangle final {
    */
   GlobeRectangle computeUnion(const GlobeRectangle& other) const noexcept;
 
+  /**
+   * @brief Computes the normalized position of the given cartographic
+   * coordinate on this globe rectangle, with (0.0, 0.0) representing southwest
+   * and (1.0, 1.0) representing northeast.
```
kring: It's worth mentioning that it interpolates linearly in longitude/latitude space. Not in ECEF or some map projection. Probably borderline obvious, but still.

CesiumRasterOverlays/src/GeoJsonDocumentRasterOverlay.cpp
```diff
@@ -159,36 +170,46 @@ struct GlobeRectangleFromObjectVisitor {
...
-          Cartographic::fromDegrees(point.x, point.y));
+  void visitWithLineWidth(
+      const std::vector<glm::dvec3>& coordinates,
+      const std::optional<LineStyle>& lineStyle) {
+    if (lineStyle && lineStyle->widthMode == LineWidthMode::Meters) {
+      const double halfWidth =
+          (lineStyle->width / ellipsoid.getRadii().x) / 2.0;
+      for (const glm::dvec3& point : coordinates) {
+        // Include the width of the line in the bounding region calculation
+        builder.expandToIncludePosition(Cartographic::fromDegrees(
```
kring: Maybe you already did, but it's worth testing that this logic still works correctly when adding +/- halfWidth causes an antimeridian wrap. I'm actually not sure if `expandToIncludePosition` will handle that correctly or not.

CHANGES.md
```diff
@@ -9,6 +9,9 @@
 ##### Fixes :wrench:
 
 - Fixed a bug in `GoogleMapTilesRasterOverlay` that tried to parse credits from an erroneous viewport service response.
+- Fixed issues with GeoJsonRasterOverlay with certain types of data.
+  - Polygons with holes hould now display correctly.
```
kring:
```suggestion
  - Polygons with holes should now display correctly.
```

kring:
I've reverted [this commit](https://github.com/CesiumGS/cesium-native/pull/1256/commits/55bbae20fe3de9a389c510aff428699494018ab1) in order to get this into today's release.

kring:
Thanks @azrogers!

yuxueliyuxl:
I'm not sure if the latest version fixes the antimeridian issue？ I use V2.18
A2222
A11111

PR #1258, opened 2025-10-07
Don't parse non-object credits from Google 2D Maps

j9liu:
This PR addresses a bug that @david-lively discovered https://github.com/CesiumGS/cesium-unity/pull/612/. Unfortunately I didn't discover it while testing because it was an assertion failure that only triggered in Debug builds.

Google's Roadmap layer consistently returns an Error 500 for one of its tiles. The credit-loading logic didn't account for the fact that an empty JSON document gets returned for this tile, and it tried to access properties on this non-object document. This solves it by simply checking whether the document is a valid object.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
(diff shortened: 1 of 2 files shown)


david-lively (approved):
LGTM.

PR #1262, opened 2025-10-14
Implement support for `KHR_gaussian_splatting` and `KHR_gaussian_splatting_compression_spz_2`

### LINKED ISSUE #1219: Implement `KHR_spz_compression`
While I've already implemented the `KHR_gaussian_splatting` extension in the [gaussian-splats](https://github.com/CesiumGS/cesium-native/tree/gaussian-splats) branch, the `KHR_spz_compression` extension - [documented here](https://github.com/CesiumGS/glTF/tree/spz-extension/extensions/2.0/Khronos/KHR_spz_compression) - will require more work to implement. In particular, we will need to integrate the [spz library](https://github.com/nianticlabs/spz) into the glTF decoding process.

azrogers:
Closes #1219. This PR adds support for the new `KHR_gaussian_splatting` extension and the `KHR_gaussian_splatting_compression_spz_2` extension. The latter is the majority of this PR - it implements SPZ decoding using the [spz library](https://github.com/nianticlabs/spz), allowing glTFs with SPZ payloads to be read. It also implements backwards compatibility for the `KHR_spz_gaussian_splats_compression` and `KHR_gaussian_splatting_compression_spz` extensions, previous iterations that have already had some tilesets created using them.

### DIFF
--- a/CesiumGltfReader/src/decodeSpz.cpp
+++ b/CesiumGltfReader/src/decodeSpz.cpp
@@ -1,7 +1,5 @@
 #include "decodeSpz.h"
 
-#include "CesiumGltf/PropertyType.h"
-
 #include <CesiumGltf/Accessor.h>
 #include <CesiumGltf/Buffer.h>
 #include <CesiumGltf/BufferView.h>
@@ -10,6 +8,7 @@
 #include <CesiumGltf/Mesh.h>
 #include <CesiumGltf/MeshPrimitive.h>
 #include <CesiumGltf/Model.h>
+#include <CesiumGltf/PropertyType.h>
 #include <CesiumGltfReader/GltfReader.h>
 #include <CesiumUtility/Assert.h>
 #include <CesiumUtility/JsonValue.h>


tools/generate-classes/package.json
```diff
@@ -5,7 +5,7 @@
...
-    "generate-gltf": "node index.js --schemas https://raw.githubusercontent.com/CesiumGS/glTF/cesium-native/specification/2.0/schema/glTF.schema.json --output ../../CesiumGltf --readerOutput ../../CesiumGltfReader --writerOutput ../../CesiumGltfWriter --extensions https://raw.githubusercontent.com/CesiumGS/glTF/cesium-native/extensions/2.0/ --namespace CesiumGltf --readerNamespace CesiumGltfReader --writerNamespace CesiumGltfWriter --config glTF.json",
+    "generate-gltf": "node index.js --schemas https://raw.githubusercontent.com/CesiumGS/glTF/cesium-native/specification/2.0/schema/glTF.schema.json --output ../../CesiumGltf --readerOutput ../../CesiumGltfReader --writerOutput ../../CesiumGltfWriter --extensions https://raw.githubusercontent.com/CesiumGS/glTF/cesium-native/extensions/2.0/ https://raw.githubusercontent.com/CesiumGS/glTF/draft-splat-spz/extensions/2.0/ --namespace CesiumGltf --readerNamespace CesiumGltfReader --writerNamespace CesiumGltfWriter --config glTF.json",
```
kring: Please open a PR to merge the draft extension into the cesium-native branch, rather than pulling it from a separate branch. Or maybe Adam already did?
https://github.com/CesiumGS/glTF/pull/88

kring: Looks like it needs an [update](https://github.com/CesiumGS/glTF/compare/cesium-native...draft-splat-spz), but you can open another similar PR.

extern/vcpkg/ports/spz/loadSpz.patch
```diff
@@ -0,0 +1,29 @@
+diff --git a/src/cc/load-spz.cc b/src/cc/load-spz.cc
+index e028445..7388d56 100644
+--- a/src/cc/load-spz.cc
++++ b/src/cc/load-spz.cc
+@@ -635,6 +635,10 @@ GaussianCloud loadSpz(const std::vector<uint8_t> &data, const UnpackOptions &o)
+   return unpackGaussians(loadSpzPacked(data), o);
+ }
+ 
++GaussianCloud loadSpz(const uint8_t *data, int32_t size, const UnpackOptions &o) {
```
kring: Let's open a PR upstream for this. It seems like a nice change that should be easy to get accepted, and then we (eventually) won't need an overlay port. spz is in vcpkg now, btw.

azrogers: PR made: https://github.com/nianticlabs/spz/pull/62

tools/generate-classes/package.json
kring: I thought I'd clarify why this is important, because it may not be obvious...

If we directly pull the extension from the branch where it's being worked, as is currently done, then every time we regenerate the glTF classes we'll get the latest updates from that branch. Some of these could break our code, and that would be pretty annoying for someone trying to work on a _different_ glTF extension. By always generating from the `cesium-native` branch, we keep control of when we take upstream updates.

CesiumGltfReader/src/decodeSpz.cpp
```diff
@@ -0,0 +1,483 @@
...
+namespace CesiumGltfReader {
+namespace {
+
+const float SH_C0 = 0.282095f;
+
+const std::string ALTERNATE_EXT_NAME1 = "KHR_spz_gaussian_splats_compression";
+const std::string ALTERNATE_EXT_NAME2 =
+    "KHR_gaussian_splatting_compression_spz";
+
+std::unique_ptr<spz::GaussianCloud> decodeBufferViewToGaussianCloud(
```
kring: I don't think there's much reason to return this as a unique_ptr, just return the spz::GaussianCloud directly. It should be more efficient that way because it avoids the heap allocation. And with return value optimization, it doesn't even require a move!

CesiumGltfReader/src/decodeSpz.cpp
```diff
@@ -0,0 +1,483 @@
...
+}
+
+void decodePrimitive(
+    GltfReaderResult& readGltf,
+    CesiumGltf::MeshPrimitive& primitive,
+    CesiumGltf::ExtensionKhrGaussianSplattingCompressionSpz2& spz) {
+  CESIUM_TRACE("CesiumGltfReader::decodePrimitive");
+  CESIUM_ASSERT(readGltf.model);
+
+  // TODO: handle different accessor component types
```
kring: Does this need to be addressed? It should probably at least log a warning.

CesiumGltfReader/src/decodeSpz.cpp
```diff
@@ -0,0 +1,483 @@
...
+    }
+  }
+}
+
+CesiumGltf::ExtensionKhrGaussianSplattingCompressionSpz2*
+addExtensionFromJsonValue(
+    const std::string& extName,
+    CesiumGltfReader::GltfReaderResult& readGltf,
+    CesiumGltf::ExtensionKhrGaussianSplatting& splatting,
+    CesiumUtility::JsonValue* pKhrJson) {
```
kring: This isn't allowed to be nullptr, and isn't modified, so why not make it a const reference?

CesiumGltfReader/src/decodeSpz.cpp
```diff
@@ -0,0 +1,483 @@
...
+                       : static_cast<int32_t>(it->second.getUint64());
+  return &ext;
+}
+
+// Maps attribute names from older versions of the extension to the names from
+// the current version of the extension.
+void fixAttributeNames(CesiumGltf::MeshPrimitive& primitive) {
+  std::vector<std::string> attributesToConvert;
+  attributesToConvert.reserve(primitive.attributes.size());
+  for (std::pair<const std::string, int32_t>& attribute :
```
kring:
```suggestion
  for (const std::pair<const std::string, int32_t>& attribute :
```

azrogers:
@kring Took care of the review comments!

PR #1264, opened 2025-10-15
Add `AzureMapsRasterOverlay`

### LINKED ISSUE #1159: Add support for Azure Maps raster overlays
Bing Maps is deprecated. Users using it via Cesium ion will still be able to access it through 2028, but people that were using it directly are running out of time. Adding an `AzureMapsRasterOverlay` should be very straightforward and would be helpful to these users.

j9liu:
Closes #1159.

This PR adds `AzureMapsRasterOverlay` to stream 2D maps from the Azure Maps API. The implementation is a blend between `BingMapsRasterOverlay` and `GoogleMapTilesRasterOverlay`.

It also fixes a bug with on-screen credits in `GoogleMapTilesRasterOverlay`. There is some light refactoring as well to accommodate common raster overlay code.

References:

- [Get Map Tileset API](https://learn.microsoft.com/en-us/rest/api/maps/render/get-map-tileset?view=rest-maps-2025-01-01&tabs=HTTP)
- [Get Map Tile API](https://learn.microsoft.com/en-us/rest/api/maps/render/get-map-tile?view=rest-maps-2025-01-01&tabs=HTTP)
- [Get Map Attribution API](https://learn.microsoft.com/en-us/rest/api/maps/render/get-map-attribution?view=rest-maps-2025-01-01&tabs=HTTP)

### DIFF
--- a/CesiumRasterOverlays/src/AzureMapsRasterOverlay.cpp
+++ b/CesiumRasterOverlays/src/AzureMapsRasterOverlay.cpp
@@ -227,7 +227,7 @@ AzureMapsRasterOverlay::createTileProvider(
        pCreditSystem,
        pPrepareRendererResources,
        pLogger,
-       &sessionParameters = this->_sessionParameters](
+       sessionParameters = this->_sessionParameters](
           const std::shared_ptr<IAssetRequest>& pRequest,
           const std::span<const std::byte>& data) -> CreateTileProviderResult {
     JsonObjectJsonHandler handler{};
--- a/CesiumRasterOverlays/src/GoogleMapTilesRasterOverlay.cpp
+++ b/CesiumRasterOverlays/src/GoogleMapTilesRasterOverlay.cpp
@@ -337,13 +337,14 @@ GoogleMapTilesRasterOverlay::createNewSession(
           {{"Content-Type", "application/json"}},
           requestPayloadBytes)
       .thenInMainThread(
-          [this,
-           asyncSystem,
+          [asyncSystem,
            pAssetAccessor,
            pCreditSystem,
            pPrepareRendererResources,
            pLogger,
-           pOwner](std::shared_ptr<IAssetRequest>&& pRequest)
+           pOwner,
+           newSessionParameters = this->_newSessionParameters](
+              std::shared_ptr<IAssetRequest>&& pRequest)
               -> Future<CreateTileProviderResult> {
             const IAssetResponse* pResponse = pRequest->response();
             if (!pResponse) {
@@ -450,9 +451,9 @@ GoogleMapTilesRasterOverlay::createNewSession(
                     pCreditSystem,
                     pPrepareRendererResources,
                     pLogger,
-                    this->_newSessionParameters->apiBaseUrl,
+                    newSessionParameters->apiBaseUrl,
                     session,
-                    this->_newSessionParameters->key,
+                    newSessionParameters->key,
                     maximumZoomLevel,
                     static_cast<uint32_t>(tileWidth),
                     static_cast<uint32_t>(tileHeight),


CesiumUtility/include/CesiumUtility/Uri.h
```diff
@@ -356,6 +356,14 @@ class Uri final {
   static std::string
   setPath(const std::string& uri, const std::string& newPath);
 
+  /**
+   * @brief Ensures that the given string URL ends with a slash, modifying
+   * in-place if necessary. Useful when the string is used as a base URL.
+   *
+   * @param url The URL to which to append a trailing slash.
+   */
+  static void ensureTrailingSlash(std::string& url);
```
kring: This class has a lot of static methods, perhaps for largely historical reasons. But I think this one is much better as an instance method on `Uri`. After all, the very first thing it does is construct a `Uri` instance!

CesiumUtility/include/CesiumUtility/Uri.h
```diff
@@ -356,6 +356,14 @@ class Uri final {
   static std::string
   setPath(const std::string& uri, const std::string& newPath);
 
+  /**
+   * @brief Ensures that the given string URL ends with a slash, modifying
```
kring:
```suggestion
   * @brief Ensures that the given string URL's path ends with a slash, modifying
```

CesiumUtility/test/TestUri.cpp
```diff
@@ -294,4 +294,14 @@ TEST_CASE("Uri::getExtension") {
   CHECK(Uri("http://example.com/a/b/c/.hidden").getExtension() == "");
   CHECK(Uri("http://example.com/a/b/c/.").getExtension() == "");
   CHECK(Uri("http://example.com/a/b/c/..").getExtension() == "");
-}
\ No newline at end of file
+}
+
+TEST_CASE("Uri::ensureTrailingSlash") {
+  std::string url = "noslash.com";
```
kring: Perhaps a bit of a nitpick, but it's best to use example.com instead of URLs that are - or might someday be - owned by someone.

CesiumUtility/test/TestUri.cpp
```diff
@@ -294,4 +294,14 @@ TEST_CASE("Uri::getExtension") {
...
+  Uri::ensureTrailingSlash(url);
+  CHECK_EQ(url, "noslash.com/");
+
+  url = "slash.com/";
+  Uri::ensureTrailingSlash(url);
+  CHECK_EQ(url, "slash.com/");
```
kring: Should add some tests for complete URLs. Because these don't have a scheme, they're probably interpreted as relative, and `slash.com` ends up being the path rather than the domain/host. That won't break anything, but it's kind of misleading.

CesiumRasterOverlays/include/CesiumRasterOverlays/AzureMapsRasterOverlay.h
```diff
@@ -0,0 +1,239 @@
...
+  /**
+   * @brief Refresh a previously-created tile provider using a new key.
+   *
+   * Calling this method on a tile provider that was not created by this @ref
+   * AzureMapsRasterOverlay will lead to undefined behavior.
+   *
+   * @param pProvider The previously-created tile provider.
+   * @param newKey The new key to use.
+   */
+  CesiumAsync::Future<void> refreshTileProviderWithNewKey(
```
kring: This was part of the old token refresh system, and isn't needed anymore. I should have removed it from the existing tile providers.

CesiumUtility/test/TestUri.cpp
j9liu: I'll change them both to `http://www.example.com`, but let me know if you meant something else by this.

PR #1265, opened 2025-10-17
Clean up dependencies, especially zlib-ng.

kring:
This PR cleans up some weirdness I noticed while investigating the Android build failure in CesiumGS/cesium-unreal#1748.

We had our own custom find script for zlib-ng, even though it has its own. Perhaps we thought we needed it because we had the name of the import library wrong? Or perhaps the built-in one was added in a more recent vcpkg release? Either way, it's not needed anymore, so I removed it and fixed the library name (`zlib-ng::zlib-ng` -> `zlib-ng::zlib`).

Other changes:
* Removed a duplicate `find_package` for curl. Looks like a merge error.
* Fixed a typo.
* Sorted the find_package lines alphabetically.
* Updated `Config.cmake.in` with some recent third-party library changes. @timoore I'm not especially well equipped to test this one, so you should take a look and make sure I haven't broken something.

There's a `cesium-unreal` branch with the same name to demonstrate this still works ok in Unreal.


timoore (approved):
Looks good, a positive change.

Issue #1260, opened 2025-10-08
Add support for reporting credits per `TilesetViewGroup`


kring:
By calling `getSnapshot` on a `CreditSystem`, clients can get the current set of active credits, as well as the set of credits that were removed since the last snapshot. However, this only works across _all_ `TilesetViewGroup`s. There's currently no easy way to get this information per view group. This is unfortunate, because some clients would like to, for example, show separate credits in each viewport.

To support such clients, let's add a `getCreditsSnapshot` method to `TilesetViewGroup`. This method can walk the `_previousFrameCredits` and `_currentFrameCredits` collections to determine which credits are active, and which were removed in the current frame.

PR #1270, opened 2025-11-10
Update TilesetViewGroup to have information about their associated credits

ludovic-cheval-bentley-systems:
By calling getSnapshot on a CreditSystem, clients can retrieve the current set of active credits. However, this snapshot represents credits across all TilesetViewGroups.

Some clients need to display credits separately for each viewport. To support this, new isCreditReferenced methods have been added to both CreditReferencer and TilesetViewGroup. These allow clients to check whether credits obtained from the CreditSystem apply to a specific TilesetViewGroup.

Similarly, certain clients must distinguish credits originating from raster overlays from other credits. To enable this, a dedicated CreditReferencer has been introduced within TilesetViewGroup to reference raster overlay credits. The existing CreditReferencer continues to reference all credits associated with the TilesetViewGroup.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### v0.54.0 - 2025-11-17
+
+##### Additions :tada:
+
+- Cesium Native can now be built with Emscripten.
+
 ### v0.53.0 - 2025-11-03
 
 ##### Breaking Changes :mega:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -194,7 +194,7 @@ endif()
 include("cmake/defaults.cmake")
 
 project(cesium-native
-    VERSION 0.53.0
+    VERSION 0.54.0
     LANGUAGES CXX C
 )
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cesium-native",
-  "version": "0.53.0",
+  "version": "0.54.0",
   "description": "Cesium 3D Geospatial for C++",
   "main": "index.js",
   "directories": {


kring:
Thanks again for the PR @ludovic-cheval-bentley-systems. It looks good, but I have a couple of questions about use-cases.

1. Can you elaborate on why you need to separate credits from raster overlays versus "everything else"? I can see why it'd be useful to be able to display credits per dataset (whether a Tileset, raster overlay, or something else), but I'm less convinced it's useful to separate out the credits from all raster overlays.

2. In #1260, I suggested adding a `getCreditsSnapshot` method to `TilesetViewGroup`. Would that be a worse interface from your perspective, or was it just more difficult to implement?

kring:
Thanks @ludovic-cheval-bentley-systems!

> Since we can load 3D Tiles with different data sources (for example, Google Maps imagery applied on Cesium ION terrain), we need to distinguish these credits'sources for legal compliance and ensure we display which credit belongs to which source (instead of a soup of credentials from undistinguished sources).

That's not a legal requirement we've run into anywhere else, as far as I know. But if you do need to separate requirements by data source, then the implementation in this PR is insufficient anyway, because it doesn't allow you to separate credits that come from two different tilesets or from two different raster overlays. Perhaps you're getting away with it for now because you are only displaying one tileset and one raster overlay. I think that's very likely to change in the future, though.

Let me think about how it might be possible to provide credits both per view and per data source. It's not entirely straightforward to do it efficiently...

> However, that approach means looping through all TileSet credits for every TileSetViewGroup — and then looping again through each group’s credit snapshot in our application.

Ok, so it sounds like the interface in this PR is more convenient for your application, even if we put aside any difficulties in implementation getCreditSnapshot on the cesium-native side?

PR #1273, opened 2025-11-20
Add CreditSources to CreditSystem, refactor RasterOverlay::createTileProvider

### LINKED ISSUE #1260: Add support for reporting credits per `TilesetViewGroup`

kring:
This PR incorporates and builds on #1270.
CC @ludovic-cheval-bentley-systems

Fixes #1260

### CreditSources

**Before**: It's impossible to tell which Tileset / RasterOverlay / etc. produced a Credit.
**This PR**: You can ask the `CreditSystem` for a credit's `CreditSource` and compare it against the `CreditSource` returned by the `getCreditSource` method on a `Tileset` or `RasterOverlayTileProvider`.

**Before**: Identical HTML from two different sources is treated as a single Credit.
**This PR**: Identical HTML from two different sources is treated as two different Credits.

**Before**: Once a Credit is created with `CreditSystem::createCredit`, it is _never_ destroyed.
**This PR**: When a `CreditSystem` is destroyed, all `Credits` associated with it are invalidated. Attempting to use `CreditSystem::getHtml` or similar methods with an invalidated `Credit` will produce an error string.

#### Implementation

Credits are implemented using a "generational arena", which allows them to have stable identity without allocating them individually on the heap.

A Credit's identity is an index, rather than its location in memory. Allocating a new Credit is a simple matter of assigning it the next available index. A Credits can be de-allocated by marking its record unused. If someone still has a `Credit` with that now-deallocated index, we can _safely_ (unlike with a pointer!) recognize that it is no longer valid.

But that's only if every new Credit allocation gets a new index, which means our set of credit records could grow without bound if we're creating and destroying many different credits. So, to support deallocation, we can add a "free list", `CreditSystem::_unusedCreditRecords`. When a `Credit` is destroyed, its index is added to the list of unused records. Later, when we need a new one, we grab an existing one from the free list.

But this is vulnerable to what's known as the ABA problem. Initially, a particular Credit index refers to a credit with the HTML `Foo`. Later, Credit `Foo` is deallocated and its record is marked unused. If we look up that Credit index now, we'll correctly conclude that it is no longer valid.

But then we allocate a new Credit with the HTML `Bar`, and it is allocated at the index that previously held `Foo`. Now if we look up that old Credit that was supposed to point to `Foo`, we get `Bar` instead. Oops!

This is where the "generational" aspect comes in. Every Credit has not just an index but also a `generation`. Every Credit record has a `generation` as well. When a Credit is deallocated, in addition to marking it unused, we also increment its `generation`. This way, whenever we use the index in a Credit to look up its record, we can check that the Credit and the Record have matching generations. If not, we know the Credit refers to something that has been deallocated even if something new has been allocated in its place.

#### Creating CreditSources

For the most part, creating CreditSources is pretty straightforward. Each Tileset or RasterOverlayTileProvider creates its own. It can be stored in an embedded field or on the heap. The only requirement is that its pointer not change. Unlike `Credit`, `CreditSource` _does_ base its identity on its location in memory.

However, as usual, `IonRasterOverlay` makes things a bit more complicated. The `RasterOverlayTileProvider` created by `IonRasterOverlay` is actually a wrapped around an underlying TMS, Bing, Google, or Azure `RasterOverlayTileProvider`. Users need to be able to call `getCreditSource()` on the wrapper and get a credit source that matches the credits created by the underlying tile provider.

I can think of two ways this can be done:

1. Make `getCreditSource` virtual so that the `IonRasterOverlay` version can delegate to the underlying tile provider.
2. Allow the credit source to be passed into a `RasterOverlayTileProvider` at construction time so that the underlying tile provider can be forced to use the one created by the wrapper tile provider.

I've gone with (2), and that's the reason for the raster overlay refactoring in this PR.

### Raster Overlay Refactoring

The biggest change is that `RasterOverlay::createTileProvider` changed from taking a bunch of parameters to instead taking just one, of type `CreateRasterOverlayTileProviderParameters`. This allows us to pass more information to `createTileProvider` in the future without breaking changes.

I also changed the `RasterOverlayTileProvider` and `QuadtreeRasterOverlayTileProvider` constructors to match. This is a breaking change that required changing every raster overlay.

One subtle difference... Previously, `RasterOverlayTileProvider` was constructed with an "owner" parameter. This was either the `RasterOverlay` that created it directly, or, in the case of `IonRasterOverlay`, it was the `IonRasterOverlay` even though the tile provider was actually created by a `BingMapsRasterOverlay` or similar.

Now, that same mechanism is found in the `pOwner` property of `CreateRasterOverlayTileProviderParameters`. But now, it is optional (nullptr by default). Effectively, if the owner isn't specified, then it is the same as the RasterOverlay that created the tile provider, which arrives in a separate `pCreator` parameter.

So there were some cases where constructors were using the `pOwner` property to get options from the raster overlay, such as whether to show credits on the screen. I could have changed this to resolve the owner (using the pCreator plus the optional pOwner). But instead, I just switched to using `pCreator` directly. The logic is that the only place this is used is `IonRasterOverlay`, and it just passes the options through to the aggregated tile provider verbatim. So the new approach is easier and its equivalent.

Really, this whole pattern where tile providers reach back out to the raster overlay for some things is not great, but that's a problem for another time.


kring:
The duplicate credit problem is fixed. It turned out to be more elegant than I expected. Credits from different sources are still different credits, but now, when you call `getSnapshot`, you can pass a parameter telling it how to treat credits with identical HTML strings. The default value for the parameter preserves the behavior in previous releases. The implementation doesn't require any extra hashing or string lookups beyond the string comparisons that were already happening in `createCredit`.

CesiumUtility/include/CesiumUtility/CreditReferencer.h
```diff
@@ -106,9 +106,17 @@ class CreditReferencer {
    */
   void releaseAllReferences() noexcept;
 
+  /**
+   * @brief Tests if a credit is referenced by this referencer.
+   *
+   * @param credit to test if it is referenced.
```
j9liu:
```suggestion
   * @param credit The credit to test if it is referenced.
```

CesiumRasterOverlays/include/CesiumRasterOverlays/CreateRasterOverlayTileProviderParameters.h
```diff
@@ -0,0 +1,41 @@
...
+ * @brief Parameters passed to \ref RasterOverlay::createTileProvider.
+ */
+struct CESIUMRASTEROVERLAYS_API CreateRasterOverlayTileProviderParameters {
+  /**
+   * @brief The external interfaces for use by the raster overlay tile provider.
+   */
+  RasterOverlayExternals externals;
+
+  /**
+   * @brief The overlay that owns this overlay.
```
j9liu: Should this be...?

```suggestion
   * @brief The overlay that owns this tile provider.
```

Or...?

```suggestion
   * @brief The overlay that owns the overlay that created this tile provider.
```

CesiumUtility/include/CesiumUtility/CreditSystem.h
```diff
@@ -57,18 +155,67 @@ struct CreditsSnapshot {
...
+  Credit createCredit(
+      const CreditSource& source,
+      std::string&& html,
+      bool showOnScreen = false);
+
+  /**
+   * @brief Inserts a credit string.
+   *
+   * @param source The source of the credit. This should be an instance created
+   * and owned by a tileset, raster overlay, or other data source.
```
j9liu: **NITPICK**: would `@copydoc` work here to avoid the duplicate documentation?

CHANGES.md
```diff
@@ -1,5 +1,21 @@
 # Change Log
 
+### ? - ?
+
+##### Breaking Changes :mega:
+
+- `RasterOverlay::createTileProvider` now receives a reference to `CreateRasterOverlayTileProviderParameters` instead of a large number of individual parameters.
+- The constructor parameters for `RasterOverlayTileProvider` and `QuadtreeRasterOverlayTileProvider` have changed.
+- The `getCredit` method has been removed from `RasterOverlayCreditProvider`. Use `getCredits` instead.
+- Removed unused property `RasterOverlayOptions::subTileCacheBytes`.
```
j9liu: This is to address https://github.com/CesiumGS/cesium-native/issues/1274 ? (If so, let's link it so it auto-closes)

CesiumUtility/test/TestCreditSystem.cpp
j9liu, line 1: Not blocking (and not a fault of this PR), but it seems like `TestCreditSystem` should be in `CesiumUtility` tests instead?

Cesium3DTilesSelection/test/TestCreditSystem.cpp
```diff
@@ -198,3 +199,285 @@ TEST_CASE("Test setting showOnScreen on credits") {
...
+    std::unique_ptr<CreditSource> pTempSourceA =
+        std::make_unique<CreditSource>(creditSystem);
+    Credit credit0 = creditSystem.createCredit(*pTempSourceA, html0);
+
+    creditSystem.addCreditReference(credit0);
+    const CreditsSnapshot& snapshot0 = creditSystem.getSnapshot();
+    CHECK(snapshot0.currentCredits.size() == 1);
+    CHECK(snapshot0.currentCredits[0] == credit0);
+
+    // Remove the credit reference, which will add it to the list of "no longer
```
j9liu: `_creditsToNoLongerShowThisSnapshot` got removed, so this comment is outdated, right?

Cesium3DTilesSelection/test/TestCreditSystem.cpp
```diff
@@ -198,3 +199,285 @@ TEST_CASE("Test setting showOnScreen on credits") {
...
+        // credit1. So credit1 should be shown before credit0.
+        REQUIRE(snapshot.currentCredits.size() == 2);
+        CHECK(snapshot.currentCredits[0] == credit1);
+        CHECK(snapshot.currentCredits[1] == credit0);
+        CHECK(snapshot.removedCredits.empty());
+      }
+    }
+  }
+
+  // Refeference count for sorting is the sum of collapsed credit reference
```
j9liu:
```suggestion
  // Reference count for sorting is the sum of collapsed credit reference
```

CHANGES.md
kring: Oops I pulled this change out of this PR (because I realized I'd prefer to fix that property rather than removing it) and forgot to remove it from the changelog.

kring:
> Is this expected behavior?

No, there were a couple of bugs. The worst of which was that the credit filtering system would sometimes filter out a referenced credit in favor of an unreferenced one. It should be fixed now. Good catch!

Issue #1034, opened 2024-12-09
Fails to decode meshopt when glTF has external bin file


lilleyse:
The first tileset (.glb) works, the second tileset (.gltf + .bin) doesn't. Both use meshopt compression. There might be some ordering issue where it tries to decode meshopt before it resolves the external buffers.

[box-rotated.zip](https://github.com/user-attachments/files/18066635/box-rotated.zip)
[box-rotated-json.zip](https://github.com/user-attachments/files/18066636/box-rotated-json.zip)

```
LogCesium: Loading tileset from URL http://localhost:8070/v1/static/box-rotated-json/tileset.json
LogCesium: Loading tileset from URL http://localhost:8070/v1/static/box-rotated-json/tileset.json done
LogCesium: Warning: [2024-12-09 14:05:35.687] [warning] [ErrorList.h:107] Warning when loading http://localhost:8070/v1/static/box-rotated-json/tiles/0/0/0/0.gltf:
- The EXT_meshopt_compression extension has a bufferView that extends beyond its buffer.
- The EXT_meshopt_compression extension has a bufferView that extends beyond its buffer.
- The EXT_meshopt_compression extension has a bufferView that extends beyond its buffer.
- The EXT_meshopt_compression extension has a bufferView that extends beyond its buffer.
LogCesium: Warning: http://localhost:8070/v1/static/box-rotated-json/tiles/0/0/0/0.gltf mesh 0 primitive 0: Invalid position buffer
```

Georeferencing coordinates in case it's helpful:

* Latitude: 40.022374
* Longitude: -79.886655
* Height: 300

kring:
Yeah, your explanation is correct. It looks like the problem is that `TilesetContentManager` calls [loadTileContent](https://github.com/CesiumGS/cesium-native/blob/v0.42.0/Cesium3DTilesSelection/src/TilesetContentManager.cpp#L1023), which decodes meshtop but doesn't load external buffers, followed by [resolveExternalData](https://github.com/CesiumGS/cesium-native/blob/v0.42.0/Cesium3DTilesSelection/src/TilesetContentManager.cpp#L575), which loads external buffers but doesn't decode meshopt.

It can be fixed by calling `GltfReader::postProcessGltf` after external buffers are resolved. Or perhaps a little refactoring to avoid the need for a redundant `postProcessGltf` is in order!

PR #1280, opened 2025-11-25
Read a glTF file's external data before postprocessing.

timoore:
An new method `CesiumGltfReader::readGltfAndExternalData` has been added which reads external data and then does postprocessing.

Addresses #1034

### DIFF
--- a/Cesium3DTilesSelection/src/ImplicitOctreeLoader.cpp
+++ b/Cesium3DTilesSelection/src/ImplicitOctreeLoader.cpp
@@ -23,6 +23,7 @@
 #include <CesiumGltf/Ktx2TranscodeTargets.h>
 #include <CesiumGltfReader/GltfReader.h>
 #include <CesiumUtility/Assert.h>
+#include <CesiumUtility/IntrusivePointer.h>
 
 #include <glm/ext/matrix_double4x4.hpp>
 #include <spdlog/logger.h>
--- a/Cesium3DTilesSelection/src/ImplicitQuadtreeLoader.cpp
+++ b/Cesium3DTilesSelection/src/ImplicitQuadtreeLoader.cpp
@@ -23,6 +23,7 @@
 #include <CesiumGltf/Ktx2TranscodeTargets.h>
 #include <CesiumGltfReader/GltfReader.h>
 #include <CesiumUtility/Assert.h>
+#include <CesiumUtility/IntrusivePointer.h>
 
 #include <glm/ext/matrix_double4x4.hpp>
 #include <spdlog/logger.h>
--- a/Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
+++ b/Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
@@ -23,7 +23,6 @@
 #include <Cesium3DTilesSelection/TilesetContentLoaderResult.h>
 #include <Cesium3DTilesSelection/TilesetExternals.h>
 #include <Cesium3DTilesSelection/TilesetMetadata.h>
-#include <Cesium3DTilesSelection/TilesetSharedAssetSystem.h>
 #include <CesiumAsync/AsyncSystem.h>
 #include <CesiumAsync/HttpHeaders.h>
 #include <CesiumAsync/IAssetAccessor.h>


cmake/macros/configure_cesium_library.cmake
```diff
@@ -2,7 +2,7 @@ function(configure_cesium_library targetName)
     if (MSVC)
         target_compile_options(${targetName} PRIVATE /W4 /WX /wd4201 /bigobj /w45038 /w44254 /w44242 /w44191 /w45220)
     else()
-        target_compile_options(${targetName} PRIVATE -Werror -Wall -Wextra -Wconversion -Wpedantic -Wshadow -Wsign-conversion -Wno-unknown-pragmas)
+        target_compile_options(${targetName} PRIVATE -Wall -Wextra -Wconversion -Wpedantic -Wshadow -Wsign-conversion -Wno-unknown-pragmas)
```
kring: I'm assuming you meant to only change this locally, and accidentally committed it. I'll revert.

CHANGES.md
```diff
@@ -7,13 +7,15 @@
 - `RasterOverlay::createTileProvider` now receives a reference to `CreateRasterOverlayTileProviderParameters` instead of a large number of individual parameters.
 - The constructor parameters for `RasterOverlayTileProvider` and `QuadtreeRasterOverlayTileProvider` have changed.
 - The `getCredit` method has been removed from `RasterOverlayCreditProvider`. Use `getCredits` instead.
+- Changed the converters in Cesium3DTilesContent to call `GltfReader::readGltfAndExternalData`. The `TilesetContentManger` does not call `resolveExternalData` anymore.
```
kring:
```suggestion
- Custom functions registered with `GltfConverter` can no longer expect that external data in the glTF will be automatically loaded by the caller. If they want external data in the glTF to be loaded as well, they should use `GltfReader::readGltfAndExternalData`.
```

Cesium3DTilesContent/src/BinaryToGltfConverter.cpp
```diff
@@ -36,7 +36,26 @@ CesiumAsync::Future<GltfConverterResult> BinaryToGltfConverter::convert(
...
+  return _gltfReader
+      .readGltfAndExternalData(
+          gltfBinary,
+          assetFetcher.asyncSystem,
+          assetFetcher.requestHeaders,
+          assetFetcher.pAssetAccessor,
+          assetFetcher.baseUrl,
+          options)
+      .thenInWorkerThread(
+          [&assetFetcher](CesiumGltfReader::GltfReaderResult&& gltfResult) {
```
kring: It's not safe to lambda capture a parameter by reference if the lambda will outlive the function invocation.

Cesium3DTilesContent/src/BinaryToGltfConverter.cpp
```diff
@@ -36,7 +36,26 @@ CesiumAsync::Future<GltfConverterResult> BinaryToGltfConverter::convert(
...
+            if (gltfResult.model) {
```
kring: Looks like the `convertImmediate` function above has been inlined here, so it can be removed.

Cesium3DTilesSelection/src/TilesetContentManager.cpp
```diff
@@ -566,85 +565,16 @@ postProcessContentInWorkerThread(
...
-          return tileLoadInfo.asyncSystem
-              .createResolvedFuture(TileLoadResult::createFailedResult(
-                  result.pAssetAccessor,
-                  nullptr))
-              .thenPassThrough(std::move(tileLoadInfo));
-        }
-
-        result.contentKind = std::move(*gltfResult.model);
+  return asyncSystem
+      .runInWorkerThread([result = std::move(result),
```
kring: `postProcessContentInWorkerThread` is already running in a worker thread, so this `runInWorkerThread` is unnecessary.

CesiumGltfReader/src/GltfReader.cpp
```diff
@@ -405,6 +405,53 @@ GltfReaderResult GltfReader::readGltf(
...
 
+CesiumAsync::Future<GltfReaderResult> GltfReader::readGltfAndExternalData(
+    const std::span<const std::byte>& data,
+    const CesiumAsync::AsyncSystem& asyncSystem,
+    const std::vector<CesiumAsync::IAssetAccessor::THeader>& headers,
+    const std::shared_ptr<CesiumAsync::IAssetAccessor>& pAssetAccessor,
+    const std::string& baseUrl,
+    const GltfReaderOptions& options) const {
+  CesiumAsync::HttpHeaders httpHeaders;
+  httpHeaders.insert(headers.begin(), headers.end());
```
kring:
```suggestion
  CesiumAsync::HttpHeaders httpHeaders(headers.begin(), headers.end());
```

kring:
I've addressed all of my comments above, but there's a more subtle problem remaining. Previously, `TilesetContentManager` passed in the `GltfSharedAssetSystem` when resolving external data. Now, that isn't happening. This is slightly tricky to fix because the new load mechanism is invoked by the `TilesetContentLoaders`, which don't have a readily-available shared asset system. Will just have to weave one through, I guess.

CesiumGltfReader/src/GltfReader.cpp
timoore: Thanks @kring for fixing my broken code!

cmake/macros/configure_cesium_library.cmake
timoore: Yes. I had previously discussed making this an option in order to deal with compilation environments that are a bit too strict for vcpkg and everything, but that should be a separate PR, of course.

PR #1283, opened 2025-12-01
Add `bool` and `std::string_view` overloads for `PropertyArrayCopy`

j9liu:
This PR adds missing `bool` and `std::string_view` overloads to `PropertyArrayCopy` so they can be used by runtimes.

I also found that much of the `PropertyView` code was done before `PropertyArrayCopy` existed, so I replaced what I could to simplify the implementation.

### DIFF
--- a/CesiumGltf/include/CesiumGltf/PropertyArrayView.h
+++ b/CesiumGltf/include/CesiumGltf/PropertyArrayView.h
@@ -244,7 +244,7 @@ public:
   PropertyArrayCopy(const std::vector<bool>& values) noexcept
       : _storage(), _view() {
     size_t numberOfElements = values.size();
-    size_t sizeInBytes = (numberOfElements / 8) + 1;
+    size_t sizeInBytes = (numberOfElements + 7) / 8;
     this->_storage.resize(sizeInBytes);
 
     for (size_t byteIndex = 0, valueIndex = 0; byteIndex < sizeInBytes;


CesiumGltf/include/CesiumGltf/PropertyArrayView.h
```diff
@@ -219,6 +219,103 @@ template <> class PropertyArrayView<bool> {
...
+
+  /**
+   * @brief Constructs a bool array copy from a vector.
+   *
+   * @param values The vector containing the values.
+   */
+  PropertyArrayCopy(const std::vector<bool>& values) noexcept
+      : _storage(), _view() {
+    size_t numberOfElements = values.size();
+    size_t sizeInBytes = (numberOfElements / 8) + 1;
```
azrogers: If an array contains a multiple of eight elements, this will always allocate an extra byte, right? Is that intentional?

If not, I think `(numberOfElements + 7) / 8` will give the correct results.

CesiumGltf/include/CesiumGltf/PropertyArrayView.h
```diff
@@ -282,6 +379,167 @@ template <> class PropertyArrayView<std::string_view> {
...
+      copyDataWithOffsetType.template operator()<uint32_t>();
+    } else {
+      offsetType = PropertyComponentType::Uint64;
+      copyDataWithOffsetType.template operator()<uint64_t>();
+    }
+
+    this->_storage.resize(stringData.size() + offsetData.size());
+    std::memcpy(this->_storage.data(), stringData.data(), stringData.size());
+    std::memcpy(
+        this->_storage.data() + static_cast<int32_t>(stringData.size()),
```
azrogers, line 456: Is this cast needed? `size_t` should be able to be added to a pointer unless there's something I'm missing.

CesiumGltf/include/CesiumGltf/PropertyArrayView.h
```diff
@@ -282,6 +379,167 @@ template <> class PropertyArrayView<std::string_view> {
...
+        offsetData.data(),
+        offsetData.size());
+
+    this->_view = PropertyArrayView<std::string_view>(
+        std::span<const std::byte>(this->_storage.begin(), stringData.size()),
+        std::span<const std::byte>(
+            this->_storage.begin() + static_cast<int32_t>(stringData.size()),
```
azrogers, line 463: Same here.

azrogers (changes requested):

azrogers:
@j9liu Mostly looks good, just a few small changes (or just questions, if incorrect).

CesiumGltf/include/CesiumGltf/PropertyArrayView.h
j9liu: This was sadly required by the Emscripten build on CI. It wasn't happy with my initial `size_t` or even `int64_t` cast :/

CesiumGltf/include/CesiumGltf/PropertyArrayView.h
j9liu: Good catch, thanks!

PR #1286, opened 2025-12-02
Don't overwrite `styles` parameter of `WebMapServiceRasterOverlay`

### LINKED ISSUE #1285: `WebMapServiceRasterOverlay` adds query parameters even if they already exist in the base URL
As reported here:
https://community.cesium.com/t/cesiumwmsrasteroverlay-missing-request-parameters/44251/3

We should only add query parameters that don't already exist.

azrogers:
Fixes #1285. There are some parameters we need to set - `crs`, `request`, `TRANSPARENT`, `service`, `bbox`, etc. - but the `styles` parameter we are currently setting even if it's already present in the base URL. This change should make sure that it's not overwritten if already present in the URL, as well as making sure that no query parameters show up twice in the URL.

### DIFF
--- a/CesiumRasterOverlays/src/WebMapServiceRasterOverlay.cpp
+++ b/CesiumRasterOverlays/src/WebMapServiceRasterOverlay.cpp
@@ -15,6 +15,7 @@
 #include <CesiumRasterOverlays/RasterOverlayTile.h>
 #include <CesiumRasterOverlays/RasterOverlayTileProvider.h>
 #include <CesiumRasterOverlays/WebMapServiceRasterOverlay.h>
+#include <CesiumUtility/ErrorList.h>
 #include <CesiumUtility/IntrusivePointer.h>
 #include <CesiumUtility/Math.h>
 #include <CesiumUtility/Uri.h>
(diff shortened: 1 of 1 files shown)


CesiumRasterOverlays/src/WebMapServiceRasterOverlay.cpp
```diff
@@ -201,44 +200,52 @@ class WebMapServiceTileProvider final
...
+              ErrorList::error(
+                  fmt::format("Failed to parse URL {}", this->_url))});
+    }
+
+    UriQuery query(uri);
+    // Constants
+    query.setValue("request", "GetMap");
+    query.setValue("TRANSPARENT", "TRUE");
+    query.setValue("crs", "EPSG:4326");
+    query.setValue("service", "WMS");
```
kring: I think in general we should allow the user to specify alternate values for most of these:

1. `crs`: It's somewhat common to use a CRS that is _roughly equivalent_ to EPSG:4326. GDA2020 in Australia, for example. We shouldn't prevent this from working.
2. `TRANSPARENT`: If the WMS is satellite imagery, you wouldn't want it to be transparent. The main advantage of transparent=false is that the server can return JPG, which will be smaller. Also not sure why this is capitalized. It shouldn't matter I think?
3. `request`: Ok not much legitimate reason to let the user specify this one, so I'm ok with hard-coding it. But if we do let the user specify, be aware that it's super common for users to paste a "request=GetCapabilities" URL into a WMS URL field, and in that case we would definitely want to replace it with GetMap.
4. `service`: Again I can't think of much reason the user would want something different here. But hey maybe they know what they're doing, right? The aforementioned GetCapabilities URL will still have `service=WMS` fwiw.

kring (commented):
Looks good @azrogers. Just one comment. Also please check the CI lint step.

CesiumRasterOverlays/src/WebMapServiceRasterOverlay.cpp
azrogers: Primarily it's capitalized because it was capitalized in the previous version of the code. The WMS spec does capitalize it, but it *also* capitalizes every other parameter, so I'm not sure why this is the only one that we brought over as capitalized.

azrogers:
@kring Should be ready for another look. Also changed the `GetCapabilities` part of the WMS overlay - there's nothing there that we are overwriting that we shouldn't be, but it'll help prevent URLs with duplicate query parameters.

PR #1287, opened 2025-12-05
Replace bad `begin()` use with `data()`

j9liu:
Mistake I made in #1283 where for some reason, I used `.begin()` instead of `.data()` on a `std::vector<std::byte>`. For some reason this doesn't cause CI to fail here, but it results in compilation errors for non-Windows platforms in Cesium for Unreal ([example](https://github.com/CesiumGS/cesium-unreal/actions/runs/19967266724/job/57262296692)).

### DIFF
--- a/CesiumGltf/include/CesiumGltf/PropertyArrayView.h
+++ b/CesiumGltf/include/CesiumGltf/PropertyArrayView.h
@@ -476,11 +476,12 @@ public:
       : _storage(rhs._storage), _view() {
     // Reconstruct spans so they point to this copy's data.
     size_t valueSpanSize = rhs._view._values.size();
+    size_t offsetSpanSize = this->_storage.size() - valueSpanSize;
     this->_view = PropertyArrayView<std::string_view>(
         std::span<const std::byte>(this->_storage.data(), valueSpanSize),
         std::span<const std::byte>(
-            this->_storage.begin() + static_cast<int32_t>(valueSpanSize),
-            this->_storage.end()),
+            this->_storage.data() + static_cast<int32_t>(valueSpanSize),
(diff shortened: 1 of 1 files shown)


j9liu:
@timoore could you please review?

j9liu:
@azrogers I got overzealous in the GitHub web editor and replaced more instances than I should have 😅 but whereas those constructors take two iterators, the first one is supposed to take a data pointer. I accidentally gave it an iterator instead.

We'll see if this does the job for Unreal CI: https://github.com/CesiumGS/cesium-unreal/actions/runs/19970583611

j9liu:
Hmm... looks like Android is still failing in the Unreal build, so I'll have to figure out what else to change. :/

CesiumGltf/include/CesiumGltf/PropertyArrayView.h
```diff
@@ -458,9 +458,9 @@ template <> class PropertyArrayCopy<std::string_view> {
         offsetData.size());
 
     this->_view = PropertyArrayView<std::string_view>(
-        std::span<const std::byte>(this->_storage.begin(), stringData.size()),
+        std::span<const std::byte>(this->_storage.data(), stringData.size()),
         std::span<const std::byte>(
-            this->_storage.begin() + static_cast<int32_t>(stringData.size()),
+            this->_storage.data() + static_cast<int32_t>(stringData.size()),
```
timoore, line 463: Quick question: why is the offset cast to `int32_t`?

j9liu: Without it, CI fails for Emscripten.

PR #1291, opened 2025-12-16
Don't let SharedAssetDepot cache loads that throw

### LINKED ISSUE #1290: SharedAssetDepot caches errors
After CesiumGS/cesium-unity#632, we sometimes abort in-flight network requests. The problem is that if we happen to abort a request for a resource managed by the `SharedAssetDepot`, then that error will be cached and the request will never be retried.

So every time we try to load something that uses that shared asset again, we'll get the same error we got the first time: `Error creating asset: Request was canceled because the Unity AppDomain is reloading.` Which seems pretty broken because the AppDomain isn't reloading anymore.

I think the solution here is to not cache shared asset loads that return an error (that is, the factory function Future rejects).

It's less clear what should happen when the Future resolves to an error, though. For example, if the HTTP server returns 404, should we cache that? On the one hand, maybe the error is transient. Someone forgot to copy the file up there, and it's annoying to have to restart the process (or at least call `invalidate`) to get the app to try again. On the other hand, avoiding repeated requests for resources that don't exist is a useful function of the SharedAssetDepot. For example, `QuadtreeRasterOverlayTileProvider` would be significantly less efficient for incomplete quadtrees with unknown tile availability (very common scenario) if we didn't cache 404s.

So for the time being I suggest:

1. Factory function Future rejects -> do not cache
2. Factory function Future resolves (even if it resolves to an error) -> cache

kring:
Fixes #1290

While trying to reproduce https://github.com/CesiumGS/cesium-unity/pull/642#pullrequestreview-3573386635, I ran into a problem a few times where raster overlay images got stuck not loading after an AppDomain reload. This PR should avoid that problem.


kring:
Testing this further in Unity, I noticed that it still wasn't working quite right. The problem was that two places we used SharedAssetDepot, ion raster overlays and quadtree tile images, were catching future rejection and turning it into a regular, resolved value representing an error. This, in turn, meant that the failure was still being cached.

So, I've now removed the code that does that. I think this is still a good change, but it could mean a change of behavior in some cases.

We generally don't use promise rejection to indicate normal errors. There are various reasons for this, but one major one is the same reason we don't throw exceptions in normal control paths: exceptions should be reserved for exceptional circumstances. So pretty much the only time a future is rejected is when:

1. A network request fails badly. Not like "I got a 404" but rather "there doesn't seem to be a network at all."
2. After CesiumGS/cesium-unity#632, when a network request is canceled because Unity is reloading an AppDomain.
3. There are developer errors, like an access violation.

So if we request a raster overlay quadtree tile, and it fails complete (no network!), then before this PR, we would have used a parent image. If the parent image worked, that is, which seems pretty unlikely if the network is broken. In this PR, we just let the error stand, so the tile will be missing a raster overlay in the area of the tile.

On the plus side, in this PR, refreshing the tileset will retry that failed image. Before this PR, we would use that parent image forever (until the SharedAssetDepot ejects the failed image from the cache).

kring:
This PR now builds on #1295 so merge that first.

PR #1292, opened 2025-12-16
Handle refresh tokens in `CesiumIonClient::Connection`

azrogers:
Cesium ion login tokens (_not_ the tokens used to access assets, but the OAuth2 tokens used to access account features) will soon be given expiration dates. This is a change from the current behavior where the access tokens never expire. This means we'll need to handle refreshing access tokens from refresh tokens when the access tokens expire, as well as handling in our runtimes when access tokens have expired and no valid refresh token is available.

This change handles the first part of that. Refresh tokens are obtained from the authorize step and are transparently used to obtain new access tokens as needed, in a pattern similar to the one already used in `CesiumITwinClient`. Any issues with the tokens will be returned as 403 errors. 

The next step is to implement in our runtimes that when a 403 error is returned from ion, we should trigger a popup that prompts the user to sign in to their ion account again.

### DIFF
--- a/CesiumIonClient/src/LoginToken.cpp
+++ b/CesiumIonClient/src/LoginToken.cpp
@@ -1,7 +1,6 @@
 #include <CesiumClientCommon/JwtTokenUtility.h>
 #include <CesiumIonClient/LoginToken.h>
 #include <CesiumUtility/ErrorList.h>
-#include <CesiumUtility/JsonHelpers.h>
 #include <CesiumUtility/Result.h>
 
 #include <rapidjson/document.h>


azrogers:
Apparently `modp_b64` requires padded base64 strings, and JWT tokens are base64 strings without padding. I was confused why all the tests were passing but the testing against ion was failing to decode the token, until I realized that all the tests are using tokens generated by `modp_b64`, so they include padding. Added a new test working from a pre-generated dummy token and added padding to the b64 string to fix it. 

I'm rethinking my approach of handling the 403 errors on the runtimes end. Having dug into the Cesium ion editor code in [my WIP branch](https://github.com/CesiumGS/cesium-unreal/tree/ion-refresh-token), I think it will be a lot more of a hassle to add a check for 403 errors to every place we make a Cesium ion call than to find some way of handling it within the `Connection` itself. I'm thinking some sort of callback...

CHANGES.md
```diff
@@ -2,18 +2,25 @@
 
 ### ? - ?
 
-##### Fixes :wrench:
+##### Breaking Changes :mega:
 
-- `CesiumRasterOverlays::WebMapServiceRasterOverlay` will no longer attempt to add a default `styles` parameter to the URL if the user has already specified one.
+- `CesiumIonClient::Connection::authorize` now returns a `CesiumUtility::Result<Connection>`. This removes the previous behavior of throwing an exception when authorization failed. 
 - `SharedAssetDepot` will no longer cache asset loads that fail with an exception / `Future` rejection, allowing them to be retried. Other types of load failures are cached as before.
```
j9liu, line 8: Should this still be under "Fixes"?

CesiumClientCommon/src/JwtTokenUtility.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+  std::string payloadSegment(parts[1]);
+  size_t payloadLength = payloadSegment.length();
+  // modp_b64 requires padded base64 strings, which JWT tokens are not.
+  if (payloadLength % 4 != 0) {
+    const size_t newSize = (payloadLength / 4 + 1) * 4;
+    payloadSegment.resize(newSize, '=');
+    for (size_t i = payloadLength; i != newSize; i++) {
+      payloadSegment[i] = '=';
+    }
```
j9liu: This for loop seems redundant when we're already calling `payloadSegment.resize(newSize, '=');`, no?

CesiumIonClient/include/CesiumIonClient/Connection.h
```diff
@@ -97,7 +104,7 @@ class CESIUMASYNC_API Connection {
    * @return A future that resolves to a Cesium ion {@link Connection} once the
```
j9liu: Should this `@return` be tweaked to account for `CesiumUtility::Result`?

CesiumIonClient/test/MockIonAssetAccessor.cpp
```diff
@@ -0,0 +1,188 @@
...
+MockIonAssetAccessor::request(
+    const CesiumAsync::AsyncSystem& asyncSystem,
+    const std::string& verb,
+    const std::string& url,
+    const std::vector<THeader>& headers,
+    const std::span<const std::byte>& body) {
+  std::vector<std::byte> bodyBuffer(body.begin(), body.end());
+  CesiumUtility::Uri uri(url);
+
+  return handleApiServer(asyncSystem, verb, uri, headers, bodyBuffer);
```
j9liu: For style consistency:

```suggestion
  return this->handleApiServer(asyncSystem, verb, uri, headers, bodyBuffer);
```

CesiumIonClient/include/CesiumIonClient/LoginToken.h
```diff
@@ -0,0 +1,63 @@
...
+   * @returns A `Result` containing either the parsed \ref LoginToken
+   * or error messages.
+   */
+  static CesiumUtility::Result<LoginToken>
+  parse(const std::string& tokenString);
+
+  /**
+   * @brief Is this token currently valid?
+   *
+   * For the token to be valid, it must be before its expiration date.
```
j9liu: **NIT**:

```suggestion
   * @brief Returns whether this token is currently valid.
   *
   * The token is valid up until its expiration date.
```

CesiumIonClient/src/LoginToken.cpp
```diff
@@ -0,0 +1,51 @@
...
+          .count();
+  return currentTimeSinceEpoch < this->_expires;
+}
+
+LoginToken::LoginToken(const std::string& token, int64_t expires)
+    : _token(token), _expires(expires) {}
+
+int64_t LoginToken::getExpirationTime() const { return _expires; }
+
+const std::string& LoginToken::getToken() const { return _token; }
```
j9liu:
```suggestion
int64_t LoginToken::getExpirationTime() const { return this->_expires; }

const std::string& LoginToken::getToken() const { return this->_token; }
```

CesiumIonClient/include/CesiumIonClient/LoginToken.h
```diff
@@ -0,0 +1,63 @@
...
+   */
+  bool isValid() const;
+
+  /**
+   * @brief Returns the number of seconds since the Unix epoch representing the
+   * time that this token expires.
```
j9liu:
```suggestion
   * @brief Returns the time that this token expires, represented as
   * a number of seconds since the Unix epoch.
```

CesiumIonClient/include/CesiumIonClient/LoginToken.h
```diff
@@ -0,0 +1,63 @@
...
+   *
+   * @param token The full token string.
+   * @param expires A UNIX timestamp representing the point in time that this
+   * token stops being valid.
+   */
+  LoginToken(const std::string& token, int64_t expires);
+
+private:
+  std::string _token;
+  int64_t _expires;
```
j9liu: **NIT**: I think `expirationTime` serves better as a noun, unless `expires` is commonly used for this measure

CesiumIonClient/src/LoginToken.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+  return Result<LoginToken>(LoginToken(tokenString, expires));
+}
+
+bool LoginToken::isValid() const {
+  if (this->_expires == -1) {
+    // If the value is -1, that suggests we have no expiration date, so the
+    // token is always valid.
+    return true;
+  }
```
j9liu: This feels hand-wavey since there's nothing alluding to an optional expiration time in the `LoginToken` constructor. Maybe consider _all_ negative numbers as no expiration date, or otherwise make this number a `std::optional`?

CesiumIonClient/include/CesiumIonClient/Connection.h
```diff
@@ -161,33 +172,78 @@ class CESIUMASYNC_API Connection {
       const CesiumIonClient::ApplicationData& appData,
       const std::string& apiUrl = "https://api.cesium.com");
 
+  /**
+   * @brief Creates a connection to Cesium ion using the provided access and
+   * refresh tokens.
```
j9liu:
```suggestion
   * @brief Creates a connection to Cesium ion using the provided access and
   * refresh tokens, while also caching the provided client ID and redirect path.
```

CHANGES.md
kring: It's under "Breaking Changes", isn't it?

CesiumClientCommon/src/JwtTokenUtility.cpp
kring: Oops, good catch! I switched to using the resize overload to make the loop unnecessary, and then failed to remove the loop.

CesiumIonClient/include/CesiumIonClient/Connection.h
kring: I went with some different wording here. See what you think.

CHANGES.md
j9liu: Oh I was referring to the "`SharedAssetDepot` will no longer cache asset loads that fail with an exception / `Future` rejection" entry. It looked like it was originally categorized as a fix, so I was wondering if it should move as well. Unless my holiday break brain is completely misinterpreting this diff 😆 

Image

CesiumIonClient/include/CesiumIonClient/Connection.h
j9liu: Your changes are even better, thank you! :pray:

j9liu (approved):
Thanks for the changes @kring ! Approved, but I'll test this out in a runtime before merging.

CHANGES.md
kring: Your diff screenshot looks strange. It says the SharedAssetDepot item isn't in a section at all? That's not what I see looking at the diff in this PR.

But anyway, I understand your point now and it makes sense. That item moved from "Fixes" to "Breaking Changes". I think it's legitimately a breaking change, though, so we should leave it where it is in this PR. The distinction between a fix and a breaking change can sometimes be pretty subtle. It's pretty much always possible to imagine some poor developer relying on old broken behavior in such a way that what seems like a clear "fix" is really "breaking" for them. But in this case, the old behavior wasn't "obviously wrong", we just think the new behavior is better, and we have reasons for doing it, even though it has the potential to introduce subtle bugs into existing code.

So I'm not sure why it went into Fixes in the first place, but let's leave it where it is in this PR.

PR #1295, opened 2025-12-19
Enable warnings-as-errors on each target rather than globally.

### LINKED ISSUE #1294: Setting CMAKE_COMPILE_WARNING_AS_ERROR=ON is overly aggressive
In #1289, we changed how cesium-native enables warning-as-errors. Rather than an explicit compiler option set on each target, we simply set `CMAKE_COMPILE_WARNING_AS_ERROR=ON`, and CMake handles the rest.

Unfortunately, this has some side-effects because it enables warnings-as-errors with a much larger scope than the previous approach. When cesium-native is included in another project via `add_subdirectory`, we end up inadvertently enabling warnings-as-errors for _all_ targets in the CMake build, not just the cesium-native ones.

This breaks the cesium-unity build because there are warnings in our own code (we should fix this and enable warnings-as-errors) and in third-party code like tidy-html5 (probably not worth fixing).

I think the right approach is to not touch `CMAKE_COMPILE_WARNING_AS_ERROR` at all, but instead set the `COMPILE_WARNING_AS_ERROR` property on each our targets to ON. We could make it an option, but I don't think there's much need. Warnings-as-errors can be disabled entirely with the `--compile-no-warning-as-error` CMake command-line option.

kring:
Fixes #1294


azrogers:
Thanks @kring!

PR #1298, opened 2026-01-07
Add `AsyncSystem::all` override for futures that resolve to different types

jaadelgren:
Adds an override for `AsyncSystem::all` that accepts futures that resolve to different types. The existing `all` implementations allow merging multiple futures, but they must all resolve to the same type.

The returned future resolves to a tuple containing all the types of the futures passed to `all`.

Futures that resolve to `void` are currently unsupported. I explored a few options but opted to open the PR first, figuring a discussion may be warranted. First off, note that you can't create a tuple with void in it, e.g. `std::tuple<int, void>` is not valid (interestingly, you could create a type alias of this, but you can't actually create the tuple). Additionally, async++ doesn't support passing void tasks to `when_all`, so some sort of workaround would be required. A few options that come to mind:
 - Don't support void types and leave things as-is
 - Don't support void types in `all`, but add another function that allows merging void futures with non-void futures (then callers could aggregate all void futures, call `AsyncSystem::all(std::vector<Future<void>>)`, and merge the result with the future returned from `AsyncSystem::all(Futures&&...)`
 - Require callers to pass all void futures _first_, and use some sort of recursion + void check to aggregate these and then merge with the result of non-void futures (in practice likely builds on the previous bullet point)
 - Try to get something working where e.g. `AsyncSystem::all(Future<int>, Future<void>, Future<double>)` would resolve to `std::tuple<int, double>` (a slightly more complicated version of the previous bullet point)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@
 
 - Added an overload for `AsyncSystem::all` that supports multiple futures that resolve to different types. Futures that resolve to `void` are currently unsupported by this overload.
 
+##### Fixes :wrench:
+
+- Fixed a bug that could cause an assertion failure or crash when destroying a `Cesium3DTilesSelection::Tileset` very soon after creating it using the constructor taking a custom `TilesetContentLoader` or `TilesetContentLoaderFactory`.
+
 ### v0.56.0 - 2026-01-05
 
 ##### Breaking Changes :mega:
--- a/Cesium3DTilesSelection/src/Tileset.cpp
+++ b/Cesium3DTilesSelection/src/Tileset.cpp
@@ -72,7 +72,7 @@ Tileset::Tileset(
       _distances(),
       _childOcclusionProxies(),
       _pTilesetContentManager{
-          new TilesetContentManager(
+          TilesetContentManager::createFromLoader(
               _externals,
               _options,
               std::move(pCustomLoader),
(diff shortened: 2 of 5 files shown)


CesiumAsync/include/CesiumAsync/AsyncSystem.h
```diff
@@ -250,6 +251,66 @@ class CESIUMASYNC_API AsyncSystem final {
         std::forward<std::vector<SharedFuture<T>>>(futures));
   }
 
+  /**
+   * @brief The value type of the Future returned by
+   * {@link AsyncSystem::all(Futures&&... futures) const}.
```
jaadelgren, line 256: Unfortunately I needed to add the parameter name to get doxygen to generate a valid link.

jaadelgren:
> @jaadelgren just confirming, would you like eyes on this PR even though it's marked draft?

Yeah I figured there might be some discussion around handling void, not sure it's ready to be merged as-is. If you prefer things to be marked "ready" before discussion begins in earnest that's fine, just let me know :slightly_smiling_face:

kring:
This looks great, thanks @jaadelgren! I think it's fine to merge it as-is and add support for `Future<void>` in a separate PR. And only when you or someone else has a need for it, I don't think there's any urgency.

j9liu:
@jaadelgren just a drive-by comment: can you update CHANGES.md for us before we merge?

PR #1299, opened 2026-01-13
Account for `nullptr` root tile in `Tileset::loadMetadata`

j9liu:
Sometimes `getRootTileAvailableEvent` will resolve with a `nullptr` root tile if there are errors while loading (e.g., attempting to load a tileset from an invalid URL). This is desired behavior, even if the name is slightly misleading for the situation.

The `loadMetadata` function will now account for this possibility to avoid a `nullptr` deference error.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Fixed a bug in `Tileset::loadMetadata` that did not account for cases where the root tile could be `nullptr`, e.g., after attempting to load a tileset from an invalid URL.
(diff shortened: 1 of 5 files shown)


kring:
> Wait I'm pretty confused. I noticed that the promise is supposed to reject if something goes wrong loading the root tile

Not exactly. It only rejects in the `catch` handler, which is when something quite catastrophic goes wrong. A normal load failure (e.g., request for the tileset.json returns a 404, or JSON parsing of the tileset.json response fails) will not invoke that catch continuation.

PR #1300, opened 2026-01-15
Make sure TilesetContentManager stays alive for async work when constructed with a TilesetContentLoader.

kring:
Fixes CesiumGS/cesium-unreal#1783

Previously, a `TilesetContentManager` constructed with a loader (for example, when using an `EllipsoidTilesetLoader`) could end up being destroyed before the async work begun in the constructor completed, leading to a crash.

In theory this is easily solved by grabbing an `IntrusivePointer` to the `TilesetContentManager` while the async work is in progress, as we do elsewhere. But this is a sketchy thing to do in a constructor. The code that called the `TilesetContentManager` constructor has no possible way to take its own `IntrusivePointer` to the instance before the constructor returns. So if the constructor itself creates an IntrusivePointer, and if there's _any possibility_ that that IntrusivePointer is destroyed before the caller constructs theirs, we risk `new TilesetContentManager(...)` returning a pointer to an instance that has already been deleted. 😱 

The cleanest way to deal with this is to use factory functions that return `IntrusivePointer<TilesetContentManager>`, and don't let anyone call the constructor directly. That's what this PR does. This isn't a breaking changes because `TilesetContentManager` is not part of the public API.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
+- Fixed a bug that could cause an assertion failure or crash when destroying a `Cesium3DTilesSelection::Tileset` very soon after creating it using the constructor taking a custom `TilesetContentLoader` or `TilesetContentLoaderFactory`.
+
 ### v0.56.0 - 2026-01-05
 
 ##### Breaking Changes :mega:


Cesium3DTilesSelection/test/TestTilesetContentManager.cpp
```diff
@@ -1724,11 +1724,13 @@ TEST_CASE("Test the tileset content manager's post processing for gltf") {
...
     IntrusivePointer<TilesetContentManager> pManager =
-        new TilesetContentManager{
+        TilesetContentManager::createFromLoader(
             externals,
             {},
             std::move(loaderResult.pLoader),
-            std::move(loaderResult.pRootTile)};
+            std::move(loaderResult.pRootTile));
+
+    pManager->waitUntilIdle(5000.0);
```
j9liu, line 1733: Is this needed now for any particular reason?

kring: Yeah, the test will fail without it. Previously, the test was calling `loadTileContent` without waiting for the root tile to become available first. This was always dodgy, but it kind of worked previously. In this PR, it will assert or crash.

kring:
This is ready for another look.

Issue #1301, opened 2026-01-15
`CesiumGeospatial::GlobeRectangle::computeIntersection` fails when one rectangle crosses the antimeridian and the other exactly touches it.


azrogers:
If I try to calculate an intersection of two rectangles with GlobeRectangle::computeIntersection, and one of those rectangles crosses the antimeridian and the other rectangle just touches the antimeridian, it will fail to find an intersection - even though they should obviously be overlapping.

You can [see a demonstration of this in CesiumJS](https://sandcastle.cesium.com/index.html#c=vZNNj9MwEIb/yiintEqcLGilpZutQKVCSHypIE65TOLZ1sLxVLbTahftf1/na6EFhLhwiO2ZvM9Y79hWzZ6thzmggxU51TZwa7mBMqr7qIyuS1ua7svmQE4r41OpHFaawHDamtaRTA9oHcyzSVqzcR4Oio5k4QYMHcfi4mufi6fyKzYelSFbRrPrUxpPwQ3VHs1WU/xMvEggzcVFGPt1WA7wAFZ/AtP36Hfi09uezq/C9Fzklx2fX53vbkk+gaHeYEWQ8corcgKljL93agCDDS1CvzYkwVs0Trd10IUKI15GyaB8yixgZAFqZiuVQU9uAZhM6SYkrEK9mFysWLMVm/VrcVR+90rvdxjn4nI2Eg/9/HDuYmuJzP/3Uf3Nx5vNev3hH5ywJqF5G58fqQi3kawLoWITYwLV7Ac4er1nbr5wfOZ8kEVJVDh/p2nZbf1SDY+htToWIvPU7HVnKKva+ht5UbsOM0U2IYVUB1Dy5je3GWqNzoU/t63Wn9V96N6yyIL+BNOMoWnbjweyGu86ye5i+W5ICiGKLIS/Up5ZV2h/qvgI), because the intersection code is the same. Even though the two rectangles are clearly overlapping in the image, the intersection calculation returns undefined. If you change the west coordinate of rectangle b to some other value (even one very close to -Math.PI, like -3.1415926535), suddenly it can correctly find the intersection.

[...]

PR #1302, opened 2026-01-15
Fix more GeoJSON rasterization issues

### LINKED ISSUE #1275: GeoJSON Rasterization issues at anti-meridian
As reported by @yuxueliyuxl in https://github.com/CesiumGS/cesium-native/pull/1256#issuecomment-3561368443:

> [LineStringTestData.zip](https://github.com/user-attachments/files/23704100/LineStringTestData.zip)
I'm not sure if the latest version fixes the antimeridian issue

Image

Image

> I tested it under v2.21.0 Preview for Unreal Engine 5.7 - 2025-11-17, and the issue still exists. I have provided test data.

azrogers:
The sequel to #1256. This pull request fixes a few standing issues with GeoJSON rasterization:

- The width of the style was not being factored into the bounding box of a GeoJSON geometry primitive. This could cause issues when the width of a line means it should be considered for rendering in other tiles besides the ones the line itself is passing through, causing inconsistent rendering. However, the most common way for this issue to crop up is that a fully vertical or fully horizontal line (a line around the equator or a line around the meridian, for example) would not show up, because its bounding box height or width would be zero without the style's width being factored in. This was noted in #1256.
- Geometry near the antimeridian would not render correctly, because the use of `expandToIncludePosition` would mean the bounding box would never encompass the full globe no matter how many points were in your document. GlobeRectangle calculation has been changed to handle this (which has introduced other far more minor issues as a result - see below). This means this PR closes #1275.
- GeoJSON points, features, feature collections, and geometry collections were being included in the quadtree, even though they will never be rendered (even if geometry they contain will). Besides wasting some memory and computation time, it meant that primitives could be rendered multiple times - first in the FeatureCollection, then each geometry individually. [This caused issues with transparency.](https://community.cesium.com/t/color-randomization-of-polygon-style-affected-by-alpha-in-cesiumvectorstyle/44640)

The above issues have been fixed. In the course of doing so, we've introduced a few small edge cases which we've decided are at this point not worth the time to fix. For future reference, they are:

- Geometry set to use `LineWidthMode::Meters` will have its width factored in when building the quadtree instead of when rendering the quadtree. This is more efficient but can cause issues with geometry near the antimeridian. This is because we grow the bounding rectangle of the geometry based on the width in meters, but this can cause it to grow past the antimeridian and produce a rectangle which `GlobeRectangle::computeUnion` will mangle. The solution here is to either come up with a smarter way of growing the bounding rectangle that produces a rectangle `computeUnion` will accept, or handling `LineWidthMode::Meters` the same as handling `LineWidthMode::Pixels` and do it all at render time.
- The GeoJSON specification says that geometry *should* be cut at the antimeridian, but this is different from saying that geometry *must* be cut at the antimeridian. The way that other renderers like MapBox allow you to specify geometry crossing the antimeridian is to use coordinates outside the bounds of [-180, 180]. So, for example, you could create a line from -170 to -190 and this would work. This PR does not support doing this for two reasons: one, `radiansToPoint` in `VectorRasterizer.cpp` doesn't understand these coordinates, and two, #1301 means that raster overlay tiles adjacent to the antimeridian will not be considered to include the document bounding rectangle that crosses the antimeridian, even though it should.

### DIFF
--- a/CesiumRasterOverlays/src/GeoJsonDocumentRasterOverlay.cpp
+++ b/CesiumRasterOverlays/src/GeoJsonDocumentRasterOverlay.cpp
@@ -359,7 +359,7 @@ struct GeoJsonChildVisitor {
           feature.geometry.get(),
           data,
           documentRegionBuilder,
-          featureStyle ? *featureStyle : style,
+          featureStyle.value_or(style),
           ellipsoid);
     }
   }


CesiumNativeTests/src/checkFilesEqual.cpp
```diff
@@ -0,0 +1,22 @@
...
+void checkFilesEqual(
+    const std::filesystem::path& fileA,
+    const std::filesystem::path& fileB) {
+  const std::vector<std::byte>& bytes = readFile(fileA);
+  const std::vector<std::byte>& bytes2 = readFile(fileB);
+
+  REQUIRE(bytes.size() == bytes2.size());
+  for (size_t i = 0; i < bytes.size(); i++) {
+    CHECK(bytes[i] == bytes2[i]);
+  }
```
j9liu: Is the `==` operator not overloaded for `std::byte` here?

```suggestion
  REQUIRE(bytes == bytes2);
```

CHANGES.md
```diff
@@ -1,5 +1,14 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
+- Various `CesiumRasterOverlays::GeoJsonDocumentRasterOverlay` improvements:
```
j9liu: **NITPICK:**

```suggestion
- Fixed various bugs with `CesiumRasterOverlays::GeoJsonDocumentRasterOverlay`:
```

CesiumRasterOverlays/src/GeoJsonDocumentRasterOverlay.cpp
```diff
@@ -216,7 +359,8 @@ struct GeoJsonChildVisitor {
           feature.geometry.get(),
           data,
           documentRegionBuilder,
-          featureStyle ? *featureStyle : style);
+          featureStyle ? *featureStyle : style,
```
j9liu: No need to change this, but for future reference, there's a [`value_or` function](https://en.cppreference.com/w/cpp/utility/optional/value_or.html) on `std::optional` that does this logic for us:

```suggestion
          featureStyle.value_or(style),
```

CesiumNativeTests/src/checkFilesEqual.cpp
azrogers: I'll be honest, I had no idea that the vector `==` operator actually compared contents!

PR #1305, opened 2026-01-23
Add mapped raster tiles in the progress percentage. 

rbois-bentley:
This allows apps that rely on this progress percentage to continue updating their views until all raster tiles have been properly loaded.


Cesium3DTilesSelection/src/TilesetViewGroup.cpp
j9liu, line 1: Could you add a note about this to `CHANGES.md`? You could put it under "Fixes" since raster overlay tiles were not originally accounted for in the percentage.

Cesium3DTilesSelection/src/TilesetViewGroup.cpp
```diff
@@ -157,6 +157,18 @@ void TilesetViewGroup::finishFrame(
                         updateResult.tilesKicked +
                         this->_tilesAlreadyLoadingOrUnloading;
 
+  // Add the per-tile mapped raster tiles in the progress percentage.
+  for (const Tile::ConstPointer& pTile : updateResult.tilesToRenderThisFrame) {
+    const std::vector<RasterMappedTo3DTile>& mappedRasterTiles =
+        pTile->getMappedRasterTiles();
+    for (const RasterMappedTo3DTile& mappedRasterTile : mappedRasterTiles) {
```
j9liu: There's actually a similar loop happening on line 205. To avoid the duplicate passes, can we merge this logic into that loop below? (We may have to shift around other lines to compute the percentage at the very end, but that's fine.)

j9liu (changes requested):
Thanks for the PR @rbois-bentley ! This change makes sense, I just have a few comments.

Cesium3DTilesSelection/src/TilesetViewGroup.cpp
rbois-bentley: @j9liu Thanks for the review, I'll look into merging the logic...

rbois-bentley:
> @rbois-bentley Just checking in, is this ready for another look?

@j9liu Yes, it's ready. Thanks!

PR #1307, opened 2026-02-16
glTF modifier: fixed case of raster overlays

Julot94:
GltfModifier: made it compatible with raster overlays (the initial revision of Gltf modifier was just ignoring them - so as soon as a modification was re-triggered, any cut-out polygon applied to the model would just be lost)
    
- raster overlays (computed as post-process) need to be recomputed *after* the modification step.    
- in the case of the re-apply path (see _reapplyGltfModifier_), we need to restore the initial tile bounding volume, because the post-process can lead to set an empty bounding volume, which would trigger plenty of false positive errors when re-running the post-process on the tile.
- for up-sampled tiles, the re-apply path consists in up-sampling the modified parent (and not run the modifier on the current up-sampled tile...)
- in the the re-apply path, we must take the result of the modification step into account: if the modification step fails (which can now happen in case of up-sampling...), we still keep the current model, but we should notify the next step to avoid a crash when trying to set null resources to the rendered tile
- added a shared-mutex in TileRenderContent to fix a potential crash when the game thread is replacing the parent tile's content (replaceWithModifiedModel) while a worker thread is currently up-sapling it for one of its children

### DIFF
--- a/Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileContent.h
+++ b/Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileContent.h
@@ -251,6 +251,24 @@ public:
    */
   void replaceWithModifiedModel() noexcept;
 
+  /**
+   * @brief Returns whether this tile is currently being up-sampled.
+   * It should only be called by the main thread.
+   */
+  bool isBeingUpSampled() const noexcept;
+
+  /**
+   * @brief Increment the current up-sampling task count.
+   * It should only be called by the main thread.
+   */
+  void incrementUpSamplingTaskCount() const noexcept;
+
+  /**
+   * @brief Decrement the current up-sampling task count.
+   * It should only be called by the main thread.
+   */
+  void decrementUpSamplingTaskCount() const noexcept;
+
 private:
   CesiumGltf::Model _model;
   void* _pRenderResources;
@@ -258,6 +276,7 @@ private:
   GltfModifierState _modifierState;
   std::optional<CesiumGltf::Model> _modifiedModel;
   void* _pModifiedRenderResources;
+  mutable int32_t _activeUpSamplingTaskCount;
 
   CesiumRasterOverlays::RasterOverlayDetails _rasterOverlayDetails;
   std::vector<CesiumUtility::Credit> _credits;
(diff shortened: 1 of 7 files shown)


Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileContent.h
```diff
@@ -251,13 +252,21 @@ class CESIUM3DTILESSELECTION_API TileRenderContent {
...
   CesiumGltf::Model _model;
   void* _pRenderResources;
 
   GltfModifierState _modifierState;
   std::optional<CesiumGltf::Model> _modifiedModel;
   void* _pModifiedRenderResources;
+  /** replaceWithModifiedModel is called by the main thread, while a worker
+   * thread can be upsampling the same tile - this shared mutex is used to
+   * ensure both operations are mutually exclusive. */
+  mutable std::shared_mutex _modelMutex;
```
kring: Both operations are initiated (and completed) in the main thread, so I don't think there's any need for a mutex. A simple boolean should suffice.

kring:
Thanks for the PR @Julot94! This is trickier than I originally anticipated.

My initial thought is that we'd really like to treat upsampling as just another way of loading a tile. So, just like with normal loading, we would always upsample first, and _then_ apply the GltfModifer to the upsampled tile.

As you know, one problem with that is we have to be careful to sequence the upsample and the GltfModifier run. Both need the model as input, and the latter destroys and replaces the model, so doing both at the same time will result in undefined behavior. I think this problem is relatively easy to solve, though, and in fact you've done so in this branch. My main suggestion there is that you avoid the use of a mutex. A simple boolean should suffice, because both operations are initiated and complete in the main thread.

Another problem is that we usually don't upsample a tile until later, after the GltfModifier has already run on it. So we'll end up upsampling from a tile that has already been GltfModified, and then GltfModifying the result of that upsample, too. But I think this is only a little worse than the usual GltfModifier situation where each new version starts with the results of the previous one. A GltfModifier implementation just needs to be non-lossy and produce deterministic results on re-application.

So that seems preferable to me over the alternative, which I believe you've implemented here, where GltfModifier only runs on the real tiles, and then the modified tiles are upsampled. But maybe there's some reason the alternative is required or preferred? What do you think?

Julot94:
Hello @kring ,
I updated the PR, following your suggestions

j9liu:
Sorry for the delay on this @Julot94 ! Are you able to merge in main, and add an entry to `CHANGES.md` so we can get this rolling again?

Julot94:
Hello @j9liu ,
I will have a look in a couple of days. I had almost forgotten this PR ;-)

Julot94:
Hello @j9liu 
I have added a note in CHANGES.md, and rebased it on main (no conflict), so we can try to merge this PR

PR #1310, opened 2026-02-18
Added support in the glTF codegen to force optional props as required

weegeekps:
With the new `KHR_gaussian_splatting` specification we're recommending that writers include the optional values `projection` and `sortingMethod` for maximum interoperability. Both of these values are marked optional because they can be ignored by renderers in some cases, but we should follow this advice despite not using them ourselves.

This PR adds support so we can mark optional props as required in the code generator configuration file. It also includes an updated `ModelJsonWriter.cpp` with this already applied for `KHR_gaussian_splatting`.

### DIFF
--- a/tools/generate-classes/generate.js
+++ b/tools/generate-classes/generate.js
@@ -51,19 +51,19 @@ function generate(options, schema, writers) {
   // Allows us to force specific optional props as required to meet occasional extension recommendations.
   const requireOptionalProperties = thisConfig.requireOptionalProperties;
   if (Array.isArray(requireOptionalProperties) && requireOptionalProperties.length > 0) {
-    const validProps = Object.keys(schema.properties || {});
-    for (const prop of requireOptionalProperties) {
-      if (!validProps.includes(prop)) {
+    const validProperties = Object.keys(schema.properties || {});
+    for (const property of requireOptionalProperties) {
+      if (!validProperties.includes(property)) {
         // Skip invalid properties.
         continue;
       }
 
-      if (required.includes(prop)) {
-        // Don't add props to list if they're already in `required`.
+      if (required.includes(property)) {
+        // Don't add properties to list if they're already in `required`.
         continue;
       }
 
-      required.push(prop);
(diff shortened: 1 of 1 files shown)


tools/generate-classes/generate.js
```diff
@@ -48,6 +48,25 @@ function generate(options, schema, writers) {
 
   const required = schema.required || [];
 
+  // Allows us to force specific optional props as required to meet occasional extension recommendations.
+  const requireOptionalProps = thisConfig.requireOptionalProps;
```
j9liu: **NIT**: We avoid abbreviating / shortening words where it's not cumbersome to spell them out. Could you do a pass for `prop` => `property`?

```suggestion
  const requireOptionalProperties = thisConfig.requireOptionalProperties;
```

This is admittedly not in the Cesium Native style guide, hence the "NIT" prefix, but it should be.

j9liu (commented):
Thanks @weegeekps ! This looks great to me, just one comment about conventions for our codebase.

tools/generate-classes/generate.js
weegeekps: Fixed.

tools/generate-classes/generate.js
```diff
@@ -48,6 +48,25 @@ function generate(options, schema, writers) {
...
+        // Skip invalid properties.
+        continue;
+      }
+
+      if (required.includes(prop)) {
+        // Don't add props to list if they're already in `required`.
+        continue;
+      }
+
+      required.push(prop);
```
j9liu:
```suggestion
    const validProperties = Object.keys(schema.properties || {});
    for (const property of requireOptionalProperties) {
      if (!validProperties.includes(property)) {
        // Skip invalid properties.
        continue;
      }

      if (required.includes(property)) {
        // Don't add properties to list if they're already in `required`.
        continue;
      }

      required.push(property);
```

weegeekps: Wow, I must be blind. :facepalm:

Issue #1321, opened 2026-03-11
Assess support for `EXT_mesh_primitive_restart`


j9liu:
`EXT_mesh_primitive_restart` is a recent extension that is meant to reduce draw calls for continuous drawing modes (e.g., line strip or loop) by batching primitives together. The max index value (which is typically illegal) is otherwise used to separate primitives from each other.

Support for "primitive restart" varies across graphics APIs, so it's not clear whether or not our supported runtimes can really benefit from this. But ultimately, we could just have a setting that toggles the batching behavior in Native if we encounter the extension. 

- [Spec](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Vendor/EXT_mesh_primitive_restart)
- [CesiumJS implementation](https://github.com/CesiumGS/cesium/pull/12792/changes)

j9liu:
Note: this may be affected by https://github.com/CesiumGS/glTF/pull/100

j9liu:
I was initially doing some handling directly in Cesium for Unreal's `getIndices` function, but realized there is  _some_ shared code with Cesium for Unity. Per the Unity [docs](https://docs.unity3d.com/6000.4/Documentation/ScriptReference/MeshTopology.html) there are no modes for `LINE_LOOP`, `TRIANGLE_STRIP` or `TRIANGLE_FAN`. The rewriting we do for the indices (to render as regular lines or triangles) may actually be better to put in Cesium Native.

This will be done in `GltfReader` with the option to pick which primitives' are converted depending on the client.

PR #1334, opened 2026-03-27
geoJSON tilesets with glTF translation

timoore:
Encode geoJson features to glTF geometry : points, lines, and triangles. Support the 3D Tiles MAXAR_content_geojson extension, which allows tilesets with GeoJSON content.


Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetOptions.h
```diff
@@ -55,6 +55,12 @@ struct CESIUM3DTILESSELECTION_API TilesetContentOptions {
    * shader.
    */
   bool applyTextureTransform = true;
+
+  /**
+   * @brief Whether to translate vector feature data, e.g. GeoJSON content, to
+   * glTF, or leave it in its native format.
+   */
+  bool translateVectorFeatures = true;
```
lilleyse, line 63: I believe this option can be removed now?

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -1155,15 +1212,16 @@ TilesetJsonLoader::loadTileContent(const TileLoadInput& loadInput) {
             } else {
               // not a renderable content, then it must be external tileset
```
lilleyse, line 1213: Maybe update this comment to something like:

```suggestion
              // Otherwise must be JSON content, either an external tileset or GeoJSON
```

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -770,27 +772,11 @@ TileLoadResult parseExternalTilesetInWorkerThread(
     const std::shared_ptr<CesiumAsync::IAssetAccessor>& pAssetAccessor,
     std::shared_ptr<CesiumAsync::IAssetRequest>&& pCompletedRequest,
     ExternalContentInitializer&& externalContentInitializer,
-    const CesiumGeospatial::Ellipsoid& ellipsoid) {
+    const CesiumGeospatial::Ellipsoid& ellipsoid,
+    rapidjson::Document& tilesetJson) {
```
lilleyse, line 776: Since `tilesetJson` is later moved I think rvalue reference would make more sense here.

```suggestion
    rapidjson::Document&& tilesetJson) {
```

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -770,27 +772,11 @@ TileLoadResult parseExternalTilesetInWorkerThread(
...
   // create external tileset
```
lilleyse, line 777: This comment can be removed now that the parsing code has moved elsewhere

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -841,6 +827,77 @@ TileLoadResult parseExternalTilesetInWorkerThread(
...
+    const std::shared_ptr<spdlog::logger>& pLogger,
+    const std::shared_ptr<CesiumAsync::IAssetAccessor>& pAssetAccessor,
+    std::shared_ptr<CesiumAsync::IAssetRequest>&& pCompletedRequest,
+    ExternalContentInitializer&& externalContentInitializer,
+    const CesiumGeospatial::Ellipsoid& ellipsoid,
+    bool translateToGltf) {
+  const CesiumAsync::IAssetResponse* pResponse = pCompletedRequest->response();
+  const auto& responseData = pResponse->data();
+
+  rapidjson::Document tilesetJson;
```
lilleyse, line 843: Calling this variable `tilesetJson` might be misleading. Maybe simply `json`?

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -841,6 +827,77 @@ TileLoadResult parseExternalTilesetInWorkerThread(
...
+  tilesetJson.Parse(
+      reinterpret_cast<const char*>(responseData.data()),
+      responseData.size());
+  if (tilesetJson.HasParseError()) {
+    SPDLOG_LOGGER_ERROR(
+        pLogger,
+        "Error when parsing tileset JSON, error code {} at byte offset {}",
```
lilleyse, line 850:
```suggestion
        "Error when parsing JSON, error code {} at byte offset {}",
```

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -841,6 +827,77 @@ TileLoadResult parseExternalTilesetInWorkerThread(
...
+        std::move(pCompletedRequest));
+  }
+  if (const auto typeIt = tilesetJson.FindMember("type");
+      typeIt != tilesetJson.MemberEnd()) {
+    auto geoJson = CesiumVectorData::GeoJsonDocument::fromGeoJson(tilesetJson);
+    if (!(geoJson.value.has_value() && !geoJson.errors.hasErrors() &&
+          translateToGltf)) {
+      return TileLoadResult::createFailedResult(
+          pAssetAccessor,
+          std::move(pCompletedRequest));
```
lilleyse, line 864: If there's an error it should probably be logged

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -841,6 +827,77 @@ TileLoadResult parseExternalTilesetInWorkerThread(
...
+    } else {
+
```
lilleyse, line 866: Remove whitspace

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -841,6 +827,77 @@ TileLoadResult parseExternalTilesetInWorkerThread(
...
+            std::nullopt,
+            pAssetAccessor,
+            std::move(pCompletedRequest),
+            {},
+            TileLoadResultState::Success,
+            ellipsoid};
+      } else {
+        return TileLoadResult::createFailedResult(
+            pAssetAccessor,
+            std::move(pCompletedRequest));
```
lilleyse, line 885: Same comment, any errors should be logged

CesiumVectorData/include/CesiumVectorData/GltfConverter.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include <CesiumGltf/Model.h>
+#include <CesiumVectorData/GeoJsonDocument.h>
+
+namespace CesiumVectorData {
+
+/**
+ * @brief The result of converting a GeoJSON file to a glTF model.
+ */
+struct CESIUMVECTORDATA_API ConverterResult {
```
lilleyse, line 11: Is it possible to use the `Result` class?

I see that `GltfConverterResult` is still its own class but I think we wanted to start using `Result` for new code.

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,511 @@
...
+  int32_t makeBufferView(int32_t buffer, int target);
+  int32_t makeAccessor(
+      int32_t bufferView,
+      int64_t byteOffset,
+      int64_t count,
+      int componentType = Accessor::ComponentType::FLOAT,
+      const std::string& accessorType = Accessor::Type::VEC3);
+};
+
+double GltfConverterImpl::heightOffset = 10.0;
```
lilleyse, line 59: I assume this was for testing and can be removed?

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,511 @@
...
+  glm::dmat4 enuToFixedFrame;
+  GltfConverterImpl(const GeoJsonDocument& inGeoJson)
+      : geoJson(inGeoJson), enuToFixedFrame(1.0) {}
+  static double heightOffset;
+  // Gather a feature type from GeoJson into a mesh in a node.
+  int32_t gatherLines();
+  int32_t gatherPolygons();
+  int32_t gatherPoints();
+  // Create a buffer view from a whole buffer
+  int32_t makeBufferView(int32_t buffer, int target);
```
lilleyse, line 50: Prefer fixed width integer types. Same comment throughout this file.

```suggestion
  int32_t makeBufferView(int32_t buffer, int32_t target);
```

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,511 @@
...
+      int64_t byteOffset,
+      int64_t count,
+      int componentType = Accessor::ComponentType::FLOAT,
+      const std::string& accessorType = Accessor::Type::VEC3);
+};
+
+double GltfConverterImpl::heightOffset = 10.0;
+
+int32_t GltfConverterImpl::makeBufferView(int32_t buffer, int target) {
+  int32_t bufferViewIndex = static_cast<int32_t>(model.bufferViews.size());
```
lilleyse, line 62: From the style guide:

> Use constructor notation for very safe integer conversions (where the range is checked) rather than static_cast.

Same comment for the rest of the file.

```suggestion
  int32_t bufferViewIndex = int32_t(model.bufferViews.size());
```

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,511 @@
...
+        numCoords++;
+        sum += coord;
+      }
+    }
+  }
+  if (numCoords > 0) {
+    sum /= static_cast<double>(numCoords);
+  }
+  return sum;
+}
```
lilleyse, line 156: Could you use `BoundingRegionBuilder` for this? That would be safer if the data spans across the antimeridian

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,511 @@
...
+      positions.end(),
+      outPositions.begin(),
+      [&toLocal](const glm::dvec3& cartoDegrees) {
+        CesiumGeospatial::Cartographic cartographic(Cartographic::fromDegrees(
+            cartoDegrees.x,
+            cartoDegrees.y,
+            cartoDegrees.z + GltfConverterImpl::heightOffset));
+        auto cartesian =
+            CesiumGeospatial::Ellipsoid::WGS84.cartographicToCartesian(
+                cartographic);
```
lilleyse, line 175: Is GeoJSON guaranteed to be WGS84? Might be better to pass in an ellipsoid than hardcode WGS84 here. Same comment below.

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,511 @@
...
+  }
+  if (cartoCoordinates.empty()) {
+    return -1;
+  }
+  std::vector<glm::dvec3> localPositions(cartoCoordinates.size());
+  transformIntoFrame(enuToFixedFrame, cartoCoordinates, localPositions);
+  int32_t bufferIndex = static_cast<int32_t>(model.buffers.size());
+  model.buffers.emplace_back();
+  std::vector<std::byte>& bytes = model.buffers.back().cesium.data;
+  bytes.resize(localPositions.size() * sizeof(glm::vec3));
```
lilleyse, line 210: Might be good to use `glm::fvec3` just to be explicit.

Same comment throughout.
```suggestion
  bytes.resize(localPositions.size() * sizeof(glm::fvec3));
```

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,511 @@
...
+    model.accessors[static_cast<size_t>(accessorIndex)].max =
+        positionMaxVector(stringCoords);
+    accessorByteOffset +=
+        model.accessors[static_cast<size_t>(accessorIndex)].count * 4 * 3;
+    elementCount += lineStringItr->coordinates.size();
+    linesMesh.primitives.emplace_back();
+    linesMesh.primitives.back().attributes["POSITION"] = accessorIndex;
+    linesMesh.primitives.back().mode = MeshPrimitive::Mode::LINE_STRIP;
+    linesMesh.primitives.back().material = 0;
+  }
```
lilleyse, line 243: It looks like each line string is converted to a separate primitive. That could result in a lot of draw calls. Could all line strings be batched into a single primitive? Either:

* Use `LINE_STRIP` with [`KHR_primitive_restart`](https://github.com/KhronosGroup/glTF/pull/2569) once we've verified that primitive restart is supported in all the native runtimes (including Godot)
* Convert to `LINES`

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,511 @@
...
+      linesAccessor.min = positionMinVector(stringCoords);
+      linesAccessor.max = positionMaxVector(stringCoords);
+      accessorByteOffset += linesAccessor.count * 4 * 3;
+      elementCount += lineStringCoords.size();
+      linesMesh.primitives.emplace_back();
+      linesMesh.primitives.back().attributes["POSITION"] = accessorIndex;
+      linesMesh.primitives.back().mode = MeshPrimitive::Mode::LINE_STRIP;
+      linesMesh.primitives.back().material = 0;
+    }
+  }
```
lilleyse, line 267: This is kinda similar to the code above. Possible to break out a common helper function?

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,511 @@
...
+  ConverterResult result;
+  GltfConverterImpl converter(geoJson);
+  const GeoJsonObject& root = converter.geoJson.rootObject;
+  glm::dvec3 centroid = computeCentroid(root);
+  //  local to global cartesian
+  converter.enuToFixedFrame = GlobeTransforms::eastNorthUpToFixedFrame(
+      Ellipsoid::WGS84.cartographicToCartesian(
+          Cartographic::fromDegrees(centroid.x, centroid.y, centroid.z)));
+  result.model.emplace();
+  converter.model.asset.version = "2.0";
```
lilleyse, line 486: Best practice would be to add `scene` and `scenes` to the glTF

```
  "scenes": [
    {
      "nodes": [
        0
      ]
    }
  ],
  "scene": 0
```

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,511 @@
...
+  model.nodes.back().mesh = meshIndex;
+  CesiumGltfContent::GltfUtilities::setNodeTransform(
+      model.nodes.back(),
+      enuToFixedFrame);
+  return nodeIndex;
+}
+
+std::vector<uint64_t>
+triangulatePolygon(const std::vector<std::vector<glm::dvec3>>& polygonIn) {
+  using N = uint64_t;
```
lilleyse, line 279: Since this function isn't templated I would just get rid of `N` and use `uint64_t` directly. Or `uint32_t` since the indices are downcasted to `uint32_t` later.

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,511 @@
...
+  using Point = std::array<double, 2>;
+  std::vector<std::vector<Point>> polygon;
+
+  for (const auto& ring : polygonIn) {
+    polygon.emplace_back();
+    for (const auto& coord : ring) {
+      polygon.back().push_back(Point{{coord[0], coord[1]}});
+    }
+  }
```
lilleyse, line 288: For performance, some of these vectors could be reserved in advance

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,511 @@
...
+  CesiumGltf::MaterialPBRMetallicRoughness& pbr =
+      material.pbrMetallicRoughness.emplace();
+  pbr.metallicFactor = 0.0;
+  pbr.roughnessFactor = 1.0;
+  pbr.baseColorFactor[0] = 0.0;
+  pbr.baseColorFactor[1] = 0.0;
+  pbr.baseColorFactor[2] = 0.0;
+  material.doubleSided = true;
+  // International orange
+  material.emissiveFactor = {0xff / 255.0, 0x4f / 255.0, 0.0};
```
lilleyse, line 497: Does setting the emissive have any impact on lighting in Unreal?

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,511 @@
...
+  result.model.emplace();
+  converter.model.asset.version = "2.0";
+  CesiumGltf::Material& material = converter.model.materials.emplace_back();
+  CesiumGltf::MaterialPBRMetallicRoughness& pbr =
+      material.pbrMetallicRoughness.emplace();
+  pbr.metallicFactor = 0.0;
+  pbr.roughnessFactor = 1.0;
+  pbr.baseColorFactor[0] = 0.0;
+  pbr.baseColorFactor[1] = 0.0;
+  pbr.baseColorFactor[2] = 0.0;
```
lilleyse, line 494: Also add `pbr.baseColorFactor[3] = 1.0;`?

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,511 @@
...
+  BufferView& bufferView = model.bufferViews.emplace_back();
+  bufferView.buffer = buffer;
+  bufferView.byteOffset = 0;
+  bufferView.byteLength = static_cast<int64_t>(
+      model.buffers[static_cast<size_t>(buffer)].cesium.data.size());
+  bufferView.target = target;
+  return bufferViewIndex;
+}
+
+int32_t GltfConverterImpl::makeAccessor(
```
lilleyse, line 72: These helper functions (`makeBufferView`, `makeAccessor`) could possibly be removed. Some areas call the helper functions but other areas just create the objects inline (which I think is clean enough).

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,511 @@
...
+  return triangulatePolygon(polygonIn.coordinates);
+}
+
+int32_t GltfConverterImpl::gatherPolygons() {
+  const GeoJsonObject& root = geoJson.rootObject;
+  std::vector<glm::dvec3> cartoCoordinates;
+  // GeoJson polygons contain an outer counter and possible inner contours. The
+  // first and last coordinates must be identical. An optimization would be to
+  // only include that coordinate once, but for now we will put both values into
+  // the big array of coordinates.
```
lilleyse, line 306: The implementation looks ok, but it would be good to get visual confirmation that holes are working. I found an example here: https://mapsam.com/geojson-ring-visual/

lilleyse (commented):
Thanks @timoore! A few comments from me, the only major one being about polyline batching.

Also, when you have a chance could you test this out in Unreal, Unity, and Godot?

CesiumVectorData/src/GltfConverter.cpp
lilleyse: Never mind, 1.0 is the default for all channels. Still, might not hurt to be explicit.

lilleyse:
Oh and a meta comment, @timoore could you update the PR description now that the PR is further along?

CesiumVectorData/src/GltfConverter.cpp
timoore: My intent is to use the helper functions throughout. The repetitive verbosity was driving me crazy .

CesiumVectorData/src/GltfConverter.cpp
timoore: It can be, but is there a way in the native runtimes to displace a tileset's elevation? It has been very handy to have this cheat to ensure that the features are visible over terrain.

CesiumVectorData/src/GltfConverter.cpp
timoore: I don't think primitive restart is supported in any of the native runtimes, but I'm happy to be wrong about that. I hesitated to implement it that way, both for forcing clients to deal with the glTF extension and for friction in testing. I agree that using LINES where possible may be good compromise.

CesiumVectorData/src/GltfConverter.cpp
timoore: > Could you use `BoundingRegionBuilder` for this? That would be safer if the data spans across the antimeridian

If the objective is to use the center of the furthest extents as the base for the tile transform, and not the average of the cartographic positions, then yes. On reflection that probably makes the most sense.

CesiumVectorData/src/GltfConverter.cpp
timoore: That would be completely new in cesium-native, as far as I can tell.  Although I've just learned that the definition of `glm::vec3` can be changed by a macro (who knew?), `vec3` would seem to be universally understood as having float element types, at least in the desktop world, referencing GLSL. I'm happy to have my mind changed if you think this should be introduced across all our code.

CesiumVectorData/src/GltfConverter.cpp
lilleyse: The transform tool in the unreal/unity editor should work.

CesiumVectorData/src/GltfConverter.cpp
lilleyse: Oh, maybe `glm::fvec3` is just a `tilers` convention. If cesium-native is using `glm::vec3` everywhere then fine to stick with that.

CesiumVectorData/src/GltfConverter.cpp
lilleyse: @j9liu and I were looking at this on Friday, it seems like Unity and Godot support `LINE_STRIP` but Unreal does not. Tested with the sample model in https://github.com/KhronosGroup/glTF/pull/2569. CC https://github.com/CesiumGS/cesium-native/issues/1321.

Fine to use `LINES` here.

CesiumVectorData/src/GltfConverter.cpp
lilleyse: Good with me to use the helpers throughout.

CesiumVectorData/src/GltfConverter.cpp
timoore: It is guaranteed to be WGS84, but I agree that the ellipsoid should come from the tileset options.

j9liu:
@timoore FYI I'm putting this in the release milestone since this is looking close to merge. Also please remember to update the PR description as @lilleyse pointed out!

CesiumVectorData/src/GltfConverter.cpp
timoore: No.

CesiumVectorData/src/GltfConverter.cpp
timoore: I confess to laziness in the current code, where I'm assigning an obnoxious orange to the emissive channel as well as the baseColorFactor.

timoore:
@lilleyse This is ready for another look.

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -124,6 +126,21 @@ struct ExternalContentInitializer {
...
+    const rapidjson::Value& extensionsRequired = extensionsRequiredIt->value;
+    if (extensionsRequired.IsArray()) {
+      for (const auto& v : extensionsRequired.GetArray()) {
+        if (v.IsString() &&
+            !std::strcmp(v.GetString(), "MAXAR_content_geojson")) {
+          return CesiumGeometry::Axis::Z;
+        }
+      }
+    }
+  }
```
lilleyse: Since we have full control over the GeoJSON to glTF conversion it might be cleaner if the up axis conversion happens in `GltfConverter`. [`PntsToGltfConverter`](https://github.com/CesiumGS/cesium-native/blob/9741cf270a1a578ef26c9f061397b5cb934da7c2/Cesium3DTilesContent/src/PntsToGltfConverter.cpp#L1461-L1465) does this by setting an up axis conversion on the root node. `GltfConverter` could do something like

```c++
  CesiumGltfContent::GltfUtilities::setNodeTransform(
      converter.model.nodes[rootNodeIndex],
      CesiumGeometry::Transforms::Z_UP_TO_Y_UP * converter.enuToFixedFrame);
```

CesiumVectorData/include/CesiumVectorData/GltfConverter.h
```diff
@@ -0,0 +1,32 @@
...
+
+namespace CesiumVectorData {
+
+/**
+ * @brief The result of converting a GeoJSON document to glTF.
+ */
+using ConverterResult = CesiumUtility::Result<CesiumGltf::Model>;
+
+/**
+ * @brief Functional for converting GeoJSON documents to glTF.
```
lilleyse:
```suggestion
 * @brief Converts a GeoJSON document to glTF.
```

CesiumVectorData/include/CesiumVectorData/GltfConverter.h
```diff
@@ -0,0 +1,32 @@
...
+ */
+class CESIUMVECTORDATA_API GltfConverter {
+public:
+  /**
+   * @brief Convert geoJSON document to a result containing a glTF Model
+   * object.
+   *
+   * @param geoJson The GeoJSON document.
+   * @param ellipsoid The ellipsoid for GeoJSON coordinates
```
lilleyse:
```suggestion
   * @param ellipsoid The ellipsoid for GeoJSON coordinates.
```

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,531 @@
...
+  Accessor& accessor = this->model.accessors.emplace_back();
+  accessor.bufferView = bufferView;
+  accessor.byteOffset = byteOffset;
+  accessor.componentType = componentType;
+  accessor.count = count;
+  accessor.type = accessorType;
+  return accessorIndex;
+}
+
+glm::dvec3 positionMin(const std::span<glm::dvec3> coords) {
```
lilleyse: From the [C++ guide](https://cesium.com/learn/cesium-native/ref-doc/style-guide.html):

> Don't use const for by-value parameters

Instead, to make the span immutable, put const in the type.

```suggestion
glm::dvec3 positionMin(std::span<const glm::dvec3> coords) {
```

Same comment below.

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,531 @@
...
+      this->ellipsoid);
+  int32_t bufferIndex = int32_t(this->model.buffers.size());
+  this->model.buffers.emplace_back();
+  std::vector<std::byte>& bytes = this->model.buffers.back().cesium.data;
+  bytes.resize(localPositions.size() * sizeof(glm::vec3));
+  glm::vec3* position32 = reinterpret_cast<glm::vec3*>(bytes.data());
+  for (const auto& localPosition : localPositions) {
+    *position32++ =
+        glm::vec3(localPosition.x, localPosition.y, localPosition.z);
+  }
```
lilleyse, line 251: Could this be a memcpy instead?
Same comment for polyons and points.

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,531 @@
...
+
+  int32_t nodeIndex = int32_t(this->model.nodes.size());
+  this->model.nodes.emplace_back();
+  this->model.nodes.back().mesh = meshIndex;
+  return nodeIndex;
+}
+
+std::vector<uint64_t>
+triangulatePolygon(const std::vector<std::vector<glm::dvec3>>& polygonIn) {
+  using N = uint64_t;
```
lilleyse: See previous comment: https://github.com/CesiumGS/cesium-native/pull/1334/changes#r3101256422

> Since this function isn't templated I would just get rid of `N` and use `uint64_t` directly. Or `uint32_t` since the indices are downcasted to `uint32_t` later.

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,531 @@
...
+  using Point = std::array<double, 2>;
+  std::vector<std::vector<Point>> polygon;
+
+  for (const auto& ring : polygonIn) {
+    polygon.emplace_back();
+    for (const auto& coord : ring) {
+      polygon.back().push_back(Point{{coord[0], coord[1]}});
+    }
+  }
```
lilleyse, line 309: See previous comment: https://github.com/CesiumGS/cesium-native/pull/1334/changes#r3101262175

> For performance, some of these vectors could be reserved in advance

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,531 @@
...
+  const GeoJsonObject& root = converter.geoJson.rootObject;
+  Cartographic centroid = computeCentroid(root);
+  //  local to global cartesian
+  converter.enuToFixedFrame = GlobeTransforms::eastNorthUpToFixedFrame(
+      ellipsoid.cartographicToCartesian(centroid));
+  converter.model.asset.version = "2.0";
+  Material& material = converter.model.materials.emplace_back();
+  MaterialPBRMetallicRoughness& pbr = material.pbrMetallicRoughness.emplace();
+  // International orange
+  std::array orange{0xff / 255.0, 0x4f / 255.0, 0.0};
```
lilleyse: I would use white so that it doesn't inadvertently affect styling later

(e.g. CesiumJS's [`colorBlendMode`](https://cesium.com/learn/cesiumjs/ref-doc/Cesium3DTileset.html#colorBlendMode) is a `HIGHLIGHT` by default, so white would be the best choice)

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -0,0 +1,531 @@
...
+  BufferView& bufferView = this->model.bufferViews.emplace_back();
+  bufferView.buffer = buffer;
+  bufferView.byteOffset = 0;
+  bufferView.byteLength =
+      int64_t(this->model.buffers[size_t(buffer)].cesium.data.size());
+  bufferView.target = target;
+  return bufferViewIndex;
+}
+
+int32_t GltfConverterImpl::makeAccessor(
```
lilleyse, line 79: Should `gatherPolygons` and `gatherPoints` also use `makeAccessor`?

lilleyse (commented):
Thanks @timoore, this is getting close.

Major things left would be writing tests and updating CHANGES.md

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
timoore: Ok. I thought it was more modèrne to use Z up in tilesets, as well as being a natural representation of the GeoJson source data, but I can change that.

CesiumVectorData/include/CesiumVectorData/GltfConverter.h
timoore: My bad, it isn't a functional any more.

CesiumVectorData/src/GltfConverter.cpp
timoore: No, because there is a double-to-float conversion. But `transformIntoFrame()` could do the conversion and store the results as float, perhaps directly into the glTF buffer.

CesiumVectorData/src/GltfConverter.cpp
timoore: Yes.

CesiumVectorData/src/GltfConverter.cpp
lilleyse: Oh right. No change needed here then.

lilleyse:
Looks good. Thanks @timoore and thanks @j9liu for the last minute polygon fix!

timoore:
@lilleyse Here's a test with the GeoJson at https://mapsam.com/geojson-ring-visual/ , which you mentioned. It looks fine, so the tessellation seems to be working. I'll add a test using the tileset I created to wrap that sample GeoJson.

image

PR #1337, opened 2026-03-31
Fix FeatureCollection visitor to check geometry style

### LINKED ISSUE #672: GeoJSONDocumentRasterOverlay styling not applying

baruchInsert-tech:
Fix GeoJSON FeatureCollection visitor not reading geometry styles
                                                     
The GeoJsonFeatureCollection visitor in GeoJsonDocumentRasterOverlay was checking the style on the Feature wrapper object instead of the geometry inside it. This caused programmatically-set styles via SetStyle() on geometry objects to be silently ignored when the root object is a FeatureCollection (the common case for GeoJSON documents).

Additionally, the style fallback chain incorrectly fell back to
pFeature->style a second time instead of collection.style, meaning collection-level styles were never consulted.

The GeoJsonFeature visitor already handled this correctly by checking feature.geometry->getStyle(). This change aligns the FeatureCollection visitor with that behavior.

Fixes CesiumGS/cesium-unity#672

## Author Checklist

- [x] I have submitted a Contributor License Agreement (only needed once).
- [x] I have updated CHANGES.md with a short summary of my change (for user-facing changes).

### DIFF
--- a/CesiumRasterOverlays/test/TestGeoJsonDocumentRasterOverlay.cpp
+++ b/CesiumRasterOverlays/test/TestGeoJsonDocumentRasterOverlay.cpp
@@ -81,6 +81,96 @@ CesiumGltf::ImageAsset rasterizeOverlayTile(
   }
   return *pTile->getImage();
 }
+
+CesiumGltf::ImageAsset rasterizeOverlayTileFromDocument(
+    const GlobeRectangle& rectangle,
+    const glm::dvec2& imageSize,
+    GeoJsonDocument&& document,
+    const GeoJsonDocumentRasterOverlayOptions& overlayOptions) {
+  AsyncSystem asyncSystem(
+      std::make_shared<CesiumNativeTests::SimpleTaskProcessor>());
+
+  IntrusivePointer<GeoJsonDocumentRasterOverlay> pOverlay;
+  pOverlay.emplace(
+      asyncSystem,
+      "overlay0",
+      std::make_shared<GeoJsonDocument>(std::move(document)),
+      overlayOptions);
+
+  std::shared_ptr<CesiumAsync::IAssetAccessor> pAssetAccessor =
+      std::make_shared<CesiumNativeTests::SimpleAssetAccessor>(
+          std::map<
+              std::string,
+              std::shared_ptr<CesiumNativeTests::SimpleAssetRequest>>());
+
+  IntrusivePointer<ActivatedRasterOverlay> pActivated = pOverlay->activate(
+      RasterOverlayExternals{
+          .pAssetAccessor = pAssetAccessor,
+          .pPrepareRendererResources = nullptr,
+          .asyncSystem = asyncSystem,
+          .pCreditSystem = nullptr,
+          .pLogger = spdlog::default_logger()},
+      Ellipsoid::WGS84);
+
+  pActivated->getReadyEvent().waitInMainThread();
+
+  REQUIRE(pActivated->getTileProvider() != nullptr);
+  GeographicProjection projection(Ellipsoid::WGS84);
+  const CesiumGeometry::Rectangle& tileRect = projection.project(rectangle);
+
+  IntrusivePointer<RasterOverlayTile> pTile =
+      pActivated->getTile(tileRect, imageSize);
+  pActivated->loadTile(*pTile);
+  while (pTile->getState() != RasterOverlayTile::LoadState::Loaded) {
+    asyncSystem.dispatchMainThreadTasks();
+  }
+  return *pTile->getImage();
+}
+
+bool imageHasPixelWithColor(
+    const CesiumGltf::ImageAsset& image,
+    uint8_t r,
+    uint8_t g,
+    uint8_t b,
+    uint8_t a) {
+  REQUIRE(image.channels == 4);
+  REQUIRE(image.bytesPerChannel == 1);
+  for (int32_t y = 0; y < image.height; y++) {
+    for (int32_t x = 0; x < image.width; x++) {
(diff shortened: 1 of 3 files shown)


baruchInsert-tech:
@azrogers please take a look when available 


This means cesium-unreal would also exhibit this issue if a user sets a style on the geometry object within a FeatureCollection and refreshes. It likely hasn't been reported there because users either haven't hit this exact code path, or they set styles at a different level (e.g., on the Feature or the default style, which do work).

CesiumRasterOverlays/test/TestGeoJsonDocumentRasterOverlay.cpp
```diff
@@ -81,6 +81,96 @@ CesiumGltf::ImageAsset rasterizeOverlayTile(
   }
   return *pTile->getImage();
 }
+
+CesiumGltf::ImageAsset rasterizeOverlayTileFromDocument(
```
azrogers: This seems to be most of the contents of `rasterizeOverlayTile`, only taking an existing document instead of a path to a GeoJSON file. I would recommend changing the implementation of `rasterizeOverlayTile` to call `rasterizeOverlayTileFromDocument` after the document is loaded, to avoid duplicating the code.

CesiumRasterOverlays/test/TestGeoJsonDocumentRasterOverlay.cpp
```diff
@@ -81,6 +81,96 @@ CesiumGltf::ImageAsset rasterizeOverlayTile(
...
+    uint8_t r,
+    uint8_t g,
+    uint8_t b,
+    uint8_t a) {
+  REQUIRE(image.channels == 4);
+  REQUIRE(image.bytesPerChannel == 1);
+  for (int32_t y = 0; y < image.height; y++) {
+    for (int32_t x = 0; x < image.width; x++) {
+      size_t idx = static_cast<size_t>((y * image.width + x) * 4);
+      if (idx + 3 < image.pixelData.size() &&
```
azrogers: I would raise the `idx + 3 < image.pixelData.size()` check up to the top of the function, with something like `REQUIRE(image.width * image.height * image.channels * image.bytesPerChannel == image.pixelData.size())`. If the image doesn't have enough data for the pixels it's supposed to contain that indicates a different issue than the image not having a pixel of that color, so it should be raised as such.

azrogers:
Looks like the core issue here is a copy-and-paste error on my part. Thanks for the fix! A few suggestions for the tests file and a merge conflict that needs resolved, but if that's taken care of we can get this into the next release!

CesiumRasterOverlays/test/TestGeoJsonDocumentRasterOverlay.cpp
baruchInsert-tech: Thanks! Refactored rasterizeOverlayTile to load the document and then delegate to rasterizeOverlayTileFromDocument, removing the duplicated setup/activation/tile-load code

CesiumRasterOverlays/test/TestGeoJsonDocumentRasterOverlay.cpp
baruchInsert-tech: Hoisted it to a REQUIRE(image.width * image.height   * image.channels * image.bytesPerChannel == image.pixelData.size()) at the top of imageHasPixelWithColor, and dropped   the per-pixel idx + 3 < size() guard from the loop.

baruchInsert-tech:
> Looks like the core issue here is a copy-and-paste error on my part. Thanks for the fix! A few suggestions for the tests file and a merge conflict that needs resolved, but if that's taken care of we can get this into the next release!

Done!

azrogers:
Thanks @baruchInsert-tech!

Issue #36, opened 2020-11-04
Consider exposing fewer of the internal Tile functions to the user


javagl:
At the time of writing this, the [`Tile` class has >30 functions that are `public`](https://github.com/CesiumGS/cesium-native/blob/7260ddca4d8ec2bcab0aa87f3c34b5938d01b556/Cesium3DTiles/include/Cesium3DTiles/Tile.h#L125). 

I'm listing them here (omitting possible `const` versions), as an overview, and from a **very** short glance (just to have an estimate!), only the first half of this list should actually be `public`:

        Tileset* getTileset();
        TileContext* getContext();
        Tile* getParent();
        gsl::span<Tile> getChildren();
        const BoundingVolume& getBoundingVolume() const;
        const std::optional<BoundingVolume>& getViewerRequestVolume() const;
        const std::optional<BoundingVolume>& getContentBoundingVolume() const;
        double getGeometricError() const;
        TileRefine getRefine() const;
        const glm::dmat4x4& getTransform() const;
        TileContentLoadResult* getContent();
        void* getRendererResources() const;
        bool isRenderable() const;

        void prepareToDestroy();
        void setContext(TileContext* pContext);
        void setParent(Tile* pParent);
        void createChildTiles(size_t count);
        void createChildTiles(std::vector<Tile>&& children);
        void setBoundingVolume(const BoundingVolume& value);
        void setViewerRequestVolume(const std::optional<BoundingVolume>& value);
        void setGeometricError(double value);
        void setRefine(TileRefine value);
        void setTransform(const glm::dmat4x4& value);
        const TileID& getTileID() const;
        void setTileID(const TileID& id);
        void setContentBoundingVolume(const std::optional<BoundingVolume>& value);
        LoadState getState() const;
        TileSelectionState& getLastSelectionState();
        void setLastSelectionState(const TileSelectionState& newState);
        void loadContent();
        bool unloadContent();
        void update(uint32_t previousFrameNumber, uint32_t currentFrameNumber);

The second part (particularly, basically all `set...` functions) are only or mainly used for the initialization and setup, and maintaining the tiles structure during the rendering process. Specifically, most of them are only called by the `Tileset` (and few of them by other internal classes). 

One could consider to reduce the size of the "surface" of this class: Users should not call things like `tile->setParent(nullptr)`, and thus, should preferably not be *able* to do this.

(*Some* of the function may also only be relevant for certain use-cases. This should be pointed out in the documentation)

[...]

PR #1342, opened 2026-04-09
Extract `Tile` selection algorithms from `Tileset` into free functions

calebbuffa:
## Description

This PR breaks tile selection algorithms out of private methods in the `Tileset` class and into free functions. The purpose is to reduce the surface area of `Tileset` and make the selection algorithms independently testable.

#### 1) — `tileStateUpdater` callback: decouple traversal from loading

The traversal helpers (`_visitTileIfNeeded`, etc.) previously called `_pTilesetContentManager->updateTileContent()` directly, creating a compile-time dependency on `TilesetContentManager` inside the selection algorithm. A `std::function<void(Tile&)> tileStateUpdater` field was added to `TilesetFrameState`. `Tileset::updateViewGroup` injects the implementation via a lambda; traversal calls the callback. The traversal has no `#include` dependency on `TilesetContentManager`.

**Files:** `TilesetFrameState.h`, `Tileset.cpp`

#### 2) — `selectTiles()`: pure free function for tile LOD selection

The entire traversal algorithm (`_visitTileIfNeeded`, `_visitTile`, `_renderLeaf`, `_renderInnerTile`, `_visitVisibleChildrenNearToFar`, `_kickDescendantsAndRenderTile`, `_loadAndRenderAdditiveRefinedTile`, frustum/fog cull, occlusion check, SSE computation) was extracted from `Tileset` private methods into free functions in a new translation unit `TilesetSelection.cpp`. The public entry point is:

```cpp
selectTiles(
    TileSelectionContext& context,   // options + externals + scratch buffers
    TilesetFrameState& frameState,
    Tile& rootTile,
    ViewUpdateResult& result,
);
```

`Tileset::updateViewGroup` is now a thin coordinator: it builds `frameState` and calls `selectTiles()`. The private traversal method declarations and `TraversalDetails`/`CullResult` structs are gone from `Tileset.h`.

**Files:** `TilesetSelection.h` (new), `TilesetSelection.cpp` (new), `Tileset.h`, `Tileset.cpp`

---

## Issue number or link

N/A.

---

## Author checklist

- [x] I have submitted a Contributor License Agreement (only needed once).

---

## Testing plan

**Regression:** All existing tests in `TestTilesetSelectionAlgorithm.cpp` exercise the full selection pipeline via `Tileset::updateViewGroup`, which now delegates to `selectTiles()`. These cover: replace/additive refinement, SSE thresholds, frustum/fog culling, multi-frustum, `forbidHoles`, LOD transition fade-out, `unconditionallyRefine`.

**New isolation test (`TestTilesetSelection.cpp`):**

- *`selectTiles is callable as a free function`* — constructs a minimal `EllipsoidTilesetLoader`-backed tileset, calls `selectTiles()` directly (bypassing `Tileset::updateViewGroup`), and asserts at least one tile is selected with no tiles kicked.
- *`selectTiles result matches updateViewGroup result`* — runs both `updateViewGroup` and `selectTiles()` on the same warmed-up tileset and asserts tile render count and visit count are identical between the two paths.

**Build verification:** `cmake --build build --target Cesium3DTilesSelection` succeeds with zero errors.

---

## Remaining Tasks

Things that could be done: 

- `TilesetViewGroup::startNewFrame`/`finishFrame` still take `const Tileset&`, which prevents fully async-free unit tests of `selectTiles`. Decoupling that is a natural follow-up once this PR lands.


Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tile.h
```diff
@@ -684,7 +687,10 @@ class CESIUM3DTILESSELECTION_API Tile final {
   void setMightHaveLatentChildren(bool mightHaveLatentChildren) noexcept;
 
   // Position in bounding-volume hierarchy.
-  Tile* _pParent;
+  // Non-owning: the child observes the parent but does not extend its lifetime.
+  // Reference-count bookkeeping (addReference/releaseReference) is performed
+  // explicitly in Tile.cpp whenever the child's own reference count changes.
+  TileObserver<Tile> _pParent;
```
kring: I don't think this is necessary.

We follow the C++ Core Guidelines, which explicitly states that raw pointers are non-owning:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rr-ptr

At one point, a new `observer_ptr` was considered for the C++ standard. Here's Bjarne arguing against it:
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1408r0.pdf

Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tile.h
```diff
@@ -697,11 +703,13 @@ class CESIUM3DTILESSELECTION_API Tile final {
   TileRefine _refine;
   glm::dmat4x4 _transform;
 
-  // tile content
+  // tile content — both fields are [main-thread-only]; access via
+  // getContent() / getState() / setState() which enforce the threading
+  // contract in debug builds via MainThreadOnly<T>.
   CesiumUtility::DoublyLinkedListPointers<Tile> _unusedTilesLinks;
-  TileContent _content;
+  MainThreadOnly<TileContent> _content;
```
kring: This entire class is "main thread only". I think it's misleading to call out these two fields explicitly.

It's actually not strictly required to always access a `Tile` on the main thread. However, the caller is responsible for ensuring that a single instance is only used from one thread at a time. So it's valid to create it on one thread and "transfer" it to another (we do this during loading of external tilesets). It's also valid - but unusual! - to call `updateViewGroup` and `loadTiles` on a different thread each time, so long as they're only called from one thread at a time. In practice you'll hit assertion failures in `ReferenceCountedNonThreadSafe` in debug builds if you do this, though.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/ThreadSafety.h
```diff
@@ -0,0 +1,124 @@
...
+ * MainThreadOnly<TileContent> _content;
+ * MainThreadOnly<TileLoadState> _loadState;
+ *
+ * // Access (debug: asserts main thread; release: direct access):
+ * TileContent& c = _content.get();
+ * ```
+ *
+ * @tparam T The wrapped value type.
+ */
+template <typename T> class MainThreadOnly {
```
kring: This would be more appropriate in CesiumUtility.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileHierarchy.h
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * @brief Owns the root of the tile tree.
+ *
+ * All Tile instances are either held here (as the root) or owned
+ * by their parent via `Tile::_children`. This is the single point of
+ * tree ownership — it is non-copyable and non-shared.
+ *
+ * Access must occur on the main thread. [main-thread-only]
+ */
+class TileHierarchy {
```
kring: I don't think this class adds any value. It has the exact same semantics as `std::unique_ptr<Tile>`.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetSelection.h
```diff
@@ -0,0 +1,54 @@
...
+ * @param scratchDistances Caller-owned scratch buffer.
+ * @param scratchOcclusionProxies Caller-owned scratch buffer.
+ * @return The frame's render result and statistics.
+ */
+CESIUM3DTILESSELECTION_API ViewUpdateResult selectTiles(
+    const TileSelectionContext& ctx,
+    const TilesetFrameState& frameState,
+    Tile& rootTile,
+    std::vector<double>& scratchDistances,
+    std::vector<const TileOcclusionRendererProxy*>& scratchOcclusionProxies);
```
kring: Did you consider putting these in `TileSelectionContext` instead? I think that might be a little cleaner. Especially considering we might need more scratch vectors in the future and we ideally wouldn't bother the caller with that.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetSelection.h
```diff
@@ -0,0 +1,54 @@
...
+ * @param frameState Per-frame view parameters and the view group to update.
+ * @param rootTile Root of the tile hierarchy to traverse.
+ * @param scratchDistances Caller-owned scratch buffer.
+ * @param scratchOcclusionProxies Caller-owned scratch buffer.
+ * @return The frame's render result and statistics.
+ */
+CESIUM3DTILESSELECTION_API ViewUpdateResult selectTiles(
+    const TileSelectionContext& ctx,
+    const TilesetFrameState& frameState,
+    Tile& rootTile,
```
kring: Sure would be nice if this could be const! If I remember correctly, the main reason it can't be is because of the `lodTransitionFadePercentage`, which doesn't work correctly with multiple view groups anyway. Outside the scope of this PR, though.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileUnloadQueue.h
```diff
@@ -0,0 +1,68 @@
...
+#include <Cesium3DTilesSelection/Tile.h>
+
+namespace Cesium3DTilesSelection {
+
+/**
+ * @brief Tracks tiles whose content may be evicted when memory is low.
+ *
+ * Tiles at the head are least-recently-used; tiles at the tail are
+ * most-recently-used. All methods must be called from the main thread.
+ * [main-thread-only]
```
kring: I think this class plays by our thread safety rules just fine. There's no need to further restrict it to "main thread only".
https://cesium.com/learn/cesium-native/ref-doc/thread-safety-rules.html

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileUnloadQueue.h
```diff
@@ -0,0 +1,68 @@
...
+  /** @brief Returns true when `tile` is currently in the candidate list.
+   * [main-thread] */
+  bool contains(const Tile& tile) const noexcept {
+    return _queue.contains(tile);
+  }
+
+  /**
+   * @brief Directly removes `tile` from the list.
+   * Prefer markIneligible; this exists for sites that call remove()
+   * unconditionally. [main-thread]
```
kring: This comment is inaccurate. The implementations of `markIneligible` and `remove` are identical.

Cesium3DTilesSelection/src/TileOverlaySystem.h
```diff
@@ -0,0 +1,63 @@
...
+ * This is the single point of ownership for all overlay-related resources.
+ * In the current design tiles hold references into data owned here.
+ *
+ * The upsampler is a `TilesetContentLoader` specialisation whose `setOwner()`
+ * must be called by the containing `TilesetContentManager` after construction,
+ * because the owner back-reference is to the manager, not the system.
+ *
+ * [main-thread-only]
+ */
+class TileOverlaySystem {
```
kring: I really don't see much value in this kind of "holder" class.

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,895 @@
...
+};
+
+struct CullResult {
+  bool shouldVisit = true;
+  bool culled = false;
+};
+
+enum class VisitTileAction { Render, Refine };
+
+// Small helpers (previously in anonymous namespace in Tileset.cpp)
```
kring: No need to document the code history in comments.

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,895 @@
...
+
+TileSelectionState getPreviousState(
```
kring, line 98: It's a bit of a stylistic thing I suppose, but I usually like to put the main entry point function near the top of the file so it's easy to find, followed by other functions in a sensible order so that someone unfamiliar with the code can read from top to bottom with an increasing level of detail as they go. The downside, of course, is that function declarations are required. We haven't been completely consistent about this in this code base, though.

kring (commented):
Thanks for the PR @calebbuffa!

I love the way you've broken the tileset selection algorithm into a separate, testable function.

Some of the other changes seem mostly unrelated to this important core, though, and I have some quibbles with them. See below.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetSelection.h
calebbuffa: Agreed! I will make an issue to follow up on this!

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetSelection.h
calebbuffa: Good idea, I moved the scratch vectors into the ctx

Cesium3DTilesSelection/src/TilesetSelection.cpp
calebbuffa: moved `selectTiles` to the top!

Cesium3DTilesSelection/include/Cesium3DTilesSelection/ThreadSafety.h
calebbuffa: I went ahead and deleted `MainThreadOnly`

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileUnloadQueue.h
calebbuffa: Thanks for catchign this!

calebbuffa:
@kring thank you for the feedback! I got a little too carried away. I stripped the changes just to `selectTiles` and the `TilesUnloadQueue`. Thanks for refocusing!

kring:
I asked Copilot to review this PR, specifically looking for any changes that were inadvertently introduced when 
moving code from Tileset.cpp to TilesetSelection.cpp. Both observations look correct to me. However, I suggest moving the orphaned comment into the new file instead of deleting it.

---

I reviewed the code moved from `Tileset.cpp` to `TilesetSelection.cpp` for inadvertent changes. The core traversal logic appears to be a faithful move — no inadvertent behavioral differences found.

Two issues:

**1. Orphaned doc comment in `Tileset.h`**

When `_loadAndRenderAdditiveRefinedTile` was removed, its full `/** ... */` doc comment was accidentally left behind. It now sits immediately before `_unloadCachedTiles`, making it appear to document that function:

```cpp
private:
  /**
   * @brief When called on an additive-refined tile, queues it for load and adds
   * it to the render list.
   * ...
   * @return false The non-additive-refined tile was ignored.
   */
  // ← _loadAndRenderAdditiveRefinedTile declaration was deleted, but this comment wasn't

  void _unloadCachedTiles(double timeBudget) noexcept;
```

This happened because the deletion straddled a diff hunk boundary — the `/**` and `@brief` lines fell in the first hunk, the closing `@param`/`@return`/`*/` lines in the second. The function declaration was correctly deleted but the entire comment body survived as context lines in both hunks.

Fix: delete the orphaned `/** ... */` block from `Tileset.h`.

**2. Unnecessary `const_cast` introduced in `checkOcclusion`**

In `TilesetSelection.cpp`, the new `checkOcclusion` accesses `ctx.childOcclusionProxies` via a spurious `const_cast`:

```cpp
auto& childOcclusionProxies =
    const_cast<std::vector<const TileOcclusionRendererProxy*>&>(
        ctx.childOcclusionProxies);
```

`childOcclusionProxies` is declared as `std::vector<const TileOcclusionRendererProxy*>&` (a non-const reference member) in `TraversalContext`. Reference members retain their mutability when accessed through a `const` struct reference in C++, so the cast is unnecessary. The original `_checkOcclusion` in `Tileset.cpp` modified `this->_childOcclusionProxies` directly without any cast.

Fix: remove the `const_cast` and use `ctx.childOcclusionProxies` directly.

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -424,11 +413,23 @@ const ViewUpdateResult& Tileset::updateViewGroup(
...
   if (!frustums.empty()) {
     viewGroup.startNewFrame(*this, frameState);
-    this->_visitTileIfNeeded(frameState, 0, false, *pRootTile, result);
+    TileSelectionContext ctx{
+        this->_options,
+        this->_externals,
+        this->_distances,
+        this->_childOcclusionProxies};
+    ViewUpdateResult selectionResult = selectTiles(ctx, frameState, *pRootTile);
+    result = std::move(selectionResult);
```
kring, line 432: This is a small step backwards. Previously, the `ViewUpdateResult` was reused each frame, allowing us to avoid heap-allocating each of the vectors it holds. Now, a new ViewUpdateResult is created every time, and then moved into the persistent one. It's a pretty big benefit to avoid any heap allocations in the steady-state, every-frame code path, because heap allocations are somewhat slow and doing them repeatedly can lead to heap fragmentation. It's a little ugly, but a reasonable solution to this is pass ViewUpdateResult into the function as scratch in the same way as the other two.

kring, line 432: I just noticed we were already heap allocating `fogDensities` every frame, which is also not great. It's outside the scope of this pull request, but it would be nice to fix that, too.

kring, line 432: After I noticed this new heap allocation, I asked Copilot to look for any such problems. It didn't find any others, but it noticed that this isn't just a performance problem. There are actual bugs introduced by allocating a new `ViewUpdateResult` each frame as well:

---

I also checked for new heap allocations introduced by the PR. Three issues:

**1. Per-frame vector reallocation** (performance regression)

`selectTiles()` constructs a fresh `ViewUpdateResult` (with empty vectors) each frame. `startNewFrame()` carefully calls `.clear()` on `tilesToRenderThisFrame` and `tileScreenSpaceErrorThisFrame` to preserve their capacity across frames. The PR discards that capacity preservation — the fresh vectors in `selectTiles` always start at capacity 0 and must reallocate as tiles are added. After warmup, the old code had zero vector reallocations per frame for these; the PR regresses to O(log N) reallocations per frame.

**2. `tilesFadingOut` persistence lost** (correctness bug)

When `enableLodTransitionPeriod = true`, `startNewFrame()` intentionally does **not** clear `tilesFadingOut`, so tiles mid-fade persist across frames for `_updateLodTransitions` to keep advancing their percentage. Because the PR replaces `_updateResult` wholesale via `result = std::move(selectionResult)`, those persisting tiles are discarded every frame. Tiles that aren't re-added by the current traversal (because their state is already `Culled`, not `Rendered`) vanish from `tilesFadingOut` after one frame, cutting their fade-out short.

**3. `frameNumber` reset** (minor bug)

`startNewFrame()` increments `_updateResult.frameNumber`. The subsequent `result = std::move(selectionResult)` replaces `_updateResult` with a default-constructed `ViewUpdateResult` (`frameNumber = 0`), discarding the increment. `frameNumber` will always be 0 in the returned result, breaking `DebugTileStateDatabase`.

All three stem from the same root cause: `selectTiles` constructs and returns a `ViewUpdateResult` by value rather than populating the existing one in-place. The fix would be to change `selectTiles` to accept a `ViewUpdateResult&` output parameter (matching the original `_visitTileIfNeeded` pattern), or to preserve and restore the affected fields around the move.

As a side note, the `tileStateUpdater` lambda captures only `this` (one pointer), so standard small-buffer optimisation should prevent any heap allocation from the new `std::function` member in `TilesetFrameState` — that part is fine.

kring (commented):
A few more comments. Other than this, this PR looks good to me! I'll let @azrogers or @j9liu merge it, though.

j9liu:
Hi @calebbuffa! We're interested in getting this into the next release, are you able to address Kevin's comments above within the next few days?

calebbuffa:
@j9liu yes! Sorry got a little sidetracked the last few days. I'll fix the PR by end of week!

j9liu:
Thanks for the updates @calebbuffa! Just for transparency, I think I'll actually defer merging this to after our May 1 release if that's alright. We'll still review this soon, but it would help to have additional time to stress-test the changes, just in case there are any surprises in our runtimes.

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+  const TilesetFrameState& frameState;
+  std::vector<double>& distances;
+  std::vector<const TileOcclusionRendererProxy*>& childOcclusionProxies;
+};
+
+struct TraversalDetails {
+  bool allAreRenderable = true;
+  bool anyWereRenderedLastFrame = false;
+  uint32_t notYetRenderableCount = 0;
+};
```
j9liu, line 98: I think it would still be valuable to preserve comments that were here when this was defined in `Tileset.h` (minus allusions to outdated implementation details).

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetSelection.h
```diff
@@ -0,0 +1,49 @@
...
+ * Populates `result` with the tiles to render this frame. The view group in
+ * `frameState` also receives load queue updates.
+ *
+ * @param ctx Configuration, external dependencies, and scratch buffers.
+ * @param frameState Per-frame view parameters and the view group to update.
+ * @param rootTile Root of the tile hierarchy to traverse.
+ * @param result Filled with the selected tiles and statistics for this frame.
+ */
+CESIUM3DTILESSELECTION_API void selectTiles(
+    const TileSelectionContext& ctx,
```
j9liu: Hm, I can't find where we mention it in the style guide, but in general we avoid abbreviations for better code readability. Since this is public API, it's better to be overly explicit and write out everything unless truly cumbersome.

```suggestion
 * @param context Configuration, external dependencies, and scratch buffers.
 * @param frameState Per-frame view parameters and the view group to update.
 * @param rootTile Root of the tile hierarchy to traverse.
 * @param result Filled with the selected tiles and statistics for this frame.
 */
CESIUM3DTILESSELECTION_API void selectTiles(
    const TileSelectionContext& context,
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileUnloadQueue.h
```diff
@@ -0,0 +1,68 @@
...
+   */
+  void remove(Tile& tile) noexcept { _queue.remove(tile); }
+
+  /** @brief Returns the least-recently-used tile, or nullptr. [main-thread] */
+  Tile* head() noexcept { return _queue.head(); }
+
+  /** @brief Returns the tile following `tile` in LRU order. [main-thread] */
+  Tile* next(Tile& tile) noexcept { return _queue.next(tile); }
```
j9liu: Per our [style guide](https://cesium.com/learn/cesium-native/ref-doc/style-guide.html#other):

> Use `this->` when accessing member variables and functions. This helps distinguish member variables from local variables and member functions from anonymous namespace functions.

If we could `this->_queue` throughout the file that'd be great!

Cesium3DTilesSelection/src/Tileset.cpp
```diff
@@ -424,11 +413,22 @@ const ViewUpdateResult& Tileset::updateViewGroup(
...
+      frustums,
+      std::move(fogDensities),
+      [this](Tile& tile) {
+        this->_pTilesetContentManager->updateTileContent(tile, this->_options);
+      }};
 
   if (!frustums.empty()) {
     viewGroup.startNewFrame(*this, frameState);
-    this->_visitTileIfNeeded(frameState, 0, false, *pRootTile, result);
+    TileSelectionContext ctx{
```
j9liu, line 426: Avoid abbreviations unless the full word(s) are truly cumbersome.

```suggestion
    TileSelectionContext context{
```

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+    Tile& tile,
+    ViewUpdateResult& result) {
+  addCurrentTileToTilesFadingOutIfPreviouslyRendered(viewGroup, tile, result);
+  addCurrentTileDescendantsToTilesFadingOutIfPreviouslyRendered(
+      viewGroup,
+      tile,
+      result);
+}
+
+bool isVisibleFromCamera(
```
j9liu, line 209: IMO it's valuable to keep the previous comments around, even for the more straightforward private functions.

```suggestion
/**
 * @brief Returns whether a tile with the given bounding volume is visible for
 * the camera.
 *
 * @param viewState The {@link ViewState}
 * @param boundingVolume The bounding volume of the tile
 * @param forceRenderTilesUnderCamera Whether tiles under the camera should
 * always be considered visible and rendered (see
 * {@link Cesium3DTilesSelection::TilesetOptions}).
 * @return Whether the tile is visible according to the current camera
 * configuration
 */
bool isVisibleFromCamera(
```

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+      viewState.getPositionCartographic();
+  std::optional<GlobeRectangle> maybeRectangle =
+      estimateGlobeRectangle(boundingVolume, ellipsoid);
+  if (position && maybeRectangle) {
+    return maybeRectangle->contains(position.value());
+  }
+  return false;
+}
+
+bool isVisibleInFog(double distance, double fogDensity) noexcept {
```
j9liu, line 241: Preserve documentation:

```suggestion
/**
 * @brief Returns whether a tile at the given distance is visible in the fog.
 *
 * @param distance The distance of the tile bounding volume to the camera
 * @param fogDensity The fog density
 * @return Whether the tile is visible in the fog
 */
bool isVisibleInFog(double distance, double fogDensity) noexcept {
```

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+    double sse,
+    bool culled) noexcept {
+  return culled ? !ctx.options.enforceCulledScreenSpaceError ||
+                      sse < ctx.options.culledScreenSpaceError
+                : sse < ctx.options.maximumScreenSpaceError;
+}
+
+bool isLeaf(const Tile& tile) noexcept { return tile.getChildren().empty(); }
+
+bool mustContinueRefiningToDeeperTiles(
```
j9liu, line 356: Preserve documentation:

```suggestion
/**
 * @brief Determines if we must refine this tile so that we can continue
 * rendering the deeper descendant tiles of this tile.
 *
 * If this tile was refined last frame, and is not yet renderable, then we
 * should REFINE past this tile in order to continue rendering the deeper tiles
 * that we rendered last frame, until such time as this tile is loaded and we
 * can render it instead. This is necessary to avoid detail vanishing when
 * the camera zooms out and lower-detail tiles are not yet loaded.
 *
 * @param tile The tile to check, which is assumed to meet the SSE for
 * rendering.
 * @param lastFrameSelectionState The selection state of this tile last frame.
 * @return True if this tile must be refined instead of rendered, so that we can
 * continue rendering deeper tiles.
 */
bool mustContinueRefiningToDeeperTiles(
```

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+    const BoundingVolume& boundingVolume,
+    const Ellipsoid& ellipsoid,
+    bool forceRenderTilesUnderCamera) {
+  if (viewState.isBoundingVolumeVisible(boundingVolume)) {
+    return true;
+  }
+  if (!forceRenderTilesUnderCamera) {
+    return false;
+  }
+  const std::optional<CesiumGeospatial::Cartographic>& position =
```
j9liu, line 220: I think this TODO got lost in the move:

```suggestion
  // TODO: it would be better to test a line pointing down (and up?) from the
  // camera against the bounding volume itself, rather than transforming the
  // bounding volume to a region.
  const std::optional<CesiumGeospatial::Cartographic>& position =
```

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+    Tile& rootTile,
+    ViewUpdateResult& result) {
+  TraversalContext tctx{
+      ctx.options,
+      ctx.externals,
+      frameState,
+      ctx.scratchDistances,
+      ctx.scratchOcclusionProxies};
+
+  visitTileIfNeeded(tctx, 0, false, rootTile, result);
```
j9liu: NITPICK: `TraversalContext` seems largely duplicative with `TileSelectionContext` so there's not much gain from its definition. Maybe just pass the `TilesSelectionContext` + `frameState` separately?

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+            previousState.getResult(),
+            *pTile,
+            result);
+      });
+
+  std::vector<Tile::ConstPointer>& renderList = result.tilesToRenderThisFrame;
+  std::vector<double>& sseList = result.tileScreenSpaceErrorThisFrame;
+  renderList.erase(
+      renderList.begin() +
+          static_cast<std::vector<Tile*>::iterator::difference_type>(
```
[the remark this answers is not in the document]
j9liu: Ignore Copilot, this was here before this PR :)

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+    }
+  }
+}
+
+TileOcclusionState
+checkOcclusion(const TraversalContext& ctx, const Tile& tile) {
+  const std::shared_ptr<TileOcclusionRendererProxyPool>& pOcclusionPool =
+      ctx.externals.pTileOcclusionProxyPool;
+  if (!pOcclusionPool) {
+    return TileOcclusionState::NotOccluded;
```
j9liu, line 485: Worth marking that this is an intentional decision in the face of the missing parameter:

```suggestion
  // No occlusion pool, treat everything as unoccluded.
    return TileOcclusionState::NotOccluded;
```

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+    traversalDetails.allAreRenderable &= childTraversal.allAreRenderable;
+    traversalDetails.anyWereRenderedLastFrame |=
+        childTraversal.anyWereRenderedLastFrame;
+    traversalDetails.notYetRenderableCount +=
+        childTraversal.notYetRenderableCount;
+  }
+  return traversalDetails;
+}
+
+TraversalDetails visitTile(
```
j9liu, line 810: This function has several internal comments that got lost in the move; could you restore them?

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+
+enum class VisitTileAction { Render, Refine };
+
+TraversalDetails visitTileIfNeeded(
+    const TraversalContext& ctx,
+    uint32_t depth,
+    bool ancestorMeetsSse,
+    Tile& tile,
+    ViewUpdateResult& result);
+
```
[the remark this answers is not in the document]
j9liu: Not the end of the world, but I guess just remove the duplicate one that appears later 🤷

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+  }
+
+  if (ctx.options.preloadAncestors && !queuedForLoad) {
+    addTileToLoadQueue(ctx, tile, TileLoadPriorityGroup::Preload, tilePriority);
+  }
+
+  return traversalDetails;
+}
+
+TraversalDetails visitTileIfNeeded(
```
j9liu, line 1027: This function had several comments / TODOs that got lost in the move. Could you restore them?

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+      getPreviousState(ctx.frameState.viewGroup, tile).getResult();
+
+  bool isRenderable = tile.isRenderable();
+  bool wasRenderedLastFrame =
+      lastFrameResult == TileSelectionState::Result::Rendered;
+
+  if (!wasRenderedLastFrame &&
+      lastFrameResult == TileSelectionState::Result::Refined) {
+    if (tile.getRefine() == TileRefine::Add) {
+      wasRenderedLastFrame = true;
```
j9liu, line 576: Worth keeping the comment around:
```suggestion
      // An additive-refined tile that was refined was also rendered.
      wasRenderedLastFrame = true;
```

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+    } else {
+      ctx.frameState.viewGroup.getTraversalState().forEachPreviousDescendant(
```
j9liu: Keep this comment around:

```suggestion
      // With replace-refinement, if any of this refined tile's children were
      // rendered last frame, but are no longer rendered because this tile is
      // loaded and has sufficient detail, we must treat this tile as rendered
      // last frame, too. This is necessary to prevent this tile from being
      // kicked just because _it_ wasn't rendered last frame (which could cause
      // a new hole to appear).
      ctx.frameState.viewGroup.getTraversalState().forEachPreviousDescendant(
```

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+}
+
+TraversalDetails visitVisibleChildrenNearToFar(
+    const TraversalContext& ctx,
+    uint32_t depth,
+    bool ancestorMeetsSse,
+    Tile& tile,
+    ViewUpdateResult& result) {
+  TraversalDetails traversalDetails;
+  for (Tile& child : tile.getChildren()) {
```
j9liu, line 787: This TODO got lost:
```suggestion
  // TODO: actually visit near-to-far, rather than in order of occurrence.
  for (Tile& child : tile.getChildren()) {
```

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+          tile,
+          TileLoadPriorityGroup::Normal,
+          tilePriority);
+    }
+    return true;
+  }
+  return false;
+}
+
+bool kickDescendantsAndRenderTile(
```
j9liu, line 678: This function has several internal comments that got lost in the move; could you restore them?

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+
+TileOcclusionState
+checkOcclusion(const TraversalContext& ctx, const Tile& tile) {
+  const std::shared_ptr<TileOcclusionRendererProxyPool>& pOcclusionPool =
+      ctx.externals.pTileOcclusionProxyPool;
+  if (!pOcclusionPool) {
+    return TileOcclusionState::NotOccluded;
+  }
+
+  const TileOcclusionRendererProxy* pOcclusion =
```
j9liu, line 490: The internal comments in this function got lost in the move; could you restore them?

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+  }
+
+  const auto& frustums = ctx.frameState.frustums;
+  const auto& fogDensities = ctx.frameState.fogDensities;
+  const auto& distances = ctx.distances;
+
+  bool isFogCulled = true;
+  for (size_t i = 0; i < frustums.size(); ++i) {
+    if (isVisibleInFog(distances[i], fogDensities[i])) {
+      isFogCulled = false;
```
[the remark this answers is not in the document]
j9liu: `CESIUM_ASSERT` isn't a bad idea here, since that internal assumption is no longer guaranteed outside of `Tileset.cpp`

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -0,0 +1,887 @@
...
+TraversalDetails visitVisibleChildrenNearToFar(
+    const TraversalContext& ctx,
+    uint32_t depth,
+    bool ancestorMeetsSse,
+    Tile& tile,
+    ViewUpdateResult& result);
+
+// Frustum and fog culling
+
+void frustumCull(
```
j9liu, line 374: This function has several internal comments that got lost in the move; could you restore them?

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetSelection.h
```diff
@@ -0,0 +1,49 @@
...
+ * @brief Runs the tile selection / LOD traversal algorithm.
+ *
+ * Populates `result` with the tiles to render this frame. The view group in
+ * `frameState` also receives load queue updates.
+ *
+ * @param ctx Configuration, external dependencies, and scratch buffers.
+ * @param frameState Per-frame view parameters and the view group to update.
+ * @param rootTile Root of the tile hierarchy to traverse.
+ * @param result Filled with the selected tiles and statistics for this frame.
+ */
```
j9liu, line 45: Maybe worth marking as `@private` API for now; not sure if access to this function might confuse users

```suggestion
 * @private
 */
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileUnloadQueue.h
```diff
@@ -0,0 +1,50 @@
...
+
+#include <Cesium3DTilesSelection/Tile.h>
+
+namespace Cesium3DTilesSelection {
+
+/**
+ * @brief LRU list of tiles eligible for content eviction.
+ *
+ * Head is least-recently-used, tail is most-recently-used.
+ */
```
j9liu, line 13: Maybe worth marking as `@private` API for now, might not have public API value for users.

```suggestion
 * Head is least-recently-used, tail is most-recently-used.
 * 
 * @private
 */
```

j9liu (changes requested):
Thank you for the work on this PR @calebbuffa! It might look like there's a lot of comments but I promise they're all minor.

The biggest note is that we lost a lot of helpful internal comments that we should preserve for future knowledge. Otherwise, it's just a matter of style / minor tweaks.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetSelection.h
```diff
@@ -0,0 +1,49 @@
...
+CESIUM3DTILESSELECTION_API void selectTiles(
+    const TileSelectionContext& ctx,
+    const TilesetFrameState& frameState,
+    Tile& rootTile,
+    ViewUpdateResult& result);
```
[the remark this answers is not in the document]
j9liu: Probably enough to leave an "OUTDATED" comment on the PR description, not the end of the world :shrug:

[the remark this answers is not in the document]
calebbuffa: done! updated the PR

calebbuffa:
@j9liu thank you for the feedback! ~~I believe I fixed all of your comments - but please let me know if I missed any!~~

I mangled one of the files in git diffs accidentally it looks like, I'm reverting the changes and will re-update.

javagl:
Stumbling over this, it looks like it addresses https://github.com/CesiumGS/cesium-native/issues/36 to some extent, alongside some of the suggestions from https://github.com/CesiumGS/cesium-native/issues/126 (even though that suggested (pseudocode) separation of `TraversealState` and `TraversalResult` seems to be reflected in the `TileSelectionContext` here)

j9liu (approved):
Thanks for all the changes @calebbuffa! Looks great, and I spot-checked this in Unreal to confirm things are working as expected. Merging now!

PR #1357, opened 2026-05-01
Move the source for the KTX-Software overlay port to the CesiumGS org.

### LINKED ISSUE #1303: KTX overlay port pulls from my personal fork of KTX-Software
See here:
https://github.com/CesiumGS/cesium-native/blob/v0.56.0/extern/vcpkg/ports/ktx/portfile.cmake#L3

This isn't catastrophic, but it's poor practice (on my part). I did it because I was actively pushing changes into the upstream project at the time, and so I expected the need for the fork to only be temporary. For various reasons, that has taken longer than expected, and may take awhile longer still. So we may want to move to a CesiumGS fork, at least.

For reference, here's the PR that I opened upstream:
https://github.com/KhronosGroup/KTX-Software/pull/1075

kring:
Fixes #1303


j9liu:
@timoore could you please review and merge at your earliest convenience? (we're starting to PR chain 😅)

timoore (approved):
Not much to review, as we've already been building with this version of KTX.

Issue #1208, opened 2025-06-30
Support for 3D Tiles Styling language


azrogers:
One issue we're running into with vector rasterization is that styling loaded GeoJSON documents is done by manipulating the style properties on the document itself. This works great from C++ and is just about as performant as can be, but really falters when it comes to Unreal - trying to run an Unreal Blueprint that iterates over thousands or tens of thousands or hundreds of thousands of GeoJSON objects results in a noticeable freeze when the document is loaded (on the order of several seconds for a modest dataset like the Philadelphia Streets dataset).

It would be nice if we could implement the [3D Tiles Styling language](https://github.com/CesiumGS/3d-tiles/tree/main/specification/Styling) as used in CesiumJS for this purpose. The upside here would be that this is already a tested and capable solution for the problem we're facing, and would give users pretty much all the tools they need to style a GeoJSON document.

The downside is that it requires a JavaScript interpreter.

Now, the 3D Tiles Styling language only uses a very constrained set of JavaScript features, so we don't need to bring in the entire kitchen sink to implement it. And that's very good news, as v8 would add something like 35MB to our binary size for a feature most users wouldn't need. Implementing our own interpreter for the language isn't beyond the pale of possibility, but it would be a significant amount of work.

Luckily, there are plenty of projects out there trying to run JavaScript in very limited environments - particularly in embedded environments, where even 1MB would be considered a gigantic binary.

Some interesting possibilities I've found:
- [duktape](https://duktape.org/)
- [JerryScript](https://jerryscript.net/)

Both of these libraries are meant to run with less than 200KB of flash memory, and both seem to support regex which is necessary for 3D Tiles Styling language support.

r-veenstra:
I would also request this for regular geometry too (do we want to open a seperate issue or just use this one)?

I'm increasingly working with metadata from sources such as iTwin's, there are plenty of use cases where we want to highlight or hide certain aspects of a tileset based on metadata.

To some extent this can be done via `CesiumFeaturesMetadata` component but it is kind of clunky to use in comparison to styling in JS.

It also has a substantial limitation of only working with shader compatible data like floats.  Metadata often features many strings or other data types, so it is not uncommon to setup styling with filters such as `${name} === 'Crown Entertainment Complex'` as shown [here](https://cesium.com/learn/cesiumjs-learn/cesiumjs-3d-tiles-styling/#apply-a-basic-style)

PR #1365, opened 2026-05-11
Add `VectorTilesRasterOverlay`

azrogers:
This PR adds support for rendering vector tilesets as a raster overlay. It uses a rasterization approach similar to the existing `GeoJsonDocumentRasterOverlay`, but instead of working from a single document, it works from a 3D Tiles tileset.

**Future work after this PR is merged**
- How do we handle styling here? A single default style for the entire tileset only gets us so far. In the `GeoJsonDocumentRasterOverlay` we got away with letting users customize the styles on elements directly in the document before the raster overlay was loaded, so users can accomplish any kind of metadata styling or other sort of styling they're looking for. But we'll need a different approach for a 3D Tiles tileset. Is it finally time to handle style expressions? (#1208)
- The `EXT_mesh_polygon` extension (see KhronosGroup/glTF#2570) will continue to evolve, possibly in ways that break the current implementation of this extension. We will need to keep up-to-date with the changes and implement them as necessary.
- This implementation is just blindly grabbing all the geometry it can find in the tileset and trying to interpret it as vector data. We should implement `3DTILES_content_gltf_vector` (CesiumGS/3d-tiles#838) as it matures.
- We are currently doing a relatively inefficient full selection of the vector tiles tileset. This has some drawbacks, [listed in Kevin's comment below](https://github.com/CesiumGS/cesium-native/pull/1365#issuecomment-4474418112). A more intelligent selection algorithm that takes into account the previous selections we've done would be ideal.
- At the moment, the only tilesets we are seeing are tilesets that are exclusively vector data. However, it's possible that we will eventually see tilesets that combine vector data with regular geometry. In these cases where a vector tiles overlay is draped over the same tileset, we should use a reference to that tileset instead of creating a separate instance.


j9liu:
Hey @kring! Are you able to take a glance from an architectural level for this PR?

Cesium3DTilesSelection/include/Cesium3DTilesSelection/RasterOverlayCollection.h
```diff
@@ -229,6 +229,8 @@ class CESIUM3DTILESSELECTION_API RasterOverlayCollection final {
    */
   size_t size() const noexcept;
 
+  void tick() noexcept;
```
kring: This doesn't appear to be used. The `Tileset` just calls `ActivateRasterOverlay::tick` directly.

CesiumRasterOverlays/include/CesiumRasterOverlays/ActivatedRasterOverlay.h
```diff
@@ -259,6 +259,10 @@ class CESIUMRASTEROVERLAYS_API ActivatedRasterOverlay
    */
   bool loadTileThrottled(RasterOverlayTile& tile);
 
+  void tick();
+
+  bool isTickable() const noexcept;
```
kring: Not used, other than by the ticking system in `RasterOverlayCollection`, which is itself unused.

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
```diff
@@ -0,0 +1,840 @@
...
+
+    const CesiumGeospatial::GlobeRectangle tileRectangle =
+        CesiumGeospatial::unprojectRectangleSimple(
+            this->getProjection(),
+            overlayTile.getRectangle());
+
+    // part 1 - selecting from scratch
+    if (this->_pTilesetContentManager->getRootTile() == nullptr) {
+      return this->_pTilesetContentManager->getRootTileAvailableEvent()
+          .thenInWorkerThread([this,
```
kring: Two problems here:
1. I think the root tile available event can resolve while the root tile is still nullptr, e.g., if there's an error. If it does, this will be an endless promise cycle.
2. It's not valid to call `loadTileImageFromTileset` in a worker thread, because it accesses the `TilesetContentManager`.

CesiumVectorOverlays/include/CesiumVectorOverlays/VectorTilesRasterOverlay.h
```diff
@@ -0,0 +1,40 @@
...
+   * @param asyncSystem The async system to use.
+   * @param name The user-given name of this polygon layer.
+   * @param document The GeoJSON document to use for the overlay.
+   * @param vectorOverlayOptions Options to configure this
+   * GeoJsonDocumentRasterOverlay.
+   * @param overlayOptions Options to use for this RasterOverlay.
+   */
+  VectorTilesRasterOverlay(
+      const std::string& name,
+      const std::string& url, // todo: accept factory instead of url
```
kring: Probably should also have an asset ID / token overload.

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
```diff
@@ -0,0 +1,744 @@
...
+        this->_pPrepareRendererResources,
+        parameters.externals.asyncSystem,
+        parameters.externals.pCreditSystem,
+        parameters.externals.pLogger,
+        nullptr,
+        TilesetSharedAssetSystem::getDefault(),
+        nullptr};
+
+    this->_pTilesetContentManager =
+        TilesetContentManager::createFromUrl(externals, TilesetOptions{}, url);
```
kring: What if we had a single tileset that had both normal meshes and vector data? For example, what if Google Photorealistic 3D Tiles also contained vector roads or points of interest?
I think in that case it would be really important to have just a single `TilesetContentManager`. Otherwise, we'd end up loading every tile twice.

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
```diff
@@ -0,0 +1,744 @@
...
+
+    return loadTileImageFromTileset(
+        overlayTile.getRectangle(),
+        tileRectangle,
+        textureSize);
+  }
+
+  // TODO: how does this get called?
+  virtual void tick() override {
+    this->_pTilesetContentManager->processWorkerThreadLoadRequests(
```
kring: Is anything ever calling `unloadCachedBytes`? I can't see it, and without that the tileset's memory usage will grow without bound.

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
```diff
@@ -0,0 +1,744 @@
...
+                               tileRectangle,
+                               textureSize]() mutable {
+            return this->loadTileImageFromTileset(
+                rectangle,
+                tileRectangle,
+                textureSize);
+          });
+    }
+
+    return loadTileImageFromTileset(
```
kring:
```suggestion
    return this->loadTileImageFromTileset(
```

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
```diff
@@ -0,0 +1,744 @@
...
+
+    visitIfNeeded(*pRootTile, loadInfo);
+
+    if (!loadInfo.tileLoadTasks.empty()) {
+      // Add load request and re-run this function when all those tiles are
+      // loaded. We need to re-run since some of those tiles might have been
+      // external content and now we need to check their children.
+      return this
+          ->addLoadRequest(this->getAsyncSystem(), loadInfo.tileLoadTasks)
+          .thenImmediately([this, tileRectangle, textureSize]() mutable {
```
kring: You can't call findAndLoadTiles anywhere but the main thread, so this should be `thenInMainThread`.

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
```diff
@@ -0,0 +1,744 @@
...
+      requestedTiles.insert(&*task.pTile);
+    }
+
+    LoadRequest request{
+        asyncSystem.createPromise<void>(),
+        std::move(requestedTiles)};
+    CesiumAsync::Future<void> future = request.promise.getFuture();
+
+    std::scoped_lock<std::mutex> lock(
+        this->_pSharedTileSelectionState->loadRequestsMutex);
```
kring: I can't see any reason to use a mutex. All accesses of `loadRequests` are (and must be) on the main thread.

CesiumVectorOverlays/include/CesiumVectorOverlays/VectorTilesRasterOverlay.h
```diff
@@ -0,0 +1,81 @@
...
+   * @param asyncSystem The async system to use.
+   * @param name The user-given name of this layer.
+   * @param source The source of the tileset.
+   * @param vectorOptions Options to configure this
+   * VectorTilesRasterOverlay.
+   * @param overlayOptions Options to use for this RasterOverlay.
+   */
+  VectorTilesRasterOverlay(
+      const std::string& name,
+      const TilesetSource& source,
```
kring: I'd suggest an overload rather than a variant here.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetContentManager.h
```diff
@@ -1,9 +1,8 @@
 #pragma once
 
-#include "RasterOverlayUpsampler.h"
-
 #include <Cesium3DTilesSelection/CesiumIonTilesetContentLoaderFactory.h>
```
kring, line 5: I'd prefer not to make `TilesetContentManager` part of the public API. Almost everything you need it for can also be achieved through `Tileset`. Expand and refactor that class, if necessary.

I guess it's worth considering what is the actual separation of concerns here. Originally, there was no TilesetContentManager; everything was in Tileset. When Bao [introduced](https://github.com/CesiumGS/cesium-native/pull/508) the TilesetContentManager, he described it like this:

> Previously Tileset and Tile are used to set the state machine of Tile. Now only TilesetContentManager will take care of that. Its responsibilities consists of loading tile content (by using TilesetContentLoader), managing tile loading state machine, and unloading tile. No ones except the TilesetContentManager can set the state of the tile now

So the content manager handled tile content state transitions (loading/unloading, basically) and the Tileset handled the rest. Where "the rest" was mostly the selection algorithm and the public entry points to the content manager.

Now with https://github.com/CesiumGS/cesium-native/pull/1342, we're talking about moving the selection algorith out of Tileset, too.

What's left? Not much I guess, but I think it's still serving a useful purpose as the public API representation of a Tileset, even if almost everything it does is delegated. By exposing TilesetContentManager, we a) need to give a lot more thought to its public API, because very little has gone into it, and b) are committing to maintaining it (yes, our deprecation policy doesn't strictly require we avoid breakage, but we still should avoid it wherever possibe).

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
kring: There is an interesting possibility here... rather than strictly running the selection algorithm in the main thread, it is theoretically possible to run it in a worker thread instead. However, in that case, we need to be very careful that it's always called from the same worker thread. We could do that with a thread pool with one thread, as we do with the request cache. Also, this will break down if a VectorTilesRasterOverlay ever uses a Tileset that is also used for normal rendering.

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
```diff
@@ -0,0 +1,773 @@
...
+              primitive.mode));
+        }
+      });
+
+  return {content, errors};
+}
+
+struct LoadRequest {
+  CesiumAsync::Promise<void> promise;
+  std::set<Tile*> requestedTiles;
```
kring: I think it might be better to use a sorted `std::vector<Tile*>` here instead of a `std::set`. Similar time complexity, far fewer memory allocations and better cache locality.

kring:
If I can try to summarize how this works...

Every time we need a raster image for a terrain tile, we call `findAndLoadTiles`. It returns a future that eventually resolves to the vector tiles that need to be rendered, once all such tiles have been completely loaded.

`findAndLoadTiles` does a full traversal of the vector tile tree from the root. If any selected tiles still need to be loaded, then it calls itself recursively after the future returned by the first invocation resolves.

The futures returned by `findAndLoadTiles` are resolved by the `tick` method. This is necessary because we don't currently have a future that is resolved when a tile finishes loading. (In the normal SSE-based selection, we essentially poll for completion instead.)

The `tick` method is called every frame. It loops over all tiles that have a load in flight, and for each that has completed, it loops over all the "requests" (one per raster tile that is loading) to see if that tile uses that now-completed tile. Once a raster tile has no vector tile loads remaining, the request's promise is resolved. That will trigger another `findAndLoadTiles` (a full traversal) to make sure that there aren't now _new_ tiles that we need to wait for.

This is potentially a lot of traversals. For vector tilesets with a small number of tiles, it probably isn't too bad. But I think it'll get pretty expensive for bigger ones.

Another more subtle problem is that a single call to `loadTileImage` can, in a tileset that uses lots of external tilesets or implicit subtrees, trigger a significant amount of loading. This loading is required in order to render the tile, so that's fine. Except that this loading process, once set in motion, will continue to completion even if the viewer moved away and this raster tile isn't even needed anymore.

I don't have any easy solution to this. Probably the ideal approach is more polling-based and incremental. That is, perhaps `tick` could a) check which (relevant) raster tiles are still in flight and advanced them by loading additional vector tiles if necessary, and b) do incremental traversals starting from a known point instead of traversing the entire vector tile tree from the root each time. It's not simple, though. (A) works against our raster overlay system, unfortunately. There's no way to "complete" a raster image load other than by providing the final image or failing it forever.

So I'd say it's reasonable to move forward with the current approach, and just keep an eye on how much time we're spending in this process as users start to work with larger vector tilesets.

kring (commented):
Some smaller comments below.

CesiumVectorOverlays/include/CesiumVectorOverlays/VectorTilesRasterOverlay.h
kring: You added one after I wrote this but before I submitted the review. Resolving.

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
azrogers: Can you elaborate what you mean by this? Do you mean using the same `TilesetContentManager` as the tileset the overlay is attached to?

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
azrogers: If the root tile is still `nullptr` by the time it gets to `findAndLoadTiles`, it will return an empty set of tiles and so return an empty image. I don't think there's a way for it to get stuck in a loop.

CesiumVectorData/include/CesiumVectorData/VectorStyle.h
```diff
@@ -106,17 +122,29 @@ struct VectorStyle {
...
+  PointStyle point;
 
   /**
    * @brief Default constructor for VectorStyle.
    */
   VectorStyle() = default;
 
   /**
-   * @brief Initializes style information for all types.
+   * @brief Initializes style information for point and polygon types.
```
j9liu: typo?

```suggestion
   * @brief Initializes style information for line and polygon types.
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
```diff
@@ -205,6 +205,13 @@ class CESIUM3DTILESSELECTION_API Tileset final {
    */
   const Tile* getRootTile() const noexcept;
 
+  /**
+   * @brief Gets the root tile of this tileset.
+   *
+   * This may be `nullptr` if there is currently no root tile.
+   */
+  Tile* getRootTile() noexcept;
+
```
j9liu, line 214: Maybe a good idea to discourage consumers of cesium-native from using this... would `@private` work for our documentation?

j9liu:
@kring just optionally requesting a review if you want to take another look. I'm doing a pass myself anyways :)

CesiumRasterOverlays/include/CesiumRasterOverlays/ActivatedRasterOverlay.h
```diff
@@ -259,6 +259,14 @@ class CESIUMRASTEROVERLAYS_API ActivatedRasterOverlay
    */
   bool loadTileThrottled(RasterOverlayTile& tile);
 
+  /**
+   * @brief Ticks the underlying raster overlay. This should be called
+   * frequently, such as once per frame, so that raster overlays can perform
+   * necessary updates. `Tileset` will call this method on all attached raster
```
j9liu: Worth linking?

```suggestion
   * necessary updates. \ref Cesium3DTilesSelection::Tileset will call this method on all attached raster
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
```diff
@@ -438,6 +445,21 @@ class CESIUM3DTILESSELECTION_API Tileset final {
...
+   * creating tile children, handling tile unloading, and more.
+   *
+   * @note This is automatically called as part of the normal tile selection
+   * process, such as when using \ref updateViewGroup or \ref
+   * updateViewGroupOffline. This method should only be called manually when
+   * tile selection is being driven by code external to the `Tileset`.
+   *
+   * @param tile The tile to update.
+   */
+  void updateTileContent(Tile& tile);
```
j9liu, line 461: Worth marking @private?

CesiumVectorOverlays/include/CesiumVectorOverlays/VectorTilesRasterOverlay.h
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * @brief HTTP headers to attach to requests made for this tileset.
+   */
+  std::vector<CesiumAsync::IAssetAccessor::THeader> requestHeaders;
+};
+
+/**
+ * @brief A raster overlay made from rasterizing a \ref
+ * CesiumVectorData::GeoJsonDocument.
+ */
```
j9liu, line 26: Forgot to change the copy-pasted comment, I assume

CHANGES.md
```diff
@@ -2,10 +2,15 @@
...
 
+##### Breaking Changes :mega:
+
+- `CesiumRasterOverlays::GeoJsonDocumentRasterOverlay` has been moved to the `CesiumVectorOverlays` library and namespace.
+
 ##### Additions :tada:
 
 - Added support for [`EXT_mesh_primitive_edge_visibility`](https://github.com/KhronosGroup/glTF/pull/2479) in `CesiumGltf`, `CesiumGltfReader`, and `CesiumGltfWriter`.
 - Added support for [`3DTILES_content_conditional`](https://github.com/CesiumGS/3d-tiles/pull/834) in `Cesium3DTiles`, `Cesium3DTilesReader`, and `Cesium3DTilesWriter`.
+- Added `CesiumVectorOverlays::VectorTilesRasterOverlay` supporting vector data loaded from 3D Tiles tilesets.
```
j9liu, line 13: Add a note that points from vector data are now rendered / stylable too?

CesiumVectorOverlays/include/CesiumVectorOverlays/VectorTilesRasterOverlay.h
```diff
@@ -0,0 +1,81 @@
...
+      const VectorTilesRasterOverlayOptions& vectorOptions = {},
+      const CesiumRasterOverlays::RasterOverlayOptions& overlayOptions = {});
+
+  /**
+   * @brief Creates a new VectorTilesRasterOverlay from a Cesium ion asset.
+   *
+   * @param name The user-given name of this layer.
+   * @param ionAssetID The Ion asset ID of the tileset.
+   * @param ionAccessToken The Ion access token to use to access the tileset.
+   * @param ionAssetEndpointUrl The Ion asset endpoint URL to use to access the
```
j9liu: ion should be decapitalized (outside of code conventions like class or variable names)

```suggestion
   * @param ionAssetID The ion asset ID of the tileset.
   * @param ionAccessToken The ion access token to use to access the tileset.
   * @param ionAssetEndpointUrl The ion asset endpoint URL to use to access the
```

CesiumVectorData/src/VectorRasterizer.cpp
```diff
@@ -243,8 +418,12 @@ void VectorRasterizer::drawGeoJsonObject(
...
       }
     }
-    void operator()(const GeoJsonPoint& /*catchAll*/) {}
-    void operator()(const GeoJsonMultiPoint& /*catchAll*/) {}
+    void operator()(const GeoJsonPoint& point) {
+      rasterizer.drawPoints({point.coordinates}, style.point);
+    }
+    void operator()(const GeoJsonMultiPoint& points) {
+      rasterizer.drawPoints(points.coordinates, style.point);
+    }
```
j9liu, line 426: If we have some data on hand, it would be nice to add a unit test for this.

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
```diff
@@ -0,0 +1,817 @@
...
+
+  glm::dmat4x4 rootTransform =
+      CesiumGltfContent::GltfUtilities::applyRtcCenter(model, gltfTransform);
+  rootTransform = CesiumGltfContent::GltfUtilities::applyGltfUpAxisTransform(
+      model,
+      rootTransform);
+
+  CesiumUtility::ErrorList errors;
+
+  VectorRenderContent* content = new VectorRenderContent();
```
j9liu:
```suggestion
  VectorRenderContent* pContent = new VectorRenderContent();
```

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
```diff
@@ -0,0 +1,817 @@
...
+        std::move(requestedTiles)};
+    CesiumAsync::Future<void> future = request.promise.getFuture();
+    this->_pSharedTileSelectionState->loadRequests.emplace_back(
+        std::move(request));
+
+    return future;
+  }
+
+  bool meetsSse(const Tile& tile, LoadTileImageInformation& loadInfo) {
+    // TODO: is this SSE calculation actually correct
```
j9liu: Are we confident enough in removing this TODO?

j9liu (commented):
Thanks @azrogers for the PR! I have some comments but hopefully nothing too big to address.

I didn't get to pore through `VectorTilesRasterOverlay.cpp` as much as I wanted, but hopefully that's okay given @kring's previous look at the file?

Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
azrogers: I think it's perfectly reasonable that a cesium-native consumer might want to use loadTiles directly and would need a way to access and modify the tiles and to update the tile content. I think it's fine to leave public with the explanation that it does not normally need to be called.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tileset.h
azrogers: Same for `updateTileContent` - I think it's fine to leave it open.

j9liu (approved):
Thanks @azrogers! Looking good, just waiting on CI to finish before merge.

PR #1386, opened 2026-06-03
fix: preserve MXCSR around earcut triangulation to support UE5 builds…

Johnly1986:
… UE5 enables SSE DAZ (Denormals Are Zero) and FTZ (Flush To Zero) MXCSR flags on worker threads in packaged Development/Shipping builds. These flags cause mapbox::earcut's floating-point geometry predicates (signed area, point-in-triangle, etc.) to flush denormal intermediate values to zero, producing incorrect triangulation results for polygons with near-collinear vertices.

## Author checklist

## Reviewer checklist

- [x] Review and run all parts of the test plan on this branch and verify it matches expectations.
    - [x] If the issue is a bug please make sure you can reproduce the bug in the main branch and then checkout this branch to make sure it actually solved the issue.
- [x] Review the code and make sure you do not have any remaining questions or concerns. You should understand the code change and the chosen approach. If you are not confident or have doubts about the code, please do not hesitate to ask questions.
- [x] Review the unit tests and make sure there are no missing tests or edge cases.
- [x] Review documentation changes and updates to `CHANGES.md` to make sure they accurately cover the work in this PR.

### DIFF
--- a/CesiumGeospatial/src/CartographicPolygon.cpp
+++ b/CesiumGeospatial/src/CartographicPolygon.cpp
@@ -16,6 +16,13 @@
 #include <utility>
 #include <vector>
 
+
+#if defined(_MSC_VER)
+#include <intrin.h>
+#elif defined(__x86_64__) || defined(__i386__)
+#include <x86intrin.h>
+#endif
+
 using namespace CesiumGeometry;
 
 namespace CesiumGeospatial {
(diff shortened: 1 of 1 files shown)


azrogers:
@Johnly1986 Thanks for the PR! Can you run down the author checklist in the template there? Particularly we need you to fill out the Contributor License Agreement - otherwise we can't take a look at your code!

j9liu:
Hi @Johnly1986, can you make sure to fill out the Contributor License Agreement? We're not seeing your username in our database, which prevents us from giving this a review.

Johnly1986:
Hi @j9liu , I've just signed the CLA — could you give it a recheck when you get a moment? Thanks! 🙏

j9liu:
Confirmed that we received your CLA, thanks @Johnly1986! We'll take a look when we get the chance.

kring:
@Johnly1986 since this is an Unreal-specific problem (mostly), would it be reasonable to change the mode at the start of each task in cesium-unreal and restore it at the end? So here:
https://github.com/CesiumGS/cesium-unreal/blob/main/Source/CesiumRuntime/Private/UnrealTaskProcessor.cpp

My thinking is:
1. Calling `_mm_setcsr` could have a bit of overhead, so doing it less often is good. Probably minor because `triangulatePolygon` is a pretty chunky call.
2. Perhaps other cesium-native computations are affected by this as well, we just haven't noticed yet.
3. It's better not to put Unreal-specific concerns in cesium-native if we have a good alternative.

j9liu:
Hey @Johnly1986, did you see Kevin's comment above? It seems like it will affect the approach taken in this PR.

Issue #1355, opened 2026-04-29
GeoJSON tilesets: encode the metadata schema


timoore:
The metadata schema in the MAXAR_content_geojson extension should be translated into an EXT_structural_metadata schema.

j9liu:
Closed by #1396

PR #1396, opened 2026-06-16
Convert MAXAR_content_geojson schema to glTF schema

timoore:
## Description

This is not very useful on its own, but is a stepping stone on the way to reading and transcoding the metadata included with GeoJson features in these tilesets. Also, I think the new  `getJson` function, which fetches and parses a JSON file, could be generally useful; however, it introduces an inner call to `thenInWorkerThread` and I wonder if that is generally a good approach.

## Issue number or link

#1355 

## Author checklist

- [X] I have submitted a [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) (only needed once).
- [X] I have done a full self-review of my code.

## Reviewer checklist

- [x] Review the code and make sure you do not have any remaining questions or concerns. You should understand the code change and the chosen approach. If you are not confident or have doubts about the code, please do not hesitate to ask questions.
- [x] Review the unit tests and make sure there are no missing tests or edge cases.


Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -1032,14 +1081,34 @@ TilesetJsonLoader::createLoader(
   TileExternalContent* pExternal =
       result.pRootTile->getContent().getExternalContent();
   CESIUM_ASSERT(pExternal);
+  std::string foreignSchemaUrl;
```
j9liu: **NIT**: not the end of the world, but this would be consistent with the name of the field the variable's read into

```suggestion
  std::string externalSchemaUrl;
```

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
```diff
@@ -1032,14 +1081,34 @@ TilesetJsonLoader::createLoader(
...
+          pExternal->metadata.metadata->getGenericExtension(
+              ExtensionMaxarContentGeoJson::ExtensionName);
+      if (extensionJson) {
+        if (const JsonValue* pSchemaUriValue =
+                extensionJson->getValuePtrForKey("propertiesSchemaUri")) {
+          if (pSchemaUriValue->isString()) {
+            foreignSchemaUrl = CesiumUtility::Uri::resolve(
+                tilesetJsonUrl,
+                pSchemaUriValue->getString());
+          }
```
j9liu: We should autogenerate this from `metadataEntity.MAXAR_content_geojson.schema.json`:

https://github.com/CesiumGS/3d-tiles/commit/3c0447dc20a92343132d18d00555374b2dc48e3a#diff-8da5b6ddf47331e5e0c3332e30c49ea23e3067816f401d996cfa579b7492c437

Though, since there are two places where the extension occurs, we'll need to deconflict them similar to `EXT_structural_metadata` in glTF, where we have `ExtensionModelExtStructuralMetadata` and `ExtensionMeshPrimitiveExtStructuralMetadata`. The names should be something like `ExtensionTilesetMaxarContentGeojson` and `ExtensionMetadataEntityMaxarContentGeojson`.

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -507,4 +511,61 @@ ConverterResult GltfConverter::convert(
...
+        } else if (propType == "Integer") {
+          metaClass.componentType = ClassProperty::ComponentType::INT32;
+        }
+        geoJsonClass.properties[*propName] = std::move(metaClass);
+      }
+    }
+  }
+  schema.classes[classKey] = std::move(geoJsonClass);
+  return schema;
+}
```
j9liu: Let's autogenerate a struct/reader for `MaxarGeoJsonProperties` from the `geojsonproperties.schema.json`: 
https://github.com/CesiumGS/3d-tiles/commit/3c0447dc20a92343132d18d00555374b2dc48e3a#diff-46dbe1c71e56099271a285827c6db18ce4918dcab4dd23f87517f09ee0d2a187

Even though the types will still have to be converted to `CesiumGltf::Schema`'s, it's more robust and scalable to auto-generate a reader class, than to manually check and compare the structure of properties.

j9liu (changes requested):
@timoore I took an early look; the main feedback I have is to take advantage of our auto-generation tools to create classes/readers from the schemas, which will make the code more robust and scalable. Otherwise, the overall approach seems fine!

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
timoore: Great suggestion, as this code is pretty awkward.

CesiumVectorData/src/GltfConverter.cpp
timoore: Another great suggestion; it didn't occur to me that we could use the autogeneration tools for reading "3rd party" Json content.

CesiumVectorData/include/CesiumVectorData/GltfConverter.h
```diff
@@ -34,6 +39,8 @@ class CESIUMVECTORDATA_API GltfConverter {
       const CesiumGeospatial::Ellipsoid& ellipsoid);
   /**
    * @brief Convert MAXAR_content_geojson schema into a glTF schema.
+   * @param schemaJson The schema supplied with a GeoJson tileset
```
j9liu: Since this is doc (not code) JSON should be capitalized.

```
   * @param schemaJson The schema supplied with a GeoJSON tileset
```

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
j9liu: @timoore this is marked resolved even though it wasn't applied / you hadn't responded, was this intentional?

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -511,58 +515,41 @@ ConverterResult GltfConverter::convert(
...
+  if (!schemaReadResult.value || !schemaReadResult.errors.empty()) {
+    ErrorList errorList = {schemaReadResult.errors, schemaReadResult.warnings};
+    return {errorList};
   }
-  std::string classKey =
-      JsonHelpers::getStringOrDefault(schemaJson, "semantic", "geoJsonClass");
-  Class geoJsonClass;
-  geoJsonClass.name = *className;
-  auto propertiesIt = schemaJson.FindMember("properties");
-  if (propertiesIt == schemaJson.MemberEnd()) {
```
j9liu: I don't think we should error out here; `name` is not a required property of the schema: 

https://github.com/CesiumGS/3d-tiles/blob/cesium-native/extensions/MAXAR_content_geojson/schema/geojsonproperties.schema.json

CesiumVectorData/include/CesiumVectorData/GltfConverter.h
```diff
@@ -34,6 +39,8 @@ class CESIUMVECTORDATA_API GltfConverter {
...
+   * @returns A result object that includes the converted schema and any errors.
    */
   static ConvertSchemaResult
   convertSchema(const rapidjson::Document& schemaJson);
```
j9liu: Now that we have the generated classes, this should be changed to `Cesium3DTiles::ExtensionSchemaMaxarContentGeojson` rather than an arbitrary `rapidjson::Document`. The reading step before it is passed in to `convertSchema`.

j9liu (commented):
This is looking a lot better @timoore ! The generated classes are a huge improvement.

I have a few small comments left, but shouldn't be anything major.

Cesium3DTilesSelection/src/TilesetJsonLoader.cpp
j9liu: Never mind, I saw that this got addressed in #1402 .

CesiumVectorData/src/GltfConverter.cpp
j9liu: For some reason GitHub is showing this on the old version of this file, but it corresponds to lines 528-530 in the new version.

```
  const auto& maxarSchema = *schemaReadResult.value;
  if (!maxarSchema.name) {
    return {ErrorList::error("No schema class name")};
  }
```

CesiumVectorData/src/GltfConverter.cpp
j9liu, line 1: Also would be great to get a unit test for this!

timoore:
@j9liu , ready for another look.

j9liu (approved):
Looks good, thanks Tim!

PR #1400, opened 2026-06-22
Rasterize Point and MultiPoint in GeoJsonDocumentRasterOverlay

### LINKED ISSUE #1403: GeoJsonDocumentRasterOverlay does not rasterize Point / MultiPoint geometry
### Description

`CesiumVectorOverlays::GeoJsonDocumentRasterOverlay` does not actually rasterize
`Point` and `MultiPoint` geometry, even though the changelog for v0.61.0 states
that the document overlay can display points. Points and multi-points are
silently dropped and never appear in the rasterized output.

### Steps to reproduce

1. Build a `GeoJsonDocument` containing `Point` and/or `MultiPoint` geometry.
2. Create a `GeoJsonDocumentRasterOverlay` from that document with a
   `VectorStyle` whose `point` style is set (fill and/or outline).
3. Rasterize the overlay.

### Expected behavior

Points and multi-points are rendered as filled / outlined circles using
`VectorStyle::point`, consistent with the v0.61.0 changelog claim that the
document overlay can display points.

### Actual behavior

No point geometry is drawn. The points are dropped before reaching the
rasterizer, so the output contains nothing for them.

### Root cause

The rasterizer itself already knows how to draw points:
`VectorRasterizer::drawGeoJsonObject` routes `GeoJsonPoint` / `GeoJsonMultiPoint`
to `drawPoints(..., style.point)`. The geometry just never reaches it.

When `GeoJsonDocumentRasterOverlay` builds its quadtree, `addPrimitivesToData`
runs `RectangleAndLineWidthFromObjectVisitor` to compute a bounding rectangle for
each object and only stores objects that produce one (`if (rect)`). The visitor's
`operator()` for `GeoJsonPoint` and `GeoJsonMultiPoint` are empty, so points
yield no rectangle and are dropped before ever reaching the rasterizer.

Additionally, `QuadtreeGeometryData::calculateBoundingRectangleForTileSize` has
no `Point` / `MultiPoint` branch, so even if points were stored, their per-tile
bounds would fall into the line branch and be sized by `style.line` instead of
`style.point`.

### Affected version

v0.61.0 (the v0.61.0 changelog already claims the document overlay can display
points, but it cannot).

### Notes

A fix is proposed in PR #1400, which implements the `Point` / `MultiPoint`
visitor operators, adds a `Point` / `MultiPoint` branch to
`calculateBoundingRectangleForTileSize`, and adds render tests for `Point`,
outlined `Point`, and `MultiPoint`.

baruchInsert-tech:
## Summary

`CesiumVectorOverlays::GeoJsonDocumentRasterOverlay` did not actually rasterize `Point` and `MultiPoint` geometry, even though the changelog for v0.61.0 states that it can display points. This PR makes the document overlay genuinely render points, fixing that latent gap.

## The problem

The rasterizer already knows how to draw points: `VectorRasterizer::drawGeoJsonObject` routes `GeoJsonPoint`/`GeoJsonMultiPoint` to `drawPoints(..., style.point)`. The geometry just never got there.

When `GeoJsonDocumentRasterOverlay` builds its quadtree, `addPrimitivesToData` runs `RectangleAndLineWidthFromObjectVisitor` to compute a bounding rectangle for each object and only stores objects that produce one (`if (rect)`). The visitor's `operator()` for `GeoJsonPoint` and `GeoJsonMultiPoint` were **empty**, so points yielded no rectangle and were dropped before ever reaching the rasterizer.

## The fix

- Implement the `GeoJsonPoint` / `GeoJsonMultiPoint` visitor operators so they build a (degenerate) bounding rectangle at each coordinate and bump `maxLineWidthPixels` by the point's pixel footprint (`2 * radius`, plus the outline width when outlined), mirroring the existing line-width padding.
- Add a `Point`/`MultiPoint` branch to `QuadtreeGeometryData::calculateBoundingRectangleForTileSize` so per-tile bounds are padded by that same pixel footprint (otherwise points fell into the line branch and were sized by `style.line`).
- Update the now-stale comment in `addPrimitivesToData` that claimed points are intentionally not rasterized.

Points then flow through the existing machinery and are rasterized as filled/outlined circles using `VectorStyle::point`. No new drawing code was needed.

## Tests

Added three render tests to `TestGeoJsonDocumentRasterOverlay.cpp`, each rasterizing an in-code document and asserting the expected colors appear:

- `Point` rendered as a filled circle (fill color present).
- `Point` with an outline (both fill and outline colors present).
- `MultiPoint` (fill color present).

All three pass locally (RelWithDebInfo), and `CesiumVectorOverlays` + `cesium-native-tests` build clean.

## Changelog

Added a *Fixes* entry under the unreleased section. Note the v0.61.0 entry already (incorrectly) claimed the document overlay could display points; this PR makes that true.

issue #1403

@azrogers 
@j9liu


j9liu:
Hi @baruchInsert-tech, sorry for the delay in getting around to this!

Can you merge in `main` here, and also fix the location of the changelog entry? We'll give this a review once that's done.

baruchInsert-tech:
> Hi @baruchInsert-tech, sorry for the delay in getting around to this!
> 
> Can you merge in `main` here, and also fix the location of the changelog entry? We'll give this a review once that's done.

Done! Thank U!

CesiumVectorOverlays/test/TestGeoJsonDocumentRasterOverlay.cpp
```diff
@@ -552,3 +565,84 @@ TEST_CASE("GeoJsonDocumentRasterOverlay FeatureCollection falls back to "
...
+              LineWidthMode::Pixels},
+          PolygonStyle{std::nullopt, std::nullopt},
+          PointStyle{
+              5.0,
+              ColorStyle{Color{0, 255, 0, 255}, ColorMode::Normal},
+              std::nullopt}},
+      Ellipsoid::WGS84,
+      0};
+
+  CesiumGltf::ImageAsset image = rasterizeOverlayTileFromDocument(
```
j9liu: There was a recent breaking change with `ImageAsset`; it is now under `CesiumImage`.

CesiumVectorOverlays/src/GeoJsonDocumentRasterOverlay.cpp
```diff
@@ -418,11 +454,12 @@ void addPrimitivesToData(
...
-  // GeometryCollections have no bounding box. For the first two, this is
-  // because they are not rasterized by this overlay. For the rest, though they
-  // may contain geometry, they do not themselves have anything to render. We
-  // can save some effort by ignoring them all now.
+  // Features, FeatureCollections, and GeometryCollections produce no bounding
+  // box of their own: though they may contain geometry, they have nothing to
+  // render directly, so the visitor leaves `rect` empty and we skip them here
+  // (their children are handled by GeoJsonChildVisitor below). Points and
+  // MultiPoints now DO produce a bounding box and are rasterized as
+  // filled/outlined circles.
```
j9liu: **NIT** for minor wording tweaks:

```suggestion
  // Features, FeatureCollections, and GeometryCollections produce no bounding
  // box of their own: though they may contain geometry, they themselves have
  // nothing to to render. Their children are handled by GeoJsonChildVisitor below.
```

CesiumVectorOverlays/src/GeoJsonDocumentRasterOverlay.cpp
```diff
@@ -102,6 +102,17 @@ struct QuadtreeGeometryData {
...
+        this->pObject->isType<GeoJsonPoint>() ||
+        this->pObject->isType<GeoJsonMultiPoint>()) {
+      // A point's pixel footprint is its radius (plus its outline width) in
+      // every direction. Encode that as an equivalent line width (the full
+      // diameter) so the shared pixel padding math below applies to it too.
+      activeLineStyle.width = 2.0 * this->pStyle->point.radius;
+      if (this->pStyle->point.outline) {
+        activeLineStyle.width += this->pStyle->point.outline->width;
+      }
+      activeLineStyle.widthMode = LineWidthMode::Pixels;
```
j9liu, line 115: Is it correct to hardcode this as `Pixels`?

I noticed that if you set `Width Mode` to meters, and the meters is greater than the point radius, then the points appear like this:

Image

This doesn't seem correct, though I can't tell from a glance which line of code is the culprit.

j9liu (changes requested):
Thanks for the PR @baruchInsert-tech! I can confirm that points now render as a result of this PR, though there is a slight issue with outline rendering (see the comments).

Let me know when these comments are addressed -- and please push the commits regularly, _without_ `--force`, so that the diff is more obvious upon review. Thank you!

CesiumVectorOverlays/test/TestGeoJsonDocumentRasterOverlay.cpp
baruchInsert-tech: Done

CesiumVectorOverlays/src/GeoJsonDocumentRasterOverlay.cpp
baruchInsert-tech: Done

CesiumVectorOverlays/src/GeoJsonDocumentRasterOverlay.cpp
baruchInsert-tech: Fixed

j9liu: I appreciate the new unit test / changes for this, but this sadly wasn't fixed...

However, I did test whether point outlines work for `VectorTilesRasterOverlay`, and it appears to be a problem there too. This does point to a more underlying mistake that can be out of scope for this PR. At the same time, we'd really like for a human to verify the results of any PR changes before we get to it, to avoid this back-and-forth!

baruchInsert-tech: Thanks!
Given you flagged it as reasonably out of scope, I've narrowed this PR back to the verified change , Point/MultiPoint now rasterize and reverted the bounds-padding attempt, which only affected tile coverage and never touched the actual rendering. I opened a separate issue  for the shared meters-mode point-outline bug in VectorRasterizer.
https://github.com/CesiumGS/cesium-native/issues/1413

And point taken on verifying renders before pushing, that's on me; I'll eyeball the output next time rather than round-tripping through review.

j9liu (approved):
Looks good, thanks @baruchInsert-tech !

PR #1402, opened 2026-06-24
load Maxar feature properties as structural metadata

### LINKED ISSUE #1356: GeoJSON tilesets: encode metadata in GeoJSON features
Using the tileset's metadata schema, the metadata in each GeoJSON feature should be encoded in the glTF.

timoore:
Load the feature data in MAXAR_content_geojson tiles as EXT_structural_metadata, along with the accompanying schema.

This depends on #1396, so either merge it first or withdraw it.

This is marked as draft awaiting some tests and verification that the schema is complete.


CesiumVectorData/include/CesiumVectorData/GltfConverter.h
```diff
@@ -32,11 +34,15 @@ class CESIUMVECTORDATA_API GltfConverter {
    *
    * @param geoJson The GeoJSON document.
    * @param ellipsoid The ellipsoid for GeoJSON coordinates.
+   * @param pSchema the converted MAXAR_content_geojson schema for the GeoJSON
+   * feature properties.
```
j9liu: We could word this more generically:
```suggestion
   * @param pSchema The @ref CesiumGltf::Schema to use for converting feature properties in the GeoJSON.
```

CesiumVectorData/include/CesiumVectorData/GltfConverter.h
```diff
@@ -32,11 +34,15 @@ class CESIUMVECTORDATA_API GltfConverter {
...
    * @returns A result object that includes the glTF Model and any errors.
    */
   static ConverterResult convert(
       const GeoJsonDocument& geoJson,
-      const CesiumGeospatial::Ellipsoid& ellipsoid);
+      const CesiumGeospatial::Ellipsoid& ellipsoid,
+      const CesiumUtility::IntrusivePointer<CesiumGltf::Schema>& pSchema);
```
j9liu: Perhaps `pSchema` should be optional (default `nullptr`) in case no schema is specified.

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -255,6 +260,191 @@ struct GltfConverterImpl {
...
+          }
+        }
+        metadataSize += pPropRep->buffer.size();
+        offsetsSize += pPropRep->offsets.size();
+      }
+    }
+    metadataBuffer.cesium.data.resize(metadataSize);
+    offsetsBuffer.cesium.data.resize(offsetsSize * sizeof(size_t));
+    size_t buffOffset = 0;
+    size_t offsetOffset = 0;
```
j9liu: An extra word might make this more understandable from a glance 😅 
```
    size_t offsetDataOffset = 0;
```

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -255,6 +260,191 @@ struct GltfConverterImpl {
...
+          if (offset == std::numeric_limits<size_t>::max()) {
+            offset = lastValidOffset;
+          } else {
+            lastValidOffset = offset;
+          }
+        }
+        metadataSize += pPropRep->buffer.size();
+        offsetsSize += pPropRep->offsets.size();
+      }
+    }
```
j9liu, line 322: Since we're going through and compacting the buffer data anyway to get its true size, it would be great if the offset type was downcasted from `size_t` to a smaller unsigned int when possible.

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -255,6 +260,191 @@ struct GltfConverterImpl {
...
+      if (!pGeoJsonObject) {
+        continue;
+      }
+      const GeoJsonFeature& feature = pGeoJsonObject->get<GeoJsonFeature>();
+      if (feature.properties) {
+        for (auto propIt = feature.properties->begin();
+             propIt != feature.properties->end();
+             ++propIt) {
+          const auto& [name, value] = *propIt;
+          if (value.isString()) {
```
j9liu: This is executing based on the type of `JsonValue`, rather than comparing against what's defined in `pSchema`. We should use the schema to decide how we handle these `JsonValue`s, and do our best to convert if a `JsonValue` is somehow the wrong type

CesiumVectorData/src/GltfConverter.cpp
```diff
@@ -255,6 +260,191 @@ struct GltfConverterImpl {
...
+            BufferView::Target::ARRAY_BUFFER,
+            int64_t(buffOffset),
+            int64_t(pPropRep->buffer.size()));
+        std::memcpy(
+            metadataBuffer.cesium.data.data() + buffOffset,
+            pPropRep->buffer.data(),
+            pPropRep->buffer.size());
+        for (size_t& offset : pPropRep->offsets) {
+          offset += buffOffset;
+        }
```
j9liu: I don't know if this is correct.[ Per the spec](https://github.com/CesiumGS/glTF/blob/320d97585c4ae7915c5fce65f34fb822203608ef/extensions/2.0/Vendor/EXT_structural_metadata/schema/propertyTable.property.schema.json#L29-L36), the `stringOffsets` are relative to the `offsets` buffer view it uses, not the underlying buffer data.

j9liu (commented):
Thanks for opening this PR @timoore. Ultimately my concern is the correctness of the conversion, which is difficult to verify from reading alone. Unit tests would greatly help with that.

We should test some data in Unreal to ensure they work with metadata picking/styling.

CesiumVectorData/include/CesiumVectorData/GltfConverter.h
timoore: I agree, but I'm not sure what the optional behavior should be. If there's not schema, should we create one based on the features that are present and still give access to the metadata?

CesiumVectorData/src/GltfConverter.cpp
timoore: I agree, but I think it's barely in scope for the work I'm doing. I will look at doing that once the rest seems to working well.

CesiumVectorData/src/GltfConverter.cpp
timoore: You're right!

CesiumVectorData/src/GltfConverter.cpp
timoore: I've reworked this to use the schema to drive the parsing and do some conversion too when it make sense.

CesiumVectorData/test/TestGltfConverter.cpp
```diff
@@ -70,4 +86,101 @@ TEST_CASE("Conversion from geoJSON to glTF") {
               featureIdView);
         });
   }
+  SUBCASE("GeoJSON metadata") {
+    const auto* metadata =
```
j9liu:
```
    const auto* pMetadata =
```

CesiumVectorData/include/CesiumVectorData/GltfConverter.h
j9liu, line 1: Please add a changelog entry for this!

j9liu (commented):
Thanks @timoore ! Just a few small comments:

j9liu (approved):
Thank you @timoore. Per our offline discussion we'll use the `TilesetContentLoader` workaround for now; I opened #1406 as an item to address later.

Just waiting on CI to complete before merge (will rerun the Windows 2022 fluke).

PR #1405, opened 2026-06-24
Fix WebP library dependencies for CesiumImage and CesiumGltfReader

timoore:
CesiumImage should depend directly on WebP, while CesiumGltfReader does not anymore. If not, programs that use CesiumImage directly, without CesiumGltfReader, won't link.


j9liu (approved):
Thanks @timoore for the fix!

PR #1430, opened 2026-07-30
Allow custom styling for VectorTilesRasterOverlay.

azrogers:
This PR allows the user to specify a `VectorStylingProvider` that will be called to obtian styling information for the primitives loaded by the `VectorTilesRasterOverlay`. This allows per-element styling with a great degree of flexibility.


CesiumVectorData/include/CesiumVectorData/VectorStyle.h
```diff
@@ -149,5 +162,48 @@ struct VectorStyle {
    * @brief Initializes all styles to the given color.
    */
   VectorStyle(const CesiumUtility::Color& color);
+
+  /** @brief Checks if two VectorStyle instances are equal. */
+  bool operator==(const VectorStyle& rhs) const;
+};
+} // namespace CesiumVectorData
+
+/** @brief Hash implementation for \ref CesiumVectorData::ColorStyle. */
```
j9liu: Prefer `@ref` according to the [style guide](https://cesium.com/learn/cesium-native/ref-doc/style-guide.html#documentation), and apply this to the other instances of `\ref` in this doc too.

```suggestion
/** @brief Hash implementation for @ref CesiumVectorData::ColorStyle. */
```

CesiumVectorData/include/CesiumVectorData/VectorRasterizer.h
```diff
@@ -125,6 +125,16 @@ class VectorRasterizer {
       const std::vector<CesiumGeospatial::Cartographic>& points,
       const PointStyle& style);
 
+  /**
+   * @brief Draws a set of points to the canvas.
+   *
+   * @param points The set of points to draw.
+   * @param style The @ref PointStyle to use when drawing the points.
```
j9liu:
```suggestion
   * @param styles The @ref PointStyle instances to use when drawing the points.
```

CesiumVectorData/src/VectorRasterizer.cpp
```diff
@@ -395,6 +395,43 @@ void VectorRasterizer::drawPoints(
...
+
+  for (size_t i = 0; i < points.size(); i++) {
+    BLPoint point = radiansToPoint(
+        points[i].longitude,
+        points[i].latitude,
+        this->_bounds,
+        this->_context);
+    // clang-tidy does not understand that we *are* in fact checking these
+    // optionals
+    if (styles[i]->point.fill) {
```
j9liu: It seems like there's some assumptions here:

- `styles.size() == points.size()`
- `styles[i] != nullptr` for all `i`
- `styles[i].point.hasValue()`

We should add `if` statements to guard for these, or at least add asserts.

CesiumVectorData/include/CesiumVectorData/VectorRasterizer.h
j9liu: Also, it seems like `styles.size()` must equal `points.size()`, where `style[i]` is applied to `points[i]`. Perhaps we should call it out here.

CesiumVectorOverlays/include/CesiumVectorOverlays/VectorStylingProvider.h
```diff
@@ -0,0 +1,96 @@
...
+#include <CesiumGeospatial/Cartographic.h>
+#include <CesiumGltf/Model.h>
+#include <CesiumVectorData/VectorStyle.h>
+
+namespace CesiumVectorOverlays {
+
+/**
+ * @brief An interface for providing styling information for vector features.
+ */
+class CESIUMVECTOROVERLAYS_API VectorStylingProvider {
```
j9liu, line 16: Add a unit test for this, to confirm that a simple style can be successfully applied?

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
```diff
@@ -112,15 +160,28 @@ CesiumUtility::Result<VectorRenderContent*> vectorizeModel(
           return;
         }
 
+        const CesiumGltf::ExtensionExtMeshFeatures* pMeshFeatures =
+            primitive.getExtension<CesiumGltf::ExtensionExtMeshFeatures>();
+        const CesiumGltf::FeatureId* pFeatureIdSet =
+            pMeshFeatures != nullptr && !pMeshFeatures->featureIds.empty()
+                ? &pMeshFeatures->featureIds[0]
+                : nullptr;
```
j9liu, line 171: This could be simplified using `model.getSafe`.

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
```diff
@@ -64,18 +69,51 @@ using namespace CesiumUtility;
...
+        CesiumGltf::getFeatureIdAccessorView(
+            model,
+            primitive,
+            static_cast<int32_t>(*pFeatureIdSet->attribute));
+    return {std::visit(
+        CesiumGltf::FeatureIdFromAccessor{index},
+        featureIdAccessor)};
+  } else if (pFeatureIdSet->texture) {
+    return CesiumUtility::Result<int64_t>(ErrorList::warning(
+        "Feature ID textures for vector primitives is not yet supported."));
```
j9liu:
```suggestion
    return CesiumUtility::Result<int64_t>(ErrorList::warning(
        "Feature ID textures for vector primitives are not yet supported."));
```

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
```diff
@@ -242,7 +325,115 @@ CesiumUtility::Result<VectorRenderContent*> vectorizeModel(
         }
       });
 
-  return {pContent, errors};
+  // Apply per-element styling if the user specified a styling provider.
```
j9liu: Move this comment down after the `if  (!pStylingProvider)` ?

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
```diff
@@ -132,14 +193,21 @@ CesiumUtility::Result<VectorRenderContent*> vectorizeModel(
...
               // Primitive restart.
               if (polyline.size() >= 2) {
+                const CesiumUtility::Result<int64_t> featureIdResult =
+                    getFeatureId(pFeatureIdSet, model, primitive, startIndex);
+                errors.merge(featureIdResult.errors);
+                polylineFeatureIds.emplace_back(
+                    featureIdResult.value.value_or(-1));
                 pContent->polylines.emplace_back(std::move(polyline));
                 polyline.clear();
+                startIndex = -1;
```
j9liu, line 213: Wonder if there's a nice way to reorganize this so there aren't two `polyline.size() >=2` checks... but doesn't have to be in scope for this PR.

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
```diff
@@ -242,7 +325,115 @@ CesiumUtility::Result<VectorRenderContent*> vectorizeModel(
...
+                           pContent->pointStyles.resize(
+                               pContent->points.size(),
+                               &pContent->defaultStyle);
+                           return false;
+                         }
+
+                         for (const auto& style : result) {
+                           if (style.has_value()) {
+                             pContent->pointStyles.emplace_back(
+                                 &*pContent->uniqueStyles.insert(*style).first);
```
j9liu: **NIT**: For readability help? (Same with the others in the file)

```suggestion
                                 &(*pContent->uniqueStyles.insert(*style).first));
```

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
```diff
@@ -311,26 +502,39 @@ class VectorTilesPrepareRendererResources : public IPrepareRendererResources {
...
+               tileLoadResult.ellipsoid,
+               transform,
+               this->defaultStyle,
+               this->_pStylingProvider)
+        .thenInWorkerThread([tileLoadResult = std::move(tileLoadResult),
+                             this](CesiumUtility::Result<VectorRenderContent*>&&
+                                       vectorizationResult) mutable {
+          if (vectorizationResult.errors.hasErrors()) {
+            if (vectorizationResult.value) {
+              delete *vectorizationResult.value;
```
j9liu, line 525: I see this was already in the code, but for the future -- would be good to use smart pointers instead or otherwise document any specific reasons for why we're not using them.

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
```diff
@@ -596,19 +806,22 @@ class VectorTilesRasterOverlayTileProvider final
...
+                rasterizer.drawPolyline(
+                    pVectorContent->polylines[i],
+                    pVectorContent->polylineStyles[i]->line);
               }
 
               if (!pVectorContent->points.empty()) {
                 rasterizer.drawPoints(
                     pVectorContent->points,
-                    pVectorContent->style.point);
+                    pVectorContent->pointStyles);
```
j9liu, line 832: **NIT**: why do points have their own multi-style override when polylines/polygons use `for` loops?

j9liu (changes requested):
Thanks for the PR @azrogers! I still have to try this out in Unreal, but I wanted to give the Native part a pass to get things moving.

Also remember to update CHANGES.md!

CesiumVectorOverlays/include/CesiumVectorOverlays/VectorStylingProvider.h
```diff
@@ -0,0 +1,96 @@
...
+   * size as `featureIds`, and each point corresponds to the feature ID at the
+   * same index.
+   * @returns A future that resolves to a vector of optional VectorStyle
+   * objects, one for each point feature. The vector should be the same size as
+   * `featureIds and `points`. If `std::nullopt` is provided for a feature, the
+   * feature will use the default style. If the returned vector is empty or does
+   * not match the size of `featureIds` and `points`, the default style will be
+   * used for all features.
+   */
+  virtual CesiumAsync::Future<
```
j9liu, line 35: Are we using the full `VectorStyle` here for any particular reason? You could restrict this to `PointStyle` for points, `LineStyle` for lines, etc?

CesiumVectorData/src/VectorRasterizer.cpp
azrogers: styles[i]->point is not an optional so we don't need to check if it has the value. It contains optionals but we are already checking them. The rest I'll add checks for.

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
azrogers: I think model.getSafe would only simplify the check to whether pMeshFeatures->featureIds is empty. We still would need to check if it's nullptr and still need to handle the case where it's not present.

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
azrogers: `vectorizationResult.value` is the `pRenderResources` for this tile, and we don't/can't use smart pointers for those as the lifetime is handled by the `TilesetContentManager`.

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
azrogers: Well, polylines and polygons functions in the `VectorRasterizer` were implemented with one call per object, so the functions can already be called with a distinct style for each object. The points function was implemented with one call for all objects, so we need an overload that can specify a distinct style for each object. You might argue that this is an inconsistent API, and I think you would be correct in arguing that, but I didn't want to go back and redo those methods here.

CesiumVectorOverlays/include/CesiumVectorOverlays/VectorStylingProvider.h
azrogers: There *was* a reason, but as it turns out it was pretty easy to change things around so that reason no longer applies.

CesiumVectorOverlays/src/VectorTilesRasterOverlay.cpp
j9liu: Sorry, my intent was to say this could be slightly more streamlined as:

```suggestion
        const CesiumGltf::FeatureId* pFeatureIdSet =
            pMeshFeatures != nullptr ? model.getSafe(&pMeshFeatures->featureIds, 0) : nullptr;
```

Yes, the `nullptr` check is still necessary.

j9liu:
Sorry, one more thing -- I noticed that the Editor isn't exiting properly because `UCesiumRasterOverlay::IsReadyForFinishDestroy()` indefinitely returns false. I'm seeing that `this->_overlaysBeingDestroyed` never gets decremetned -- is there something new that prevents the vector raster overlay from destroying properly?

azrogers:
Added a test for styling that uses the PhillyStressTest tileset I created for the vector-tiles-benchmark branch, since it has all three types of primitives. Will investigate that editor exit hang bug.

j9liu:
> Sorry, one more thing -- I noticed that the Editor isn't exiting properly because UCesiumRasterOverlay::IsReadyForFinishDestroy() indefinitely returns false. I'm seeing that this->_overlaysBeingDestroyed never gets decremetned -- is there something new that prevents the vector raster overlay from destroying properly?

This strangely appears to go away after deleting the `extern/build` folder and rebuilding cesium-native. Observed by both @azrogers and me :shrug:

j9liu (approved):
Thanks @azrogers !

PR #1435, opened 2026-08-14
Fixed area paging

timoore:
## Author checklist

- [X ] I have submitted a [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) (only needed once).
- [ ] I have done a full self-review of my code.

## Reviewer checklist

    - [ ] Code should follow the [C++ Style Guide](https://cesium.com/learn/cesium-native/ref-doc/style-guide.html)
- [ ] Verify that the [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) has been submitted, if needed.


CesiumGeometry/include/CesiumGeometry/AxisAlignedBox.h
```diff
@@ -133,4 +134,20 @@ struct CESIUMGEOMETRY_API AxisAlignedBox final {
   static AxisAlignedBox fromPositions(const std::vector<glm::dvec3>& positions);
 };
 
+/**
+ * @brief Test if two axis aligned boxes intersect.
+ */
```
j9liu: Add `@param` and `@returns` for completeness

CesiumGeometry/include/CesiumGeometry/AxisAlignedBox.h
```diff
@@ -133,4 +134,20 @@ struct CESIUMGEOMETRY_API AxisAlignedBox final {
...
+bool CESIUMGEOMETRY_API
+intersects(const AxisAlignedBox& b0, const AxisAlignedBox& b1);
+
+/**
+ * @brief Return the equivalent OrientedBoundingBox for an AxisAlignedBox. The
+ * axes will be the principal axes.
+ *
+ * @param box the source bounding box
+ *
+ * @returns An oriented bounding box
```
j9liu:
```suggestion
 * @param box The source axis-aligned box.
 *
 * @returns The equivalent oriented bounding box.
```

CesiumGeometry/include/CesiumGeometry/OrientedBoundingBox.h
```diff
@@ -147,4 +147,10 @@ class CESIUMGEOMETRY_API OrientedBoundingBox final {
   glm::dvec3 _lengths;
 };
 
+/**
+ * @brief Test if two oriented bounding boxes intersect.
+ */
+bool CESIUMGEOMETRY_API
+intersects(const OrientedBoundingBox& b1, const OrientedBoundingBox& b2);
```
j9liu, line 158: Add `@param` and `@returns` for completeness

CesiumGeometry/src/AxisAlignedBox.cpp
```diff
@@ -23,4 +24,20 @@ AxisAlignedBox::fromPositions(const std::vector<glm::dvec3>& positions) {
 
   return AxisAlignedBox(min.x, min.y, min.z, max.x, max.y, max.z);
 }
+
+bool intersects(const AxisAlignedBox& b0, const AxisAlignedBox& b1) {
+  // Do all the axes overlap?
```
j9liu: **NIT**:

```suggestion
  // Check if all the axes overlap.
```

CesiumGeometry/include/CesiumGeometry/OrientedBoundingBox.h
j9liu, line 158: We should add unit tests for this.

CesiumGeometry/include/CesiumGeometry/AxisAlignedBox.h
```diff
@@ -133,4 +134,20 @@ struct CESIUMGEOMETRY_API AxisAlignedBox final {
...
+ */
+OrientedBoundingBox toOrientedBoundingBox(const AxisAlignedBox& box);
```
j9liu, line 155: We should add unit tests for these.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/GeneralCullingVolume.h
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * @brief A generalization of @ref CesiumGeometry::CullingVolume.
+ *
+ * CesiumGeometry::CullingVolume represents orthographic and perspective view
+ * frustums; GeneralCullingVolume includes @ref
+ * Cesium3DTilesSelection::BoundingVolume too, allowing filtering on
+ * e.g. geographic areas.
+ */
+using GeneralCullingVolume =
```
j9liu: Could go one step further...

```suggestion
using Volume =
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/ViewState.h
```diff
@@ -176,6 +194,11 @@ class CESIUM3DTILESSELECTION_API ViewState final {
     return this->_projectionMatrix;
   }
 
+  /**
+   * @brief Gets the selection measure.
+   */
+  double getSelectionMeasure() const { return this->_selectionMeasure; }
```
j9liu: So far we're strictly using this as a check against geometric error. So instead of "selection measure", we should give this a more precise name.

Additionally, this property should be an `optional` to reflect when a `ViewState` was not constructed with one. It gives more meaning than a ` > 0.0` check. A value of 0.0 would refine to the very maximum level of detail, which could be a valid/desired use case!

An off-the-dome suggestion:

```suggestion c++
  std::optional<double> getGeometricErrorThreshold() const { return this->_geometricErrorThreshold; }
```

Cesium3DTilesSelection/src/ViewState.cpp
```diff
@@ -254,6 +216,9 @@ double ViewState::computeDistanceSquaredToBoundingVolume(
 double ViewState::computeScreenSpaceError(
     double geometricError,
     double distance) const noexcept {
+  if (this->_viewportSize.y == 0.0) {
+    return geometricError;
+  }
```
j9liu: Change to a check for `this->_geometricErrorThreshold`?

The comment for this function should be updated to reflect how the output changes for a `ViewState` constructed by bounding volume, instead of by camera.

Cesium3DTilesSelection/src/BoundingVolume.cpp
```diff
@@ -277,4 +278,118 @@ OrientedBoundingBox getOrientedBoundingBoxFromBoundingVolume(
...
+          return intersectsImpl(bv0.toOrientedBoundingBox(), bv1);
+        } else if constexpr (std::is_same_v<
+                                 decltype(bv1),
+                                 const BoundingCylinderRegion&>) {
+          return intersectsImpl(bv1.toOrientedBoundingBox(), bv0);
+        }
+        return false;
+      },
+      volume0,
+      volume1);
```
j9liu, line 390: This looks rather heavy handed compared to the `struct Visitor { bool operator(...) }` that we normally do. Unless there's some obvious advantage to this that I'm missing, we should use a more readable `struct Visitor` than a beefy lambda with many template checks.

Cesium3DTilesSelection/src/ViewState.cpp
```diff
@@ -119,85 +121,45 @@ ViewState::ViewState(
...
+  LocalHorizontalCoordinateSystem enu{
+      *_positionCartographic,
+      LocalDirection::East,
+      LocalDirection::North,
+      LocalDirection::Up,
+      1.0,
+      ellipsoid};
+  _viewMatrix = enu.getLocalToEcefTransformation();
+  _position = positionFromView(_viewMatrix);
+  _direction = directionFromView(_viewMatrix);
```
j9liu: Use `this->` per our [style guide.](https://cesium.com/learn/cesium-native/ref-doc/style-guide.html#other)

Cesium3DTilesSelection/src/ViewState.cpp
```diff
@@ -119,85 +121,45 @@ ViewState::ViewState(
...
 }
-} // namespace
+
+namespace {} // namespace
```
j9liu: Delete?

Cesium3DTilesSelection/src/ViewState.cpp
```diff
@@ -119,85 +121,45 @@ ViewState::ViewState(
...
-    operator()(const BoundingCylinderRegion& boundingCylinderRegion) noexcept {
-      return Cesium3DTilesSelection::isBoundingVolumeVisible(
-          boundingCylinderRegion,
-          viewState._cullingVolume);
-    }
-  };
-
-  return std::visit(Operation{*this}, boundingVolume);
+  return Cesium3DTilesSelection::isBoundingVolumeVisible(
+      _cullingVolume,
```
j9liu: Use `this->` per our [style guide.](https://cesium.com/learn/cesium-native/ref-doc/style-guide.html#other)

```suggestion
      this->_cullingVolume,
```

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -1125,7 +1133,18 @@ TraversalDetails visitTileIfNeeded(
   }
 
   double tileSse = computeSse(context, frameState, tile);
-  bool meetsSse = meetsSseThreshold(context, tileSse, cullResult.culled);
+  auto fixedErrorIt = std::min_element(
```
j9liu: This variable name could be more specific:

```suggestion
  auto lowestGeometricErrorThresholdIt = std::min_element(
```

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -1125,7 +1133,18 @@ TraversalDetails visitTileIfNeeded(
...
+      frameState.frustums.begin(),
+      frameState.frustums.end(),
+      [](const ViewState& a, const ViewState& b) {
+        return a.getSelectionMeasure() < b.getSelectionMeasure();
+      });
```
j9liu, line 1154: We really should turn this into a check for `optional`s before comparing values. Not all view states (and in most cases, none) will be using this geometric error metric.

Cesium3DTilesSelection/src/TilesetSelection.cpp
```diff
@@ -1125,7 +1133,18 @@ TraversalDetails visitTileIfNeeded(
...
+  bool meetsSse = meetsSseThreshold(
+      context,
+      tileSse,
+      cullResult.culled,
+      tile,
+      fixedErrorIt->getSelectionMeasure());
```
j9liu: Instead of shoving the `fixedError` check into `meetsSseThreshold`, can we change this to something like:

```c++
std::optional<double> geometricErrorThreshold; // get it above
bool meetsSelectionCriteria = geometricErrorThreshold
         ? meetsGeometricErrorThreshold()
         : meetsSseThreshold();
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/GeneralCullingVolume.h
```diff
@@ -0,0 +1,33 @@
...
+namespace Cesium3DTilesSelection {
+
+/**
+ * @brief A generalization of @ref CesiumGeometry::CullingVolume.
+ *
+ * CesiumGeometry::CullingVolume represents orthographic and perspective view
+ * frustums; GeneralCullingVolume includes @ref
+ * Cesium3DTilesSelection::BoundingVolume too, allowing filtering on
+ * e.g. geographic areas.
+ */
```
j9liu, line 18: IMO `CesiumGeometry::CullingVolume` should be renamed to `CesiumGeometry::Frustum` so this can just be `CullingVolume`. But in lieu of that bigger change, I would tweak this comment:

```suggestion
/**
 * @brief A generalization of a culling volume for selecting and filtering tiles.
 * 
 * Whereas @ref CesiumGeometry::CullingVolume represents orthographic and perspective view
 * frustums, @ref Cesium3DTilesSelection::BoundingVolume allows for acting on more general volume shapes.
 */
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/GeneralCullingVolume.h
```diff
@@ -0,0 +1,33 @@
...
+using GeneralCullingVolume =
+    std::variant<CesiumGeometry::CullingVolume, BoundingVolume>;
+
+/**
+ * @brief test if a general culling volume intersects a bounding volume.
+ *
+ * @param cullingVolume the general culling volume.
+ * @param boundingVolume the bounding volume.
+ * @return true if volumes intersect.
```
j9liu:
```suggestion
 * @brief Tests if a general culling volume intersects a bounding volume.
 *
 * @param cullingVolume The general culling volume.
 * @param boundingVolume The bounding volume.
 * @return `true` if the volumes intersect, false otherwise.
```

j9liu (changes requested):
Thanks @timoore! I didn't check the math closely, my comments are mostly about organization, naming, and cleanup.

Please also update CHANGES.md for the public-facing changes here!

CesiumGeometry/include/CesiumGeometry/OrientedBoundingBox.h
timoore: There is a test for this in Cesium3DTilesSelection/test/TestBoundingVolume.cpp, but it would be good to have them in CesiumGeometry as well.

azrogers:
@timoore The PR looks good, but there are some remaining comments from Janine besides just the documentation changes. Can you take a look at those? (And if you've already taken a look at those, can you respond to them / resolve them?)

Cesium3DTilesSelection/include/Cesium3DTilesSelection/ViewState.h
timoore: I've gone with the use of optional, but I'm not going to implement your suggested semantic for 0 fixed error yet. The tile selection algorithm is a bit too involved easily handle that change.

Cesium3DTilesSelection/src/BoundingVolume.cpp
timoore: The advantage I saw was not having to write out the cross product of the different volume types. This way, the strategy (Is one volume a bounding sphere? Then test against that; An oriented bounding box, test that; etc) is written out directly instead of being scattered across 36 `operator()` functions. Maybe you could get away with fewer templated functions, but I do think this way is more explicit.

Cesium3DTilesSelection/src/ViewState.cpp
timoore: Feels very strange inside of a constructor...

azrogers:
@timoore The changes you made look good! If you can just add those requested tests for the functions you added to `AxisAlignedBox`, I think it's good to go!

timoore:
> @timoore The changes you made look good! If you can just add those requested tests for the functions you added to `AxisAlignedBox`, I think it's good to go!

I think you mean `OrientedBoundingBox`. There tests for the new intersection functions, but they are in Cesium3DTilesSelection/test/TestBoundingVolume.cpp.

azrogers:
Thanks @timoore!

Issue #1143, opened 2025-03-31
Query tilesets by bounding volume at a particular LOD


kring:
`Tileset::updateView` effectively allows you to query for a set of tiles that meets a particular screen-space error from the viewpoint of a camera with a particular perspective frustum. After #1125, this will be extended to multiple independent views. And after #1135, it will be extended to orthographic views.

It would be useful to further extend this idea to allow querying of a set of tiles within a particular bounding box or sphere at a particular level-of-detail.

More generally, we should make it easy for users to implement these kinds of queries external to the `Tileset`, while still leveraging the progressive loading that `Tileset` does for views and height queries.

CC @Lee-Bull

PR #1442, opened 2026-09-11
Level paging

timoore:
## Author checklist

- [X] I have submitted a [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) (only needed once).
- [X] I have done a full self-review of my code.

## Remaining Tasks

Unit tests

## Reviewer checklist

    - [ ] Code should follow the [C++ Style Guide](https://cesium.com/learn/cesium-native/ref-doc/style-guide.html)
- [ ] Verify that the [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) has been submitted, if needed.


j9liu (changes requested):
Thanks for the PR @timoore!

It's true that we should move towards a generalizable criteria for tile selection. However, I think this interface still needs some development before we can truly release it. As it stands, the simple geometric error approach introduced in #1435 is now being convoluted by an "error measure" function that could potentially replace the geometric error computation, even though it is being used in the same logical paths throughout tile selection. 

Instead of requiring `computeErrorMeasure` and `meetsErrorThreshold` to be implemented, it would be cleaner and more efficient to implement a `shouldRefineTile`. This function could expose `ViewState`, `Tile` info, etc. to allow for more flexible refinement logic. The default behavior would be to compute the SSE of a tile.

However, the goal of this feature was to implement level-of-detail selection based on the depth of the tree. As we had previously discussed offline, the cleanest approach with our current architecture would be to put this as a parameter on `TilesetOptions`, then to check for the depth throughout the tile selection code. Unless we're confident within the time constraints that we could do this interface right (which personally, I am not), we should pivot to the simpler approach. We can always come back to abstracting this interface later.

timoore:
> Thanks for the PR @timoore!
> 

> However, the goal of this feature was to implement level-of-detail selection based on the depth of the tree. As we had previously discussed offline, the cleanest approach with our current architecture would be to put this as a parameter on `TilesetOptions`, then to check for the depth throughout the tile selection code. Unless we're confident within the time constraints that we could do this interface right (which personally, I am not), we should pivot to the simpler approach. We can always come back to abstracting this interface later.

I went with attaching this property to the view instead of the tileset because, it seems, changing it requires reloading the tileset in the current Cesium Native and and client implementations. Maybe this isn't a big deal -- changing the screen space error already requires a reload -- but changing these level properties, geometric error, area for paging etc. is something that should be fast to execute.

j9liu:
> changing it requires reloading the tileset in the current Cesium Native and and client implementations

@timoore This isn't true of Unreal. [`ACesium3DTileset::updateOptionsFromTilesetProperties`](https://github.com/CesiumGS/cesium-unreal/blob/8e45cd9f1be83da72edba2bf32558ffa187831ca/Source/CesiumRuntime/Private/Cesium3DTileset.cpp#L1912) allows configuring of cesium-native `TilesetOptions` without a tileset reload.

This _is_ true of Unity, but it's because of our lack of implementation. https://github.com/CesiumGS/cesium-unity/pull/412 was opened to address this, but we never got the signed CLA to review and merge.

timoore:
> > changing it requires reloading the tileset in the current Cesium Native and and client implementations
> 
> @timoore This isn't true of Unreal. [`ACesium3DTileset::updateOptionsFromTilesetProperties`](https://github.com/CesiumGS/cesium-unreal/blob/8e45cd9f1be83da72edba2bf32558ffa187831ca/Source/CesiumRuntime/Private/Cesium3DTileset.cpp#L1912) allows configuring of cesium-native `TilesetOptions` without a tileset reload.
> 
> This _is_ true of Unity, but it's because of our lack of implementation. [CesiumGS/cesium-unity#412](https://github.com/CesiumGS/cesium-unity/pull/412) was opened to address this, but we never got the signed CLA to review and merge.

Okay then! I really thought that the tileset options were immutable after tileset creation.

I'm not sure if moving the level check into the rest of the tileset selection code really simplifies things: the level is completely equivalent to a tile geometric error, and is in fact the  same thing for implicit tilesets. I'm also not sure about the interface presented to the clients of Cesium Native. I will take another look at this.

j9liu:
> the level is completely equivalent to a tile geometric error

They are related mathematically, but I don't believe this should manifest in the public API.

The related feature requests (https://github.com/CesiumGS/cesium-unreal/issues/1811 and https://github.com/CesiumGS/cesium-native/issues/1143, to name a few) imply control over which LOD is chosen. Although geometric error correlates to LOD, something like `geometricErrorThreshold` or `desiredGeometricError` makes it hard to estimate to get the LOD you want, without digging into the details of a tileset.

The easiest way to identify LODs, without regard for a tileset's internal organization, would be an integer level. If we allow users to specify an integer level on the `TilesetOptions` in some way, then there's more freedom to do what we need under-the-hood, even if it means we convert back to geometric error. We should keep in mind that not all tilesets are implicit, and some may have varying tree depth / geometric error that we don't anticipate.

Feel free to propose alternatives but I think this is easiest, especially to meet the October release milestone.

timoore:
Here's a summary of the offline meeting attended by @j9liu, @azrogers, @laurenfrederick, and myself. We decided that the current approach is fine for short-term needs to get new functionality into Cesium Native, but probably needs more work for a long-term solution. The long-term issue is making explicit the interaction of a ViewState with ViewGroup objects. Multiple view states can be passed to a view group for selection, but the "most detailed" tile always wins and is returned in the resulting list of tiles, so an error measure predicate might end up being a no-op.

We talked through how the measure handler supports tile selection by depth in the tileset and agreed that tests and documention are needed to illustrate this for a user. These are incoming.

I explained that the tileset selection algorithm relies on some kind of value that can be computed, compared, and then passed down in when recursively visiting tiles, and that it makes sense to keep using a `double` for this until we understand the algorithm better.

The structure of the current code tries to keep the existing SSE path fast while allowing novel selection methods to be specified by the Cesium Native user. It would be good to verify that the existing algorithm can be implemented completely with error measure handler; it might be necessary to pass the ViewState object to the handler as an additional argument to really make that possible. Again, more tests and documentation will make that clearer.

j9liu:
Thanks @timoore for the writeup. I also remember that we agreed to replace `ViewState`'s `geometricErrorThreshold` entirely with the new `ErrorMeasureHandler`, because the latter makes it redundant. Is that correct?

timoore:
> Thanks @timoore for the writeup. I also remember that we agreed to replace `ViewState`'s `geometricErrorThreshold` entirely with the new `ErrorMeasureHandler`, because the latter makes it redundant. Is that correct?

Yup, that's right! Thanks for noting that.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/ViewState.h
```diff
@@ -115,18 +145,19 @@ class CESIUM3DTILESSELECTION_API ViewState final {
...
    * constructor does not specify a viewport, and so doesn't use Screen Space
-   * Error (SSE) as a selection criteria.
+   * Error (SSE) as a selection criteria. Instead, a user-supplied functor
+   * object calculates a measure used as a standin for SSE.
    *
    * @param boundingVolume The geographic viewing volume
-   * @param geometricErrorThreshold Value used for selection as an alternative
-   * to screen space error.
+   * @param errorMeasureHandler `std::shared_ptr` to object that implements
+   * measure calculation
```
j9liu:
```suggestion
   * @param errorMeasureHandler A `std::shared_ptr` to the ErrorMeasureHandler
   * used to compute an error for tiles in view.
```

CHANGES.md
```diff
@@ -5,6 +5,7 @@
 ##### Additions :tada:
 
 - The `CesiumVectorOverlays::VectorTilesRasterOverlay` now supports styling via specifying a `VectorStylingProvider` through the `pStylingProvider` parameter in `VectorTilesRasterOverlayOptions`.
+- `Cesium3DTilesSelection::ViewState` can be supplied an object that computes tile error measure.
```
j9liu: Since we removed `getGeometricErrorThreshold`, we should announce this as a breaking change and encourage people to adopt the new interface instead.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/ViewState.h
```diff
@@ -13,10 +13,40 @@
...
 
+class Tile;
+
+/**
+ * @brief class for customizing calculation and comparison of the selection
+ * measure for tiles. Usually this value is the screen space error, but it be
+ * another criteria such as a fixed geometric error or the depth of a tile in
+ * the tileset.
+ */
+class CESIUM3DTILESSELECTION_API ErrorMeasureHandler {
```
j9liu: This name could be improved to be more obviously related to `ViewState` (alternative suggestions welcome):

```suggestion
 * @brief A handler for calculating a numerical error metric for tiles
 * seen by a @ref ViewState to influence their selection. This is usually 
 * screen space error, but can be implemented to meet other criteria 
 * such as a fixed geometric error or tile depth.
class CESIUM3DTILESSELECTION_API ViewStateErrorMetric {
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/ViewState.h
```diff
@@ -115,18 +145,19 @@ class CESIUM3DTILESSELECTION_API ViewState final {
...
    * associated with a geographic area, as opposed to a viewing projection. This
    * constructor does not specify a viewport, and so doesn't use Screen Space
-   * Error (SSE) as a selection criteria.
+   * Error (SSE) as a selection criteria. Instead, a user-supplied functor
+   * object calculates a measure used as a standin for SSE.
    *
    * @param boundingVolume The geographic viewing volume
-   * @param geometricErrorThreshold Value used for selection as an alternative
-   * to screen space error.
+   * @param errorMeasureHandler `std::shared_ptr` to object that implements
```
j9liu:
```suggestion
   * @param errorMeasureHandler A `std::shared_ptr` to the metric used to compute an error for tiles in view.
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/ViewState.h
```diff
@@ -243,6 +267,26 @@ class CESIUM3DTILESSELECTION_API ViewState final {
   double computeScreenSpaceError(double geometricError, double distance)
       const noexcept;
 
+  /**
+   * @brief Computes the screen space error from a given geometric error
+   *
+   * Computes the screen space error (SSE) that results from the given
+   * geometric error, when it is viewed with this camera from the given
+   * distance.
```
j9liu, line 275:
```suggestion
   * @brief Computes the screen space error for a given tile.
   *
   * Computes the screen space error (SSE) that results from the given
   * tile's geometric error when viewed with this ViewState from the
   * specified distance.
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/ViewState.h
```diff
@@ -243,6 +267,26 @@ class CESIUM3DTILESSELECTION_API ViewState final {
...
+   *
+   * The given distance will be clamped to a small positive value if
+   * it is negative or too close to zero.
+   *
+   * @param tile the tile
+   * @param distance The viewing distance
+   * @param depth level in the tileset traversal
```
j9liu: Future note: these `@param`s should be capitalized in our docs.

```suggestion
   * @param tile The tile.
   * @param distance The tile's distance from the ViewState.
   * @param The tile's depth relative to the tileset.
```

Cesium3DTilesSelection/src/ViewState.cpp
```diff
@@ -211,17 +215,23 @@ double ViewState::computeDistanceSquaredToBoundingVolume(
...
+    uint32_t depth) const noexcept {
+  if (this->_errorMeasureHandler) {
+    return this->_errorMeasureHandler->computeErrorMeasure(
+        tile,
+        distance,
+        depth);
+  } else {
+    return this->computeScreenSpaceError(tile.getGeometricError(), distance);
+  }
+}
```
j9liu, line 230: If `_errorMeasureHandler` is defined then this is no longer screen space error we're computing, so this is somewhat dishonest. Could we rename for `computeTileError` or something similar?

j9liu (changes requested):
Thanks @timoore ! This looks a lot cleaner, and I have no functional feedback. However, there's some doc suggestions and a few naming changes just so we're precise about what we mean by "error" around the code. This will help with with future maintainability.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/ViewState.h
timoore: I decided to go with `ViewStateMeasureDelegate` and changed the function names too. I thought about putting it inside of the `ViewState` class, but I'm not to keen on that in general.

Cesium3DTilesSelection/include/Cesium3DTilesSelection/ViewState.h
```diff
@@ -115,18 +145,19 @@ class CESIUM3DTILESSELECTION_API ViewState final {
...
+   * @param measureDelegate `std::shared_ptr` to the delegate that implements
+   * measure calculation
    * @param ellipsoid The ellipsoid that will be used to compute the
    * {@link ViewState#getPositionCartographic cartographic position} and other
    * parameters for tile selection.
    */
   ViewState(
       const BoundingVolume& boundingVolume,
-      double geometricErrorThreshold,
+      std::shared_ptr<ViewStateMeasureDelegate> measureDelegate,
```
j9liu, line 160:
```suggestion
      std::shared_ptr<ViewStateMeasureDelegate> pMeasureDelegate,
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/ViewState.h
```diff
@@ -254,7 +298,7 @@ class CESIUM3DTILESSELECTION_API ViewState final {
   Cesium3DTilesSelection::GeneralCullingVolume _cullingVolume;
   glm::dmat4 _viewMatrix;
   glm::dmat4 _projectionMatrix;
-  std::optional<double> _geometricErrorThreshold;
+  std::shared_ptr<ViewStateMeasureDelegate> _measureDelegate;
```
j9liu, line 301:
```suggestion
  std::shared_ptr<ViewStateMeasureDelegate> _pMeasureDelegate;
```

Cesium3DTilesSelection/include/Cesium3DTilesSelection/ViewState.h
```diff
@@ -13,10 +13,40 @@
...
+ * in the tileset.
+ */
+class CESIUM3DTILESSELECTION_API ViewStateMeasureDelegate {
+public:
+  /**
+   * @brief compute the error measure, a generalization of screen space error.
+   * @param tile the tile being tested
+   * @param distance distance from the ViewState's position to the tile
+   * @param depth the depth (level) of the tile in the tileset.
+   * @return the error measure
```
j9liu, line 38:
```suggestion
   * @brief Compute the error measure, a generalization of screen space error.
   * @param tile The tile being tested
   * @param distance Distance from the ViewState's position to the tile
   * @param depth The depth (level) of the tile in the tileset.
   * @return The error measure
```

Cesium3DTilesSelection/src/ViewState.cpp
timoore: I've held off on this one only because it's the name of the function called in the tileset selection algorithm, even if "error" is now more general than screen space error. Maybe `computeTileError` is the thing.

PR #1448, opened 2026-09-18
Handle Draco zero faces and short point mappings

bjornblissing:
## Description

Hardens Draco mesh decoding against malformed data that can otherwise trigger
undefined behavior or leave stale compressed accessors in the decoded model.

The change handles two cases in `CesiumGltfReader/src/decodeDraco.cpp`:

- A Draco mesh with zero faces no longer attempts to access its first face while
  constructing the decoded index buffer. The zero-length buffer and zero-count
  index accessor are retained, and index copying is skipped safely.
- A non-identity Draco attribute mapping that contains fewer entries than the
  mesh has points is rejected before Draco is indexed by point number. A warning
  is emitted and the corresponding attribute is removed from the primitive,
  rather than copying out-of-bounds data or leaving the original compressed
  accessor advertised after the Draco extension is removed.

This keeps the resulting glTF model internally consistent even when decoding
crafted or corrupted Draco bitstreams.

## Issue number or link

N/A

## Author checklist

## Testing plan

Exercise Draco decoding with malformed or edge-case meshes covering:

1. A mesh that decodes successfully but contains zero faces. Verify that
   decoding does not access a nonexistent face and produces a zero-count index
   accessor with an empty buffer.
2. A mesh containing a non-identity attribute mapping shorter than the decoded
   point count. Verify that decoding emits a warning and removes the affected
   attribute from the primitive instead of reading beyond the mapping.
3. A normal Draco mesh with identity and complete attribute mappings. Verify
   that valid decoded attributes and indices are still copied as before.

The changes are defensive checks in the Draco decoding path; no user-facing
format or API changes are introduced.


CesiumGltfReader/src/decodeDraco.cpp
j9liu, line 1: I think it's worth mentioning these fixes in `CHANGES.md`, e.g., something like "fixed potential out-of-bounds access while decoding Draco-compressed glTFs."

CesiumGltfReader/src/decodeDraco.cpp
```diff
@@ -226,6 +235,14 @@ void copyDecodedAttribute(
   CESIUM_ASSERT(readGltf.model.has_value());
   CesiumGltf::Model& model = readGltf.model.value();
 
+  if (!pAttribute->is_mapping_identity() &&
+      pAttribute->indices_map_size() < pMesh->num_points()) {
```
j9liu, line 239: Would this comment be correct? I'm not familiar with how Draco works, but a small explainer would help:

```suggestion
  // For explicit index mapping, the remapped indices must match the
  // number of points in the mesh.
  if (!pAttribute->is_mapping_identity() &&
      pAttribute->indices_map_size() < pMesh->num_points()) {
```

j9liu (commented):
Thanks @bjornblissing for the PR! I just have two small comments that are doc-related.

CesiumGltfReader/src/decodeDraco.cpp
bjornblissing: Added a note in the `changes.md`

CesiumGltfReader/src/decodeDraco.cpp
bjornblissing: Added a comment explaining the check.

j9liu (approved):
Thanks @bjornblissing !

